PR #6, opened 2020-02-10
WIP Getting simple box example to work

### LINKED ISSUE #5: Render a simple mesh
We left master in a broken state during last friday's refactoring session; let's get it to the point where can render a simple mesh again.

almarklein:
closes #5

TODO:
* [x] vertex buffers
* [ ] index buffers
* [ ] uniforms
* [x] let material define compute shaders
* [ ] python-shader support for matrices (and multiplying with vectors)
* [ ] take camera into account
* [ ] take world transform into account
* [ ] Skim over all todo's that were added


Korijn:
Resolved #5

almarklein:
Ok, things are more or less working. The triangle example and the qt-embedding example work again. The box example draws *something* on my machine, but that's because I have made changes to wgpu that I have yet to push :)

PR #8, opened 2020-02-12
Triangle material

Korijn:
Refactored a bit; pushed the "special" stuff down the data model into a `TriangleMaterial` so the renderer and examples can just lean on the Mesh data model and no longer need anything special.

I removed all the `get_renderer_info_*` methods because all they do is return data in dicts whereas the classes they're defined on are already data classes. No need to add another layer of data models. Duck typing will be just as hard on the data classes as it will be on dictionaries.


visvis2/renderers/_wgpu.py
```diff
@@ -38,16 +39,21 @@ def traverse(self, obj):
...
-
         device = self._device
 
-        info = wobject.get_renderer_info_wgpu()
-        if not info:
+        # object type determines pipeline composition
+        if not isinstance(wobject, Mesh):
             return None, None, None
 
+        if not wobject.material.dirty and hasattr(wobject, "_pipeline_info"):
```
almarklein: nice!

visvis2/material/_base.py
```diff
@@ -1,2 +1,6 @@
 class Material:
-    pass
+    def __init__(self):
+        self.dirty = True
+        self.shaders = {}
+        self.uniforms = {}
+        self.primitiveTopology = None
```
almarklein: I can imagine that having a method to get these info can be beneficial in some cases, e.g. to to some JIT logic. But let's see, maybe this will hold up just fine :)

Korijn: this _is_ the info :)

almarklein: Haha, I get that. I meant *with a getter* to support the info to be more "dynamic" where needed. Anyway, we might never need it, so let's not overdesign now.

PR #12, opened 2020-02-14
Remove throw_loop

Korijn:
I feel like this is a pointless effort and a waste of CPU cycles :) users should just not create loops

### DIFF
--- a/tests/objects/test_world_object.py
+++ b/tests/objects/test_world_object.py
@@ -1,0 +1,133 @@
+from math import pi
+from unittest.mock import Mock, call
+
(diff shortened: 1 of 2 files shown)


almarklein:
If found it to be nice. Users have a thing for shooting one's own foot. But perhaps this is not the best place to check it; we could also check it in the renderer where whe create the flat list/iter of world objects.

Korijn:
That makes more sense. I wanted to open another PR to transform the graph into an ordered list anyway. There's a bunch of stuff to tackle there.

PR #13, opened 2020-02-14
Implement index buffers and uniform buffers

almarklein:
* [x] Implement index buffers
* [x] Implement uniform buffers (expimental approach)
* [x] Diverge
* [x] Diverge some more, try stuff, move things around, etc.
* [x] Converge
* [x] Refactor / clean up
* [x] put world-transform in per-object uniform buffer
* [x] Merge / release wgpu with needed changes: https://github.com/almarklein/wgpu-py/pull/62
* [x] Merge / release python-shader with needed changes: https://github.com/almarklein/python-shader/pull/21


visvis2/geometry/_base.py
```diff
@@ -1,5 +1,9 @@
...
+
+
 class Geometry:
-    def __init__(self):
-        self.index = None
-        self.vertex_data = []
+    def __init__(self, index=None, *vertex_data):
+        # todo: don't cast if an array is already given
+        self.index = None if index is None else np.array(index, np.uint32)
+        self.vertex_data = [np.array(array, np.float32) for array in vertex_data]
```
Korijn: why is this a list instead of a dict? don't you need to know if something is for `position` or `color` for example?

visvis2/geometry/_base.py
```diff
@@ -1,5 +1,9 @@
+import numpy as np
+
+
 class Geometry:
-    def __init__(self):
-        self.index = None
-        self.vertex_data = []
+    def __init__(self, index=None, *vertex_data):
+        # todo: don't cast if an array is already given
```
Korijn: use `np.asarray` then :) easy fix

visvis2/renderers/_wgpu.py
```diff
@@ -43,10 +43,10 @@ def compose_pipeline(self, wobject):
 
         # object type determines pipeline composition
         if not isinstance(wobject, Mesh):
-            return None, None, None
+            return None
 
-        if not wobject.material.dirty and hasattr(wobject, "_pipeline_info"):
-            return wobject._pipeline_info
+        if not wobject.material.dirty and hasattr(wobject, "_wgpu_info"):
+            return wobject._wgpu_info
```
Korijn: like it

visvis2/geometry/_base.py
almarklein: Because binding in the shader is based on integers. Unless we put some sauce on top to bind with strings ...

Korijn: :tomato:

visvis2/geometry/_base.py
Korijn: Nvm

visvis2/material/_triangle.py
```diff
@@ -1,7 +1,11 @@
 from ._base import Material
 
 import wgpu  # only for flags/enums
-from python_shader import python2shader, RES_INPUT, RES_OUTPUT
+from python_shader import python2shader, RES_INPUT, RES_OUTPUT, RES_UNIFORM
+from python_shader import Struct, vec3
+
+
+uniform_type = Struct(color=vec3)
```
almarklein: See how we define the uniform structure once, and then use it both in the shader and the material!

visvis2/material/_triangle.py
```diff
@@ -33,4 +38,15 @@ def __init__(self):
...
         }
+
         self.primitive_topology = wgpu.PrimitiveTopology.triangle_list
+
+        # Instantiate uniforms. This produces a ctypes struct object.
+        # The renderer will use it to create a mapped buffer, create a
+        # new ctypes object with the same type, mapped onto that buffer,
+        # and copy the original data over. So we can just assign fields
+        # of our uniforms object and it Just Works!
+        self.uniforms = uniform_type(color=(1, 0, 0))
```
almarklein: I think I kinda like this :)

almarklein: An alternative is to use a numpy array with a structured dtype.

visvis2/material/_triangle.py
```diff
@@ -1,7 +1,11 @@
 from ._base import Material
 
 import wgpu  # only for flags/enums
-from python_shader import python2shader, RES_INPUT, RES_OUTPUT
+from python_shader import python2shader, RES_INPUT, RES_OUTPUT, RES_UNIFORM
+from python_shader import Struct, vec3
```
Korijn: The only concern I have is that this struct is tightly coupled to python-shader. We are going to lean on it very heavily I guess but we should keep in mind that it should remain possible for users to bring their own shaders. They might want to use glsl2spirv type shaders for example.

almarklein: Not necessarily. The only requirement (but we should formalize this some more, if this is the way we want to go) is that `.uniforms` is a ctypes object. Instantiating the `Struct(color=vec3)` is just a convenient way to create such a ctypes object.

And if numpy is a dependency, it might make more sense to use a numpy object instead.

Korijn: So then why exactly do you need to import it from python shader?

almarklein: Convenience. It allows to declare the type once, and use it both in the shader and for the `.uniforms` object. So they are guaranteed to match up. But your questions may suggest it's a bit too magical, perhaps. Maybe it would help to make it more explicit, e.g. `uniform_type.instantiate_ctypes_object()`?

Korijn: Let's keep it for now and see where it takes us :)

visvis2/renderers/_wgpu.py
```diff
@@ -83,6 +83,32 @@ def compose_pipeline(self, wobject):
...
+            index_buffer = None
+            index_format = wgpu.IndexFormat.uint32
+        else:
+            array = wobject.geometry.index
+            nbytes = array.nbytes
+            usage = wgpu.BufferUsage.INDEX
+            index_buffer = device.create_buffer_mapped(size=nbytes, usage=usage)
+            # Copy data from array to buffer
+            ctypes.memmove(index_buffer.mapping, array.ctypes.data, nbytes)
+            index_buffer.unmap()
```
Korijn: this part of the wgpu API is quite elegant! nice!

visvis2/renderers/_wgpu.py
```diff
@@ -95,7 +121,7 @@ def compose_pipeline(self, wobject):
             ctypes.memmove(buffer.mapping, array.ctypes.data, nbytes)
             buffer.unmap()
             shader_location = len(buffers)
```
Korijn: perhaps unrelated but why is `shader_location = len(buffers)`?

visvis2/renderers/_wgpu.py
```diff
@@ -95,7 +121,7 @@ def compose_pipeline(self, wobject):
...
-            buffers[shader_location] = buffer
+            # buffers[shader_location] = buffer
             vbo_des = {
                 "array_stride": 3 * 4,
```
Korijn: is this 3 * 4 because every vertex has 3 x float32 (4 bytes)?

visvis2/renderers/_wgpu.py
```diff
@@ -95,7 +121,7 @@ def compose_pipeline(self, wobject):
...
                 "stepmode": wgpu.InputStepMode.vertex,
```
Korijn: I see this enum also has the option "instance", do you know what that is?

visvis2/renderers/_wgpu.py
Korijn: Looked it up in the spec:

> Each attribute appears as a separate input in a vertex shader, each bound by a numeric location, which is specified by shaderLocation.

So you're probably counting as you expand the array. Perhaps using `enumerate()` in the for loop would be a little more explicit?

almarklein: Nice catch. I think you're right and this should just enumerate. This is probably a leftover from copying this code. I have to get the namespaces (slot-spaces?) straight.

visvis2/renderers/_wgpu.py
almarklein: Exactly

visvis2/renderers/_wgpu.py
almarklein: Instance rendering. So in that case the vertex does not change per-vertex, but per instance, I suppose :)  We should look into that once we are in a better shape. I think instance rendering can be very useful indeed.

examples/geometry_cube.py
```diff
@@ -25,12 +24,22 @@ def animate():
...
+    rot = vv.linalg.Quaternion().set_from_euler(vv.linalg.Euler(0.005, 0.01))
+    cube.rotation.multiply(rot)
 
     # actually render the scene
     renderer.render(scene, camera)
 
 
+timer = QtCore.QTimer()
+timer.setInterval(1000 * 1 / 30)
+timer.timeout.connect(canvas.update)
```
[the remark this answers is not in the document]
almarklein: Yes, that's `QWidget.update()` which requests a new draw. And `animate()` gets called on each draw.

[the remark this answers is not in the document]
almarklein: Idea: expose a cross-gui method for this.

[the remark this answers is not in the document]
Korijn: > Idea: expose a cross-gui method for this.

I would suggest trying to take some inspiration from `window.requestAnimationFrame` if possible ([docs](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)).

It has some nice properties:

- animation is paused when the canvas is not visible
- refresh rate of the GUI environment is matched if possible
- callback is passed the time delta since the last call
- you can request another call yourself; or not!

visvis2/geometry/_mesh.py
```diff
@@ -0,0 +1,18 @@
...
+class MeshGeometry(Geometry):
+    """ Geometry for a polygonial mesh.
+    """
+
+    # todo: we can use a definition like this to ensure that the shaders use the correct types
+    # todo: but then we must also validate that any binding arrays that get set match up
+    binding_def = {
+        "positions": (0, Array(vec4)),
+        "colors": (1, Array(vec4)),
+    }
```
Korijn: I think in THREE.JS you just take a regular Geometry and add the buffers + bindings as they are available. Not every mesh has a color buffer for example. Example

```js
var geometry = new THREE.BufferGeometry();
var vertices = new Float32Array( [
	-1.0, -1.0,  1.0,
	 1.0, -1.0,  1.0,
	 1.0,  1.0,  1.0,
	 1.0,  1.0,  1.0,
	-1.0,  1.0,  1.0,
	-1.0, -1.0,  1.0
] );
geometry.setAttribute( 'position', new THREE.BufferAttribute( vertices, 3 ) );
var material = new THREE.MeshBasicMaterial( { color: 0xff0000 } );
var mesh = new THREE.Mesh( geometry, material );
```

`geometry.setAttribute` defines the binding and supplies the buffer in one call here and avoids the need for a `MeshGeometry` subclass.

visvis2/geometry/_dynmesh.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+# I think this is where this thing goes, right?
+
+
+class DynamicMeshGeometry(Geometry):
+    """ A datasource, euh, I mean Geometry, specifically for
+    triangulated meshes, supporting resampling and a variety of methods
+    to query the mesh and make changes to it.
+    """
```
Korijn: I would strongly recommend against trying to support this functionality through the class hierarchy.

visvis2/material/_base.py
```diff
@@ -1,6 +1,28 @@
...
+    projection_transform=mat4,
+    physical_size=vec2,
+    logical_size=vec2,
+)
+
+
+def array_from_shader_type(spirv_type):
+    """ Get a numpy array object from a SpirV type from python-shader.
+    """
+    return np.asarray(spirv_type())
```
Korijn: should this go into python-shader?

visvis2/material/_line.py
```diff
@@ -1,7 +1,27 @@
...
 import python_shader
-from python_shader import vec3, mat4
+from python_shader import vec3, mat4, Array
+
+
+# Idea:
+# scene objects (like stdinfo, lights) are at bind group 0
+# the objects defined by the geometry are at bind group 1
+# the objects defined by the material are at bind group 2
+# that leaves (by default) one bind group to be used for something else
```
Korijn: are there always 4 bind groups?

visvis2/geometry/_dynmesh.py
ivoflipse: What would be the benefit of using a class for this anyway?

examples/geometry_cube.py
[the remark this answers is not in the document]
almarklein: Most of it already works this way. The drawing is scheduled by the GUI (and will probably not draw when hidden). You're right that we can just call `canvas.update()` inside `animate()`, so we can get rid of the timer. Tracking the rest in https://github.com/almarklein/wgpu-py/issues/63

visvis2/geometry/_dynmesh.py
almarklein: I'm trying to see how pieces like this fit into the puzzle. So how do you think that this functionality should be fit into (the eventual) visvis2?

visvis2/material/_base.py
almarklein: Still working on this part. I think this particular function not, because `python-shader` is numpy-agnostic.

visvis2/material/_line.py
almarklein: The default number of bind groups (that is requested when you create a device) is 4. You can request more, but then the hardware might say no.

visvis2/geometry/_dynmesh.py
Korijn: Let me think about this one

Korijn: Read the section "BufferGeometry" on this page: https://threejs.org/docs/#manual/en/introduction/How-to-update-things

The linked [fiddle](http://jsfiddle.net/w67tzfhx/) showcases the approach.

almarklein:
Ok, this is converging again. One major difference since say Monday, is that there is now a `BufferWrapper` class that represents a buffer, plus (where it makes sense) a numpy array. This makes it much easier to refer to buffers from different places, and I guess paved the way for the next change.

The second change is that each material implements `get_wgpu_info(obj)`, which must return a list of pipeline descriptions. The descriptions are very simple/high level. Since `obj` (and its geometry) is accessible, its trivial to describe the list of vertex buffers and bind groups here. And since this info in now in the same file as the shaders, the story with the pre-defined bind groups is no longer necessary, except that we reserve bind group zero for global/scene info.

As an example, this is the LineMaterial:
```py
class LineStripMaterial(Material):
    ...

    def get_wgpu_info(self, obj):
        geometry = obj.geometry

        n = len(geometry.positions.data)  # number of vertices

        self.positions2.set_nbytes(2 * geometry.positions.nbytes)

        return [
            {
                "compute_shader": compute_shader,
                "indices": (n, 1, 1),
                "bindings1": [geometry.positions, self.positions2],
            },
            {
                "vertex_shader": vertex_shader,
                "fragment_shader": fragment_shader,
                "primitive_topology": wgpu.PrimitiveTopology.triangle_strip,
                "indices": n * 2,
                "vertex_buffers": [self.positions2],
                "target": None,  # default
            },
        ]
```

Korijn:
I've been thinking about this and looking up examples elsewhere, but I can't convince myself that this should be allowed in a material. It doesn't fit the concept of a Material.

visvis2/geometry/_dynmesh.py
almarklein: It looks like `needsUpdate` re-uploads the whole buffer, which would be *way* to slow for our purpose. But from the point of updating, things are very different anyway, because we can use mapped data.

However, we still need to perform all that state management, and you want to wrap that up in a class. So the question is whether that class should be a subclass of geometry (which would be my guess) or something else.

This is not that important right now, but I suppose it helps us to think about these things  in the same way :)

visvis2/cameras/_perspective.py
```diff
@@ -17,9 +17,8 @@ def __init__(self, fov, aspect, near, far):
     def __repr__(self) -> str:
         return f"PerspectiveCamera({self.fov}, {self.aspect}, {self.near}, {self.far})"
 
-    def update_matrix_world(self, *args, **kwargs):
-        super().update_matrix_world(*args, **kwargs)
-        self.matrix_world_inverse.get_inverse(self.matrix_world)
+    def update_viewport_size(self, width, height):
+        self.aspect = width / height
```
Korijn: how about `set_aspect(self, width, height)`

visvis2/geometry/_base.py
```diff
@@ -1,5 +1,25 @@
...
+    For the GPU, geometry objects are responsible for providing buffers.
+    """
+
+    def __init__(self, **data):
+        for name, val in data.items():
+            val = np.asanyarray(val)
+            usage = None
+            if name.lower() in ("index", "indices"):
+                usage = wgpu.BufferUsage.STORAGE | wgpu.BufferUsage.INDEX
+            setattr(name, BufferWrapper(val, usage=usage))
```
Korijn: I like the BufferWrapper abstraction, but it's tightly bound to WGPU. It no longer allows implementation of an OpenGLRenderer

visvis2/cameras/_base.py
```diff
@@ -3,12 +3,55 @@
...
+    """
+
     def __init__(self):
         super().__init__()
 
         self.matrix_world_inverse = Matrix4()
         self.projection_matrix = Matrix4()
         self.projection_matrix_inverse = Matrix4()
 
+    def update_viewport_size(self, width, height):
```
Korijn: this is just a setter I think, so instead of `update` it should say `set`

examples/line.py
```diff
@@ -0,0 +1,47 @@
...
+        [70, 30, 0, 1],
+    ],
+    np.float32,
+)
+geometry = vv.Geometry()
+geometry.positions = vv.BufferWrapper(positions * 5, mapped=True)
+
+
+material = vv.LineStripMaterial()
+line = vv.Mesh(geometry, material)  # Mesh??
```
Korijn: in response to the comment: we use Mesh in threejs as well. its just a different type of geometry and material

examples/simple_scene.py
```diff
@@ -10,21 +10,27 @@
 
 scene = vv.Scene()
 
-t1 = vv.Mesh(vv.Geometry(), vv.TriangleMaterial())
+
+class Triangle(vv.WorldObject):
+    def __init__(self, material):
+        super().__init__()
+        self.material = material
```
Korijn: I like that you moved it here. :) it's a weird little object, with hardcoded geometry

examples/simple_scene.py
```diff
@@ -10,21 +10,27 @@
...
-camera = vv.PerspectiveCamera(45, 16 / 9, 0.1, 1000)
+camera = vv.NDCCamera()  # This material does not use a camera
 
 
 def animate():
     width, height, ratio = canvas.get_size_and_pixel_ratio()
-    camera.aspect = width / height
-    camera.update_projection_matrix()
-    # Actually render the scene
+    camera.update_viewport_size(width, height)
```
Korijn: why doesnt this need `camera.update_projection_matrix()`?

visvis2/_wrappers.py
```diff
@@ -0,0 +1,202 @@
...
+def array_from_shader_type(spirv_type):
+    """ Get a numpy array object from a SpirV type from python-shader.
+    """
+    return np.asarray(spirv_type())
+
+
+# todo: can this be generic enough, keeping the GPU bits out / optional?
+
+
+class BaseBufferWrapper:
```
Korijn: do you think it's possible to move the WGPU specific stuff into the WgpuRenderer and make it possible for an OpenGL renderer to also support BufferWrapper?

visvis2/geometry/_dynmesh.py
Korijn: needsUpdate is something you should use when you actually _want_ the whole buffer to be reuploaded, for example in the event of resizing the buffer. otherwise you dont need it

visvis2/material/_base.py
```diff
@@ -1,6 +1,97 @@
...
+
+# Definition of struct with standard info,
+# provided to each world-object as uniform at slot 0.
+stdinfo_type = Struct(
+    world_transform=mat4,
+    cam_transform=mat4,
+    projection_transform=mat4,
+    physical_size=vec2,
+    logical_size=vec2,
+)
```
Korijn: are we still allowing for non-python-shader based materials?

visvis2/material/_base.py
```diff
@@ -1,6 +1,97 @@
...
-        self.uniforms = {}
-        self.primitive_topology = None
+
+    def get_wgpu_info(self, obj):
+        """ Return a list of high level pipeline descriptions.
+        These can be compute or render pipelines.
+
+        The default implementation returns a single render pipeline description,
+        using vertex_shader, fragment_shader, and primitive_topology from the material,
+        and index and vertex buffers from the geometry.
```
Korijn: It feels really wrong to be working with WorldObject and Geometry inside of a Material. Why not solve this in the renderer?

visvis2/material/_line.py
```diff
@@ -1,32 +1,73 @@
...
+        n = len(geometry.positions.data)  # number of vertices
+
+        self.positions2.set_nbytes(2 * geometry.positions.nbytes)
+
+        return [
+            {
+                "compute_shader": compute_shader,
+                "indices": (n, 1, 1),
+                "bindings1": [geometry.positions, self.positions2],
+            },
```
Korijn: I very much appreciate your work but I think I would rather have the instance based line drawing method as [implemented in THREE.JS](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/lines/LineMaterial.js) and [documented here quite well](https://wwwtyro.net/2019/11/18/instanced-lines.html) if it means that we don't need the separate compute shader pass anymore.

visvis2/geometry/_dynmesh.py
almarklein: > otherwise you dont need it

For uniforms this is the case. But AFAIK if you change the array data in a BufferGeometry you need to set that flag to make it sync to the GPU.

visvis2/cameras/_base.py
almarklein: I agree on using `set`. I figured that there may be camera's that need both width and height. E.g. a camera that keeps its aspect ratio can confine itself to whichever dimension is (relatively) the smallest.

visvis2/material/_triangle.py
```diff
@@ -21,16 +34,34 @@ def vertex_shader(
...
+        return [
+            {
+                "vertex_shader": vertex_shader,
+                "fragment_shader": fragment_shader,
+                "primitive_topology": wgpu.PrimitiveTopology.triangle_list,
+                "indices": range(n),
+                "bindings1": [self.uniforms],
+                "target": None,  # default
+            },
+        ]
```
Korijn: I guess this whole thing should be moved to the example, just like you moved the WorldObject subclass. it doesn't belong in the library

visvis2/geometry/_dynmesh.py
Korijn: try commenting the line in the fiddle; it still works

Korijn: I think its solved by setting [https://threejs.org/docs/#api/en/core/BufferAttribute.usage](BufferAttribute.usage) to DynamicDraw

examples/simple_scene.py
almarklein: The renderer calls `update_projection_matrix()`, I guess the other examples update it twice :)

almarklein: Sry, only the `scene_in_qt_subwidget.py` still did it in the example.

visvis2/material/_base.py
almarklein: Absolutely. This is just a convenient way to define the struct, and python-based shaders have the advantage that they can use this struct type in the signature.

visvis2/material/_triangle.py
almarklein: I agree.

visvis2/material/_line.py
almarklein: I'd want to have a look at the limitations of instanced rendering (maybe they're smaller than I think). This is certainly not the final approach :)

examples/simple_scene.py
Korijn: > The renderer calls `update_projection_matrix()`, I guess the other examples update it twice :)

I realize that it did do that at some point, but it doesn't anymore. The user is required to do it because they are the ones updating the camera, so that's better than calling it every frame...

Korijn:
These latest updates are great!

visvis2/_wrappers.py
```diff
@@ -0,0 +1,204 @@
...
+
+# todo: Support for updating unmapped data. Use something like updateRange to do subBufferUpdate
+
+
+class BaseBufferWrapper:
+    """ A base buffer wrapper that can be implemented for numpy, ctypes arrays,
+    or any other kind of array.
+    """
+
+    def __init__(self, data=None, nbytes=None, usage=None, mapped=False):
```
Korijn: I see you've split this up into a base and a subclass for numpy based buffers. But aren't you missing ie. `itemsize` if you arrays are 1-dimensional?

visvis2/_wrappers.py
```diff
@@ -0,0 +1,204 @@
...
+            self._usage = usage
+        elif isinstance(usage, str):
+            usages = usage.upper().replace(",", " ").replace("|", " ").split()
+            assert usages
+            self._usage = 0
+            for usage in usages:
+                self._usage |= getattr(wgpu.BufferUsage, usage)
+        else:
+            self._usage = wgpu.BufferUsage.STORAGE | wgpu.BufferUsage.VERTEX
+            # raise ValueError("BufferWrapper usage must be int or str.")
```
Korijn: I guess the usage attribute depends on the renderer being used. Can we think of a way to allow for that? Can the bufferwrapper class transparently pass on the usage flag that is being passed to it? In other words: if it doesn't process it at all and doesn't have a default, it doesn't need to know the type of the usage flag, I think?

visvis2/_wrappers.py
```diff
@@ -0,0 +1,204 @@
...
+        """ The buffer usage flags (as an int).
+        """
+        return self._usage
+
+    @property
+    def mapped(self):
+        """ Whether the data is mapped. Mapped data can be updated in-place
+        to change the data on the GPU.
+        """
+        return self._mapped
```
Korijn: I'm trying to assess the generality of this abstraction, and in particular I'm unsure if OpenGL also supports mapped memory. The [docs seem to indicate _yes_](https://www.khronos.org/opengl/wiki/Buffer_Object#Mapping), do you agree?

visvis2/_wrappers.py
```diff
@@ -0,0 +1,204 @@
...
+    def dirty(self):
+        """ Whether the buffer is dirty (needs to be processed by the renderer).
+        """
+        return self._dirty
+
+    @property
+    def strides(self):
+        """ Stride info (as a tuple).
+        """
+        return self._get_strides()
```
Korijn: I'm trying to see if this class can support interleaved buffers. Can it be done using strides and shape only? I think you would need more configuration options...

visvis2/_wrappers.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        return self._gpu_buffer
+
+    def set_mapped(self, mapped):
+        self._mapped = bool(mapped)
+        self._dirty = True
+
+    def set_nbytes(self, n):
+        self._nbytes = n
+        self._dirty = True
```
Korijn: What would be the point of setting nbytes without updating the data?

visvis2/_wrappers.py
almarklein: Yes, I think this makes a lot of sense.

visvis2/_wrappers.py
almarklein: Interesting thought. Will have to look into this.

visvis2/_wrappers.py
almarklein: It's a bit of a "yes, but ...". I wonder to what extend this "but" also applies to wgpu. I will make an issue to track this.

visvis2/_wrappers.py
almarklein: > Does this class support interleaved buffers? 

Not really at the moment, but I think it can be made so.

The wgpu-renderer needs `array_stride` for vertex buffers (which is `BufferWrapper.strides[0]`.

visvis2/_wrappers.py
almarklein: The only place where it's used is to set the size for the helper buffer of the Line render function. Other use-cases might need to do more here indeed.

visvis2/_wrappers.py
Korijn: we can use numpy arrays and views to handle this. forcing shape convention is fine since there aren't many valid use cases (we can think of right now) for other conventions

visvis2/_wrappers.py
Korijn: it can also be necessary to set nbytes when you dont have the data yet but do want to allocate a certain buffer size.

almarklein:
Improvements to buffers noted in #19

visvis2/material/_line.py
```diff
@@ -1,6 +1,32 @@
...
+
+    uniform_type = Struct(color=vec3, thickness=f32)
+
+    def __init__(self, color=(1, 1, 1), thickness=2.0):
         super().__init__()
+
+        array = array_from_shadertype(self.uniform_type)
+        self.uniforms = BufferWrapper(array, usage="uniform", mapped=True)
+        self.set_color(color)
+        self.set_thickness(thickness)
```
Korijn: :+1: material is just a type and a bag of parameters mapped to uniforms
:-1: the types come from python-shader

almarklein: It's an implementation detail that we use that type system to specify the structure of the uniforms ;)

almarklein: I feel your point though ... I'll make an issue.

almarklein: #21

PR #14, opened 2020-02-14
Render queue

Korijn:
* `render()` starts by extracting ordered list of world objects to render and ensuring their matrices are up to date
* Sorts objects by view space z-axis
* User can override sorting order with `WorldObject.render_order`
* New function also filters out invisible (`WorldObject.visible`) and unrenderable (no `material`) world objects
* Fixed bug in `Vector3.apply_matrix4` handling division by zero
* Modified examples accordingly

### DIFF
--- a/visvis2/renderers/_wgpu.py
+++ b/visvis2/renderers/_wgpu.py
@@ -3,7 +3,7 @@ import ctypes
 import wgpu.backend.rs
 
 from ._base import Renderer
-from ..objects import Mesh, WorldObject
+from ..objects import WorldObject
 from ..cameras import Camera
 from ..linalg import Matrix4, Vector3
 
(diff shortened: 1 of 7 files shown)


visvis2/renderers/_wgpu.py
```diff
@@ -33,18 +35,37 @@ def __init__(self, canvas):
...
+
+        # next, sort them from back-to-front
+        def sort_func(wobject: WorldObject):
+            z = (
+                Vector3()
+                .set_from_matrix_position(wobject.matrix_world)
+                .apply_matrix4(proj_screen_matrix)
+                .z
+            )
+            return wobject.render_order, z
```
Korijn: `z` may need a `-` but I can't really tell until we test more complex scenes. it should prevent some overdraw :)

PR #32, opened 2020-04-17
Progress on buffers and textures

almarklein:
In this PR I'll try to focus on getting buffers and textures working and cleaning things up.

* [x] Refactor `BufferWrapper` to not used mapped data.
* [x] Get things working again.
* [x] Implement `TextureWrapper`
* [x] Implement samplers.
* [x] Working textured cube example.
* [x] Remove "Buffer" prefix.
* [x] Cleaning up.
* [x] Make non uint8 textures work (incl float).

More, maybe in here, maybe later:
* [ ] An image example.
* [ ] An image slice through a 3D texture example.
* [ ] A skybox example (cube texture view).

Definitely later:
* [ ] How are we going to do collections and instancing. Might affect how we do uniforms.
* [ ] How to define uniforms.
* [ ] How to deal with updates / marking things dirty.


Korijn:
Couple of questions:

* Could we just call them "Buffer" and "Texture"? :) They're a core part of the user-facing API of visvis2, so I think the "Wrapper" part isn't important in the name.
* The baseclass was created because you think there will be alternatives that won't be based on numpy?
* Can the `_renderer` functions move to the renderer?
* Is e.g. vertex buffer data _moved_ or _copied_ to the GPU? As in, after uploading to the GPU, can I still iterate over my vertex data in CPU land?

almarklein:
> Could we just call them "Buffer" and "Texture"? :) 

Yes, I think so. I wanted to make clear that they are not the same as a `GPUBuffer`, but these have the `GPU` prefix, so I think we're fine.

> The baseclass was created because you think there will be alternatives that won't be based on numpy?

Probably not within visvis2 itself. But it makes it possible for 3d parties to subclass from there and build buffer classes based on other array implementations. 

> Can the _renderer functions move to the renderer?

I'll try ...

> Is e.g. vertex buffer data moved or copied to the GPU? As in, after uploading to the GPU, can I still iterate over my vertex data in CPU land?

Copied, so you can indeed still access the data from CPU land. I can imagine cases where you want to push data to the GPU without having a copy on the CPU, but these are rare (RAM >> GPU mem). Probably too rare to make this a feature of the common Buffer class. Maybe a special Buffer class... later.

Korijn:
It's looking _really_ good! Awesome stuff.

PR #36, opened 2020-05-21
Dev

almarklein:
* Some more examples (multiple cubes, skybox, volume slice)
* Some tweaks along the way.
* The API for the texture view needs some more love ... but probably in another PR.

### DIFF
--- a/examples/volume_slice1.py
+++ b/examples/volume_slice1.py
@@ -30,7 +30,7 @@ im = vol[index].copy()
 tex = vv.Texture(im, dim=2, usage="sampled")
 
 geometry = vv.PlaneGeometry(200, 200, 12, 12)
-material = vv.MeshBasicMaterial(map=tex.get_view(filter="nearest"), clim=(0, 255))
+material = vv.MeshBasicMaterial(map=tex.get_view(filter="linear"), clim=(0, 255))
 plane = vv.Mesh(geometry, material)
 plane.scale.y = -1
 scene.add(plane)
--- a/examples/volume_slice1.py
+++ b/examples/volume_slice1.py
@@ -1,0 +1,67 @@
+"""
+Render slices through a volume, by uploading to a 2D texture.
+Simple and ... slow.
+"""
+
+import imageio
+import visvis2 as vv
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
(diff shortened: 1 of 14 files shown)


Korijn:
Think you forgot to add that new material?

examples/volume_slice2.py
```diff
@@ -0,0 +1,70 @@
+"""
+Render slices through a volume, by creating a 3D texture, with 2D views.
+Simple and relatively fast, but no subslices.
```
Korijn, line 3: What are subslices?

examples/volume_slice1.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def scroll(degrees):
+    global index
+    index = index + int(degrees / 15)
+    index = max(0, min(nslices - 1, index))
+    im = vol[index]
+    tex.data[:] = im
+    tex.update_range((0, 0, 0), tex.size)
+    material.dirty = 1
```
Korijn, line 50: Why is this needed?

examples/volume_slice2.py
```diff
@@ -0,0 +1,70 @@
...
+canvas = WgpuCanvasWithScroll()
+renderer = vv.renderers.WgpuRenderer(canvas)
+scene = vv.Scene()
+
+vol = imageio.volread("imageio:stent.npz")
+nslices = vol.shape[0]
+index = nslices // 2
+
+tex_size = tuple(reversed(vol.shape))
+tex = vv.Texture(vol, dim=2, size=tex_size, usage="sampled")
```
Korijn, line 33: `dim=2`?

examples/volume_slice2.py
ivoflipse: Interpolation between two slices?

almarklein: yeah, that's what I meant :)

examples/volume_slice1.py
almarklein: because of how the renderer works right now. That process should be made somewhat smarter.

examples/volume_slice2.py
almarklein: Yes, if I understood this right, the way to do this is to create a 2D texture with size `wxhxn`, with `n > 1`

Korijn:
Awesome! Love the simplicity :)

PR #41, opened 2020-06-23
Lines

almarklein:
Ads a few line materials:
* LineMaterial: the default line material
* LineThinMaterial: a line using `line_strip` primitives that is always 1 (physical pixel) in width.
* LineSegmentMaterial: draws a line between each two points.
* LineArrowMaterial: bonus: like the above, but draws little arrows, which is nice for drawing vector fields.


Korijn:
Is there also a thin segment material? Seems like it would be useful to visualize mesh normals, also :)

almarklein:
> Is there also a thin segment material?

There is now :)

pygfx/renderers/wgpu/linerender.py
```diff
@@ -25,74 +54,382 @@ def compute_shader(
...
+        #     },
+        # },
         {
-            "vertex_shader": vertex_shader,
+            "vertex_shader": vert_shader,
             "fragment_shader": fragment_shader,
             "primitive_topology": wgpu.PrimitiveTopology.triangle_strip,
-            "indices": positions1.nitems * 2,
-            "vertex_buffers": [positions2],
+            "indices": (n, 1),
```
Korijn, line 426: So you are just setting up a render pass with `n` iterations without an actual vertex buffer being assigned? Clever!! Was this possible in opengl too?

almarklein: Yes, since OpenGL 4.6 apparently. But one could always simulate it with textures and a single vertexbuffer containing indices.

Korijn: Right, i see. Anyway, pretty cool! I can see this being used to implement mesh slicing too, perhaps. Is it possible to discard vertices? Probably not eh?

Korijn:
Man, this PR really expanded my understanding of shaders! Do you think this same principle can be applied to other primitives like rectangles, circles, etc?

almarklein:
Yes, most of these are likely easier than lines.

PR #43, opened 2020-06-25
Orthograpic camera and some stuff related to 2D

almarklein:


pygfx/linalg/matrix4.py
```diff
@@ -967,7 +967,7 @@ def make_orthographic(
         te[13] = -y
         te[2] = 0
         te[6] = 0
-        te[10] = -2 * p
+        te[10] = -1 * p
         te[14] = -z
         te[3] = 0
         te[7] = 0
```
almarklein, line 973: I'm making changes here to lines that probably came from somewhere else. I see that vispy uses the same. So either the original source is wrong, or this should be fixed in the OrthographicCamera class, by massaging `far`  and `near` and multiplying the resulting matrix with a translation in `z`.

almarklein: I think it makes sense to *not* treat z the same as x and y, because `x` and `y` in NDC range between 0 and 1, while the depth is expressed in values between 0.0 and 1.0.

almarklein:
> This change seems odd to me

What change do you mean exactly?

pygfx/linalg/matrix4.py
Korijn: It was directly ported from threejs, so I'm hesitant to think they have a bug..

almarklein: Ok, I'll do some more digging. Perhaps `make_orthographic` is supposed to produce such a matrix, and we must fix it a up for the `[0..1]` range in the camera class then.

Korijn:
> > This change seems odd to me
> 
> What change do you mean exactly?

see https://github.com/pygfx/pygfx/pull/43#discussion_r446171848

almarklein:
I now understand the issue with the range of `z`: In OpenGL, the NDC coordinates are in the range -1..1 for `x`, `y`, and also `z`. For WGPU, the NDC coordinates are different. In specific, `z` is in the range 0..1, and positive `y` is up.

We can choose our own NDC in pygfx and make any corrections in the renderers as needed. But I think it makes the most sense to go with WGPU's system. Let's keep linalg as-is, and correct in the camera classes.

PR #45, opened 2020-07-02
Update to new wgpu

almarklein:
* [x] Update the way we update buffers and texturs (new wgpu API).
* [x] Refactor Buffer and Texture classes to use memoryviews, no more BaseBuffer and a numpy-specific Buffer.
* [x] Update depth texture size by comparing with the render texture view.
* [x] Set camera viewport in the renderer, so it does not have to be done in each example.
* [x] Use full-res images in the examples (no more silly 1MB limit)
* [x] Clean up examples.
* [x]  Make skybox and volume_slice2 example work. (DONE, not committed yet, needs cleaning up)

### DIFF
--- a/examples/custom_object.py
+++ b/examples/custom_object.py
@@ -83,12 +83,10 @@ camera = gfx.NDCCamera()  # This material does not actually use the camera
 
 
 def animate():
-    physical_size = canvas.get_physical_size()
-    camera.set_viewport_size(*physical_size)
     renderer.render(scene, camera)
 
 
 if __name__ == "__main__":
-    canvas.draw_frame = animate
+    canvas.request_draw(animate)
     app.exec_()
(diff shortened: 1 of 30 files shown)


Korijn:
In the case of rendering directly to an image (without a GUI), we still need to be able to set the size of the resulting image and thus the camera viewport, somehow. I understand that it is something we don't support yet, but we should at least consider the implications for the API.

almarklein:
Right now, the renderer has a reference to the canvas. When we implement offscreen rendering, the renderer can just get a size, or some sort of OffscreenCanvas thingy, whatever. Regardless, I think that one way or another, the renderer will  be aware of the size.

Korijn:
Ok!

almarklein:
The examples now use standard images from imageio. Make sure to update imageio first :)

PR #46, opened 2020-07-07
More mesh materials

almarklein:
* [x] Calculate normals based on the vertices.
* [x] MeshNormalMaterial to draw the mesh in a color based on the normal.
* [x] MeshNormalLinesMaterial to draw normal lines.
* [x] Phong shading
* [x] Ability to apply shading *and* texturing
* [x] Fix that things fail if texcoords are not present -> use a dict instead of a list for vertex buffers internally.
* [x] Mesh slice
* [x] Add support for dot product in pyshader

See #50 for what we can do more (but no not in this PR)

### DIFF
--- a/pygfx/renderers/wgpu/meshrender.py
+++ b/pygfx/renderers/wgpu/meshrender.py
@@ -198,7 +198,7 @@ def vertex_shader_normal_lines(
 
     pos = buf_pos[i].xyz
     normal = vec3(buf_normal[i * 3 + 0], buf_normal[i * 3 + 1], buf_normal[i * 3 + 2])
-    pos = pos + f32(r) * normal * 10.0  # todo: allow user to specify normal length
+    pos = pos + f32(r) * normal * 0.1  # todo: allow user to specify normal length
 
     world_pos = u_stdinfo.world_transform * vec4(pos, 1.0)
     ndc_pos = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos
--- a/pygfx/renderers/wgpu/meshrender.py
+++ b/pygfx/renderers/wgpu/meshrender.py
@@ -44,6 +44,8 @@ def mesh_renderer(wobject, render_info):
     if getattr(geometry, "texcoords", None) is not None:
         vertex_buffers.append(geometry.texcoords)
 
+    # todo: fail if no texcoords are given, make vertex_buffers a dict!
+
     # Normals
     if getattr(geometry, "normals", None) is not None:
         normal_buffer = geometry.normals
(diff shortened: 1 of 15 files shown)


pygfx/renderers/wgpu/meshrender.py
```diff
@@ -113,22 +87,146 @@ def mesh_renderer(wobject, render_info):
...
+    nn = np.zeros((npts, 3))
+    for verts in tris.T:  # note this only loops 3x (number of verts per tri)
+        for idx in range(3):  # x, y, z
+            nn[:, idx] += np.bincount(
+                verts.astype(np.int32), tri_nn[:, idx], minlength=npts
+            )
+    size = np.sqrt(np.sum(nn * nn, axis=1))
+    size[size == 0] = 1.0  # prevent ugly divide-by-zero
+    nn /= size[:, np.newaxis]
+    return nn.astype(np.float32)
```
Korijn, line 139: I'm glad you got this cool bit of code, but I think it should be part of a utils module and the user should supply normals as part of the geometry

almarklein: good point.

pygfx/renderers/wgpu/meshrender.py
```diff
@@ -113,22 +94,190 @@ def mesh_renderer(wobject, render_info):
...
+    #
+    # for p, verts in enumerate(tris):
+    #     nn[verts, :] += tri_nn[p, :]
+    #
+    nn = np.zeros((npts, 3))
+    for verts in tris.T:  # note this only loops 3x (number of verts per tri)
+        for idx in range(3):  # x, y, z
+            nn[:, idx] += np.bincount(
+                verts.astype(np.int32), tri_nn[:, idx], minlength=npts
+            )
```
Korijn, line 142: as an alternative, you could also write it like this if you keep a reference to the un-normalized `tri_nn`:

```
for i in range(3):
    np.add.at(nn, tris[:, i], tri_nn)
```

not sure if it's faster/equivalent/slower, but it's certainly a lot easier to read

pygfx/renderers/wgpu/meshrender.py
almarklein: Moved to utils. But at this momemtn the meshrenderer still uses it if the normals are not given. We could remove that feature later if we think that'd be better.

pygfx/renderers/wgpu/meshrender.py
```diff
@@ -36,23 +36,24 @@ def mesh_renderer(wobject, render_info):
...
+        # meh, we need to specify a buffer if the shader uses it
+        vertex_buffers[1] = geometry.positions
 
-    # Normals
+    # Normals. Usually it'd be given. If not, we'll calculate it from the vertices.
     if getattr(geometry, "normals", None) is not None:
         normal_buffer = geometry.normals
     else:
-        normal_data = _calculate_normals(geometry.positions.data, index_buffer.data)
+        normal_data = normals_from_vertices(geometry.positions.data, index_buffer.data)
```
Korijn, line 54: Can we add an option to toggle if this should be done automatically or if an error should be raised? Depending on the application it might be preferable to have this kind of "validation" instead of auto correcting behaviour :) otherwise this is great!

PR #53, opened 2020-07-14
Orbit and PanZoom camera controls

Korijn:
Ready to merge!

Changes:

* Orbit and panzoom camera controls
  * Right mouse button drag to pan
  * Left mouse button drag to rotate
  * Mouse wheel to zoom
  * Tuneable speeds for panning, zooming and rotating
  * Choose between zooming based on camera distance or zoom
  * Came up with an API that allows the user to remain in control over state flow and GUI glue
* Had to revise the orthographic camera control flow a little bit; previously, it would autonomously (i.e. without user input) zoom if you set ` zoom != 1`
* WorldObject.look_at utility
* Implement more functions on Spherical linalg class
* While debugging the above, I also:
  * Added an AxesHelper to debug coordinate systems
    * Since linesegments don't support vertex coloring yet, I had to create three separate objects to be able to set different colors on each axis
  * Fixed a bug in perspective and orthographic camera (top/bottom were flipped)
* Orbit camera example:

orbit

* Panzoom (zoom-to-mouse!) camera example:

panzoom

* Multi-slice example:

multislice

### DIFF
--- a/examples/multi_slice.py
+++ b/examples/multi_slice.py
@@ -29,7 +29,7 @@ class WgpuCanvasWithInputEvents(WgpuCanvas):
             app.setOverrideCursor(QtCore.Qt.BlankCursor)
 
     def mouseReleaseEvent(self, event):  # noqa: N802
-        if self.drag.get("button") == event.button():
+        if self.drag and self.drag.get("button") == event.button():
             self.drag = None
             app.restoreOverrideCursor()
 
@@ -86,8 +86,8 @@ for axis in [0, 1, 2]:
         plane.rotation.set_from_euler(gfx.linalg.Euler(0.5 * np.pi))
     scene.add(plane)
 
-camera = gfx.OrthographicCamera(200, 200)
-camera.position.set(50, 50, 50)
+camera = gfx.PerspectiveCamera(70, 16 / 9)
+camera.position.set(125, 125, 125)
 camera.look_at(gfx.linalg.Vector3())
 controls = gfx.OrbitControls(camera.position.clone(), up=gfx.linalg.Vector3(0, 0, 1))
 
--- a/examples/multi_slice.py
+++ b/examples/multi_slice.py
@@ -21,7 +21,7 @@ class WgpuCanvasWithInputEvents(WgpuCanvas):
         self.drag = None
 
     def wheelEvent(self, event):  # noqa: N802
-        controls.zoom(event.angleDelta().y() * 0.125)
+        controls.zoom(2 ** (event.angleDelta().y() * 0.0015))
 
     def mousePressEvent(self, event):  # noqa: N802
         button = event.button()
@@ -91,10 +91,13 @@ for axis in [0, 1, 2]:
         plane.rotation.set_from_euler(gfx.linalg.Euler(0.5 * np.pi))
     # else: XY plane
 
-camera = gfx.PerspectiveCamera(70, 16 / 9)
+# camera = gfx.PerspectiveCamera(70, 16 / 9)
+camera = gfx.OrthographicCamera(200, 200)
 camera.position.set(125, 125, 125)
 camera.look_at(gfx.linalg.Vector3())
-controls = gfx.OrbitControls(camera.position.clone(), up=gfx.linalg.Vector3(0, 0, 1))
+controls = gfx.OrbitControls(
+    camera.position.clone(), up=gfx.linalg.Vector3(0, 0, 1), zoom_changes_distance=False
+)
 
 
 def animate():
--- a/examples/multi_slice.py
+++ b/examples/multi_slice.py
@@ -33,27 +33,26 @@ for axis in [0, 1, 2]:
     # we could reuse the same plane geometry for all slices here
     geometry = gfx.PlaneGeometry(200, 200, 1, 1)
     if axis == 0:  # YZ plane
-        texcoords = np.array([
-            [0.5, 0., 0],
-            [0.5, 1., 0],
-            [0.5, 0., nslices],
-            [0.5, 1., nslices],
-        ], dtype=np.float32)
+        texcoords = np.array(
+            [[0.5, 0.0, 0], [0.5, 1.0, 0], [0.5, 0.0, nslices], [0.5, 1.0, nslices],],
+            dtype=np.float32,
+        )
     elif axis == 1:  # XZ plane
-        texcoords = np.array([
-            [0., 0.5, 0],
-            [1., 0.5, 0],
-            [0., 0.5, nslices],
-            [1., 0.5, nslices],
-        ], dtype=np.float32)
+        texcoords = np.array(
+            [[0.0, 0.5, 0], [1.0, 0.5, 0], [0.0, 0.5, nslices], [1.0, 0.5, nslices],],
+            dtype=np.float32,
+        )
     elif axis == 2:  # XY plane (default)
         z_idx = nslices / 2
-        texcoords = np.array([
-            [0., 0., z_idx],
-            [1., 0., z_idx],
-            [0., 1., z_idx],
-            [1., 1., z_idx],
-        ], dtype=np.float32)
+        texcoords = np.array(
+            [
+                [0.0, 0.0, z_idx],
+                [1.0, 0.0, z_idx],
+                [0.0, 1.0, z_idx],
+                [1.0, 1.0, z_idx],
+            ],
+            dtype=np.float32,
+        )
     print(texcoords)
     geometry.texcoords = gfx.Buffer(texcoords, usage="vertex|storage")
 
@@ -66,6 +65,8 @@ for axis in [0, 1, 2]:
     elif axis == 1:  # XZ plane
         plane.rotation.set_from_axis_angle(gfx.linalg.Vector3(1, 0, 0), 0.5 * np.pi)
     scene.add(plane)
+
+    # TODO: also add a mesh slice for each plane
 
 camera = gfx.OrthographicCamera(200, 200)
 camera.position.set(50, 50, 50)
--- a/examples/multi_slice.py
+++ b/examples/multi_slice.py
@@ -1,0 +1,89 @@
+"""
+Render slices through a volume, by creating a 3D texture, and sample it in the shader.
+Simple, fast and subpixel!
+"""
+
+import imageio
+import numpy as np
+import pygfx as gfx
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
+app = QtWidgets.QApplication([])
+
+canvas = WgpuCanvas()
+renderer = gfx.renderers.WgpuRenderer(canvas)
+scene = gfx.Scene()
+
+background = gfx.Background(gfx.BackgroundMaterial((0, 1, 0, 1), (0, 1, 1, 1)))
+scene.add(background)
+
+scene.add(gfx.AxesHelper(size=10))
+
+vol = imageio.volread("imageio:stent.npz")
+tex = gfx.Texture(vol, dim=3, usage="sampled")
+view = tex.get_view(filter="linear")
+
+for axis in [0, 1, 2]:
+    nslices = vol.shape[0]
+
+    # TODO: if we could set the slicing plane on the material parametrically
+    # we could reuse the same plane geometry for all slices here
+    geometry = gfx.PlaneGeometry(200, 200, 1, 1)
+    if axis == 0:  # YZ plane
+        texcoords = np.array([
+            [0.5, 0., 0],
+            [0.5, 1., 0],
+            [0.5, 0., nslices],
+            [0.5, 1., nslices],
+        ], dtype=np.float32)
+    elif axis == 1:  # XZ plane
+        texcoords = np.array([
+            [0., 0.5, 0],
+            [1., 0.5, 0],
+            [0., 0.5, nslices],
+            [1., 0.5, nslices],
+        ], dtype=np.float32)
+    elif axis == 2:  # XY plane (default)
+        z_idx = nslices / 2
+        texcoords = np.array([
+            [0., 0., z_idx],
+            [1., 0., z_idx],
+            [0., 1., z_idx],
+            [1., 1., z_idx],
+        ], dtype=np.float32)
+    print(texcoords)
+    geometry.texcoords = gfx.Buffer(texcoords, usage="vertex|storage")
+
+    material = gfx.MeshVolumeSliceMaterial(map=view, clim=(0, 255))
+    plane = gfx.Mesh(geometry, material)
+
+    # by default the plane is in XY plane
+    if axis == 0:  # YZ plane
+        plane.rotation.set_from_axis_angle(gfx.linalg.Vector3(0, 1, 0), 0.5 * np.pi)
+    elif axis == 1:  # XZ plane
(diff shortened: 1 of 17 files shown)


pygfx/linalg/spherical.py
```diff
@@ -1,7 +1,36 @@
...
+            elif y > 1:
+                y = 1
+            self.phi = math.acos(y)
+        return self
+
+    def make_safe(self) -> "Spherical":
+        # restrict phi to (epsilon, pi-epsilon)
+        self.phi = max(self._eps, min(math.pi - self._eps, self.phi))
+        # restrict theta to (0, 2*pi)
+        self.theta = self.theta % self._2pi
```
Korijn, line 36: Missing `return self`

pygfx/objects/_base.py
```diff
@@ -77,3 +79,13 @@ def update_matrix_world(
...
+
+    def look_at(self, target: Vector3):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        v = Vector3().set_from_matrix_position(self.matrix_world)
+        m = Matrix4().look_at(v, target, self.up)
+        self.rotation.set_from_rotation_matrix(m)
+        if self.parent:
+            m.extract_rotation(self.parent.matrix_world)
+            q = Quaternion().set_from_rotation_matrix(m)
+            self.rotation.premultiply(q.inverse())
```
Korijn, line 91: Could reduce gc pressure with some temp vars preallocated

examples/orbit_camera.py
```diff
@@ -0,0 +1,106 @@
...
+from PyQt5 import QtWidgets, QtCore
+from wgpu.gui.qt import WgpuCanvas
+
+
+class WgpuCanvasWithInputEvents(WgpuCanvas):
+    _flip_y = np.array([1, -1])
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.drag = {}
```
Korijn, line 18: What's the proper abstraction to decouple this?

examples/multi_slice.py
```diff
@@ -0,0 +1,83 @@
...
+                [1.0, 0.0, z_idx],
+                [0.0, 1.0, z_idx],
+                [1.0, 1.0, z_idx],
+            ],
+            dtype=np.float32,
+        )
+    print(texcoords)
+    geometry.texcoords = gfx.Buffer(texcoords, usage="vertex|storage")
+
+    material = gfx.MeshVolumeSliceMaterial(map=view, clim=(0, 255))
```
Korijn, line 61: Don't need to recreate this for every plane!

examples/panzoom_camera.py
```diff
@@ -0,0 +1,101 @@
...
+        # apply zoom and compute change in dimensions of camera frustum
+        controls.zoom(event.angleDelta().y())
+        controls.update_camera(camera)
+        camera.update_bounds()
+        view = (camera.right - camera.left, camera.top - camera.bottom)
+        # compute new mouse position relative to widget center
+        relpos = tuple(fracpos[i] * view[i] for i in range(2))
+        # compute delta and pan accordingly
+        delta = tuple(relpos[i] - relpos_old[i] for i in range(2))
+        controls.pan(*delta)
```
Korijn, line 37: would be nice to embed this logic in a controls class. should probably do that.

pygfx/cameras/_orthographic.py
```diff
@@ -33,29 +31,35 @@ def __repr__(self) -> str:
...
+        self.height = float(height)
+        self._view_aspect = self.width / self.height
+        self.update_bounds()
 
-    def update_projection_matrix(self):
+    def update_bounds(self):
         # Get the reference width / height
+        # TODO: when can this lead to a change? unclear to me
         width = self.width / self.zoom
         height = self.height / self.zoom
```
Korijn, line 43: if `self.zoom` is a single number, how could it ever result in a new aspect ratio?

pygfx/objects/_base.py
```diff
@@ -77,3 +83,13 @@ def update_matrix_world(
         if update_children:
             for child in self._children:
                 child.update_matrix_world()
+
+    def look_at(self, target: Vector3):
```
almarklein, line 87: Oh sweet, so we can orient any object this way 😮

pygfx/objects/_base.py
```diff
@@ -12,6 +12,10 @@ class WorldObject:
     into a single empty world object.
     """
 
+    _v = Vector3()
+    _m = Matrix4()
+    _q = Quaternion()
```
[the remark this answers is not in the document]
Korijn: Yes

pygfx/controls/_orbit.py
```diff
@@ -0,0 +1,97 @@
+from ..linalg import Vector3, Matrix4, Quaternion, Spherical
+
+
+class OrbitControls:
+    _m = Matrix4()
+    _v = Vector3()
+    _origin = Vector3()
+    _orbit_up = Vector3(0, 1, 0)
+    _s = Spherical()
```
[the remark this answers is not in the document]
Korijn: Well. I don't know about that. Is it necessary to be thread safe?

pygfx/controls/_orbit.py
```diff
@@ -0,0 +1,97 @@
...
+    def zoom(self, delta: float) -> "OrbitControls":
+        if self.zoom_type == "distance":
+            self.distance -= delta * self.zoom_speed
+            if self.distance < 0:
+                self.distance = 0
+        elif self.zoom_type == "zoom":
+            delta = delta * self.zoom_speed * 0.01
+            if self.zoom_ < 1.0:
+                delta *= self.zoom_
+            self.zoom_ += delta
```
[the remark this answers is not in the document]
Korijn: Zooming out happens with zoom < 1.0, and if you keep incrementing by the same number of every tick of the scroll wheel, you'll make bigger and bigger jumps zooming out. It's very uncomfortable. Multiplying by `self.zoom_` makes the steps smaller and smaller to compensate.

pygfx/controls/_orbit.py
```diff
@@ -0,0 +1,97 @@
...
+        return self
+
+    def pan(self, x: float, y: float) -> "OrbitControls":
+        self._v.set(x, -y, 0).multiply_scalar(self.pan_speed).apply_quaternion(
+            self.rotation
+        )
+        self.target.sub(self._v)
+        return self
+
+    def rotate(self, x: float, y: float) -> "OrbitControls":
```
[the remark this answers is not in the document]
Korijn: no, they mean phi and theta :) haha. it depends on what convention you are used to

pygfx/controls/_orbit.py
```diff
@@ -0,0 +1,97 @@
...
+        if eye is None:
+            eye = Vector3(50.0, 50.0, 50.0)
+        if target is None:
+            target = Vector3()
+        if up is None:
+            up = Vector3(0.0, 1.0, 0.0)
+        self.look_at(eye, target, up)
+        self.rotate_speed = 0.02
+        self.zoom_speed = 0.125
+        self.pan_speed = 1.0
```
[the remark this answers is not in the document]
Korijn: alright, will remove them

examples/orbit_camera.py
```diff
@@ -0,0 +1,94 @@
...
+    def mousePressEvent(self, event):  # noqa: N802
+        button = event.button()
+        mode = self._drag_modes.get(button, None)
+        if self.drag or not mode:
+            return  # drag is already initiated, or unknown button pressed
+        self.drag = {
+            "mode": mode,
+            "button": button,
+            "start": (event.x(), event.y()),
+        }
```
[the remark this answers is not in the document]
Korijn: black made me do it :)

pygfx/objects/_base.py
Korijn: yes! it also properly takes into account any parent transforms

examples/panzoom_camera.py
```diff
@@ -0,0 +1,101 @@
...
+tex = gfx.Texture(im, dim=2, usage="sampled")
+geometry = gfx.PlaneGeometry(512, 512)
+material = gfx.MeshBasicMaterial(map=tex.get_view(filter="linear"), clim=(0, 255))
+plane = gfx.Mesh(geometry, material)
+plane.scale.y = -1
+scene.add(plane)
+
+camera = gfx.OrthographicCamera(512, 512)
+camera.position.set(0, 0, 500)
+controls = gfx.OrbitControls(camera.position.clone(), zoom="zoom")
```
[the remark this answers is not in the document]
Korijn: yeah. I'll address a number of these comments simultaneously by introducing that

pygfx/cameras/_orthographic.py
Korijn: yes, but I'd say that we don't need to recompute the aspect ratio then, making the function a lot simpler

examples/multi_slice.py
```diff
@@ -0,0 +1,83 @@
...
+material = gfx.MeshVolumeSliceMaterial(map=view, clim=(0, 255))
+
+for axis in [0, 1, 2]:
+    nslices = vol.shape[0]
+
+    # TODO: if we could set the slicing plane on the material parametrically
+    # we could reuse the same plane geometry for all slices here
+    # TODO: there seems to be a problem with the texcoords... only one plane looks correct
+    # TODO: why is the third axis not in [0..1] range?
+    # TODO: also add a mesh slice for each plane
```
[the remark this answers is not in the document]
almarklein: Done. I don't think the material should have a slicing pane. I like how it simply uses the texcoords ....

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
almarklein: Shouldn't the zoom be multiplicative instead of additive all the time?

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
almarklein: I can live with `phi` and `theta`, but `x` and `y` seem a bit off ;)

examples/multi_slice.py
```diff
@@ -6,74 +6,95 @@
...
     scene.add(plane)
 
 camera = gfx.OrthographicCamera(200, 200)
 camera.position.set(50, 50, 50)
 camera.look_at(gfx.linalg.Vector3())
+controls = gfx.OrbitControls(camera.position.clone(), up=gfx.linalg.Vector3(0, 0, 1))
 
 
 def animate():
+    # todo: should the control set it directly?
```
Korijn, line 96: no, then the user would lose control over the control flow

pygfx/cameras/_orthographic.py
almarklein: I don't follow, you mean we don't need to call this method if only the zoom changes?

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
almarklein: Eventually someone will want to us threads ...

examples/multi_slice.py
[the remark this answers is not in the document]
Korijn: Thanks for fixing this: I also updated the PR description

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
Korijn: can you explain a little further what you mean?

pygfx/cameras/_orthographic.py
Korijn: we have this:

    def set_viewport_size(self, width, height):
        self.width = float(width)
        self.height = float(height)
        self._view_aspect = self.width / self.height

and this:

        width = self.width / self.zoom
        height = self.height / self.zoom
        # Increase either the width or height, depending on the view size
        aspect = width / height

and I don't see how that second `aspect` could ever be different from `self._view_aspect`

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
Korijn: Well, I don't expect users to instantiate many controls objects, so I think making them instance local is fine here.

However we also have this approach in the linalg package. So there's that..

examples/multi_slice.py
Korijn: I kind of had to violate my own policy in the zoom_to_point code.. don't like it. :/ I'll see if I can implement it without having to mutate the camera

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
Korijn: I've created #54 to track this

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
almarklein: I mean that perhaps we should always do `self.zoom *= delta`. It does not feel right that the behavior of the zoom depends on the zoom level.

examples/panzoom_camera.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.drag = None
+
+    def wheelEvent(self, event):  # noqa: N802
+        dim = self.width(), self.height()
+        pos = event.x(), event.y()
+        delta = event.angleDelta().y() * 0.00125
+        view = camera.right - camera.left, camera.top - camera.bottom
+        controls.zoom_to_point(delta, pos, dim, view)
```
almarklein, line 25: If we provide the control a ref to the camera (instead of only its initial position), then we e.g. would not have to provide the view here ...

pygfx/controls/_orbit.py
```diff
@@ -0,0 +1,75 @@
...
+        self.rotation.set_from_rotation_matrix(
+            self._m.look_at(self._v, self._origin, self.up)
+        )
+        return self
+
+    def zoom(self, delta: float) -> "OrbitControls":
+        self.distance -= delta
+        if self.distance < 0:
+            self.distance = 0
+        return self
```
almarklein, line 63: This does not work for orho camera's ;)

Korijn: it was not intended to be used for ortho cams :p

pygfx/cameras/_orthographic.py
```diff
@@ -33,29 +31,23 @@ def __repr__(self) -> str:
...
+        self.width = float(width)
+        self.height = float(height)
         width = self.width / self.zoom
         height = self.height / self.zoom
-        # Increase eihter the width or height, depending on the view size
-        aspect = width / height
-        if aspect < self._view_aspect:
-            width *= self._view_aspect / aspect
-        else:
-            height *= aspect / self._view_aspect
```
almarklein, line 47: Why is this removed? The ortho camera is broken atm, because the content does not scale as the window is resized. Not sure if the missing of these lines are the only problem.

pygfx/controls/_orbit.py
almarklein: But ortho cameras are fine for 3D too, so IMO the orbit controls should work with them.

pygfx/cameras/_orthographic.py
Korijn: Haha. Well. Remember this discussion? https://github.com/pygfx/pygfx/pull/53#pullrequestreview-457309496

pygfx/controls/_orbit.py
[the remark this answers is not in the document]
almarklein: I changed this: the parameter for zoom functions is always a multiplicative value (i.e. a value of 1 does nothing). This makes the zooming feel consistent at every scale.

PR #55, opened 2020-08-02
Fix panning for camera controls

almarklein:
* the control classes get more convenience api that is targeted at mouse stuff, but gui independent.
* the panning logic makes sure that the object under the mouse stays under the mouse.
* the gui proxy classes in the examples became a bit smaller.

### DIFF
--- a/pygfx/controls/_orbit.py
+++ b/pygfx/controls/_orbit.py
@@ -1,12 +1,35 @@
-from ..linalg import Vector3, Matrix4, Quaternion, Spherical
+from typing import Tuple
+
+from ..linalg import Vector3, Vector4, Matrix4, Quaternion, Spherical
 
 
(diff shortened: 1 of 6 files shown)


pygfx/controls/_orbit.py
```diff
@@ -1,12 +1,35 @@
...
+    pos2 = Vector3(pos2.x / pos2.w, pos2.y / pos2.w, pos2.z / pos2.w)
+    pos1.multiply_scalar(0.02 / canvas_size[0])
+    pos2.multiply_scalar(0.02 / canvas_size[1])
+    return pos1, pos2  # now they're vecs, really
+
+
 class OrbitControls:
     _m = Matrix4()
-    _v = Vector3()
+    _v = Vector3()  # todo: euhm, this vector is shared between all instances? :P
```
almarklein, line 32: eek?

PR #57, opened 2020-08-05
Refactoring the wgpu renderer internals and how it deals with updates

almarklein:
* [x] Adds an example that draws 1000 small plots. This puts a strain on the amount of work that Python must do each draw.
* [x] Put the `stdinfo` uniform into a singleton uniform that is reused by all objects.
* [x] The `world_transform` that was previously in `stdinfo` is now in a per-wobect uniform. This is what makes the 1000-plot example fast.
* [x] Investigate the need for collections. We don't seem to need them!
* [x] All shaders now uses binding set 0 for uniforms, and 1 for textures and storage buffers (just for consistency).
* [x] The ortho camera can be set to not maintain aspect ratio.
* [x] Add (basic) support for instancing, including an example.
* [ ] Would be nice to let those meshes in the instanced mesh example rotate ...
* [ ] Only update world-uniform when needed.
* [ ] Smarter updating overall.

### DIFF
--- a/examples/collection_line.py
+++ b/examples/collection_line.py
@@ -1,0 +1,88 @@
+"""
+Display multiple lines in a collection for performance.
+"""
+
+import time  # noqa
+import numpy as np
+
+import pygfx as gfx
+
+from PyQt5 import QtWidgets, QtCore
+from wgpu.gui.qt import WgpuCanvas
+
+app = QtWidgets.QApplication([])
+
+
+class WgpuCanvasWithInputEvents(WgpuCanvas):
+    _drag_modes = {QtCore.Qt.RightButton: "pan"}
+    _mode = None
+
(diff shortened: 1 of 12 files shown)


Korijn:
I'm not entirely sure about the API for instancing here (as in, I don't fully understand how it works), but, let's merge this and keep the progress going!

almarklein:
> I'm not entirely sure about the API for instancing here (

I replicated the API from ThreeJS. An `InstancedMesh` is like a normal mesh, but it has a count and a `set_matrix_at` method. You can imagine it has an object that has `n` sub-objects that are all replica's, but positioned individually.

ThreeJS also has `InstancedBufferAttribute`, so you can also give each instance e.g. a different color. We could do something along those lines as well.

PR #59, opened 2020-08-19
Implement proper update mechanism

almarklein:
This continues where #57 ended.

Updating is more automated, and is done on a more fine-grained level. Buffers and textures can be updated without re-creating the pipeline (and vice versa).

* [x] Implement a mechanic so that each render function can invalidate the pipeline when a certain attribute of wobject/geometry/material is set. I think it makes sense to let the render function decide, because that's the part that knows ...
* [x] Agree on the approach/implementation of the above.
* [x] Buffer and texture resources have a `versionflag` instead of a `dirty` flag, to help with detecting updates.
* [x] The renderer works with all of the above!
* [x] Fixes that the uniform buffer holding the wobject matrix was never updated. 
* [x] Now make sure to only upload it when needed (because it slows things down).
* [x] Prevent CPU load to kick in for simple viz.
  * [x] Examples with interaction request a draw when needed.
  * [x] Animated examples will be frame limited via https://github.com/pygfx/wgpu-py/pull/125


pygfx/objects/_base.py
```diff
@@ -1,7 +1,31 @@
...
+        else:
+            if not callbacks:
+                self._set_attr_callbacks.pop(name)  # Make it quicker next time
+            for cb in callbacks:
+                cb(self, name)
+
+    def _listen(self, name, fn):
+        # Private, because only intended for private use
+        callbacks = self._set_attr_callbacks.setdefault(name, weakref.WeakSet())
+        callbacks.add(fn)
```
almarklein, line 25: @Korijn What do you think of this? It's a base class for WorldObject, Material and Geometry. It allows the render functions to be notified when a specific attribute is set.

In practice, render functions currently listen for e.g. `wobject.geometry`, `wobject.material`, `geometry.posisions` etc, `material.map`.

Any other ideas/approaches?

Korijn: At first glance this looks good. I dislike that it's implemented through a base class (TrackableObject) but I understand how it's convenient wrt cleanup/garbage collection since the listener lifetimes are tied to the lifetime of the object. :) I made the same design decision in observ for the same reasons.

Give me some time to read up on solutions in other libraries so I can compare them to your approach.

Korijn:
> I'd love a way to only draw when needed. It's silly that having an idle figure open causes my fans to kick in...

I'd say that is already supported? Example: https://github.com/pygfx/pygfx/blob/master/examples/volume_slice2.py

I guess you could fully automate it, if you were able to fully observe any and all changes to the scene graph and were able to detect if any of those changes should trigger a redraw. But that's very hard. It's why I wrote observ, too. I think it should be solved in the GUI/application logic, just like we do in the example I linked above.

Or do you have some clever idea that would simplify this problem? :)

almarklein:
I think you're right. Any change originates from somewhere, and that's where a new draw should be requested. There are some examples that request a new draw from the animate function. It would be nice to be able to limit that to a certain fps Similar to JS requestAnimationFrame. 

If we end up in a state where we fully observe when the renderer needs to make any updates to wgpu objects, then we might be able to detect when a new draw is needed automatically. But we'll see if that happens.

An example with observ would be nice!

pygfx/objects/_base.py
Korijn: I think the approach here suffices. We'll see if we run into any limits in the future.

almarklein: Looking at what the renderers subscribe to, it seems to be safe to say that (apart from the geometry or material being reset) any attribute being set that is a resource (buffer of texture), requires a pipeline reset. Implementing this is easier and less intrusive - no callback stuff.

almarklein: The only case I can think of is e.g. a material with a property that causes the thing to be rendered in a different style, e.g. with a different shader. But one could argue that you'd need to implement a different Material subclass instead of using a property.

almarklein:
Quite happy with the current state. WorldObject's, Materials, and geometry have a `versionflag` that is bumped when the "structure" changes, i.e. when the pipeline needs to be rebuild. The Textures and buffer have a `versionflag` that is bumped when the data is updates, i.e. when new data must be uploaded.

Let's rename `versionflag` though, as it's not pep8. We could do `version_flag`, or `data_version`, `update_count`, `update_nr`. We could also use a different names for wobject/material/geometry and texture/buffer. Thoughts @Korijn ?

Korijn:
How about `revision` or shorter `rev`?

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -212,78 +207,108 @@ def _record_render(self, command_encoder, q):
...
         scene.traverse(visit)
-
-        # next, sort them from back-to-front
-        def sort_func(wobject: WorldObject):
-            z = (
-                Vector3()
-                .set_from_matrix_position(wobject.matrix_world)
-                .apply_matrix4(proj_screen_matrix)
-                .z
-            )
```
Korijn, line 236: This sorting was here to try and minimize overdraw. Why did you decide to remove it?

examples/multi_slice.py
```diff
@@ -106,7 +106,6 @@ def animate():
     plane.position.z = t * vol.shape[0] * 0.5
     plane.geometry.texcoords.data[:, 2] = (t + 1) / 2
     plane.geometry.texcoords.update_range(0, plane.geometry.texcoords.nitems)
-    material.dirty = 1  # todo: we should not have to mark the *material* dirty!
 
     controls.update_camera(camera)
     renderer.render(scene, camera)
```
Korijn, line 111: Just wondering if this example works correctly now?

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: Because I think we should make it so that the order of rendering does not matter (order independent transparency). Blending happens per-fragment, not per-object (e.g. does not help for objects that intersect).

But we don't have that yet, so guess I'll turn it back on.

examples/multi_slice.py
almarklein: It does. When the code calls `plane.geometry.texcoords.update_range(...)`, the material sets its version flag, causing the renderer to upload.

pygfx/renderers/wgpu/_wgpurenderer.py
Korijn: * We don't have order independent transparency yet right?
* "Blending happens per-fragment" is kind of the point; if you render objects closest to the camera first, it's more likely that a portion of fragments can be skipped entirely thanks to the depth buffer being filled with closer objects sooner rather than later

Or am I missing the point here?

almarklein: My point is that ordering objects is not a guarantee that the rendering occurs correctly. The z-position of an object is e.g. its center of mass, or the center of its bounding box, but its vertices can be anywhere. So even though object A can be behind object B according to the sorting, there can still be many fragments of B that are in front of A.

Your point of performance is valid; you're doing sorting on the CPU to save work for the GPU. But it will depend heavily on the situation whether it will help or only load the CPU more (especially in Python ;))

almarklein: Something like a `render_order` flag or integer `level` (ThreeJS has something like that, right?) would therefore be better IMO. It gives the user a way to influence drawing order, but its much less costly (we can put the world-objects into bins instead of a single list when we collect them).

almarklein: Restored the sorting code. I don't see a noticeable difference for the collections example, so I guess the performance overhead is not that big.

Korijn: I'd like to discuss this more in our next video call :) thanks anyway

almarklein:
Ready from my end.

PR #61, opened 2020-08-26
add WorldObject.matrix_auto_update, set_matrix and apply_matrix

### LINKED ISSUE #60: API for transforming a world object
Regardless of how we implement transformations, there are (at least) two ways that the WorldObject API can be:

* Using `wobject.position`, `wobject.rotation` and `wobject.scale` as we do now. One can set either, and they are composed "under the hood". This is similar to how one can set these properties in a tool like blender.
* Using `wobject.transform`, e.g. `ob.transform = Translation(3,4,5) * Rotation.from_euler(...)`.

The former is nice, as one can individually set and update the translation and rotation. The latter is nice, because one can compose the current transform with a new transform.

Korijn:
Closes #60

Adds `WorldObject.matrix_auto_update` property

The example shows one way to use it; the helper functions `set_matrix` and `apply_matrix` will try to keep the pos/rot/scale properties in sync.

Another option that is not demonstrated here is to just directly set the properties yourself, and forgo the pos/rot/scale property synchronization. You would replace `cube.set_matrix(rot)` in the example with:

```python
cube.matrix.copy(rot)
cube.matrix_world_dirty = True
```

### DIFF
--- a/examples/manual_matrix_update.py
+++ b/examples/manual_matrix_update.py
@@ -1,0 +1,55 @@
+"""
+Example showing transform control flow without matrix auto updating.
+"""
+
+import numpy as np
+import imageio
+import pygfx as gfx
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
+app = QtWidgets.QApplication([])
+
+canvas = WgpuCanvas()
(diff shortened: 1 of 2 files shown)


almarklein:
LGTM!

PR #62, opened 2020-08-28
Improved support for volumes

almarklein:
This started out as a way to have `VolumeSliceMaterial` with a `plane` property analog to the `MeshSliceMaterial`.

* [x] Add support for 3D texcoords/textures to the mesh renderer.
* [x] Fixes and other improvements in PyShader ...
* [x] Make BoxGeometry produce the same data as ThreeJS does, and test it.
* [x] Yet more fixes in PyShader ;)
* [x] Remove `MeshVolumeSliceMaterial`.
* [x] Add `Volume` world object class? 
* [x] Add `VolumeSliceMaterial` with a `plane` attribute.
* [x] Do a new release of PyShader.
* [x] Figure out + implement reference frame of `plane` property.


almarklein:
So we have `MeshSliceMaterial.plane` and `VolumeSliceMaterial.plane`, which I think should behave the same way. So what happens when we have a volume that is anisotropic and/or has an offset or rotation in space? I think these would be applied using the transform of the volume-world-object. (A mesh would typically not need these transforms.) With the transforms applied, the mesh and volume will be aligned. I think that in this scenario we still want to be able to apply the same value for the `plane` property. Which means that the `plane` property (abcd) should be in the local coordinate frame of the object *after* applying its transform. Agreed @Korijn?

almarklein:
BTW, the mesh slice on the stent mesh (marching cubes of stent volume) looks like crap. Will need to look at that in another PR.

Korijn:
> So we have `MeshSliceMaterial.plane` and `VolumeSliceMaterial.plane`, which I think should behave the same way. So what happens when we have a volume that is anisotropic and/or has an offset or rotation in space? I think these would be applied using the transform of the volume-world-object. (A mesh would typically not need these transforms.) With the transforms applied, the mesh and volume will be aligned. I think that in this scenario we still want to be able to apply the same value for the `plane` property. Which means that the `plane` property (abcd) should be in the local coordinate frame of the object _after_ applying its transform. Agreed @Korijn?

In my opinion, the API should be consistent throughout the application and not make special exceptions for Volume.

**First question**: in which coordinate frame is the material's plane parameter defined? Let's make sure we are on the same page here. I see the following possibilities:

* Data/local space;
  * position (0,0,0) corresponds to the first voxel's **center**
  * with shape (n,n,n), position (n,n,n) corresponds to the last voxel's center
  * so that with spacing of (1,1,1) the lower corner of the volume's bounding box is at (-0.5,-0.5,-0.5)
  * (`-0.5` is defined as the shift from voxel center to corner in data/local space)
* UV/texture space;
  * I'm honestly not sure if this is how textures work, tell me if this is correct :)
  * (0,0,0) corresponds to the first voxel's **corner** rather than center
  * (1,1,1) corresponds to the last voxel's **corner** rather than center
  * so that with spacing of (1,1,1) the lower corner of the volume's bounding box is at (0, 0, 0)
* World space (post-world transform; position/spacing/orientation)

Whichever we choose, it should be clearly documented and consistently implemented, since there is a risk of the data being displayed with a slight offset.

I _think_ my preference would be UV/texture? We are already providing UV coordinates for other materials to do texture mapping, right? If this is very inconvenient/hard for users, my second preference would be data/local space, where (0,0,0) maps to the first voxel's **center**.

**Second question**: in which coordinate frame is the volume geometry defined? I think it should match the bounding box of the data as defined in data/local space in the list above. So the box geometry's lower corner is at `-0.5,-0.5,-0.5` and the higher is `shape[:]+0.5`.

I think the answer to your **original question** can then be derived. We can define a transformation matrix from data/local space to world space which incorporates the position (ImagePositionPatient), orientation (imageOrientationPatient) and scaling (PixelSpacing) parameter's of the volume. I think that maps very nicely to pygfx architecture.

pygfx/objects/_volume.py
```diff
@@ -0,0 +1,40 @@
+from ._base import WorldObject
+from ..geometries import BoxGeometry
+from ..resources import Buffer
+
+
+class Volume(WorldObject):
```
Korijn, line 6: Didn't we agree this would be a subclass of `Mesh`? I'm a little surprised to see it being a subclass of `Volume`. Did you need special treatment in the renderer?

pygfx/objects/_volume.py
```diff
@@ -0,0 +1,40 @@
...
+    def material(self):
+        """The material of the volume."""
+        return self._material
+
+    @material.setter
+    def material(self, material):
+        self._material = material
+        self._set_geometry(material.map.size)
+
+    def _set_geometry(self, size):
```
Korijn, line 28: I think it would be fine if this is a public method.

In a similar fashion, a helper method to set `self.matrix` based on position, rotation and spacing would be very nice of course.

pygfx/objects/_volume.py
```diff
@@ -0,0 +1,40 @@
...
+        # Create box geometry, and map to 0..1
+        geometry = BoxGeometry(1, 1, 1)
+        geometry.positions.data[:, :3] += 0.5
+        # This is our 3D texture coords
+        geometry.texcoords = Buffer(
+            geometry.positions.data[:, :3].copy(), usage="vertex|storage"
+        )
+        # Map to volume size
+        for i in range(3):
+            geometry.positions.data[:, i] *= size[i]
```
Korijn, line 38: the boundary of the volume in local space goes from (-0.5,-0.5,-0.5) to (n-0.5,n-0.5,n-0.5) right? to account for the offset between corners and centers of voxels

pygfx/objects/_volume.py
almarklein: The renderer is not the limitation, because we can trigger on the material. But semantically it makes sense (to me) that this is not a mesh. Because it is not? Unless you consider anything that has geometry a mesh.

pygfx/objects/_volume.py
almarklein: Yes, I completely overlooked that. Will fix.

pygfx/objects/_volume.py
almarklein: Good point.

> ... a helper method to set self.matrix

This is a world object, so it already has properties for position, rotation and scaling. That should be enough, right?

almarklein: done

pygfx/objects/_volume.py
```diff
@@ -0,0 +1,53 @@
...
+        geometry.positions.data[:, :3] += 0.5
+        # This is our 3D texture coords
+        geometry.texcoords = Buffer(
+            geometry.positions.data[:, :3].copy(), usage="vertex|storage"
+        )
+        # Map to volume size
+        for i in range(3):
+            column = geometry.positions.data[:, i]
+            column *= size[i]
+            column -= 0.5  # Pixel centers are in origin and size
```
almarklein, line 50: @Korijn zoiets?

pygfx/objects/_volume.py
Korijn: If I construct a Mesh manually (pass it a correct BoxGeometry and VolumeSliceMaterial), will I get the same visual result?

pygfx/objects/_volume.py
Korijn: Yes that looks good.

pygfx/objects/_volume.py
almarklein: At the moment, the renderer only triggers on a `VolumeSliceMaterial` on a `Volume`.

My view is that you can create a `Mesh`, and then use texcoords (as we do in e.g. `multislice1.py`. Then you do more work, because you are responsible for positioning the mesh. Alternatively, you use a `Volume`, in which case the geometry is implicit (created automatically).

I think these underpin a different way of thinking. In the first case you say: I'm going to put a mesh here in my scene, and on it's surface I am going to project a slice through a volume (using 3D texture coords). In the second case you say: I'm going to put a volume here in my scene, and I am going to view just a slice through it (using the VolumeSliceMaterial, as opposes to e.g. a VolumeMipMaterial).

So yes, you get the same visual result, but I don't think it warrants that a Volume is a Mesh subclass. Anyway, that's my current view; we can talk :)

Korijn: I can live with it, it doesn't matter that much. :)

PR #69, opened 2020-09-28
Line rendering vertex coloring

Korijn:
See comment https://github.com/pygfx/pygfx/pull/69#issuecomment-702556643 for pictures

* Adds vertex coloring support on all line materials.
* Updates axes helper to leverage vertex coloring (is now a single world object instead of four).

I also wanted to make sure we have a good testcase for https://github.com/pygfx/pyshader/issues/55 as you can tell by the duplication in the shaders

### DIFF
--- a/examples/axes_helper.py
+++ b/examples/axes_helper.py
@@ -17,7 +17,7 @@ scene = gfx.Scene()
 background = gfx.Background(gfx.BackgroundMaterial((0, 1, 0, 1), (0, 1, 1, 1)))
 scene.add(background)
 
-scene.add(gfx.AxesHelper(size=10))
+scene.add(gfx.AxesHelper(size=40))
 
 camera = gfx.PerspectiveCamera(70, 16 / 9)
 camera.position.set(50, 50, 50)
--- a/examples/line_thin.py
+++ b/examples/line_thin.py
@@ -36,7 +36,6 @@ colors = np.hstack([colors, np.ones((colors.shape[0], 1))])
 colors = colors.astype("f4")
 
 geometry = gfx.Geometry(positions=positions, colors=colors)
-# material = gfx.LineThinMaterial(thickness=12.0, color=(0.8, 0.7, 0.0, 1.0))
 material = gfx.LineThinMaterial(thickness=12.0, vertex_colors=True)
 line = gfx.Line(geometry, material)
 scene.add(line)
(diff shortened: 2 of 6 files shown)


almarklein:
Fixed in a commit. I can see that this is easy to overlook. Would be good to hard-set uinform types like this in the renderfunction ...

Korijn:
I totally overlooked that :) thanks.

How would you suggest I clean this PR up (avoid all the copy paste)?

almarklein:
mmm, pyshader does not have functionality for calling into functions yet. It looks like we need it more and more. For now I'd leave it as-is.

One thing that was not yet clear to me is whether we'd need a separate material for this. We could, for instance, let the per-vertex color override the material's color if it is present on the geometry. How does ThreeJS doe this?

Korijn:
As far as I know, materials in threejs have "optional" uniforms. When you add or remove (not update!) one, you set "needsUpdate=true" on the material and in the next render call the shader is reconstructed (with templating) and recompiled.

There's also material "parameters" which do not map directly to uniforms such as "vertexColors=true/false" but also need "needsUpdate=true" to trigger the same shader reconstruction and recompilation process.

Play around with this: https://threejs.org/docs/index.html#api/en/materials/MeshBasicMaterial

And look at the source to see how that demo scene is implemented: https://github.com/mrdoob/three.js/blob/dev/docs/scenes/material-browser.html

almarklein:
Ah, its a boolean on the base `Material` class: https://threejs.org/docs/index.html#api/en/materials/Material.vertexColors

Korijn:
> Ah, its a boolean on the base `Material` class: https://threejs.org/docs/index.html#api/en/materials/Material.vertexColors

I guess if we would want to support the same feature, we would need to have a prop on material that is somehow also a `Resource` object, so as to hook into the update mechanism, right?

Perhaps I can simply call `self._bumprev` in a setter as well.

almarklein:
The `vertexColors` prop can be a bool. If you implement the setter to call `_bump_rev`, then any parent WorldObject's will be bumped as well, and thus trigger a pipeline rebuild.

Korijn:
Did that; don't support it on thin lines yet.

Korijn:
Axes helper is now a single world object with line segment material vertex coloring:

image

Added support for thin lines as well:

image

almarklein:
LGTM!

PR #71, opened 2020-10-09
Post-render passes and offscreen rendering

almarklein:
* [x] Implement multi-stage rendering, where the scene is rendered to a texture, and then this texture is rendered to the canvas-texture.
* [x] Implement queries from the render texture (RGB and depth).
* [x] It is possible to add custom post-processing render passes.
* [x] Off-screen rendering.
* [x] Implement aa by rendering to higher res and smoothing the result.
* [x] Add an example to demonstrate "picking" the depth, position and color.
* [x] Add an example to demonstrate offscreen rendering.
* [x] Add an example to demonstrate custom post-processing steps.
* [x] Make ssaa logic not care about alignment of texture sizes.

Future work:
* msaa -> can't implement this yet. I can set the `sample_count` of the pipeline and of textures, but get "The render pipeline and renderpass have mismatching sample_count", but I cannot yet set the `sample_count of the render_pass.
* Order-independent transparency.
* Fog.
* More fancy post-processing effects that can be added by the user (added a noise example here).

----

*The examples below will look a bit different now, but the ideas are still the same*

Example comparing direct drawing (what we have before this PR), and with ssaa:
afbeelding

Example setting `renderer.ssaa = 0.25` to lower the resolution of the render texture:
afbeelding

Example post-processing to add noise:
afbeelding


examples/picking_color_and_depth.py
```diff
@@ -0,0 +1,51 @@
...
+from wgpu.gui.qt import WgpuCanvas
+
+
+app = QtWidgets.QApplication([])
+
+
+class PickingWgpuCanvas(WgpuCanvas):
+    def mousePressEvent(self, event):  # noqa: N802
+        xy = event.x(), self.height() - event.y()  # renderer as origin at bottom left
+        print(renderer.get_info_at(xy))
```
almarklein, line 19: Returns a dict with info about that location.

almarklein: This should be a comment in the code :P

Korijn:
Is MSAA always implemented with an intermediate texture? Just wondering :)

almarklein:
MSAA works with multi-sampling of vertices, and does not necessarily work for all types of visualizations as it's somewhat mesh-specific. Nevertheless, we can/should support this too.

Yes, this is SSAA. We can also implement FXAA (and similar methods) in the post processing step. These try to detect sharp edges and smooth in the direction of these edges. Cheap but quite effective. And we can then combine FXAA + SSAA :)

pygfx/renderers/wgpu/rendertexture.py
```diff
@@ -243,4 +250,11 @@ def fragment_shader(
             w = kernel[abs(x)] * kernel[abs(y)]
             val += t_tex.sample(s_sam, texcoord) * w
 
+    # Add noise - adds a bit of a scenematic effect
+    # todo: use spirv builtin random number gen
+    xy = v_texcoord.xy
+    noise = fract(sin(xy @ vec2(12.9898, 78.233) + u_render.time) * 43758.5453)
+    noise = u_render.noise * noise
+    val += vec4(noise, noise, noise, 1.0)
```
Korijn, line 258: shouldn't post processing be set up as separate shader passes?

almarklein: Yes. I now made the post processing steps configurable, by creating a list of PostProcessingStep objects. Added an example that adds noise to a scene.

Another example is fog. That should be relatively easy now!

almarklein:
So off-screen rendering! Technically it's not hard, but we need to handle this special case to bypass the swap-chain stuff. The question is where to do this (in wgpy-py or pygfx), and what looks like the most elegant API. I see two options. They do not necessarily exclude one another, but let's pick one (or a 3d option?) for now.

### 1: special canvas class in wgpu-py

It has the same API as any canvas. In wgpu-py we need to add some special cases to fake a swap chain.

```py
from wgpu.gui.offscreen import WgpuCanvas
canvas = WgpuCanvas()
...
canvas.snapshot()
```

Interesting to have a canvas API with e.g. `set_logical_size`. Though `close` does not make sense, and `request_draw` is a bit iffy: it'd either directly call the draw function (not what you'd expect!), or post an event via asyncio (sigh), or give it its own little event loop thingy?

### 2: provide size instead of a canvas

I.e. make a renderer usable without a canvas. 
```py
renderer = gfx.renderers.WgpuRenderer((640, 480))
...
renderer.snapshot()
```
Not sure how to deal with logical size vs physical size. Perhaps `WgpuRenderer((640, 480, 2))` to specify pixel ratio.

Requesting a new draw ... not necessary, because you'd just call `renderer.render(scene, camera)`

### Current thoughts

I was first leaning for 1, but the request_draw thing turned me off. With 2 you actually have less code and it's more clear that you're not actually managing a window.

I'd like a `renderer.snapshot()` also when using a canvas :)

Thoughts?

Korijn:
IMHO you can solve this by separately thinking about the snapshot API and the canvas API (wgpu's GUI abstraction layer).

* `request_draw` vs `snapshot` is about offering two APIs for two different use cases: game/event loops and static image rendering, respectively.
* `wgpu.gui.offscreen.WgpuCanvas` is just canvas that is not tied to a GUI toolkit. You should be able to use it in both of the above use cases, if you wanted to.

In that sense, I think it would make most sense to implement the offscreen `WgpuCanvas` and the `canvas.snapshot` method in wgpu-py, and then _also_ provide `renderer.snapshot` as a convenience method.

almarklein:
> And `canvas.snapshot()` would return a view on the buffer I suppose?

Yeah, this would return the image (as a memoryview or numpy array), so you can then write it to e.g. png.


> Users can still directly call `canvas.update()` (which we could easily alias to `canvas.draw()`) if they wanted to force an immediate draw, instead of `canvas.request_draw()`, right?

Not quite ... The canvas API is aimed at GUI toolkits with an event loop. The `canvas.request_draw()` method is used to specify a drawing function, and also to ask for a new draw. The GUI's event loop will then schedule a new draw. The `canvas.update()` is specific to Qt, and does what `request_draw` does, except not limiting FPS.

One could off-course call the `animate()` function that you wrote. But if you cannot use the off-screen canvas as any other canvas, it defeats the purpose, I think. 


> `request_draw` vs `update+snapshot` is about offering two APIs for two different use cases: game/event loops and static image rendering, respectively.

In other words: there is no `update` (direct-draw) API on the canvas, because that function is already implemented separately (the animate function in our examples).

Another question to ask is whether an offscreen `Canvas` in wgpu-py would be generically useful. I think the answer is "not really", because if one want to render to a texture, one would ... use a texture :)  Maybe testing purposes? I think the answer is still no, because wgpu-py will have to bypass the swap chain stuff, so there would not be much shared code paths left.

Korijn:
In that case, I would also go with option 2. You proved your point. :) Regarding pixel ratio, why would it ever be something other than 1?

almarklein:
> why would it ever be something other than 1 (when off screen)?

Things like line widths are often expressed in logical pixels. So then you can create screenshot with high resolution without making everything thin.

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -68,24 +71,105 @@ def __init__(self, *, stdinfo_uniform):
...
+        follows the screens pixel ratio on high-res displays, and is 2
+        otherwise.
+
+        If the used pixel ratio causes the render texture to be larger
+        than the physical size of the canvas, SSAA is applied, resulting
+        in a smoother final image with less jagged edges. Alternatively,
+        this value can be set to e.g. 0.5 to lower* the resolution (e.g.
+        for performance during interaction).
+        """
+        return self._pixel_ratio
```
almarklein, line 155: In earlier iterations this was the ssaa property. Calling it `pixel_ratio` makes more sense, I think.

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -849,3 +962,116 @@ def _update_sampler(self, resource):
...
+                "offset": buf_offset,
+                "bytes_per_row": bytes_per_pixel,
+                "rows_per_image": 1,
+            },
+            copy_size=(1, 1, 1),
+        )
+
+    def snapshot(self, step_index=-1):
+        """Create a snapshot of the currently rendered image."""
+
```
almarklein, line 1043: There might be a better name for this method. Ideas?

almarklein:
Ready for review!

pygfx/renderers/wgpu/_wgpurenderer.py
ivoflipse: screenshot? Though ironic if there's no screen

almarklein: Haha, yeah I want to avoid the term "screen". When there is a canvas, you can also create a snapshot, but it's not exactly the same as what you see on screen; it will be the (perhaps higher resolution) version from what is rendered to screen with the ssaa pass.

almarklein: Maybe `save()`? But that is so generic...

Korijn: I think snapshot is fine 👍

pygfx/renderers/wgpu/_wgpurenderer.py
Korijn: Let's roll with it for now, we can still change our minds later

examples/offscreen.py
```diff
@@ -0,0 +1,35 @@
+"""
+Example demonstrating off-screen rendering.
+"""
+
+import imageio
+import pygfx as gfx
+
+# Instead of a canvas, we provide a size (in logical pixels). Properties
+# like line widths are usually expressed relative to this logical size.
+# We set pixel_ratio to 2 (the default) to double the resolution.
```
Korijn, line 10: I guess I don't understand why the default is 2? I would be surprised to get an image twice the size as what I requested

examples/offscreen.py
```diff
@@ -0,0 +1,35 @@
...
+
+rot = gfx.linalg.Quaternion().set_from_euler(gfx.linalg.Euler(0.5, 1.0))
+cube.rotation.multiply(rot)
+
+
+if __name__ == "__main__":
+    renderer.render(scene, camera)
+    im = renderer.snapshot()
+    print(im.shape)  # (960, 1280, 4)
+    imageio.imsave("offscreen.png", im)
```
Korijn, line 35: This example is extremely elegant. I absolutely love it. No window required!

examples/post_processing.py
```diff
@@ -0,0 +1,85 @@
...
+    noise = u_render.noise * noise
+    val += vec4(noise, noise, noise, 1.0)
+    out_color = val  # noqa - shader output
+
+
+canvas = WgpuCanvas()
+renderer = gfx.renderers.WgpuRenderer(canvas)
+scene = gfx.Scene()
+
+renderer.post_processing_steps.append(NoisyPostProcessingStep(renderer.device))
```
Korijn, line 58: How about `renderer.postfx` for short?

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -68,24 +71,105 @@ def __init__(self, *, stdinfo_uniform):
...
+        # Prepare render targets. These are placeholders to set during
+        # each renderpass, intended to keep together the texture-view
+        # object, its size, and its format. The final (canvas) texture
+        # has bgra format, because that's what hardware seems to like,
+        # and srgb for perseptive color mapping.
+        self._canvas_texture = RenderTexture(wgpu.TextureFormat.bgra8unorm_srgb)
+        self._depth_texture = RenderTexture(wgpu.TextureFormat.depth32float)
+        # We keep track of multiple render textures, one for each post
+        # processing step, plus one final texture. Will be filled as needed.
+        self._render_textures = []
```
Korijn, line 104: For post processing, wouldn't a pair of texture suffice? You can go back and forth between them for each step.

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -849,3 +961,116 @@ def _update_sampler(self, resource):
...
+        queue = self._device.default_queue
+        queue.submit([encoder.finish()])
+
+        # Collect data from the buffer
+        data = self._pixel_info_buffer.read_data()
+        raw_depth = data[0:4].cast("f")[0]
+
+        # Calculate world coordinates using the camera
+        # todo: how does this hold up when a renderer is used with multiple cameras?
+        camera = self._last_camera
```
Korijn, line 1004: I don't think it does :) you could let the user provide a camera as an (optional?) argument to make it a little more flexible

examples/offscreen.py
almarklein: Quality ... Note that the `pixel_ratio` property is also used when there is a canvas. On an hdpi screen, its default value is equal to the screen (usually 2), causing the render texture to be twice as large as the logical size. On non-hdpi screens, the default pixel ratio is 2; the render texture is thus of equal size as on an hdpi screen, and ssaa is applied to avoid those jagged edges. We could say that offscreen (without a canvas) the default pixel_ratio is 1, but I think it makes more sense to be consistent, and output high-res images.

examples/offscreen.py
almarklein: Yeah, writing this example was mostly deleting lines, and it felt great :)

examples/post_processing.py
almarklein: mmm, I can imagine situations where you want to add a post-filter for a while and then remove it later. We could add API for that, but a list object already has this API :)

That said, I have not given this API much thought. At first the SSAA step was also part of this list, but things worked better to make it "builtin". Maybe doing post-processing effects should be done completely different still. Let's make an issue to track this.

almarklein: #72

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: Great idea!

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: > Personally I think the picking function should only return the color and depth.

Yeah, that would solve that problem.

examples/offscreen.py
Korijn: I don't think anti-aliasing requirements should be on by default, especially SSAA which is the most resource costly approach. But I can get over it if you think this is best :)

examples/post_processing.py
Korijn: Fair enough! In that case I still suggest `renderer.postfx` as a shorter name for the list property :)

examples/offscreen.py
almarklein: I understand your concern. The main reasons I think it is a good idea are:
* If you connect an hdpi screen, you're rendering in high resolution. So why not just do that on normal screens too?
* If your making a quick example I'd rather have it look good.

In other words: as it is now, you get about the same quality (and performance penalty) regardless of using a high-res display or not :)

BTW: setting `pixel_ratio` to 0.5 during mouse interaction may be very effective in boosting performance, especially with volume rendering, resulting in 8x less rays to be cast. That last ssaa step will then just smooth the result, so it won't look blocky.

almarklein: You could even time the draw time, and scale `pixel_ratio` to get the highest resolution while still having smooth interaction ;)

examples/post_processing.py
almarklein: Haha, agreed!

examples/offscreen.py
Korijn: Fair enough! :)

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: done

PR #77, opened 2020-10-19
Picking

almarklein:
* [x] Give every world-object a unique id.
* [x] Implement a picking texture and retrieve info from it.
* [x] POC to pick meshes.
* Implement per-vertex (or per pixel) picking ...
  * [x] Picking for line.
  * [x] Picking for point.
  * [x] Picking for volume slice.
  * [x] Picking for mesh. POC: can pick face, edge and vertex by providing weights.
  * [x] Picking for instanced mesh
  * [x] Picking for mesh slice
* [x] Check any todo's that I added in the process ...
* [x] Run our test suite ... ahum ... test all examples to see whether I broke something.
* [x] Fix background materials (they did not draw because they paint at the furthest depth, but the initial depth buffer value was set too low. This regression was introduced in the previous PR (offscreen rendering and basic info-picking).
* [x] Consistent names for indices and sub-coords in pick info dict

### DIFF
--- a/examples/picking_mesh.py
+++ b/examples/picking_mesh.py
@@ -17,17 +17,17 @@ class PickingWgpuCanvas(WgpuCanvas):
     def mousePressEvent(self, event):  # noqa: N802
         # Get a dict with info about the clicked location
         xy = event.x(), event.y()
-        info = renderer.get_info_at(xy)
+        info = renderer.get_pick_info(xy)
         wobject = info["world_object"]
         # If a mesh was clicked ..
         if wobject and "face_index" in info:
             # Get what face was clicked
             face_index = info["face_index"]
-            weights = info["face_weights"]
+            coords = info["face_coords"]
             # Select which of the three vertices was closest
             # Note that you can also select all vertices for this face,
-            # or use the weights to select the closest edge.
-            sub_index = np.argmax(weights)
+            # or use the coords to select the closest edge.
+            sub_index = np.argmax(coords)
             # Look up the vertex index
             vertex_index = int(wobject.geometry.index.data[face_index * 3 + sub_index])
             # Change the position of that vertex
--- a/examples/picking_mesh.py
+++ b/examples/picking_mesh.py
@@ -1,5 +1,6 @@
 """
-Example showing picking a mesh.
+Example showing picking a mesh. Showing two meshes that can be clicked
+on. Upon clicking, the vertex closest to the pick location is moved.
 """
 
 # todo: if we have per-vertex coloring, we can paint on the mesh instead :D
@@ -10,9 +11,6 @@ import pygfx as gfx
 
 from PyQt5 import QtWidgets
 from wgpu.gui.qt import WgpuCanvas
-
-
-app = QtWidgets.QApplication([])
 
 
 class PickingWgpuCanvas(WgpuCanvas):
@@ -38,6 +36,7 @@ class PickingWgpuCanvas(WgpuCanvas):
             wobject.geometry.positions.update_range(vertex_index, 1)
 
 
+app = QtWidgets.QApplication([])
 canvas = PickingWgpuCanvas()
 renderer = gfx.renderers.WgpuRenderer(canvas)
 scene = gfx.Scene()
--- a/examples/picking_mesh.py
+++ b/examples/picking_mesh.py
@@ -1,0 +1,75 @@
+"""
+Example showing picking a mesh.
+"""
+
+# todo: if we have per-vertex coloring, we can paint on the mesh instead :D
+
+import numpy as np
+import imageio
+import pygfx as gfx
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
+app = QtWidgets.QApplication([])
+
+
+class PickingWgpuCanvas(WgpuCanvas):
+    def mousePressEvent(self, event):  # noqa: N802
+        # Get a dict with info about the clicked location
+        xy = event.x(), event.y()
+        info = renderer.get_info_at(xy)
+        wobject = info["world_object"]
+        # If a mesh was clicked ..
+        if wobject and "face_index" in info:
+            # Get what face was clicked
+            face_index = info["face_index"]
+            weights = info["face_weights"]
+            # Select which of the three vertices was closest
+            # Note that you can also select all vertices for this face,
+            # or use the weights to select the closest edge.
+            sub_index = np.argmax(weights)
+            # Look up the vertex index
+            vertex_index = int(wobject.geometry.index.data[face_index * 3 + sub_index])
+            # Change the position of that vertex
+            pos = wobject.geometry.positions.data[vertex_index]
+            pos[:] *= 1.1
+            wobject.geometry.positions.update_range(vertex_index, 1)
+
+
+canvas = PickingWgpuCanvas()
+renderer = gfx.renderers.WgpuRenderer(canvas)
+scene = gfx.Scene()
+
+im = imageio.imread("imageio:astronaut.png").astype(np.float32) / 255
+tex = gfx.Texture(im, dim=2, usage="sampled").get_view(
+    filter="linear", address_mode="repeat"
+)
+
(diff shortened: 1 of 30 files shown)


pygfx/objects/_base.py
```diff
@@ -87,6 +93,23 @@ def __init__(self):
         self.visible = True
         self.render_order = 0
 
+        # See if we reached max number of objects. If so, try cleanup and try again.
+        if len(_idmap) >= _idmax:
+            gc.collect()
```
Korijn, line 98: I seem tot recall this only schedules a cleanup.

pygfx/objects/_base.py
```diff
@@ -87,6 +93,23 @@ def __init__(self):
...
+            if len(_idmap) >= _idmax:
+                raise RuntimeError("Max number of object reached")
+        # Set id
+        self._id = id(self) % _idmax
+        while self._id in _idmap.values():
+            self._id = (self._id + 1) % _idmax
```
Korijn, line 104: Isn't this potentially very slow?

Korijn:
I'm pretty sure not all future (e.g. defined by users) materials will support this so I wonder if it should be OK to have materials that do not render to the picking texture. Or perhaps even disable the feature. What do you think?

pygfx/objects/_base.py
almarklein: A global counter could overflow if objects are continuously being created - you'd want to be able to re-use id's. This kinda uses uid's, derived from the object `id()`, the loop *could* take a long time, but assuming that the id's are randomly distributed, the average number of iterations is just 2 if we're halfway to the max. Maybe make it actually random (uid's) to make this assumption more realistic :D

almarklein:
Yes to both. Materials that want to support picking should write to the picking buffer. But it's optional. I think we can disable picking in the renderer by not using that picking texture. I think the shaders that write to it then simply write into the void.

examples/picking_mesh.py
```diff
@@ -0,0 +1,75 @@
...
+            # Select which of the three vertices was closest
+            # Note that you can also select all vertices for this face,
+            # or use the weights to select the closest edge.
+            sub_index = np.argmax(weights)
+            # Look up the vertex index
+            vertex_index = int(wobject.geometry.index.data[face_index * 3 + sub_index])
+            # Change the position of that vertex
+            pos = wobject.geometry.positions.data[vertex_index]
+            pos[:] *= 1.1
+            wobject.geometry.positions.update_range(vertex_index, 1)
```
almarklein, line 38: @Korijn how does this feel?

Every object (and maybe material) will have different pick info. For images/volumes it will be the pixel/voxel coordinate. For meshes it is the face index, plus weights. These weight may need some explanation: they represent how close the clicked fragment is to each vertex (0..1). It is actually quite powerful, because now you can:
* Select the face as a whole.
* Select the nearest edge.
* Select the nearest vertex (as in this example).
* Select either vertex/edge/face, depending on how far off a vertex/edge you are ...

Korijn: It's awesome! Are the weights barycentric coordinates?

pygfx/objects/_base.py
```diff
@@ -7,9 +7,12 @@
 from ..resources import Resource, Buffer
 from ..utils import array_from_shadertype
 
-# Keep track of id's
+# Keep track of id's. About the max:
+# * 2_147_483_647 (2**31 -1) max number for signed i32.
+# *    16_777_216 max integer that can be stored exactly in f32
+# *     4_000_000 max integer that can survive being passed as a varying
```
Korijn, line 13: Why?

examples/picking_mesh.py
almarklein: I had to look that up, but yeah, I think they are :)

Korijn: Looks like it, they range from 0..1 from every vertex to the center of the opposite edge. Awesome, you can do tons of cool stuff with that (like you said). It's a neat trick to just let the rasterizer interpolate those numbers!

pygfx/objects/_base.py
almarklein: I tested it. I expected simple float rounding, but apparently magic stuff happens there ... I just hope this is not too device-specific ... Worst case we can use 2 values to encode one int. But then we need two varyings, because we do not have `vec5` :)

almarklein:
Getting there ... mesh slice picking uses the same barycentric coords!

Korijn:
> Getting there ... mesh slice picking uses the same barycentric coords!

Very cool! I don't think I've ever seen picking by rendering with barycentric coords. I wonder if it can be considered to have feature parity with ray picking?

almarklein:
I think that if we'd gone with ray picking, we'd have tested what vertex is closest to the ray and then call it a day :)  Now we were somewhat forced to test the faces instead, and add the barycentric coords to figure out what vertex is closest, resulting in very high details about what was picked (e.g. also being able to pick edges). I suppose we could obtain similar picking details with ray picking, though it would probably be more computationally expensive to calculate face intersections rather than vertex distances.

Korijn:
Yeah, it's pretty awesome! Really impressive work.

The downsides I can think of at this point:

* Approach depends material support (geometry based would be independent of materials)
  * At the same time, this is instead an advantage where geometry is computed in a shader :)
* With a ray you can detect _all_ intersected objects (behind the front most geometry)

almarklein:
> With a ray you can detect all intersected objects (behind the front most geometry)

Good one. That would be really hard to get with render-picking, I think.

pygfx/objects/_base.py
almarklein: Made them actually random. And also decrease the max number of objects to 75% of the max id to prevent the set from "choking".

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -99,6 +100,7 @@ def __init__(self, canvas=None, size=None, pixel_ratio=None):
         # and srgb for perseptive color mapping.
         self._canvas_texture = RenderTexture(wgpu.TextureFormat.bgra8unorm_srgb)
         self._depth_texture = RenderTexture(wgpu.TextureFormat.depth32float)
+        self._pick_texture = RenderTexture(wgpu.TextureFormat.rgba32sint)
```
almarklein, line 103: Note to self: what about using f32 instead? Many values passed here are floats.

pygfx/renderers/wgpu/meshrender.py
```diff
@@ -61,6 +61,18 @@ def mesh_renderer(wobject, render_info):
     }
     bindings1 = {}
 
+    # todo: clean this up
```
almarklein, line 64: .

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: Encoding a float by multiplying it is good enough. The other way around is trickier.  Also, encoding multiple small fixed-point values into 1 32 bit int is easier than encoding three values in a float. So leaving this int.

almarklein:
Almost done. Let's refine (and document) how we present picking info.

Common:
* `world_object`: if an object was picked, this is set.
* `instance_index`: Can currently only be nonzero for meshes. Lets omit for objects for which instancing is not supported.

Meshes:
* `face_index`: int
* `face_weights`: tuple of three float weights (barycentric coordinates, 256 bit precision)

Points: 
* `vertex_index`: int

Lines:
* Currently we have: int `vertex_index` and float `subcoord`. 
* I propose: float `vertex_index`. It bothers me a bit that a thing called "index" is not integer, but this is more practical.

Volume:
* Now: `xyz`: tuple of 3 floats
* I propose: `voxel_index`: tuple of 3 floats

Image (if we add this):
* `pixel_index`: tuple of 2 floats

@Korijn thoughts?

Korijn:
Sounds good to me! It's the complete set of information I would want from picking.. so. No complaints here.

almarklein:
Done! The only change wrt what I posted, I renamed `face_weights` to `face_coords`.

PR #79, opened 2020-10-31
Cylinder geometry

Korijn:
image

### DIFF
--- a/examples/cylinder.py
+++ b/examples/cylinder.py
@@ -2,6 +2,7 @@
 Example showing a single geometric cylinder.
 """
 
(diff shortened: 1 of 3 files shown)


examples/cylinder.py
```diff
@@ -0,0 +1,107 @@
+"""
+Example showing a single geometric cylinder.
```
almarklein, line 2: It shows multiple cylinders, right? ;)

almarklein:
Nice!

almarklein:
Good to merge, I think.

PR #80, opened 2020-10-31
Grid helper

Korijn:
It's a little dark set against the bright axes helper so look closely :)

image


almarklein:
Nice! LGTM. 

One remark is that I think that we can also make infinite grids in a fragments shader, similar to how we have the background objects. But this a lot easier :)

Korijn:
> Nice! LGTM.
> 
> One remark is that I think that we can also make infinite grids in a fragments shader, similar to how we have the background objects. But this a lot easier :)

I'd love to learn how to implement that :)

PR #91, opened 2021-04-12
Adjust to API changes in WGPU

almarklein:
* [x] `Canvas.configure_swap_chain`.
* [x] WGPU objects have Less properties.
* [x] Other changes as we update wgpu to the latest WebGPU.
* [x] Maybe other changes as we update wgpu to the latest wgpu-native.


examples/geometry_cube_wx.py
```diff
@@ -0,0 +1,48 @@
+"""
+Example showing multiple rotating cubes. This also tests the depth buffer.
+"""
+
+import imageio
+import pygfx as gfx
+
+from PyQt5 import QtWidgets
```
Korijn, line 8: Filename says this is a wx example but it uses qt :)

Korijn:
So far it looks like everything got a little easier on the eyes! :)

examples/geometry_cube_wx.py
almarklein: I think I made a mistake when copying the example code :)  I just deleted this file for now.

PR #97, opened 2021-05-27
Use WGSL shaders

almarklein:
* [x] Rewrite shaders to WGSL
* [x] Maybe look into first basic version of templating (e.g. jinja2)
* [x] Define uniform structs with numpy dtypes instead of pyshader.
* [x] Auto-generate wgsl struct code from these, warn/fix for alignments?
* [x] Fixed phong lighting.

### DIFF
--- a/pygfx/renderers/wgpu/_wgpurenderer.py
+++ b/pygfx/renderers/wgpu/_wgpurenderer.py
@@ -21,7 +21,9 @@ from .postprocessing import RenderTexture, SSAAPostProcessingStep
 # todo: a combined transform would be nice too, for performance
 stdinfo_uniform_type = Struct(
     cam_transform=mat4,
+    cam_transform_inv=mat4,
     projection_transform=mat4,
+    projection_transform_inv=mat4,
     physical_size=vec2,
     logical_size=vec2,
 )
@@ -364,7 +366,9 @@ class WgpuRenderer(Renderer):
         # Update its data
         stdinfo_data = self._wgpu_stdinfo_buffer.data
         stdinfo_data["cam_transform"] = tuple(camera.matrix_world_inverse.elements)
+        # stdinfo_data["cam_transform_inv"] = tuple(camera.matrix_world.elements)
         stdinfo_data["projection_transform"] = tuple(camera.projection_matrix.elements)
+        # stdinfo_data["projection_transform_inv"] = tuple(camera.projection_matrix_inverse.elements)
         stdinfo_data["physical_size"] = physical_size
         stdinfo_data["logical_size"] = logical_size
         # Upload to GPU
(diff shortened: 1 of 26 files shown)


pygfx/renderers/wgpu/meshrender.py
```diff
@@ -757,523 +506,339 @@ def helpers(self):
...
-
-    if pos_index < 3:  # or n@u == 0
-        # Just return the same vertex, resulting in degenerate triangles
-        wpos1 = vec4(pos1, 1.0)
-        the_pos = u_stdinfo.projection_transform * u_stdinfo.cam_transform * wpos1
-        the_coord = vec2(0, 0)
-        segment_length = 0.0
+    # Put it together!
+    wgsl = shader.to_string()
+    with open(r"c:/dev/py/tmp.wgsl", "wb") as f:
```
ivoflipse, line 541: Probably not what you want to keep around right?

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -1039,6 +1041,14 @@ def _update_sampler(self, resource):
         )
         resource._wgpu_sampler = resource.rev, sampler
 
+    def _get_shader_module(self, key, source):
+        """Compile a shader module object, or re-use it from the cache."""
+        # todo: make this work for objects following the ShaderSourceTemplate interface
+        if key not in self._shader_cache:
+            m = self._device.create_shader_module(code=source)
+            self._shader_cache[key] = m
+        return self._shader_cache[key]
```
Korijn, line 1050: just wondering if you also need to implement code to release the resources?

pygfx/renderers/wgpu/meshrender.py
almarklein: Nice catch!

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: mmm, yeah, in practice the number of shaders used is probably finite, but it does not have to be.

pygfx/renderers/wgpu/volumerender.py
```diff
@@ -43,241 +41,287 @@ def volume_slice_renderer(wobject, render_info):
...
-                func.__annotations__["t_tex"] = tex_anno
-            fragment_shader = fragment_shader.float_version
+            raise ValueError("With VolumeSliceMaterial, geometry needs texcoords")
+        # Sampling type
+        if "norm" in material.map.format or "float" in material.map.format:
+            shader["texture_format"] = "f32"
+            if "unorm" in material.map.format:
+                shader["climcorrection"] = " * 255.0"
+            elif "snorm" in material.map.format:
+                shader["climcorrection"] = " * 255.0 - 128.0"
```
Korijn, line 51: 🙈

Korijn:
Would you mind trying `examples/helpers.py` as a quick check on the coordinate systems?

almarklein:
> Would you mind trying examples/helpers.py as a quick check on the coordinate systems?

I will have a look in a new PR. Will also add some examples to validate coordinates ...

PR #98, opened 2021-06-03
Wgsl part 2

almarklein:
Writing as a separate PR to make review easier. In this PR I want to focus on ironing out some details wrt the new shaders.

After this I'll port the other shaders.

### DIFF
--- a/pygfx/renderers/wgpu/_shadercomposer.py
+++ b/pygfx/renderers/wgpu/_shadercomposer.py
@@ -54,12 +54,11 @@ class BaseShader:
         if isinstance(struct, dict):
             dtype_struct = np.dtype([(key,) + val for key, val in struct.items()])
         elif isinstance(struct, np.dtype):
+            if struct.fields is None:
+                raise TypeError(f"define_uniform() needs a structured dtype")
             dtype_struct = struct
         else:
             raise TypeError(f"Unsupported struct type {struct.__class__.__name__}")
-
-        # todo: ensure dtype is structural
-        # todo: validate that wgsl_type exists? Or let wgpu do that?
 
         for fieldname, (dtype, offset) in dtype_struct.fields.items():
             # Resolve primitive type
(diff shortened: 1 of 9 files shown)


pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -19,13 +17,13 @@
...
-    projection_transform_inv=mat4,
-    physical_size=vec2,
-    logical_size=vec2,
+stdinfo_uniform_type = dict(
+    cam_transform=("float32", (4, 4)),
+    cam_transform_inv=("float32", (4, 4)),
+    projection_transform=("float32", (4, 4)),
+    projection_transform_inv=("float32", (4, 4)),
+    physical_size=("float32", 2),
+    logical_size=("float32", 2),
```
almarklein, line 26: I can't say I think this is particularly pretty, but at least it's generic, and familiar to (some) numpy users.

We can make up any format we like to define structs, so if you have other ideas, speak up :)

Korijn: since you're passing it along to numpy untouched, does that mean you can also use e.g. `f4` instead of `float32`?

almarklein: Yes. That would make it a bit shorter.

Korijn: Well I don't mean to say you should, but I just wanted to confirm that you can :)

pygfx/renderers/wgpu/_shadercomposer.py
```diff
@@ -0,0 +1,40 @@
...
+    def __setitem__(self, key, value):
+        self.kwargs[key] = value
+
+    def __getitem__(self, key):
+        return self.kwargs[key]
+
+    def get_code(self):
+        raise NotImplementedError()
+
+    def get_final_code(self, **kwargs):
```
Korijn, line 29: You could call this method `render`

almarklein:
Ready for review (this PR merges into the wgsl branch).

Korijn:
LGTM!

PR #100, opened 2021-06-08
some tweaks and extra validating examples

### LINKED ISSUE #99: Get coordinate frame right
Check + fix + document the coordinate frame definitions. In other words, make sure that up is up.

E.g.:
* I've read that depending on whether you're submitting wgsl or spirv, the NDC might be different.
* The different volume slice examples don't all look/scroll the same. Fix this and understand how.
* Image examples?

almarklein:
Closes #99


examples/volume_slice4.py
```diff
@@ -40,6 +40,7 @@ def mousePressEvent(self, event):  # noqa: N802
 
 camera = gfx.OrthographicCamera(128, 128)
 camera.position.set(64, 64, 128)
+camera.scale.y = -1  # in this case we tweak the camera, not the plane
```
Korijn, line 43: does that not mess up mouse interaction though?

almarklein: Thinking of CT scans, for example, you may want to flip axis to get the correct view. I think the camera scale is where this should happen. Maybe it Just Works with interaction? Otherwise I suppose interaction should take it into account, ideally.

Korijn: Well, let's merge this as is and we can refer to #101 later

Issue #73, opened 2020-10-15
Order-independent transparency


almarklein:
In #71 we laid the foundations for rendering to textures and managing those texture in the renderer. From there, the step to implementing OIT should be quite near!

## Links

* A great overview in a Bsc thesis by Felix Brüll in 2018: [Order Independent Transparency Acceleration](https://www2.in.tu-clausthal.de/~cgstore/theses/ba_felix_br%C3%BCll_2018.pdf)
* Weighted Blended Order-independent transparency (McGuire 2013): [publication](http://jcgt.org/published/0002/02/09/) [blog](http://casual-effects.blogspot.com/2015/03/implemented-weighted-blended-order.html)
* Phenomenological Transparency (McGuire 2016): [publication](https://casual-effects.com/research/McGuire2017Transparency/index.html) [blog](http://casual-effects.blogspot.com/2015/03/colored-blended-order-independent.html)
* Stochastic Transparency (Enderton 2011): [publication](http://enderton.org/eric/pub/stochtransp-tvcg.pdf)
* Adaptive Transparency (Salvi 2012): [publication](https://software.intel.com/content/www/us/en/develop/articles/adaptive-transparency.html)
* Multi Layer Alpha Blending MLAB (Salvi 2017): [publication](https://software.intel.com/content/www/us/en/develop/articles/multi-layer-alpha-blending.html)
* Multi-Layer Alpha Tracing (Brüll 2020): [publication](https://www.researchgate.net/publication/344441737_Multi-Layer_Alpha_Tracing/link/5f75d904299bf1b53e03ef42/download)
* OIT in vispy [#984](https://github.com/vispy/vispy/issues/984), [#2039](https://github.com/vispy/vispy/pull/2039), [#2230](https://github.com/vispy/vispy/pull/2230)

### Transparency is hard

Doing transparency the right way is hard. But it is also important to get right (IMO) or transparency will always be problematic for your viz library. The thing is that semi-transparent fragments must be blended with other semi-transparent fragments (and any opaque fragments) *in a fixed order*. That order can be back to front, or front to back, but it must be consistent. It is virtually impossible to make the fragments blend in a consistent order.

### A-buffer

The theoretically most correct solution for this is an A buffer. You store all drawn fragments, sort them, and then blend them together. Unfortunately, its hard to establish how many fragments should fit, and that number might be problematic memory-wise. 

### Weighted blended

[...]

PR #108, opened 2021-08-11
Attempt at implementing OIT via fragment buffer

almarklein:
This is an attempt at Order independent Transparency #73.

* [x] Change architecture to render via a buffer.
* [x] Test atomic ops and spin-lock. (looks like we need a newer Naga for this)
* [ ] Properly deal with canvas resizing.
* [ ] Do some benchmarking to see what the costs are.
* [ ] Implement WB-OIT or MLAB.
* [ ] Write examples to see result.
* [ ] Make all visuals use this approach.

## Spinlocks notes

This spinlock looks like it should work, but it does not. The reason is probably [this](https://stackoverflow.com/a/51708878/2271927): 
    
 > you are relying on one aspect of relative order: that all threads will eventually make forward progress. That is, you assume that any thread spinning on a lock will not starve the thread that has the lock of its execution resources. That threads holding the lock will eventually make forward progress and release it.
 
```wgsl
        fn write_fragment_spinlock1(index: i32, frag: Fragment) {
           let active_lock = &s_fragments.data[index].lock;
            loop {
                let old = atomicExchange(active_lock, 2);
                if (old != 2) { break; }
            }
            write_fragment_naive(index, frag);
            atomicStore(active_lock, 1);
        }
```

The next lock is an interesting alternative. I read in an nvidia presentation (that I can't find back now) that this has advantages over an actual spinlock. Unfortunately, this does not work right now because it's not yet (fully) implemented in Naga. Also see
 https://github.com/gfx-rs/naga/blob/3e1244c5cbff24506b82bed906e283260fc19849/src/valid/function.rs#L336,    https://github.com/gfx-rs/naga/issues/1413,  https://github.com/gpuweb/gpuweb/issues/2021, and  https://github.com/gpuweb/gpuweb/pull/2113
```
        fn write_fragment_retry(index: i32, frag: Fragment) {
            let active_lock = &s_fragments.data[index].lock;
            var old_lock_value = atomicLoad(active_lock);
            var new_lock_value = old_lock_value + 1;
            loop {
                write_fragment_naive(index, frag);
                let res = atomicCompareExchangeWeak(active_lock, old_lock_value, new_lock_value);
                if (res.exchanged) {
                    break;
                }
            }
        }
```

And then there's the one that actually does work, thanks to [this SO comment](https://stackoverflow.com/q/11820066/2271927):
```
        fn write_fragment_spinlock2(index: i32, frag: Fragment) {
            let active_lock = &s_fragments.data[index].lock;  // per pixel
            var done = false;
            loop {
                if (done) { break; }
                let old = atomicExchange(active_lock, 1);
                if (old != 1) {
                    write_fragment_naive(index, frag);
                    atomicStore(active_lock, 0);
                    done = true;
                }
            }
        }
```

### DIFF
--- a/examples/geometry_cube.py
+++ b/examples/geometry_cube.py
@@ -12,7 +12,8 @@ from wgpu.gui.qt import WgpuCanvas
 app = QtWidgets.QApplication([])
 
 canvas = WgpuCanvas()
-renderer = gfx.renderers.WgpuRenderer(canvas)
+canvas._target_fps = 1000
+renderer = gfx.renderers.WgpuRenderer(canvas, show_fps=True)
 scene = gfx.Scene()
 
 im = imageio.imread("imageio:bricks.jpg")
--- a/examples/transparency1.py
+++ b/examples/transparency1.py
@@ -1,0 +1,49 @@
+"""
+Example showing transparency.
+"""
+import time
+import wgpu
+import pygfx as gfx
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
+app = QtWidgets.QApplication([])
+
+canvas = WgpuCanvas()
+renderer = gfx.renderers.WgpuRenderer(canvas)
+scene = gfx.Scene()
+
+geometry = gfx.PlaneGeometry(50, 50)
+plane1 = gfx.Mesh(geometry, gfx.MeshBasicMaterial(color=(1, 0, 0, 0.4)))
+plane2 = gfx.Mesh(geometry, gfx.MeshBasicMaterial(color=(0, 1, 0, 0.4)))
+plane3 = gfx.Mesh(geometry, gfx.MeshBasicMaterial(color=(0, 0, 1, 0.4)))
+plane4 = gfx.Mesh(
+    gfx.PlaneGeometry(100, 10), gfx.MeshBasicMaterial(color=(0, 1, 0, 0.4))
+)
+
+plane1.position.set(-10, -10, 1)
+plane2.position.set(0, 0, 2)
+plane3.position.set(10, 10, 3)
+
+scene.add(plane2, plane1, plane3, plane4)
+# scene.add(plane1, plane2, plane3, plane4)
+# scene.add(plane3, plane2, plane1)
+
+
+camera = gfx.OrthographicCamera(100, 100)
+
+
+def animate():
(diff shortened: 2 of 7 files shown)


almarklein:
Interesting. The current state is slower on Intel graphics (especially clearing the buffer to zeros is a problem). But on NVidia there appears to be no significant performance. Only minor tests, only on one example, so take with grains of salt.

almarklein:
I've cleaned this up just a bit so we can park this for now. It also looks like we need a newer version of Naga for using atomics.

A small teaser for now. The result differs a lot depending on what hardware it runs on (Intel vs NVidia). Depth is taken into account, but because there is no atomicity, this fails quite often, leading to an interesting result :)

oit1

Korijn:
Can you explain why atomicity is needed? I don't follow.

almarklein:
So if two fragments for the same pixel location are writing at (almost) the same time, the one that is written earlier gets overwritten by the (slightly) later fragment. Same for any checks etc. We need these to be checked one by one (for each pixel). There's a trick for this using a spinlock based on atomicExchange.

Korijn:
So how come this is not a problem in "regular" forward shading, but now it is? I don't see why that couldn't happen with the pre-OIT situation, I guess.

almarklein:
Because there the atomicity is part of the backed-in pipeline when the fragment is blended with the frame buffer and the corresponding depth is set. With the solution in this PR we need to do that ourselves.

almarklein:
Got it working with atomic writes and depth testing! Added notes in the top post for future reference. Now I can start on the fun part :)  (but first volume rendering)

almarklein:
I've made some progress on this POC:

image

Also some more insights:
* I thought that I could get away with a single render-pass in this approach. I was wrong`*` -> we need two render passes either way.
* The spinlock is not very stable; depending on what I do in the locked state, the GPU can hang (consistently).
* Clearing the rather big fragment-buffer (which is a storage buffer), seems very expensive on Intel Integrated Graphics for some reason.

I have a new plan, will make a new issue for that.

`*` my idea was to reset the buffer when an opaque fragment is encountered. However, semi-transparent fragments in front of the opaque fragment may have already been processed, and these may also have been combined with fragments *behind* the opaque fragment already.

almarklein:
Closing - superseded by #196

PR #110, opened 2021-08-24
Improve WorldObject

### LINKED ISSUE #109: Improvements to world object

almarklein:
Closes #109

* [x] Make `.visible` a documented prop.
* [x] Make `.parent` a read-only documented prop.
* [x] Add a few docstrings.
* [ ] Implement layers.

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -53,7 +53,7 @@ class WorldObject(ResourceContainer):
     """The base class for objects present in the "world", i.e. the scene graph.
 
     Each WorldObject has geometry to define it's data, and material to define
-    its apearance. The object itself is only responsible for defining object
+    its appearance. The object itself is only responsible for defining object
     hierarchies (parent / children) and its position and orientation in the world.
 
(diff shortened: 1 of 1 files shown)


pygfx/objects/_base.py
```diff
@@ -115,28 +120,57 @@ def __init__(self):
...
-        if obj in self._children:
-            obj.parent = None
-            self._children.remove(obj)
+    def remove(self, *objects):
+        """Removes object as child of this object. Any number of objects may be removed.
+        """
+        for obj in objects:
+            if obj in self._children:
+                obj._parent = None
+                self._children.remove(obj)
```
Korijn, line 172: Just a micro optimization, but, I think this would be faster:

```
try:
    self._children.remove(obj)
    obj._parent = None
except ValueError:
    pass. # or just let it raise?
```

It would avoid one list traversal :)

almarklein: done

almarklein:
I'm looking at an alternative way to implement the overlay-use-case that `.layer` was initially meant for. This PR is good to go from my end.

PR #111, opened 2021-08-24
First version of docs

almarklein:
* [x] Setup Sphinx docs.
* [ ] Setup RTD.
* [x] Reference docs fairly complete.
* [x] Made sure all classes have a docstring.
* [ ] Make sure all methods and props have docstrings. I think we should maybe do this in subsequent PR's as we're improving things.
* [ ] Guide. I could write some, but might be interesting to let someone else do it (easier to find unclarities and weird stuff).

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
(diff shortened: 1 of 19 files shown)


almarklein:
Ready for review. Also see https://pygfx.readthedocs.io/en/latest/

@Korijn Do you agree on turning the last two tasks into issues, to be dealt with later?

PR #117, opened 2021-08-26
Multiple canvases, subplots, and overlays

### LINKED ISSUE #112: Enable use-case of overlays
More generally, rendering to a specific region of the canvas (viewport), but also support for rendering multiple "layers" on the same canvas, possibly rendering one scene over another (without depth checks between them).

A related question, which seems orthogonal but is relevant since we're talking about ownership of render targets etc: can one scene be rendered to two canvases?

### LINKED ISSUE #116: Remove support for using no canvas
People can use a wgpu offscreen canvas instead.

almarklein:
Closes #112, closes #116. Depends on https://github.com/pygfx/wgpu-py/pull/181

This touches on some of the core mechanics of the renderer and how it interacts with the canvas. This affects having multiple canvases, subplots, overlays, sub-scene rendering, and post-processing.

* [x] Support for using multiple canvases in the same application, and rendering world-objects in multiple, because the device is shared/global.
* [x] Support for overlays because you can `renderer.render()` multiple times and set a new `clear` param.
* [x] Support for subplots because of a new `region` param in `renderer.render()` which is in logical pixels and is translated to a viewport in physical pixels.
* [x] Remove post-processing system (in favor of a single submit-step).
* [x] Fix issues with submit related to texture size and format.
* [x] Allow multiple renderers per target?
* [x] Rename any leftover 'submits" to "flush to target" or something similar.
* [ ] ~Maybe provide hooks for submit step in wgpu-py.~
* [x] Maybe rethink offscreen rendering (if we allow `WgpuRenderer(texture)`).
* [x] Resolve any todo's that I added along the way.
* [x] Update docs in guide a bit. #119

## How the renderer now works (high level POV)

The renderer's `.render(scene, camera)` function renders that scene to it's internal render target. This target is an implementation details (it may change e.g. with OIT). You can call `.render()` multiple times. The color and depth are automatically cleared before the first `.render()` call.

The final result is automatically "submitted" to the texture provided by the canvas. This submit-step includes any FSAA. One can also call `.submit_to_texture()` to submit the result to another texture instead.

## Approach per use-case

For each of these at least one example is added.

### Use multiple canvases

You can now use multiple canvases. Each canvas can have exactly one renderer associated with it. You can render one scene to multiple canvases.

image

### Blending scenes

By calling `.render()` multiple times, the results should just blend, as if they were in one scene. For opaque objects this is easy. Eventually this should work with semitransparent objects too (if/when we get OIT working). Not sure what specific use-cases would be, but anyway, this is how it works :)

image

### Subplots

By rendering multiple times, but to different region rects, you can implement subplots.

image

### Overlays

By rendering multiple scenes on top of each other and calling `.clear(color=False, depth=True)` in between, the last scene is always on top.

image

### A scene in a scene

The typical example of having a surface in your scene (a TV screen if you like) that depicts another scene can now be implemented by rendering the subscene via an offscreen canvas, then submitting the result into a texture that is rendered via a mesh in the outer scene.

sceneinscene

### Post-processing 

Similar to the above we can make the outer scene just a single quad that draws the inner scene and do something to it. But the canvas and renderer can be re-used.

An alternative is to provide hooks for the submit step to allow for simple post-processing mechanics. Also e.g. fog. Maybe for later.

Before this PR there is a system for post-processing steps, basically because the submit-step has been generalized. I plan to remove that to simplify things.

image


pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -180,10 +213,10 @@ def pixel_ratio(self, value):
                 f"Rendered.pixel_ratio expected None or number, not {value}"
             )
 
-    def render(self, scene: WorldObject, camera: Camera):
-        """Main render method, called from the canvas."""
+    def render(self, scene: WorldObject, camera: Camera, clear=True, region=None):
+        """Main render method."""
```
almarklein, line 217: Todo: document the arguments.

Also, consider the name of what's now called the "region" argument. It is a rect (tuple of 4 numbers), and closely related to the "viewport", except that viewport is in physical coordinates and this argument is expressed in logical coordinates. Maybe "view" or "rect"?

Korijn:
I am loving these examples!

Korijn:
So how does the "submit to texture" thing work? It's not entirely clear to me. It seems like you are writing to a texture anyway, it's just a different one. Why would you need to call that method `to_texture()` manually passing a texture, as opposed to linking it to an offscreen canvas or renderer, so that it's always rendering to that?

almarklein:
Right now, the renderer has its own texture that its rendered to, which may be larger than the eventual texture (for aa). But I also took into account that with OIT there may not be a simple texture containing the images, but a stack of textures or a buffer with a specific layout.

When the `renderer.render()` method has been called one or more times, it has a certain internal state (stored in a high-res texture or a buffer, I want that to be an implementation detail). This state can then be "submitted" to an actual texture. This will be done automatically with the canvas' texture at the end of the draw pass. but you can also manually call `to_texture()` (should be renamed `submit_to_texture()`) to put the result in another texture.

So in short, "submitting" draws the result of a renderer into a texture, applying SSAA if the source is larger.

BTW: I am open to alternatives to the term "submit". It would be confusing to call it "render" IMO. I also considered blit, resolve, paint, gather, compose ...

almarklein:
Added a gif for the subscene example.

Korijn:
Coming up with a good name is pretty hard.. Essentially you are still rendering. It's just the final output pass, right?

almarklein:
> Essentially you are still rendering.

Yes. but I fear that calling it rendering would confuse it with what happens in `.render()`. I like `.output()` too, but I'm also looking for a term to use to talk about it in docs etc. Maybe something will surface when I write such docs :)

almarklein:
> Can you explain why submit comes to mind for you?

As in submitting the result to a texture. I'm starting to dislike the term. Maybe ``finalize()``?

I agree this needs some thought. Back to the requirements:

* We should allow multiple calls to `.render` to "stack up" to support subplots and overlays.
* I think it would be good to have the internal state an implementation detail, for the sake of handling transparency later.
* Eventually the result should be in a texture (which is provided by the canvas).
* It should also be possible to select another texture (for subscenes / custom post-processing).

The current docs of the `Renderer`:

    Provides a ``.render()`` method that can be called one or more times
    to render scene. This creates a visual representation that is stored
    internally. The final result is then rendered into a texture for display.
                                  __________
                                 |          |
        [scenes] -- render() --> | renderer | -- finalize() --> [texture]
                                 |__________|

    The internal visual representation includes things like a depth
    buffer and is typically at a higher resolution to reduce aliasing
    effects. Further, the representation may in the future accomodate
    for proper blending of semitransparent objects.

    The finalize step renders the internal representation into a texture,
    applying anti-aliasing. In the future this is also where fog is applied
    and perhaps any custom post-processing effects.

    Normally, the finalize step is applied automatically at the end of
    a draw pass to render into the screen-texture of the canvas.
    However, you can also call ``.submit_to_texture()`` to render to a
    texture of your choice.

Korijn:
Can we give this internal texture that holds the final output a name? E.g. the "render buffer"? Then we can refer to it clearly throughout docs and code. It would also help determine a clear naming scheme for our API.

`render()`  -> renders to internal buffer
`buffer_to_texture(tex)` -> renders buffer to texture

almarklein:
I looked into using an offscreen canvas' texture as we discussed. One problem with that approach is that its complicated (API-wise) to wrap the raw wgpu texture object of the canvas into a `pygfx.Texture`. So I came up with a slightly different approach:

* The canvas accepts either a canvas or a texture (view), as it's target to render to. 
* I got rid of the automatic clearing and "submitting" that was done in the draw pass invoked by a canvas.
* Instead, I made that the renderer is a context manager, and `.render()` can only be called inside such a context. Entering the context schedules a clear, and exiting the context submits!

For example, for the overlay example:
```py
    with renderer:
        renderer.render(scene1, camera1)
        renderer.clear(color=False, depth=True)
        renderer.render(scene2, camera2)
```
For the scene_in_scene example:
```py
    renderer1 = gfx.renderers.WgpuRenderer(texture)
    renderer2 = gfx.renderers.WgpuRenderer(canvas2)
    ...
    with renderer1:
        renderer1.render(scene1, camera1)
    with renderer2:
        renderer2.render(scene2, camera2)
``` 

The only downside is that we need to add `with renderer` to all our examples.

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -180,10 +233,31 @@ def pixel_ratio(self, value):
...
-
-        device = self._device
+    def __enter__(self):
+        self.clear()
+        self._context_is_active = True
+        return self
+
+    def __exit__(self, *args):
+        self._context_is_active = False
+        self._render_to_target()
```
almarklein, line 243: This may help understand.

Korijn:
Is it _required_ to use the with statement? I like how it cleans up a little bit, but I want to make sure people who want to integrate pygfx into a larger codebase can decide to make the method calls manually. E.g. because they need to be called in different parts of the application.

Another question: can you still render to an offscreen texture/canvas _after_ submitting?

almarklein:
> Is it required to use the with statement?

Yes, the `.render()` will raise an error if called outside a context. This makes sure that the renderer is always used "correctly".

I was a bit hesitant about the use of a with-statement. Maybe I still am a bit. But it seems to fit very well here: it forces a user to write code that always submits to the target, while abstracting away the submit step itself.

I also thought of having explicit calls, but 1) we wanted to abstract the "submit" step away, right? and 2) it would be easy to forget a step, and I'm afraid we'd get tons of questions asking why a user's code does not show anything :P

> but I want to make sure people who want to integrate pygfx into a larger codebase can decide to make the method calls manually.

I think one might argue that in larger codebase with render calls being made in different places, a with-statement is particularly convenient to ensure that the submit step is actually performed ;)

> can you still render to an offscreen texture/canvas after submitting? (I can't think of a use case but I'm just curious)

I'm not sure whether I understand what you mean here. You can rendering to it again, of course, but each submit step completely flushes the target.

Korijn:
> we wanted to abstract the "submit" step away, right?

Yes.

> it would be easy to forget a step, and I'm afraid we'd get tons of questions asking why a user's code does not show anything :P

Fair enough.

So I guess this all comes down to taste/preference, but I think the with statement looks weird. In my mind it immediately makes me think there's some kind of IO management that needs to be handled at risk of memory leakage. But that's not really the case here. We're just using it to _hide required method calls_.

So I thought of another alternative, you could provide settings `auto_clear` and `auto_flush` on the renderer. To guide new users to simple, sane defaults, they can be `True` by default. But it will allow power users to back out and handle the whole process themselves, and avoid the `with` statement:

```python
renderer = WgpuRenderer(auto_clear=False, auto_flush=False)
renderer.clear()  # needed because auto_clear=False
renderer.render(scene1, camera1)
renderer.clear(color=False, depth=True)
renderer.render(scene2, camera2)
renderer._render_to_target()  # needed because auto_flush=False
```

Korijn:
Looks great, I think we reached a clear, intuitive terminology and API!

examples/scene_side_by_side.py
```diff
@@ -49,11 +49,9 @@ def animate():
...
 
-    size = canvas.get_logical_size()
-    renderer.render(scene1, camera1, flush=False, region=(0, 0, size[0] / 2, size[1]))
-    renderer.render(
-        scene2, camera2, flush=False, region=(size[0] / 2, 0, size[0] / 2, size[1])
-    )
+    w, h = canvas.get_logical_size()
+    renderer.render(scene1, camera1, flush=False, region=(0, 0, w / 2, h))
+    renderer.render(scene2, camera2, flush=False, region=(w / 2, 0, w / 2, h))
     renderer.flush()
```
Korijn, line 55: why are you calling flush manually here? :)

almarklein: Then it's more of a pattern of rendering N subplots and flush at the end. Lemme make it 4 subplots so that its a bit more obvious.

Korijn: Right, so it's just one way to do it that makes sense stylistically for this example, and it works. :)

almarklein:
I'm going to have a look at whether multiple renderers can render to the same canvas. This is now restricted and I don't remember why. Maybe it was because of the hooks we used earlier. 

In the mean time this is ready for review.

examples/scene_side_by_side.py
almarklein: Exactly. I thought about making flush private, but I think I might like this style too.

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -140,20 +141,26 @@ class WgpuRenderer(Renderer):
...
 
+    Parameters:
+        target (WgpuCanvas or Texture): The target to render to, and what
+            determines the size of the render buffer.
+        pixel_ratio (float, optional): How large the physical size of the render
+            buffer is in relation to the target's physical size, for antialiasing.
+            See the corresponding property for details.
+        show_fps (bool): Whether to display the frames per second. Note that this
+            measure only makes sense if you're rendering as fast as you can and
+            if the canvas does not impose a frame rate limit.
```
Korijn, line 152: Well it also makes sense to see if you are dropping below such a limit.

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -180,10 +255,35 @@ def pixel_ratio(self, value):
...
+    ):
+        """Render a scene with the specified camera as the viewpoint.
+
+        Parameters:
+            scene (WorldObject): The scene to render, a WorldObject that
+                optionally has child objects.
+            camera (Camera): The camera object to use, which defines the
+                viewpoint and view transform.
+            region (tuple, optional): The rectangular region to draw into,
+                like the viewport, but expressed in logical coordinates.
```
almarklein, line 276: @korijn there was an open comment that I revive here: is "region" a good name for this? I want to avoid "viewport" to differentiate it from that, because this is in logical pixels. Alternatives could be "view" or "rect". Thoughts?

pygfx/renderers/wgpu/_wgpurenderer.py
```diff
@@ -140,20 +141,26 @@ class WgpuRenderer(Renderer):
     also where fog is applied, as well as any custom post-processing
     effects.
 
+    Parameters:
+        target (WgpuCanvas or Texture): The target to render to, and what
+            determines the size of the render buffer.
```
Korijn, line 146: I really like that you can just _literally_ render to a texture!

pygfx/renderers/wgpu/_wgpurenderer.py
almarklein: True. Will word it a bit differently.

pygfx/renderers/wgpu/_wgpurenderer.py
Korijn: Viewport is the traditional term, in opengl, threejs, and when I google it viewport is also the most common term. Let's just call it "viewport"?

I do wonder if people will go looking for a scissor test as well. Is that a thing in webgpu?

almarklein: > I don't really understand why you want to avoid the term because of the pixel type

To avoid confusion, but I guess you're right and we should just call it viewport.

> I do wonder if people will go looking for a scissor test as well. Is that a thing in webgpu?

Yes. Can add that too.

Korijn: Unless you see a good reason for it you can leave out the scissor test

Korijn:
LGTM!

I've done a thorough code review but the only "issues" I can see are "naming things". :) Let's save that for later.

almarklein:
I renamed region to viewport and left the scissors for when it's needed. Any other renames you were thinking about? I'd rather resolve these now, or list them in an issue, otherwise we'll probably never do it ;)

Korijn:
Things that I find a little surprising:

* canvas.draw returns a numpy array
* renderer.snapshot also returns a numpy array
* where is renderer.clear()

almarklein:
> canvas.draw returns a numpy array

Good point. I'll have a look (will mean a change in wgpu-py though).

> renderer.snapshot also returns a numpy array

I think it makes sense to be able to do a snapshot from the rendered state also when not using an offscreen canvas. But I understand how now it looks unclear what one should use ...

> where is renderer.clear()?

It is removed in favor of `.render(..., clear_color=True, clear_depth=True)`. The clearing is applies inside the render method, not when the (now removed) `clear()` method is called. That meant that there was internal state that affected how the next render call would be applied. Just adding them as args to `render()` makes things much simpler, both in the implementation and the API.

almarklein:
Arg, I added some more comments about the difference between `offscreen_canvas.draw()` and `renderer.snapshot()`. However, I accidentally to to my other PR :P  - anyway, I created #122 so we can resolve this properly in a new PR.

almarklein:
So if you're happy about clearing happening via args to `.render()` instead of `.clear()`, this should be good to go! (and otherwise we should discuss that)

Korijn:
Yes, thanks, now let's merge!

almarklein:
Woppa!

Issue #47, opened 2020-07-08
Improvements to base materials


almarklein:
Generic parameters:

* [x] opacity - 1.0-0.0 float, multiplied with alpha texture map if set. #121
* [x] clipping planes (requires check in every material's fragment shader) #124
* [ ] stencil buffer options
* [ ] alpha testing

almarklein:
@Korijn do we consider stencil-buffer and alpha-testing internal details (not pubic API) for the time being? If so, we can close this if #131 is merged.

Korijn:
Yes, we do consider them internal. We'll revisit the capping example from threejs later

PR #121, opened 2021-09-08
Material.opacity

### LINKED ISSUE #120: The meshslice does not honor depth test
The mesh slice visual draws itself behind other objects, so it looks like the depth is not set correctly, or the test is bypassed. This is in current main, so not due to changes in #117.

almarklein:
Based on #117, ~will rebase hen that is merged.~ done. Ticks one item off #47.

* [x] Implement opacity property on `Material`.
* [x] Refactor how materials classes pass input args to their supers.`*`
* [x] Add a test to ensure that the material uniforms include all fields from their supers.
* [x] Implement opacity for all material fragment shaders.
* [x] Also closes #120 (mesh slice depth issue).

`*` Mainly, make the base material class create the uniform buffer, and use kwargs in the `__init__` to set default values for properties.

### DIFF
--- a/examples/mesh_slice.py
+++ b/examples/mesh_slice.py
@@ -22,7 +22,6 @@ obj2 = gfx.Mesh(geometry, material2)
 scene.add(obj1)
 scene.add(obj2)
 
-
 camera = gfx.PerspectiveCamera(70, 2)
 camera.position.z = 4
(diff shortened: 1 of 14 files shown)


Korijn:
LGTM!! 😎

PR #124, opened 2021-09-10
Implement Material.clipping_planes

### LINKED ISSUE #84: Clipping planes
* Implementation for volumes in vispy: https://github.com/vispy/vispy/pull/2116
* Implementation for multiple visuals in vispy: https://github.com/vispy/vispy/pull/2197

A snippet of a discussion about how to define a plane in 3D (https://github.com/vispy/vispy/pull/2074):

>> We could define the plane as a normal plus a position, or as a plane with abcd, which is basically the normal plus a scalar value to shift away from the origin.
>
> Interesting - I didn't think about that. Longer term I have some other things I would like to do with the volume visual where having a position (without a normal) would be useful (localised rendering around a point) - for that reason I would prefer to keep the variables separate but again, happy to accommodate changes 🙂

almarklein:
Closes #84. Implements one point in #47.

* [x] Implement basics.
* [x] Implement uniform array to support multiple planes.
* [x] Implement for mesh.
* [x] Implement for line.
* [x] Implement for points.
* [x] Implement for volume.
* [x] Implement intersection (instead of union).
* [x] Implement global planes on the renderer or scene?

clipping1

### DIFF
--- a/pygfx/materials/_background.py
+++ b/pygfx/materials/_background.py
@@ -11,6 +11,7 @@ class BackgroundMaterial(Material):
     """
 
     uniform_type = dict(
+        clipping_planes=("float32", (0, 1, 4)),  # array<vec4<f32>,3>
         color_bottom_left=("float32", 4),
         color_bottom_right=("float32", 4),
         color_top_left=("float32", 4),
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -10,7 +10,7 @@ class Material(ResourceContainer):
 
     uniform_type = dict(
         opacity=("float32",),
-        clipping_planes=("float32", 3, 1, 4),  # array<vec4<f32>,3>
+        clipping_planes=("float32", (0, 1, 4)),  # array<vec4<f32>,3>
     )
 
     def __init__(self, *, opacity=1):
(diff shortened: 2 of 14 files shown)


Korijn:
Do we need this to be implemented on non-mesh objects?

almarklein:
> Do we need this to be implemented on non-mesh objects?

I think it makes sense to do so.

Korijn:
OK! I've never thought about it, but I guess there are uses for it.

almarklein:
Volume (3d renderings) at the very least! :)

almarklein:
We now have uniforms with fields that can be arrays, and these arrays can be resized, and thanks to the `.rev` bumping we already had in place, this automatically triggers a pipeline rebuild / recompile for that object.

That way we can support a variable amount of clipping planes. And this will come in handy in other places too, e.g. lights.

almarklein:
This is ready for review. Two remarks:

----

Would you expect a line to be clipped on a per-fragment basis, or based on the fragment-location of the centerline? As in, a line is infinitely thin, but made thicker for visualization, would it be ok for it to "bloom" over the clipped edge? See image:

image

Background: at the moment, the line is clipped per-fragment, but points are clipped based on their (center) location. For points this is mostly because how points are implemented, using the pointsize feature. I think we may need to change this to billboarding anyway, because IIRC the pointsize feature is sort of phases out.

----

The way clipping is implemented expects the camera transform to be linear. Now that I am writing this down I realize that this already breaks down for a perspective transform, haha.  Ok, maybe not quite *that* ready for review :D

Korijn:
I guess I would expect it to be clipped per fragment.

I was actually wondering about the same thing in the event that a line or point is just off screen in terms of center coordinates, but not in terms of thickness:

image

almarklein:
I think I'm leaning towards per-fragment as well. Otherwise it may feel weird that everything is neatly clipped in s straight line, except for lines/points sticking out. Ok this means that the points would need to be changed to fix that, which I created #126 for.

PR #125, opened 2021-09-13
Invisible objects also hide their children.

### LINKED ISSUE #123: WorldObject.visible behavior
It looks (to me) like properties on a Material apply to the objects that the material applies to, whereas properties of a WorldObject apply to that object and is children (and their children). This is at least the case for transforms. Should this apply to `.visible` too? And does it now?

almarklein:
Closes  #123

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -176,18 +176,19 @@ class WorldObject(ResourceContainer):
                 pass
         return self
 
-    def traverse(self, callback, include_non_visible=False):
+    def traverse(self, callback, skip_invisible=False):
         """Executes the callback on this object and all descendants.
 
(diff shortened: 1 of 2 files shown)


pygfx/objects/_base.py
```diff
@@ -176,10 +176,18 @@ def remove(self, *objects):
                 pass
         return self
 
-    def traverse(self, callback):
-        callback(self)
-        for child in self.children:
-            child.traverse(callback)
+    def traverse(self, callback, include_non_visible=False):
```
Korijn, line 179: I expect that this function will be used to implement update logic in game loops and animations. So it seems like a surprise/gotcha/bad idea to filter out invisible things by default.

Can you flip that around?

Issue #50, opened 2020-07-08
Improvements to mesh material


almarklein:
* [ ] Phong shading
* [ ] PBR shading
* [ ] Shading properties settable on the material
* [ ] Wireframe option / material
* [ ] Support texture coordinates that are 1D, 2D and 3D.
*

almarklein:
About culling: Three has a `side` property to define whether front, back or both are visible. I find "side" may be rather generic, so I have no problem calling this culling or something else.

I seem unable to find how Three defines whether CW or CCW is front facing. Am I missing something or does Three simply assume CCW?

almarklein:
All open points are covered by other issues. Closing.

PR #127, opened 2021-09-14
Implement culling for mesh material

### LINKED ISSUE #105: Flipping normals towards viewer in fragment shader
The faceForward trick that we use to flip normals in the mesh shader only works if the light comes from (about) the same direction as the eye vector. Otherwise, normals on faces that that are near-parallel to the eye vector can flip, affecting how they're lit, causing bright and dark artifacts at the edges of objects.

This means we have to make sure that all normals are correct in the first place. Not sure what to do with surfaces that can be seen from two sides, such as planes and e.g. the Klein bottle.

I also wonder if I might be overlooking something that still allows flipping normals ...

almarklein:
Implements a part of #50. Closes #105.

* [x] Implements `MeshMaterial.side`.
* [x] Decide on what to do with a winding property vs assuming CCW.
* [x] Fixes lighting to use `fron_facing` builtin.
* [x] Check the winding of all geometries.
* [x] Fixes winding of torus know geometry.
* [x] Correct front_facing (for lighting) when the camera has an axis flipped.
* [x] Correct culling when the camera has an axis flipped?

culling1

### DIFF
--- a/examples/validate_culling.py
+++ b/examples/validate_culling.py
@@ -7,7 +7,6 @@ Example test to validate winding and culling.
 
 """
 
-import numpy as np
 import pygfx as gfx
 from PyQt5 import QtWidgets
 from wgpu.gui.qt import WgpuCanvas
--- a/examples/validate_culling.py
+++ b/examples/validate_culling.py
@@ -1,66 +1,59 @@
 """
 Example test to validate winding and culling.
 
-* The red knot should look normal and well lit.
-* The green know should show the backfaces, well lit.
-* The purple and cyan knots are not lit.
+* The top red knot should look normal and well lit.
+* The top green know should show the backfaces, well lit.
+* The bottom row shows the same, but the camera looks backwards.
 
 """
 
+import numpy as np
 import pygfx as gfx
-
 from PyQt5 import QtWidgets
 from wgpu.gui.qt import WgpuCanvas
 
 
 app = QtWidgets.QApplication([])
 
-canvas = WgpuCanvas()
+canvas = WgpuCanvas(size=(600, 600))
 renderer = gfx.renderers.WgpuRenderer(canvas)
 scene = gfx.Scene()
 
 # geometry = gfx.BoxGeometry(1, 1, 1)
 geometry = gfx.TorusKnotGeometry(1, 0.3, 64, 10)
 
-# Top left
+# Create red know shown normally
 material1 = gfx.MeshPhongMaterial(color=(1, 0, 0, 1))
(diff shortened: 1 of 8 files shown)


almarklein:
While test-driving this, I wonder if we need the winding property. Since if the winding is the other way around, you can always set the `side` to "BACK". Or perhaps we can combine them and let the options for side be "BOTH", "CW", "CCW".

Korijn:
You're probably right. Let's just drop it.

Also, if you want to rename side to cull, that's fine with me. Both make sense to me.

almarklein:
No, I was wrong. The should be two separate things, because the winding is used to determine whether a face is front-facing or back-facing, which is used in e.g. lighting.

One thing however: I think the winding property should be on the geometry, because it says something how the geometry is composed, what do you think?

almarklein:
Accidental art:

art1

almarklein:
Or ... we do what Three seems to do and assume CCW. It's easy enough to do `np.fliplr` on your `geomerty.indices` if needed ...

Korijn:
> Or ... we do what Three seems to do and assume CCW. It's easy enough to do `np.fliplr` on your `geomerty.indices` if needed ...

Well.. changing a prop on a material is much cheaper than adjusting geometry buffers right?

My main concerns are:

1. I can't think of any use cases
2. It's additional complexity that needs to be supported everywhere

almarklein:
> changing a prop on a material is much cheaper than adjusting geometry buffers right?

Yes, but you could consider this part of your "data-loading process": before creating a geometry, make sure that the faces are CCW.

Anyway, if we do make it a prop, I think it should be on the geometry, since it;s sort of metadata of the geometry data.

> My main concerns are

You mean of having a prop for setting the winding?

almarklein:
Another issue: when the camera flips an axis (e.g. to show an image the right way up), the winding is reversed. I can fix this relatively easily for the lighting (flipping the front_facing flag in the shader when needed), but for the culling there is not much we can do. 

(Well, technically we could flip the winding passed in the pipeline, but that means that object-pipelines depend on the camera, which we should probably not do.)

My inclination is to just document in the `side` property, even though its somewhat weird if you don't know the details...

Korijn:
OK, let's just decide here and now that we will assume CCW and document this decision. Users will just have to prep their data accordingly.

almarklein:
Yeah. A prop can be added later if we later think its needed.

almarklein:
Result of the new example:
image

almarklein:
Ready!

Korijn:
LGTM but there are CI failures

PR #128, opened 2021-09-15
Implement flat shading

almarklein:
Flat shading is nice if you want lighting but also want to see your triangle. Or if you want a retro look.

image

### DIFF
--- a/pygfx/renderers/wgpu/meshrender.py
+++ b/pygfx/renderers/wgpu/meshrender.py
@@ -5,9 +5,10 @@ from ._shadercomposer import BaseShader
 from ...objects import Mesh, InstancedMesh
 from ...materials import (
     MeshBasicMaterial,
+    MeshPhongMaterial,
+    MeshFlatMaterial,
     MeshNormalMaterial,
     MeshNormalLinesMaterial,
-    MeshPhongMaterial,
     MeshSliceMaterial,
 )
 from ...resources import Buffer, Texture, TextureView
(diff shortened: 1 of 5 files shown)


pygfx/renderers/wgpu/meshrender.py
```diff
@@ -488,6 +491,29 @@ def helpers(self):
...
+
+            let u = dpdx(world_pos);
+            let v = dpdy(world_pos);
+            var normal = normalize(cross(u, v));
+
+            // The normal calculated above may not be oriented correctly.
+            // We have two flags: is_front and u_stdinfo.flipped_winding.
+            // Note that lighting_phong() also applies the is_front flag.
+            // Below code means: flip the normal if the XOR of these two flags is true.
+            normal = select(normal, -normal, (select(0, 1, is_front) + u_stdinfo.flipped_winding) == 1);
```
almarklein, line 511: I don't have a good way to explain this, but I tried this on the validate_culling.py example (using the flat material instead), and this is what I found makes it work :)

almarklein:
Mmm, in ThreeJS the MeshPhongMaterial has a `.flatShading` property. It feels like a different material to me, what do you think @Korijn ?

Korijn:
LGTM!

> Or if you want a retro look

It can look absolute gorgeous. You can also makes things look out of place/jarring by having a flat shaded object in an otherwise phong shaded scene.

Korijn:
> Mmm, in ThreeJS the MeshPhongMaterial has a `.flatShading` property. It feels like a different material to me, what do you think @Korijn ?

I don't think it matters much :) I'm fine with having it be a separate material.

PR #130, opened 2021-09-16
Minor improvement to mesh slice

almarklein:
The mesh slice example always showed a jaggy line for the mesh contour. This is because the mesh slice and the mesh itself are z-fighting. By giving the mesh slice a tiny weeny offset, this is avoided/reduced.

Also changed the code a bit to just work with `z`, because `w` is already divided by so can be set to 1.0.


Korijn:
LGTM!

PR #132, opened 2021-09-17
Fix doc code examples

almarklein:


almarklein:
So if you don't have the space, Sphinx decided to simply not show the sample, without displaying a warning ...

PR #134, opened 2021-09-17
Fix RTD

almarklein:
The previous PR's did not help. Setting RTD to build from this branch instead so I can fiddle without creating tons of PR's.

Todo:
* [x] Fix RTD
* [x] Set RTD to build `main` branch instead of `docs`.

### DIFF
--- a/docs/docs_requirements.txt
+++ b/docs/docs_requirements.txt
@@ -1,4 +1,5 @@
 sphinx
+pygments
 numpy
 wgpu
 jinja2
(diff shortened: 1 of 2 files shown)


almarklein:
No idea what's going on here. RTD builds the docs, but when I view them they don't include the changes I made.

image

So maybe RTD is glitching or some kind of cache issue? Will look again later.

almarklein:
It's fixed now. This PR itself is not necessary.

For the future: you can open the source of the page of the page at RTF and find the date and commit for which that page was build.

PR #136, opened 2021-09-17
Implement MeshBasicMaterial.wireframe 

almarklein:
This implements mesh wireframes, as a property on `MeshBasicMaterial`.

### Background

There are many ways that this can be done, e.g.:

1. Generate line geometry from the vertices and indices and render that using a Line object. I'd rather not do that.
2. Render as line primitive-topology (i.e. 1px lines), using some geometry-shader-like trickery in the vertex shader.
3. Render pretty lines with a given thickness, similar to how we render our LineMaterial, but with mesh geometry as input.
4. Render the mesh as you'd normally do, but discard fragments that are too far from the edge.

I *think* that ThreeJS uses a variant of (2) (because it mentions "Due to limitations of the [...] WebGL renderer on most platforms linewidth will always be 1 regardless of the set value."). But Three also has a nice [example](https://threejs.org/examples/?q=wireframe#webgl_materials_wireframe) with a custom wireframe, that seems to do something like (4).

### Considerations

This PR implements option 4, because it has several interesting advantages:
* Most of the shader code for the mesh is the same (easy to maintain).
* Many features of the mesh persist: Any lighting works as expected. It works with instanced meshes. Applying a texture still works. Culling works, Picking works, etc. 

Some disadvantages:
* Option (2) is most probably faster.
* At the edges of the object, the width of the lines can be thinner, because lines never extend past the actual object. You can see this effect in the ThreeJS example I linked (this is what tipped me of how it works).

Some implementation details: the way this works is a bit similar to how we support sub-face picking, using barycentric coodinates. Except now the barycentric coordinates are expressed as a distance in logical pixels in screen-space. In the fragment shader we can then easily check when a fragment must be discarded.

### Todo

* [x] Implement it.
* [ ] Make sure we're happy with the API.
* [x] Add an example showing a lit mesh and a wireframe together.
* [ ] Add an example showing a wireframe that has different materials for front and back. 

### Mandatory screenshots

A new example:

wireframe2

Interesting look when using tick lines and combined with flat shading:

wireframe3

### DIFF
--- a/pygfx/materials/_mesh.py
+++ b/pygfx/materials/_mesh.py
@@ -11,6 +11,7 @@ class MeshBasicMaterial(Material):
         clipping_planes=("float32", (0, 1, 4)),  # array<vec4<f32>,N>
         clim=("float32", 2),
         opacity=("float32",),
+        wireframe=("float32",),
     )
 
     def __init__(
@@ -18,6 +19,7 @@ class MeshBasicMaterial(Material):
         color=(1, 1, 1, 1),
         clim=(0, 1),
         map=None,
+        wireframe=0,
         side="BOTH",
         **kwargs,
     ):
@@ -26,6 +28,7 @@ class MeshBasicMaterial(Material):
         self.color = color
         self.clim = clim
         self.map = map
+        self.wireframe = wireframe
         self.side = side
 
     def _wgpu_get_pick_info(self, pick_value):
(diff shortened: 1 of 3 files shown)


pygfx/materials/_mesh.py
```diff
@@ -89,6 +92,25 @@ def side(self, value):
             raise ValueError(f"Unexpected side: '{value}'")
         self._bump_rev()
 
+    @property
+    def wireframe(self):
+        """Render geometry as a wireframe with the given thickness.
+        Default is 0 (i.e. render as polygons).
+        """
+        return float(self.uniform_buffer.data["wireframe"])
```
almarklein, line 100: Or should we have one prop to turn it on/off and one to define the thickness?

Korijn:
Why don't you just use PolygonMode ? It should be natively supported right?

Korijn:
https://github.com/gfx-rs/wgpu/blob/af960b91f42585c337aae1c0f69ca805122395c5/wgpu-types/src/lib.rs#L393

It says that it's literally meant for wireframe and native only!

almarklein:
It's not a feature that is exposed in WebGPU / wgpu-native. There is a mentioning in wgpu-native [here](https://github.com/gfx-rs/wgpu-native/blob/dba4399607976dada643368afb492e2b5279536a/src/device.rs#L488-L504) though.

Not sure if there's a chance that this will become part of the API. And if it would, I'm pretty sure it would be limited to 1px lines - which arguably covers most cases.

I also don't think it would matter much for performance - with the MeshBasicMaterial without using a texture, most code paths are removed, leaving a relatively short shader.

pygfx/materials/_mesh.py
Korijn: I do think a boolean flag is more natural. I doubt users will want to tweak the thickness very often, so offering that separately makes sense to me.

Korijn:
OK, makes sense! Nice work :)

pygfx/materials/_mesh.py
almarklein: done

Korijn:
LGTM!

PR #140, opened 2021-09-21
Show utility

Korijn:
This adds `gfx.show(scene, animate=None, backend=None, interactive=True)` which will attempt to show the scene using any available GUI layer, creating its own camera, controls, and event loop.

See the added `examples\show_scene.py` for a demonstration, and the `pygfx\show` subpackage for the code.

Note: it's not particularly clean and only supports Qt for now, but it works. It would be great if this also works in jupyter.

The main purpose is to just provide a very quick way to "get something on screen". This idea can be taken further, for example:

* provide a filename instead of a scene and the scene is also created automatically (simple model viewer)
* add picking and display properties on screen about clicked elements
* etc

I also switched all examples over to PySide6 since PyQt5 is superseded.

### DIFF
--- a/pygfx/show/__init__.py
+++ b/pygfx/show/__init__.py
@@ -1,0 +1,13 @@
+"""
+Quickly visualize scenes without any boilerplate.
+"""
+from .backends import get_backend
+from .loop import run
+
+
+__all__ = ("show",)
+
+
+def show(scene, animate=None, backend=None, interactive=True):
+    backend, canvas_cls, extras = get_backend(name=backend)
+    run(backend, canvas_cls, extras, scene, animate=animate, interactive=interactive)
--- a/pygfx/show/backends.py
+++ b/pygfx/show/backends.py
@@ -1,0 +1,27 @@
+import importlib
+
+
+BACKENDS_QT = ("PySide6", "PyQt6", "PySide2", "PyQt5", "PySide", "PyQt4")
+BACKENDS = ("qt", "wx", "glfw", "jupyter")
+
+
+def get_backend(name=None):
+    """Automatically detect which GUI layer and canvas can be used"""
+    backends = (name,) if name else BACKENDS
+    extras = None
+    for backend in backends:
+        try:
+            if backend == "qt":
+                for backend_qt in BACKENDS_QT:
+                    try:
+                        importlib.import_module(f"{backend_qt}.QtCore")
+                    except ImportError:
+                        pass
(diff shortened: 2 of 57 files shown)


pygfx/show/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+Quickly visualize scenes without any boilerplate.
+"""
+from .backends import get_backend
+from .loop import run
+
+
+__all__ = ("show",)
+
+
+def show(scene, animate=None, backend=None, interactive=True):
```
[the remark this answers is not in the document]
almarklein: Yes please :)

pygfx/show/backends.py
```diff
@@ -0,0 +1,27 @@
+import importlib
+
+
+BACKENDS_QT = ("PySide6", "PyQt6", "PySide2", "PyQt5", "PySide", "PyQt4")
+BACKENDS = ("qt", "wx", "glfw", "jupyter")
```
almarklein, line 5: Nitpick: no braces needed :)

pygfx/show/loop.py
```diff
@@ -0,0 +1,47 @@
...
+def run(backend, canvas_cls, extras, scene, animate=None, interactive=True):
+    """
+    Render a given scene and optional animate callback in the given backend GUI layer
+    """
+    if backend == "qt":
+        QtWidgets, _ = extras  # noqa: N806
+        app = QtWidgets.QApplication([])
+
+    camera = PerspectiveCamera(70, 16 / 9)
+    camera.position.z = 500
```
almarklein, line 20: These numbers are arbitrary ...

almarklein:
This is super useful, but also quite a lot, and I think we should think carefully about how/where to do this. E.g. I wonder if some of this should be in wgpu-py?

pygfx/show/loop.py
almarklein: I think the camera should also be an argument to `show()`.

almarklein: or ... the renderer/scene could detect the bounding box of the total scene and place the camera such that the whole scene is visible.

pygfx/show/backends.py
ivoflipse: I'd prefer to be explicit that this is a tuple of possible backends, that the braces are optional is just syntatic sugar

Korijn:
Superseded by #226

PR #142, opened 2021-09-23
Fix compilation error follow up

berendkleinhaneveld:
This is a follow up on #139 to make sure (most) examples now also work on Metal, by renaming some variables. We should probably follow up on Almar's suggestion that this could be a bug in Naga's shader generation.

Bonus: added scikit-image to list of example dependencies

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -11,3 +11,4 @@ twine
 PyQt5
 imageio
 imageio-ffmpeg
+scikit-image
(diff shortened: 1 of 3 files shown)


almarklein (approved):

almarklein:
Ik heb een taak in mijn persoonlijke todolist aangemaakt om, wanneer we de nieuwste naga hebben, dit nog eens te checken ...

almarklein:
Reported upstream: https://github.com/gfx-rs/naga/issues/1464

PR #144, opened 2021-09-28
Sphere geometry

Korijn:
Adds:

```python
SphereGeometry(
    radius=1,
    width_segments=32,
    height_segments=16,
    phi_start=0,
    phi_length=np.pi * 2,
    theta_start=0,
    theta_length=np.pi,
):
```

image

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -9,5 +9,6 @@ twine
 
 # examples
 PyQt5
+PySide6
 imageio
 imageio-ffmpeg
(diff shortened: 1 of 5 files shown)


almarklein:
Related: we should also add a sphere that draws like a quad and does the "geometry" in the frag shader, like https://github.com/vispy/vispy/pull/2209

almarklein:
Have you tried setting `material.side = "front"` to make sure that the winding is correct?

Korijn:
Good call, I had to flip them! Looks good now with side=front:

image

PR #146, opened 2021-09-30
Shader annotate uniform

almarklein:
This change is required with the new (upcoming) Naga (and already works with what we currently use).


Korijn:
LGTM!

PR #147, opened 2021-09-30
Clipping planes intersect option

### LINKED ISSUE #131: Clipping planes intersection
~Also, consider calculating the signed distance to the plane in the vertex shader.~ Nope, can't do that.

almarklein:
Adds `Material.clip_intersection`. Builds further on #124. Closes #131. I also looked into doing the clip logic in the vertex shader, but found out that we can't (see https://github.com/vispy/vispy/issues/2225).

Also adds a clipping planes example:

image

### DIFF
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -13,7 +13,7 @@ class Material(ResourceContainer):
         clipping_planes=("float32", (0, 1, 4)),  # array<vec4<f32>,3>
     )
 
-    def __init__(self, *, opacity=1, clipping_planes=None, clip_intersection=False):
+    def __init__(self, *, opacity=1, clipping_planes=None, clipping_mode="any"):
         super().__init__()
 
         self.uniform_buffer = Buffer(
@@ -22,7 +22,7 @@ class Material(ResourceContainer):
 
         self.opacity = opacity
         self.clipping_planes = clipping_planes or []
-        self.clip_intersection = clip_intersection
+        self.clipping_mode = clipping_mode
 
     def _set_size_of_uniform_array(self, key, new_length):
         """Resize the given array field in the uniform struct if the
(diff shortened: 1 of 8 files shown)


pygfx/materials/_base.py
```diff
@@ -116,4 +118,17 @@ def clipping_planes(self, planes):
             self.uniform_buffer.data["clipping_planes"][i] = planes2[i]
         self.uniform_buffer.update_range(0, 1)
 
-    # todo: clip_intersection
+    @property
+    def clip_intersection(self):
+        """Changes the behavior of clipping planes so that only their
+        intersection is clipped, rather than their union. Default is false.
```
Korijn, line 124: Isn't it the other way around? Any matches union right?

Korijn:
LGTM, if you think the docstrings is good as it is, feel free to merge.

pygfx/materials/_base.py
almarklein: If the intersection is clipped, you clip less fragments - the union is visible. And the other way around. Could use help rephrasing this :/

almarklein: So I mean that the docstring is technically correct (also see [threejs example](https://threejs.org/examples/?q=clipp#webgl_clipping_intersection), but it can be made more clear. Maybe it needs something like "clipped fragments are not clipped" in there.

Korijn: The thing that confused me was the use of the word "satisfies". It's not immediately clear what that means. How about:

> By default a fragment is visible if it is clipped by any clipping plane. With this set to True, it is visible only if it is clipped by all of the clipping planes.

almarklein: I like how removing the words from set theory make the docstring more clear. I'm wondering whether we can find a better name for the property itself (at the cost of deviating from the ThreeJS name). E.g. `clip_any = True/False` or `clip_mode = "any", "all"`.

Korijn: I really like clip_mode=any/all!

Issue #104, opened 2021-06-18
Unified approach to colormapping


almarklein:
Proposal:

### Colors 

In all cases, at some point you have a color that can consist of 1, 2, 3, 4 elements:
   * color is 1D: grayscale
   * color is 2D: gray + alpha
   * color is 3D: RGB
   * color is 4D: RGBA
   
### For meshes, lines and points

  * `geometry.texcoords` can be Nx1, Nx2, Nx3, Nx4, or omitted.
  * `material.map` dimension should match the texcoords.
  * If `material.map` is present, `geometry.texcoords` must be present too.
  * If `material.map` is not present, the `material.color` is used.

### For volumes and images

In this case there are two textures, one for the "data", and an optional one for the colormap.

   * If no colormap is suplied, the value of the volume/image is rendered as a color as describe above.
   * If a colormap is present, the data-values are used to sample into the colormap, and the `nchannes` of the data texture must match the dimensions of the colormap.

almarklein:
#### Question about per-vertex coloring
A user can implement per-vertex coloring, by setting Nx3 texcoords (RGB values), and setting a "unit' colormap. But this is somewhat awkward, and involves an unnecessary texture lookup. Do we want `geometry.vertex_colors`? (for mesh, line, points).

#### Question about the data-texture

I feel that the texture containing the data is not "material"; its data intrinsic to the object, not a detail on how it should be rendered. I'd argue that its geometry (the size of the texture defines the geometry).

In other words, the ThreeJS abstraction, where everything is a mesh, does not hold up so well for our more exotic types of data. Perhaps we should broaden the scope of geometry to "data that defines an object" or something.

Korijn:
### Per-vertex coloring

I think vertex coloring is such a common use case that we need to support it with a clear, simple API. I don't really see what would make it complicated? In threejs it's a flag on mesh materials just like wireframe rendering is. If `material.vertex_colors` is `True`, then `geometry.colors` is used.

Korijn:
### On (volume) data-textures

I think it's important that we keep our object model consistent in a conceptual sense. It's a scene graph that tells the renderer what to draw (world objects, geometry), and how to draw it (materials, cameras). It's modelled to support that specific purpose, it's not modelled after the data.

[...]

PR #151, opened 2021-10-02
Colormappable meshes

almarklein:
Implements #104 for meshes. The mesh material already had support of ND colormaps, so this is just a few tweaks and adding an example. Will add support for lines etc. in subsequent PR's.

* [x] Support for 1D, 2D, and 3D colormapping on Mesh. Was already implemented.
* [x] Add a check for matching map and texcoords. 
* [x] Fixed that 2-channel colors are interpreted as gray+alpha. 
* [x] Added two examples.

----
Example output: 

image

### DIFF
--- a/examples/colormap_channels.py
+++ b/examples/colormap_channels.py
@@ -1,0 +1,67 @@
+"""
+Example demonstrating colormaps in 4 modes: grayscale, gray+alpha, RGB, RGBA.
+"""
+
+import pygfx as gfx
+import numpy as np
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
+app = QtWidgets.QApplication([])
+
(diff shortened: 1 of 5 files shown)


Korijn:
Looks like you pushed an empty file: examples/colormap_channels.py

Korijn:
I'm pretty sure people will be looking for vertex coloring (and less commonly face coloring) support. E.g. distance/heat maps are often implemented with them.

almarklein:
Fixed the missing file. Also changed the other example to use a cylinder instead of a torus know, because it makes the demo more clear (especially the 3D colormap case).

Made issue #152 to track per-face coloring. Let's also discuss that per-vertex coloring elsewhere (FWIW we do support per-vertex coloring, but the API is a bit odd perhaps).

Issue #33, opened 2020-05-01
Progress on architecture


almarklein:
Listing some things we need to look at regarding the architecture. Will add more as things come to mind.

* [ ] How are we going to do collections and instancing? Might affect how we do uniforms.
* [ ] How to define uniforms (#21)
* [ ] How to deal with updates / marking things dirty (also see #24)

almarklein:
> Be consistent in function arguments w.r.t. x, y (z) vs tuple vs Vector3

My take: 

* For width/height viewport-related stuff, use separate args, not a size arg.
* For colors use a single arg.
* For points/positions I'm also leaning towards a single arg (but there seems to be no function that accepts a points as input).
* For sizes seperate args seem to make most sense (e.g. `Boxgeometry(width, height, depth)`). But I suppose it also depends on the context.
* In cases where both approaches seem to make sense, it's not hard to support both 🤷 

Anyway. No changes needed. Writing this down as a sort of guide for the future.

PR #153, opened 2021-10-05
Stick with thickness (as opposed to line width)

### LINKED ISSUE #88: Use "width" instead of "thickness"?
For lines etc. I think I like `width` more, and it's also more common in the python plotting ecosystem, i think.

almarklein:
Closes #88, which is one point in #33.

I originally proposed to rename `Line.thickness` for `Line.width`, but I now actually like `thickness` better. It better describes what it is, and avoids confusion with height/width. For example `Mesh.wireframe_thickness` feels better than `Mesh.wireframe_width` (to me).

So this PR cleans this up a bit. If someone feels `width` is better, we can discuss. I don't feel *that* strongly about it, as long as we use it consistently.

### DIFF
--- a/pygfx/materials/_line.py
+++ b/pygfx/materials/_line.py
@@ -46,7 +46,6 @@ class LineMaterial(Material):
             self._vertex_colors = value
             self._bump_rev()
 
-    # todo: thickness? maybe rename to width?
     @property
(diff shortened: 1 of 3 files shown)


Korijn:
LGTM

PR #154, opened 2021-10-05
Prefer props over setters

almarklein:
Checks one item off #33

We are already using properties in most places. This applies a few more fixes.


Korijn:
LGTM!

PR #155, opened 2021-10-05
Adding a color util

almarklein:
Checks one off in #33.

This is a POC / proposal to add a `Color` class to pygfx.

This makes it possible for users to provide colors in various ways. And when getting a color prop, it can be formatted and processed in various ways too. The user never needs to instantiate it directly (although she could).

We support every way to provide a color in CSS, except the "hsl(..)" format. I left that for another day. We also support color tuples as we've used them so far. ThreeJS promotes the use of using an integer to define a color, e.g. 0xff0000. I started out with supporting this, but eventually removed it, because that integer does not look like a color when shown in decimal notation, so it easily leads to confusion.

In the top of the module I left some ideas for further work, based on ThreeJS and Qt.

Todo:
* [x] Do we want this?
* [x] Implementation.
* [x] Tests.
* [x] ~Tweak and test the hex-int format.~ drop int format
* [x] Tweak the repr.
* [x] Check how CSS, ThreeJS, and Qt handle color.
* [x] Implement where we use colors.

### DIFF
--- a/pygfx/utils/color.py
+++ b/pygfx/utils/color.py
@@ -1,3 +1,4 @@
+"""Provides utilities to deal with color."""
 import ctypes
 
 
--- a/pygfx/utils/color.py
+++ b/pygfx/utils/color.py
@@ -1,9 +1,37 @@
 """Provides utilities to deal with color."""
 
+# Possible improvements:
+#
+# * Support for HSL colorspace:
+#   * Accept CSS compatible "hsl(...)" and "hsla(...)"
+#   * Color.hsl property
+#   * Color.from_hsl() classmethod
+# * Support for HSV color space?
+# * Support for HSLuv, a colorspace with uniform brightness https://www.hsluv.org/
+# * Color.__add__ another color or a scalar.
+# * Color.__mul__ another color or a scalar.
+# * Color.lerp(color, t) linear interpolate towards other color.
+# * Color.lerpHSL(color, t) same as lerp but interpolate in HSL space.
+# * Color.lighter(factor) and Color.darker(factort)
+
+# todo Color.__eq__
+# todo: Support "rgb(100%, 50%, 0%)"
+
+
 import ctypes
 
 
 F4 = ctypes.c_float * 4
+
+
+def _float_from_css_value(v, i):
+    v = v.strip()
+    if v.endswith("%"):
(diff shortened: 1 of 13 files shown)


pygfx/utils/colors.py
```diff
@@ -0,0 +1,150 @@
...
+            self._save_from_rgba(color[0], color[1], color[2], 1)
+        elif len(color) == 4:
+            self._save_from_rgba(*color)
+        else:
+            raise ValueError(f"Color tuple must have 1-4 value, got: {color}")
+
+    def _save_from_str(self, color):
+        color = color.lower()
+        if color.startswith("#"):
+            # todo: Hex RGB or RGBA
```
Korijn, line 77: are you going to implement this? I feel like it would be the most common use case, probably :)

almarklein: Yes, if we move forward with this idea :)

almarklein:
This is ready for review. The `Color` class is done. I'll now apply it whereever a property/method accepts a property.

Korijn:
LGTM.

Awesome stuff!

pygfx/utils/color.py
```diff
@@ -0,0 +1,423 @@
...
+    * `Color("red")` base color names.
+    * `Color("cornflowerblue")` CSS color names.
+    * `Color("m")` Matlab color chars.
+
+    Hex colors:
+
+    * `Color("#ff0000")` the commmon hex format.
+    * `Color("#ff0000ff")` the hex format that includes alpha.
+    * `Color("#ff0)` the short form hex format.
+    * `Color("#ff0f)` the short form hex format that includes alpha.
```
almarklein, line 65:
```suggestion
    * `Color("#ff0")` the short form hex format.
    * `Color("#ff0f")` the short form hex format that includes alpha.
```

pygfx/utils/color.py
```diff
@@ -0,0 +1,423 @@
...
+
+    CSS color functions:
+
+    * `Color("rgb(255, 0, 0")`.
+    * `Color("rgba(255, 0, 0, 1.0")`.
```
almarklein, line 70:
```suggestion
    * `Color("rgb(255, 0, 0)")`.
    * `Color("rgba(255, 0, 0, 1.0)")`.
```

PR #156, opened 2021-10-05
Clean up some WGPU abstraction leaks

almarklein:
Very much WIP.

There are some points where WGPUs-specifics are present in the main API. In a perfect world, only `renderers/wgpu` should know about wgpu. 

Places where wgpu-specifics are used:

* The `material._wgpu_get_pick_info` methods. I don't worry much about this, since it's not public API and clearly prefixed.
* The format prop in the `Buffer` and `Texture` classes. So we can come up with our own way to describe a format, but why not simply use the wgpu enums? In most cases the format is auto-determined anyway.
* The usage flags in the `Buffer` and `Texture` classes. This part is real ugly now. Options:
  * Use strings from their WGPU names, that can be combined using commas. yuk.
  * Let user use actual WGPU flags. Feels wrong.
  * Define our own flag (that's really just a copy of the wgpu flags). Shown in this PR.
  * Turn each flag-value in a boolean property.

### DIFF
--- a/pygfx/resources/_buffer.py
+++ b/pygfx/resources/_buffer.py
@@ -1,3 +1,4 @@
+import enum
 import numpy as np
 
 import wgpu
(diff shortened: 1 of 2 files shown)


Korijn:
As you noted yourself this is still leaky, and there's nothing we can really do about that without having more renderer implementations, to see if our abstractions are really general enough or not.

I think we the most we can really do at this point is design our API such that user code won't have to change when we refactor. And we already seem to have that; none of the examples need to import wgpu to accomplish their goals, right?

So I'm not sure these changes are worth it, right now. :)

pygfx/materials/_base.py
```diff
@@ -39,15 +39,15 @@ def _set_size_of_uniform_array(self, key, new_length):
...
+        assert (
+            "*" in dtype
+        ), f"uniform field {key} '{dtype}' does not look like an array"
+        n, _, subtype = dtype.partition("*")
+        assert int(n) >= 0
 
         # Adjust type definition (note that this is originally a class attr)
         self.uniform_type = self.uniform_type.copy()
-        self.uniform_type[key] = dtype[0], (new_length, shape[1], shape[2])
+        self.uniform_type[key] = f"{new_length}*" + subtype
```
ivoflipse, line 50: Why concatenate instead of including it in the f-string?

almarklein: Good point

PR #173, opened 2021-10-11
Implement inheritance for Material.uniform_type

### LINKED ISSUE #158: Let material uniform type automatically include the uniforms of the base classes
Right now, each material must define its uniform type including the fields defined in base classes. This is repetitive, but also fragile for people who implement custom materials (their materials break when we add fields to a uniform.

We must:
* Find a good way to only have to define the new fields.
* The `array_from_shadertype()` function must sort the fields to prevent alignment issues.

almarklein:
Closes #158

Needs updating to one test file, but waiting for #172 to avoid merge conflicts. Ready for review though.


Korijn:
LGTM!

It's still a bit surprising to me that these alignment issues aren't picked up/reported on by wgpu or naga.

PR #174, opened 2021-10-12
WIP OIT via weighted blended / McGuire

almarklein:
We use the existing color target as the accumulation buffer and a new (additional) render target for the revealage.

In the current state it implicitly assumes that there are no opaque fragments. One thing that should be done is to add a render step that renders the opaque fragments (and sets the depth buffer). These opaque fragments should then be blended where we also resolve the transparent objects.

Thoughts:
* It's nice that this does not require any advanced GPU trickery.
* I don't like that we'd need to draw everything twice to separate the opaque from the transparent objects. 
* One feature of this approach is that it has "smooth transitions". This is nice if you have particle systems, e.g. simulate fog with 20 or so blobs, but it can also result in unexpected results. In particular, this method performs really bad with near-opaque objects that are too close to each-other for the weighing function to make a significant difference.

image

### DIFF
--- a/examples/transparency1.py
+++ b/examples/transparency1.py
@@ -1,0 +1,49 @@
+"""
+Example showing transparency.
+"""
+import time
+import wgpu
+import pygfx as gfx
+
+from PyQt5 import QtWidgets
+from wgpu.gui.qt import WgpuCanvas
+
+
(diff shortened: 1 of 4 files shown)


almarklein:
Closing - superseded by #196

PR #183, opened 2021-10-16
Update two examples to use Binding( ... ) rather than tuples

marcdownie:


almarklein:
Thanks! I must have missed those!

Issue #176, opened 2021-10-14
The geometry abstraction


almarklein:
## Problem

A thing that bothers me a bit about our architecture is that the geometry abstraction is sometimes confusing.

For example in #175, the `width` and `color` are on the material, which means that they are details about how the points are rendered. But when a user specifies the `width` or `color` per point, its on the geometry, which means that suddenly it's data intrinsic to the object.

(Another thing is that we use properties everywhere, but on the geometry we just set attributes as needed. But let's discuss that later to prevent this getting too big.)

## The purpose of geometry

In my words:

1. Group the "intrinsic data of the object" in one place.
2. To make it easy to create multiple objects with the same geometry (instead of having to specify positions, normals, texcoords, indices, etc). 
3. (extending 1) To allow the rule that every property on the `WorldObject` also affects its children.

For (2) there are other ways to make that work. But for me (3) is the best selling point.

## Options

Some thoughts (not mutually exclusive):

### We should better define what geometry means and what its purpose is

And document it in the docstring and guide.

###  We could rename "geometry"

Or is geometry ok? Could you consider colors geometry? The name "data" is nice and short, but also *very* generic. One strong point in favor of geometry is that its consistent with ThreeJS.

### We could allow providing data in the object constructor

Resulting in less verbose code:

```py
points1 = gfx.Points(positions=..., colors=... material=...)
```

### We could make the helpers create an object instead of geometry

Instead of :
```py
cube1 = gfx.Mesh(geometry=gfx.CubeGeometry(2, 2, 2), material=...)
cube2 = gfx.Mesh(geometry=cube1.geometry, material=...)
```
we do:
``` py
cube1 = gfx.Cube(2, 2, 2, material=...)
cube2 = gfx.Mesh(geometry=cube1.geometry, material=...)
```

This makes for code that's somewhat easier on the eye, especially for newcomers.

### We could introduce a clone method
``` py
cube2 = cube1.clone(material=...)
```

Similarly we could have a clone method on the material and geometry.

Korijn:
I'm sorry but I strongly disagree with changing this abstraction. It's very simple and powerful, which is why so many people have gotten the hang of ThreeJS. If anything we should document it more clearly.

I also noticed that you proposed the concept "helpers create an object" that we already have in place actually, and so does ThreeJS. Perhaps you just haven't gotten used to it yet. See https://github.com/pygfx/pygfx/tree/main/pygfx/helpers and https://threejs.org/docs/?q=helper

[...]

PR #195, opened 2021-10-27
Using functions instead of classed for creating geometry

almarklein:
Relates to #176.

This is a proposal for using functions instead of classes to create predefined geometry. I feel this makes more sense given how the Geometry class works, especially if we'd add more predefined geometry like teapots, triceratopses, human heads, and e.g. generated geometry for solid lines.

* [x] We figured out how we want it.
* [x] Implemented `xx_geometry()` for all geometry generating code.
* [x] Updated all examples. 
* [x] The `WorldObject` now:
   * Always has a geometry and material property, but they can be None (e.g. Group).
   * Must be instantiated with these two items. (removed convenience logic in a few places, most notably `Volume`).
* [x] Includes a few fixes for the earlier `index` -> `indices` rename.


almarklein:
Ready from my end. Update the top post for a brief overview.

pygfx/geometries/_cylinder.py
```diff
@@ -134,62 +133,82 @@ def generate_cap(radius, height, radial_segments, theta_start, theta_length, up=
...
+    return cylinder_geometry(
+        radius_bottom=radius,
+        radius_top=0.0,
+        height=height,
+        radial_segments=radial_segments,
+        height_segments=height_segments,
+        theta_start=theta_start,
+        theta_length=theta_length,
+        open_ended=open_ended,
+    )
```
Korijn, line 214: `cone_geometry` could also be defined with `functools.partial` which should be much less verbose :)

pygfx/objects/_scene.py
```diff
@@ -13,6 +13,8 @@ class Background(WorldObject):
     Can be e.g. a gradient, a static image or a skybox.
     """
 
-    def __init__(self, material):
-        super().__init__()
-        self.material = material
+    def __init__(self, geometry=None, material=None):
+        # Allow one arg, as an exception
```
Korijn, line 17: this is a surprising gotcha :) why?

Korijn:
I only had two minor comments, looks good!

pygfx/objects/_scene.py
almarklein: I am afraid people will do `SomeBackground(SomeBackgroundMaterial())`, thus omitting the geometry, and getting weird errors. What I should do instead here, is check for this and raise an error. Will fix.

pygfx/geometries/_cylinder.py
almarklein: Not sure. How would that work with docstrings? Also some arguments are named differently. I'm not sure what this'd look like, but this could be because I've rarely used `functools.partial` :)  If you have a nice patch though, I am happy to try that out.

almarklein:
Ready again. Now that most these WorldObject subclasses have become very small, I moved them together into one module.  Do you want to do something with `cone_geometry`?

pygfx/geometries/_cylinder.py
Korijn: Good point, I'll look into those questions.

Partial is quite nice because it is about 10x faster at runtime than calling a function wrapped in another function. I assume it's a bit more cleverly handled internally to cpython.

almarklein: > about 10x faster

wow 🤯 ... that I did not know

Issue #196, opened 2021-10-30
Plan for OIT


almarklein:
This is a more concrete plan to solve #73. This supersedes #108 (POC via spinlock) and #174 (POC weighted blended).

## Step 1

Refactor the code such that it does two render passes, one for opaque fragments, and another for the semi-transparent fragments. In this refactoring I would like to put together the logic that defines how fragments are outputted by the shaders, so that we can also do this in different ways.

The end result will already produce better visual results, though not order-independent yet.

## Step 2

Implement weighted blended transparency (McGuire 2013), and the depth-weighted improvement (MacGuire 2016). This will make available an approach that is good enough for many use-cases.

## Step 3

Improve the result by correctly blending the first `n` layers. My initial idea was to implement a per-pixel fragment array using a spinlock, as I did in the poc in #108. However, the spinlock seems to be rather a burden, so I tried another approach that turned out to work well: use a 3d render pass to capture the foremost transparent layer.

## Necessary improvements

With the method added in step3, each object is rendered 3 times, which is a waste. We should improve performance by dividing objects into "buckets" if we know that all fragments produced by an object are either opaque or transparent. Can be done by looking at the opacity property, and by a new property set by the user.

We should also look at memory consumption of the render buffers, etc.

Some other cleanup (see todo's below).

## Possible improvements for the future

In step 3 we added a 1-layer that is properly blended, and do the rest via weighted blended. This gives sharp edges for the front-most layer, but and transitions behind that layer are not well-defined. I think that in practice this does not matter too much, because you can interact with the camera to get a better view. But of course there will be cases where this does matter. There are ways to improve on this.

One approach could be to reconsider the custom fragment buffer with spinlock again. The advantage is that it needs just two passes (the regular opaque pass, and one transparency pass), and the size of `n` only scales memory, not speed in any significant manner. The downside is that the custom RW buffer is bound to the pipeline of all wobjects. So perhaps that buffer should be shared by all renderers and be as large as the largest canvas. Anyway, this ads some complexity. The biggest hurdle though, is that we'd have to fix the glitches and incidental hanging of the spinlock.

Another approach would be to look at a hybrid depth-peeling + weighted blended. Sort of what we have, but for `n>1`. For this approach `n` will scale the number of passes. and it might need additional combination passes. On the plus side, it may re-use render targets better. With depth peeling we'd need a depth buffer attached to all wobject pipelines, so again managing this ads complexity.

## Thoughts on memory

For our renderer's internal state:

[...]

PR #198, opened 2021-11-01
OIT step 1: Refactor renderer

almarklein:
This refactors the wgpu render code to prepare for implementing OIT, as planned in #196.

Todo:
* [x] Refactor so all blending stuff is defined in a single place.
* [x] More refactoring to support 1 or 2 render passes. 
* [x] Implement render mode that we had before this.
* [x] Implement simple 2-pass render mode.
* [x] Add `renderer.blend_mode`.
* [x] Decide how to implement the two fragment entry points with the shader (templates). Depends on https://github.com/gfx-rs/naga/issues/1509 -> see #200 
* [x] Implement for all materials.
* [x] Fix example to use PySide6.
* [x] Remove the `with open("c:/dev/py/tmp.wgsl")` thing.
* [x] Self-review + check for todo's.
* [x] Relief limitation that varying assignment must be a oneliner
* [x] Cleanup


berendkleinhaneveld:
Well, you could at least call it `FragmentBlender` :)

pygfx/renderers/wgpu/_pipelinebuilder.py
```diff
@@ -0,0 +1,520 @@
...
+    alt_render_pipelines = []
+
+    # Process each pipeline info object, converting each to a more concrete dict
+    for pipeline_info in wobject._wgpu_pipeline_infos:
+        if "vertex_shader" in pipeline_info and "fragment_shader" in pipeline_info:
+            pipeline = compose_render_pipeline(shared, blender, wobject, pipeline_info)
+            if pipeline_info.get("target", None) is None:
+                render_pipelines.append(pipeline)
+            else:
+                raise NotImplementedError("Alternative render pipelines")
```
berendkleinhaneveld, line 159: This should not be here, right?

almarklein: The idea is that material renderers can chose to render to an alternative render target (multi-pass rendering). We don't need it (yet) so it has not been implemented.

almarklein: This code was already there, BTW. It just moved.

berendkleinhaneveld: Thanks for clarifying!

almarklein:
Still refactoring and cleaning up, but here are some results for three initial blend modes.

The "solid" blend mode pretends that there is no such thing as transparency. Only included as a stub. It has one benefit: it's order independent :P

image

The "minimal" blend mode uses one pass, processing opaque and transparent fragments at the same time, while depth-testing. This is basically what we had. You can see how the red plane (the first-drawn) hides the other planes behind it because of the depth test. Also the blue plane (drawn last) disappears behind the green plane. Oh, and the sphere is partially hidden too.

image

The "simple" blend mode uses two passes. One for opaque and one for transparent fragments. The looks result is much better. However, it's not order-independent, as can be seen in the second image.

image
transparency01
)

Korijn:
Just wondering if we are still sorting objects along their depth in camera space prior to rendering? That way at least the result will be deterministic, though still order-dependent.

almarklein:
I think we are.

almarklein:
I'm now happy with the approach to deal with the shaders that run (slightly changed) in multiple render passes. See #200 for details. Other than that, we just create two wgsl samples, and two wgpu shader modules: one for each render pass. I initially wanted to include all entrypoints into the same shader, but that quickly becomes complicated to implement and explain.

almarklein:
This is ready from my end.

Some updates:
* I implemented the new approach for all material-renderers! Ran into a few hurdles though:
* Since the blend mode affects the pipeline of the wobjects, the pipeline data could no longer be stored on the wobjects themselves (it would go wrong if using one object in multiple renders with different blend modes). Therefore the pipeline-data is now associated via weak-ref dicts. For details see the extensive comment in _renderer.py.
* The blender defines the `FragmentOutput`, but it does not know beforehand whether the eventual shader wants to also write depth. I added another (much simpler) wgsl post processing pass to resolve this.
* The varyings resolve post-processing now deals with multiline assignments.

Overall, this refactoring is mainly a preparation for more advanced blend modes, adding support for multi-stage rendering. Because we now have a 2 pass blend mode, even though its still simple/wrong, it already yields much better results for scenes that contain transparent objects. It also deals much better with e.g. the semitransparent edges of our lines (for aa); our lines look much better now!

Korijn:
> * Since the blend mode affects the pipeline of the wobjects, the pipeline data could no longer be stored on the wobjects themselves (it would go wrong if using one object in multiple renders with different blend modes). Therefore the pipeline-data is now associated via weak-ref dicts. For details see the extensive comment in _renderer.py.

Sounds like a big improvement to me!

> * The blender defines the `FragmentOutput`, but it does not know beforehand whether the eventual shader wants to also write depth. I added another (much simpler) wgsl post processing pass to resolve this.

How should I feel about this? :)

> * The varyings resolve post-processing now deals with multiline assignments.

Awesome!

> Overall, this refactoring is mainly a preparation for more advanced blend modes, adding support for multi-stage rendering. Because we now have a 2 pass blend mode, even though its still simple/wrong, it already yields much better results for scenes that contain transparent objects. It also deals much better with e.g. the semitransparent edges of our lines (for aa); our lines look much better now!

Let's merge and continue in another PR!

almarklein:
> How should I feel about this? :)

It means we can write:
```
var out = finalize_fragment();
out.depth = ndc_pos.z / ndc_pos.w;
```
and it Just Works.

We could make it more explicit, e.g.:
```
var out = finalize_fragment_at_depth(ndc_pos.z / ndc_pos.w;);
```

BTW, I think I may replace `add_fragment` and `finalize_fragment` with a single function. Let's revisit then. I'll add a note in the TODO of #196

Korijn:
As far as I am concerned we can merge :)

almarklein:
Wasn't sure whether @berendkleinhaneveld wanted to have a look.

berendkleinhaneveld:
Please go ahead and merge: I'll catch up later!

PR #204, opened 2021-11-23
Weighted blended OIT

almarklein:
Step 2 in #196.

* [x] Add (oprional) render-pass to combine the renderpasses.
* [x] Weighted blended without depth component in weights.
* [x] Weighted blended by depth.
* [x] Put RenderFlusher back in its own module.
* [x] Refactor to put Passes in small classes. 
* [x] Clean up.


almarklein:
Ready!

This implements the McGuire weighted blending. Unfortunately, I also did some refactoring, so the diff is awful. If you want to have a look, you better read the final `_blender.py`

Here the two new methods in action: First the normal blended, fullly OIT. Next weighted blended - you can see how the colors blend depending on depth.

Korijn:
It looks amazing!! Awesome! Good to merge and press on from my end... the refactoring all makes sense.

PR #206, opened 2021-11-25
OIT step 3: proper rendering of frontmost layer

almarklein:
One step in #196 

This implements a new 3-pass approach to improve on the weighted blended blender. The idea is to do an extra pass in which the front-most transparent layer is captured. In the combine-pass, this layer is subtracted from the accum+reveal targets, and merged into the final result using the alpha bend equation.

In more detail:
* [x] Refactored so that the number of passes is defined by the blender.
* [x] Also drop a few other assumptions about the passes concerning depth.
* [x] Implement new 3-pass approach.
* [x] Fixed that the existing weighted blended methods were too dark.
* [x] Remove that weird rule that I introduced that `pass_index` 0 must be the opaque pass. Maybe use named passes?
* [x] Have a look at the depth handling in `_renderer.py`, I think it still makes some assumptions.
* [x] Double-check that the colors are correct (the pink seems a bit bright?)
* [x] Self-review / check for any todos.
* [x] Write up some notes about possible alternatives like mlab and depth-peeling.

### DIFF
--- a/pygfx/renderers/wgpu/_blender.py
+++ b/pygfx/renderers/wgpu/_blender.py
@@ -23,20 +23,31 @@ standard_texture_des = {
 class BasePass:
     """The base pass class, defining and documenting the API that a pass must provide."""
 
-    depth_write_enabled = True
     write_pick = False
 
-    def get_pipeline_targets(self, blender):
+    def get_color_descriptors(self, blender):
         """Get the list of fragment targets for device.create_render_pipeline()."""
         # The result affects the wobject's pipeline.
         # Returning [] prevents this pipeline from being created.
         return []
 
-    def get_color_attachments(self, blender, clear):
+    def get_color_attachments(self, blender, clear_color):
         """Get the list of color_attachments for command_encoder.begin_render_pass()."""
         # The result affects the rendering, but not the wobject's pipeline.
         # Returning [] prevents this pass from running.
         return []
+
+    def get_depth_descriptor(self, blender):
+        """Get dict that has the depth-specific fields for the depth_stencil field
+        in device.create_render_pipeline().
+        """
+        return {}
+
+    def get_depth_attachment(self, blender, clear_depth):
+        """Get a dict that has the depth-specific fields for the
+        depth_stencil_attachment argument of command_encoder.begin_render_pass().
+        """
+        return {}
 
     def get_shader_code(self, blender):
         """Get the fragment-write shader code.
@@ -61,10 +72,9 @@ class OpaquePass(BasePass):
     in all multi-pass blenders.
     """
 
-    depth_write_enabled = True
     write_pick = True
 
-    def get_pipeline_targets(self, blender):
+    def get_color_descriptors(self, blender):
         bf, bo = wgpu.BlendFactor, wgpu.BlendOperation
         return [
             {
@@ -82,8 +92,8 @@ class OpaquePass(BasePass):
             },
         ]
 
-    def get_color_attachments(self, blender, clear):
-        if clear:
+    def get_color_attachments(self, blender, clear_color):
+        if clear_color:
             color_load_value = 0, 0, 0, 0
             pick_load_value = 0, 0, 0, 0
         else:
@@ -104,6 +114,20 @@ class OpaquePass(BasePass):
                 "store_op": wgpu.StoreOp.store,
             },
         ]
+
+    def get_depth_descriptor(self, blender):
+        return {
+            "format": blender.depth_format,
+            "depth_write_enabled": True,
+            "depth_compare": wgpu.CompareFunction.less,
+        }
+
+    def get_depth_attachment(self, blender, clear_depth):
+        return {
+            "view": blender.depth_view,
+            "depth_load_value": (1.0 if clear_depth else wgpu.LoadOp.load),
+            "depth_store_op": wgpu.StoreOp.store,
+        }
 
     def get_shader_code(self, blender):
         return """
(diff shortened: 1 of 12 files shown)


almarklein:
Order-independent *and* sharp edges!

Korijn:
Aren't we supposed to see a sharp line here as well? Or is this a limitation of only capturing the front-most layer?

image

brisvag:
This is amazing! Could you explain in short why sharp edges are enabled by this change? @Korijn , my guess is that only the sharp edges in the "front-most transparent layer" are affected, which is why those you pointed out are missing.

almarklein:
We covered most questions in the meeting this morning, but here are the answers for the record :)

> How do you bucket objects for 3 passes? It seems very expensive.

Right now all objects are drawn 3 times with this mode. The "bucketing" is one of the things to be added next. This will still render transparent objects twice, so yes, it's expensive, but worth it in most cases.

> Could you explain in short why sharp edges are enabled by this change?
> Aren't we supposed to see sharp lines here as well?

The sharp edges are not visible with the weighted -blended (McGuire) approach, because depth is not taken into account (well it is with `weighted_depth`, but then its only used to weigh the fragments, not to determine which one is actually in front). With this new approach the top-most transparent layer is blended separately, taking depth into account to see what fragment is the front-most, so where one object "sticks through another" you will see the transition. 

Since we only do this for the top layer, any transitions behind that top-layer are still blended using standard weighted-blended. 

> You haven't referenced any papers or articles here, is this an effect you came up with on your own?

I have not seen this exact method in literature. I took inspiration from some variants of MLAB (see #73). With MLAB you create a special fragment buffer that has space for `n` fragments. When that buffer is full you have to decide what to do with new fragments. Older methods just dropped the far-most ones. A better approach is to blend the furthest fragments to make space. Since this approach is actually not order-independent, an arguably even better approach is to use McGuire's blending for the furthest fragments.

You could say that the approach in this PR implements that, with `n=1`. Except we don't use this home-brew fragment buffer, but simply an extra render pass with normal render targets. I guess the main novelty/trick in this approach is that when the results are combined into the final color, the contribution that the front-most layer has in the weighted-blended part is first undone, and then it is applied using the normal alpha blending equations.

I think its possible to add more front-layers this way using depth peeling, but that comes with added complexity, in part because of the need to bind a depth buffer as a texture to sample from, which then becomes part of the pipeline of all world-objects.

> I'm just wondering if this effect is "correct" and if it is worth a 3-pass rendering penalty.

It's correct (if you consider the alpha blending equation a correct way to model the world) for the first layer. Since things behind the first layer are already obfuscated somewhat by that first layer, it will matter less there. Though this does not hold if the first layer has a very low alpha.

Whether this is worth it depends on the application, but I think it makes a big difference, and on modern hardware we can afford to do it in a lot of use-cases.

pygfx/renderers/wgpu/_blender.py
```diff
@@ -624,10 +735,135 @@ def _create_combination_pipeline(self, device):
...
+        # Depth buffer for the front-most pass.
+        # We could re-use the already existing depth buffer. And if we
+        # do the front-most pass first, then the final state of the
+        # depth buffer still matches the opaque fragments. However, to
+        # properly support renderer.render(clear_depth=False), we need
+        # to keep these states separate.
+        self._texture_info["frontdepth"] = (
+            wgpu.TextureFormat.depth32float,
+            usg.RENDER_ATTACHMENT,
+        )
```
almarklein, line 790: This is a bit unfortunate. I will try to address this later.

almarklein: Tracking this with #209

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -226,12 +226,13 @@ def blend_mode(self):
...
-          transparency, with depth weighting (McGuire 2013). The depth range affects the
-          (quality of the) visual result.
-        * "multilayer2": todo 2-layer MLAB + weighted for the rest?
+        * "weighted": two-pass approach that yields order independent transparency, using alpha weights.
+        * "weighted_depth": two-pass approach that yields order independent
+          transparency, with weights based on alpha and depth (McGuire 2013). Note that the depth
+          range affects the (quality of the) visual result.
+        * "weighted_plus": three-pass approach that yields order independent transparency,
+          in wich the front-most transparent layer is rendered correctly, while transparent
+          layers behind it are blended using "weighted".
```
almarklein, line 235: No need to let it hold up this PR, but any suggestions for better names are welcome :)

pygfx/renderers/wgpu/_blender.py
```diff
@@ -624,10 +735,135 @@ def _create_combination_pipeline(self, device):
...
+            // Reconstruct the color and alpha, and set final color, with premultiplied alpha
+            let avg_color = accum.rgb / max(accum.a, epsilon);
+            let alpha = 1.0 - reveal;
+            out.color = vec4<f32>(avg_color * alpha, alpha);
+
+            // Blend in the front color (front is already premultiplied)
+            let out_rgb = (1.0 - front.a) * out.color.rgb + front.rgb;
+            let out_a = (1.0 - front.a) * out.color.a + front.a;
+            out.color = vec4<f32>(out_rgb, out_a);
+        """
```
almarklein, line 863: This is the core part of this new blending method.

almarklein:
This is ready! Some notes:

* I wrote up some possible directions to go beyond only that front-layer in #196.
* I think I fixed #207, but awaiting confirmation.
* I created #208 and #209 to pick up soon after this.

Korijn:
I really want to review thoroughly but I just don't have the headspace right now. Feel free to merge.

Issue #209, opened 2021-11-29
Revist API for overlays and subpots


almarklein:
Just a quick write-up, following up on https://github.com/pygfx/pygfx/pull/206/files#r758278223
Maybe I'm overthinking this and/or can we change the behaviour without changing the API.

Relates to #208 and #76.

Currently, we support `renderer.render(... clear_depth=False)`. This means that we support the use-case where one scene is rendered, and then a second scene is rendered in the same place, and everything is blended together in the correct way.

This works correctly with all the transparency stuff, as long as each blender-pass has its own render targets. This is unfortunate, because now the `weighted_plus` method cannot share the depth buffer between passes, causing it to consume more memory. Also the `simple2` method currently shares both the color- and depth-buffers. These would need to be replicated for correct results.

The main use-cases to support with the API of `renderer.render()` are subplots and overlays. The ability to "blend" multiple scenes is not that big a requirement.

almarklein:
Closed by #213

PR #212, opened 2021-12-01
Reduce memory footprint for picking

### LINKED ISSUE #208: Can we use a smaller texture for picking?
Right now we use a rgba32sint for the picking texture. That contributes to 16 bytes per pixel, which is quite substantial. For comparison, the color and depth buffer each contribute 4 bytes per pixel. These all look like small numbers, but they add up if you create a full-screen window, especially with FSAA.

Proposal: see if can use rgba16uint instead (8 bytes per pixel).

The pick info for one fragment has 4 values, each an i32. How these are used depends on the material. For all materials, the first value is the object id.

For the mesh: 1:  the instance;  2: the face index, 3: the barycentric coords.
For the volume: 1,2,3: xyz
For the line: similar to mesh. 
For the point: similar to line, but does not use the last value.

Reducing to 16bit, we'd have `2**16 ` is 65536 values at our disposal for each value. I think I could live with limiting the number of objects in a scene to that number. Note that the current limit is not `2**32`, but ~4 million, because of the precision of f32 varyings. This is only about 8x more than `2**16`. A similar limit would be acceptable for the instances.

For the face coords or line vertex index, hower, this would be limiting, because 1 million or more vertices is not *that* rare. We could divide the bits more effective though, for instance:
* 16 bits for object id (65536)
* 14 bits for instance (16384)
* 25 bits for the face/vertex index (33 million)
* 3x3=9 bits for the barycentric coords (0..8 each)

almarklein:
Closes #208.

Changes:
* [x] Use object id's for instancing, so we don't need extra bits for that.
* [x] Provide some low-level bit juggling utils.
* [x] Reduce the used bits for picking to 64.
* [x] Use rgb16uint for the pick texture.
* [ ] Copy pick info to low-res texture, and reuse the rgba16uint for the other things. -> first address #209

This means that:
* We can now precisely allocate a certain amount of bits for each pick info field.
* The instance-index does not need any bits since its encoded via the wobject-id.
* The above points mean that we can now get away with 64bits (instead of 128), thus reducing the memory footprint of the pick buffer by 50%.
* The picking logic in the shader is also easier to read.

About the allocated bits:
* 20 bits are for the wobject id's, which means we can support up to 1M objects. This number if fixed, because the core renderer needs to get the object id.
* I have the vertex/face index 26 bits in all materials. That's up to 50M vertices. The 26 is a nice number, because it leaves 18 bits for other info, which is nicely divisible by both 2 and 3. 
* The mesh has 3 (barycentric) face coords of 6 bits each.
* The point has 2 point coords (in logical pixels) of 9 bits each (-255 - +255).
* The line has a 1 sub-coord of 18 bits (-0.5, 0.5).
* The volume has 3 tex_coords of 14 bits.

### DIFF
--- a/pygfx/materials/_line.py
+++ b/pygfx/materials/_line.py
@@ -24,7 +24,7 @@ class LineMaterial(Material):
         values = unpack_bitfield(pick_value, 20, 26, 18)
         return {
             "vertex_index": values[1],
-            "line_coord": (values[2] - 100000) / 100000.0,
+            "segment_coord": (values[2] - 100000) / 100000.0,
         }
 
     @property
--- a/pygfx/materials/_line.py
+++ b/pygfx/materials/_line.py
@@ -1,4 +1,5 @@
 from ._base import Material
+from ..utils import unpack_bitfield
 
 
 class LineMaterial(Material):
@@ -19,11 +20,12 @@ class LineMaterial(Material):
         self._vertex_colors = vertex_colors
 
     def _wgpu_get_pick_info(self, pick_value):
-        # The instance is zero while renderer doesn't support instancing
-        instance = pick_value[1]
-        vertex = pick_value[2]
-        vertex_sub = pick_value[3] / 1048576
-        return {"instance_index": instance, "vertex_index": vertex + vertex_sub}
+        # This should match with the shader
+        values = unpack_bitfield(pick_value, 20, 26, 18)
+        return {
+            "vertex_index": values[1],
+            "line_coord": (values[2] - 100000) / 100000.0,
+        }
 
     @property
     def color(self):
--- a/pygfx/materials/_line.py
+++ b/pygfx/materials/_line.py
@@ -21,10 +21,10 @@ class LineMaterial(Material):
 
     def _wgpu_get_pick_info(self, pick_value):
         # This should match with the shader
-        values = unpack_bitfield(pick_value, 20, 26, 18)
+        values = unpack_bitfield(pick_value, wobject_id=20, index=26, coord=18)
         return {
-            "vertex_index": values[1],
-            "segment_coord": (values[2] - 100000) / 100000.0,
+            "vertex_index": values["index"],
+            "segment_coord": (values["coord"] - 100000) / 100000.0,
         }
 
     @property
--- a/examples/picking_mesh.py
+++ b/examples/picking_mesh.py
@@ -23,7 +23,7 @@ class PickingWgpuCanvas(WgpuCanvas):
         if wobject and "face_index" in info:
             # Get what face was clicked
             face_index = info["face_index"]
-            coords = info["face_coords"]
+            coords = info["face_coord"]
             # Select which of the three vertices was closest
             # Note that you can also select all vertices for this face,
             # or use the coords to select the closest edge.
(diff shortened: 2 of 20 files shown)


almarklein:
This is ready. I updated the main description.

pygfx/materials/_line.py
```diff
@@ -19,11 +20,12 @@ def __init__(
         self._vertex_colors = vertex_colors
 
     def _wgpu_get_pick_info(self, pick_value):
-        # The instance is zero while renderer doesn't support instancing
-        instance = pick_value[1]
-        vertex = pick_value[2]
-        vertex_sub = pick_value[3] / 1048576
-        return {"instance_index": instance, "vertex_index": vertex + vertex_sub}
+        # This should match with the shader
+        values = unpack_bitfield(pick_value, 20, 26, 18)
```
berendkleinhaneveld, line 24: It's pretty hard to understand and review these 'magic' numbers. I know that it should resemble the layout of bits, but it's hard to verify / read.

berendkleinhaneveld: Suggestion: replace with a named definition. That also allows us to more easily change/update the bit layout if we want/need to.

almarklein: Good point. I made them kwargs instead, which makes this a bit easier to read. The numbers may still feel "magical" and need to match with those in the shader. Technically we could parametrize these, but that also adds complexity.

almarklein:
I just found a sample of Python code that I used to verify the bit packing in the shader. Let's add it here for future reference:
```py
import struct

p_pick_bits_used = np.uint32(0)
p_pick = np.array([0,0,0,0], np.uint8)

bits_per_channel = np.uint32(8)
mask = np.uint32(2**bits_per_channel-1)


def pick_pack(value, bits):
    global p_pick_bits_used, p_pick
    v = np.uint32(min(value, 2**bits))

    shift0 = p_pick_bits_used;
    shift1 = p_pick_bits_used - int(bits_per_channel);
    shift2 = p_pick_bits_used - int(bits_per_channel * 2);
    shift3 = p_pick_bits_used - int(bits_per_channel * 3);


    # print("--", p_pick_bits_used, v, shift0, shift1, shift2, shift3)
    pick_add = np.array([
        v << shift0 if shift0 >= 0 else v >> -shift0,
        v << shift1 if shift1 >= 0 else v >> -shift1,
        v << shift2 if shift2 >= 0 else v >> -shift2,
        v << shift3 if shift3 >= 0 else v >> -shift3,
    ], np.uint8);

    p_pick += np.array([
        pick_add[0] & mask if abs(shift0) < 32 else 0,
        pick_add[1] & mask if abs(shift1) < 32 else 0,
        pick_add[2] & mask if abs(shift2) < 32 else 0,
        pick_add[3] & mask if abs(shift3) < 32 else 0,

    ], np.uint8);

    p_pick_bits_used += np.uint32(bits)



items = [(7, 4), (10, 5), (117, 7), (5172, 13), (3, 2)]

for rot in range(4):
    p_pick_bits_used = np.uint32(0)
    p_pick = np.array([0,0,0,0], np.uint8)
    print("-----------------")
    items.append(items.pop(0))

    for value, bits in items:
        pick_pack(value, bits)

    print(struct.pack("<BBBB", *p_pick))
    print(p_pick_bits_used, "/", bits_per_channel*4 )


    in_buffer = struct.pack("<BBBB", *p_pick)
    raw = struct.unpack("<L", in_buffer)[0]

    for value, bits in items:
        value2 = raw & int(2**bits-1)
        print(value, value2)
        assert value == value2
        raw = raw >> bits

```

PR #213, opened 2021-12-05
Remove support for clear_depth

almarklein:
By removing support for the `clear_depth` argument of `renderer.render()`, the scope of that functions is reduced to rendering a scene, either on a fresh canvas, or (with clear_color off) rendering it *over* an existing rendered scene. The particular use-case that is dropped, is that of merging two scenes, but his use-case was somewhat artificial anyway. The dropping of this arg also excludes other techniques, but any such techniques should be solved using blenders anyway.

The advantage of removing this arg (and use-case) is that the blender's internal state can be refreshed on each draw (except for the main color buffer and pick buffer). This means that buffers can be re-used, which reduces the memory consumption.

In fact, in this PR the `WeightedPlusFragmentBlender` is updated to re-use the depth buffer instead of creating a secondary depth buffer. More optimization may be possible.

----

Another way to put it, is that this PR makes `clear_depth` standard, without a way to turn if off. This means you have two options: either you render a fresh image, or you render *over* the existing image. The option to render *in* the image is no longer possible, because that's a task of the blender. 

Todo: 
* [ ] the comments about memory usage in `_blender.py` should be updated: both #212 and this PR affect these. Need to update in last PR to merge.


Korijn:
I've come to understand that the workflow for pygfx will be to select an appropriate blending strategy for the visual result you are trying to achieve, so I guess we would need to document that.

I could also imagine (later) creating an API where users can configure the blending strategy themselves, in addition to having all these presets available to them.

PR #211, opened 2021-11-30
WIP MSAA

almarklein:
Not sure whether to proceed with this.

PR #214, opened 2021-12-05
Offset in flusher resolve

almarklein:
In #211 I closely examined the resolve step of the flusher, and found that we're missing an offset.


Korijn:
Good find :) you can merge this whenever you want!

PR #215, opened 2021-12-09
Update one example to use auto gui, plus wx support

almarklein:
* [x] Update cube example
* [x] Add examples specific to qt and wx
* [x] Fix for wx


Korijn:
On Windows 11, cube.py (with glfw) and cube_qt.py (with pyside6) work. cube_wx.py acts weird; it shows up, renders the first frame, but only shows the rotation after I resize the window, but before I release the left mouse button.

PR #217, opened 2021-12-10
Update weighted blended

### LINKED ISSUE #207: Segmentation fault when changing blend mode on macOS
When changing the blend mode in the transparency examples, I run into the following segmentation fault:

```
Selecting blend_mode blended
thread '<unnamed>' panicked at 'ColorState(1, FormatNotBlendable(R16Float))', src/device.rs:645:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[1]    79601 segmentation fault  poetry run python examples/transparency2.py
```
Here is the full backtrace: [rust_backtrace.log](https://github.com/pygfx/pygfx/files/7608737/rust_backtrace.log)

For now, the examples run fine when I change the `_texture_info["accum"]` and `_texture_info["reveal"]` in [BlendedFragmentBlender](https://github.com/pygfx/pygfx/blob/9b706a5209366c64e955517822e23aa23c013480/pygfx/renderers/wgpu/_blender.py#L523) from 16 bit to 32 bit variants.

almarklein:
Closes #207.

This is an attempt to fix #207, and maybe implement the idea of re-using the rgba16uint picking texture for the accumulation texture.

In [McGuire's blog](http://casual-effects.blogspot.com/2015/03/implemented-weighted-blended-order.html) it says that the reveal texture can be r8unorm. The extra bits of r16float help a bit, but don't seem *that* necessary - he argues that the memory it requires may not be worth it.

### DIFF
--- a/pygfx/renderers/wgpu/_blender.py
+++ b/pygfx/renderers/wgpu/_blender.py
@@ -664,7 +664,7 @@ class WeightedFragmentBlender(BaseFragmentBlender):
 
         # The reveal buffer collects the weights
         self._texture_info["reveal"] = (
-            wgpu.TextureFormat.r16float,
+            wgpu.TextureFormat.r8unorm,
             usg.RENDER_ATTACHMENT | usg.TEXTURE_BINDING,
         )
 
--- a/pygfx/renderers/wgpu/_blender.py
+++ b/pygfx/renderers/wgpu/_blender.py
@@ -653,8 +653,8 @@ class WeightedFragmentBlender(BaseFragmentBlender):
         usg = wgpu.TextureUsage
 
         # Create two additional render targets.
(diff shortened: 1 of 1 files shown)


almarklein:
I also tried using rgba16uint for the accumulation texture, because if that works, we could re-use the picking texture and reduce the memory considerably. There are two downsides: 1) we'd need an extra pass to copy the picking data to a lower-res texture; 2) I don't seem to be able to use a 16uint texture with texture binding.

In more detail: when using an uint texture, you cannot access it via a sampler. I get that. But I also seem unable to bind it as a texture. But binding it as a storage texture also does not work, because as of late the only access for a storage texture is write-only. I think that maybe our wgpu-native snapshot is in between a transition that now effects the formats that we can use as textures.

I really hope that the current change is enough to fix #207.

PR #220, opened 2021-12-14
API for object sorting and providing transparency hints

almarklein:
Todo:

* [x] A new `WorldObject.render_order` property.
* [x] A new `WorldObject.render_mask` property to tell the renderer whether the object is opaque or transparent (or let the renderer try figuring this out). 
* [x] A new `Renderer.sort_objects` property to control whether objects are sorted. 
* [x] Autodetermine in what passes an object must be rendered.
* [x] Cleanup
* [x] Remove `Blender.is_order_independent` attribute if we won't use it (to determine whether to sort or not).

The `Renderer.sort_objects` is False by default. If True, it will use the wobjects's `render_order` and distance from camera to sort.

The  `WorldObject.render_mask` is "auto" by default. In this case the renderer will try to establish that the object is either fully opaque or fully transparent. This cannot always be determined, e.g. because there is a texture mapping that has an alpha channel, or a color per vertex (which can also have alpha < 1). In these cases it will just render in both passes. The user can override this via `WorldObject.render_mask`. This can also be done for e.g. lines: setting the render_mask to "opaque" will effectively disable the aa for a line (though we still have the SSAA), making it a bit more performant when this matters.

### DIFF
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -76,7 +76,10 @@ class Material(ResourceContainer):
 
     @opacity.setter
     def opacity(self, value):
-        self.uniform_buffer.data["opacity"] = min(max(float(value), 0), 1)
+        value = min(max(float(value), 0), 1)
+        if (value == 1) != (self.uniform_buffer.data["opacity"] == 1):
+            self._bump_rev()  # rebuild pipeline if we become opaque/transparent
+        self.uniform_buffer.data["opacity"] = value
         self.uniform_buffer.update_range(0, 1)
 
     @property
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -70,7 +70,7 @@ class Material(ResourceContainer):
     def opacity(self):
         """The opacity (a.k.a. alpha value) applied to this material, expressed
         as a value between 0 and 1. If the material contains any
-        non-opaque fragments, these are simply scaled by this value.
+        non-opaque fragments, their alphas are simply scaled by this value.
         """
         return float(self.uniform_buffer.data["opacity"])
 
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -78,7 +78,7 @@ class Material(ResourceContainer):
     def opacity(self, value):
         value = min(max(float(value), 0), 1)
         if (value == 1) != (self.uniform_buffer.data["opacity"] == 1):
-            self._bump_rev()  # rebuild pipeline if we become opaque/transparent
+            self._bump_rev()  # rebuild pipeline if this becomes opaque/transparent
         self.uniform_buffer.data["opacity"] = value
         self.uniform_buffer.update_range(0, 1)
 
--- a/pygfx/renderers/wgpu/_renderer.py
+++ b/pygfx/renderers/wgpu/_renderer.py
@@ -443,10 +443,9 @@ class WgpuRenderer(Renderer):
         # I *think* that (eventually?) it should be possible to record the commands
         # and re-submit them, resulting in better performance. But if I try this now
         # it panics with 'Cannot remove a vacant resource'.
-        # If we do get this to work, we should also trigger a recording
-        # when the wobject.children change.
-        need_recording = any_has_changed or not self._blender.is_order_independent
-        need_recording = True
+        # If we do get this to work, we should trigger a new recording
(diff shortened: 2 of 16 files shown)


pygfx/renderers/wgpu/_renderer.py
```diff
@@ -266,6 +287,21 @@ def blend_mode(self, value):
...
+    @property
+    def sort_objects(self):
+        """Whether to sort world objects before rendering.
+
+        * ``True``: the render order is defined by 1) the object's distance to the
+          camera; 2) the object's ``render_order`` property; 3) the position object
+          in the scene graph (based on a depth-first search).
+        * ``False``: don't sort, the render order is defined by the scene graph alone.
+        """
+        return self._sort_objects
```
almarklein, line 299: * Could also allow setting a function object (in this PR or later).
* Could also allow `None` to sort depending on whether the `blend_mode` is order independent. I think we should not do this, because when a user changes the bend mode, the user may just as well edit the `sort_objects` as well.

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -266,6 +287,21 @@ def blend_mode(self, value):
         if isinstance(self._target, wgpu.gui.WgpuCanvasBase):
             self._target.request_draw()
 
+    @property
+    def sort_objects(self):
+        """Whether to sort world objects before rendering.
+
+        * ``True``: the render order is defined by 1) the object's distance to the
+          camera; 2) the object's ``render_order`` property; 3) the position object
+          in the scene graph (based on a depth-first search).
```
Korijn, line 296: Could you reorder the text in the docstring so that it aligns with the code? 2 comes before 1.

almarklein: Oh it does indeed!

pygfx/materials/_base.py
```diff
@@ -76,7 +76,10 @@ def opacity(self):
 
     @opacity.setter
     def opacity(self, value):
-        self.uniform_buffer.data["opacity"] = min(max(float(value), 0), 1)
+        value = min(max(float(value), 0), 1)
+        if (value == 1) != (self.uniform_buffer.data["opacity"] == 1):
+            self._bump_rev()  # rebuild pipeline if we become opaque/transparent
```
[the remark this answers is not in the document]
almarklein: Only when it changes being opaque/transparent.

Korijn:
> The WorldObject.render_mask is "auto" by default. In this case the renderer will try to establish that the object is either fully opaque or fully transparent. [...] In these cases it will just render in both passes.

So, I guess the choice of default is between "safest" and "most performant"? In Three.js I think everything is opaque by default (most performant) and users have to opt-in per object by setting transparent=True.

I also _think_ most objects are opague, in general. But that gut feeling might be totally wrong.

What do you think?

almarklein:
In ThreeJS an object is either opaque or transparent, and the user *has* to specify it with the `Object3D.transparent` flag. There is no way for the user to let ThreeJS figure it out. So the choice to make objects opaque by default is not purely for performance.

In pygfx the renderer handles opaqueness/transparency on a per-fragment basis, and we do have the luxury that we can let the renderer figure it out and get the expected result. Rendering each object both in the opaque-pass and the transparent-passes will do that. In this PR we made the renderer smarter, so that in cases where it knows that either the opaque or transparent passes are not needed, they are omitted. And for those cases where the renderer cannot establish this, and where performance matters, there is the `WorldObject.render_mask` property.

If we were to make "opaque" the default, it might be more performant, but only in the cases where the auto-detection "fails". IMO it's not a strong enough benefit to give up that everything works as expected by default.

To give some details about the autodetection for the mesh: if there is no texture, it looks at the `.opacity` and the alpha component of the `.color`. If there is a texture, it checks whether it is intensity/rgb, vs intensity_alpha/rgba. So in (I think) the large majority of cases, you'd get good performance *and* good visual results by default.

almarklein:
> I also think most objects are opague, in general. But that gut feeling might be totally wrong.

That depends a lot on the use-case, but in general I think this is very true indeed :)

Korijn:
Alright, thanks for elaborating! Good to merge as far as I am concerned :)

almarklein:
Pushed another tiny commit, but I'm pretty sure you agree with that small change. Merging!

Issue #219, opened 2021-12-13
Looks like on Intel cannot submit twice 


almarklein:
We already saw and fixed this in #216, but there are examples (e.g. scene_in_scene, subplots) where it still happens. Also the cube.py example in wgpu-py. This is a blocker for determining what to do with depth in `renderer.render()`.

Looks very much like https://github.com/gfx-rs/wgpu/issues/1672 which was closed by https://github.com/gfx-rs/wgpu/pull/2212 recently (on 25 November).

I think it'd be good to update wgpu-native + wgpy-py first. With a bit of luck the problem is then fixed and we can continue this work as planned.

almarklein:
This is fixed with the new wgpu-native.

PR #222, opened 2021-12-21
Updates for latest wgpu-py

almarklein:
* Pin to latest wgpu.
* The shader `[[block]]` has been removed.
* The total size of a uniform struct must now be a multiple of 16.
* Flush now submits its own commands, fixing the examples that used a manual flush. This is possible now that #219 is no longer an issue.
* Fix issue with line starts being all weird (shader does not like if we try indexing out of bounds?)
* Remove depth pick (caused an error related to depth clipping now, but we wanted to remove it anyway).
* Rename the simple1 and simple2 blend modes to ordered1 and "ordered2.
* I now know that recorded commands cannot be re-used - need to use render bundles for that .... mañana.

This ticks off the last items in #196!


almarklein:
I'm running into a few more quircks.

almarklein:
Ready. I'll go over all the examples, making sure they work, and converting to `wgou.gui.auto`, in a next PR.

PR #224, opened 2021-12-22
Update examples to use wgpu's auto gui

almarklein:
Now based off #222, will rebase once that is merged. 

I squashed, rebased and pushed to #227. Leaving this open for a bit, in case I made a mistake (I deleted my local branch).


Korijn:
I guess the `install_requires` for wgpu-py should also require at least `0.7.1` now :)

pygfx/controls/_orbit.py
```diff
@@ -175,3 +175,43 @@ def update_camera(self, camera: "Camera") -> "OrbitControls":
...
+        )
+
+    def handle_event(self, event, canvas, camera):
+        """Implements a default interaction mode that consumes wgpu autogui events
+        (compatible with the jupyter_rfb event specification).
+        """
+        type = event["event_type"]
+        if type == "pointer_down":
+            xy = event["x"], event["y"]
+            if event["button"] == 1:
```
Korijn, line 196: This used to be the other way around; left mouse button is rotate, right mouse button is pan.

That also aligns with Three.js: https://threejs.org/examples/?q=orbit#misc_controls_orbit

almarklein: Fixed!

PR #226, opened 2021-12-22
Show utility

### LINKED ISSUE #107: Simple viewer shortcut
What I think would be supremely valuable for prototyping and debugging purposes, is a simple viewer UI (based on ...Qt?) that can be launched with a simple single function.

For example... you're in the middle of some code, your scene is in a certain state, and you want to debug it at that point by inspecting it visually. Just write `pygfx.show(scene)` (or something like that) and a little window pops up with its own independent canvas, renderer and a default orbit camera with some simple interaction to control it.

What do you think?

### LINKED ISSUE #194: Camera look_at utility

Korijn:
Closes #107 
Closes #194

Now based off #224, will rebase once that is merged.

The idea is that you can visualize any world object immediately.

New API:
* `Geometry.bounding_box()`
* `Geometry.bounding_sphere()`
* `Geometry.from_trimesh(mesh)` (classmethod)
* `Matrix4.to_ndarray()`
* `gfx.linalg.utils.transform(vectors, matrix)`
* `gfx.linalg.utils.aabb_to_sphere(aabb)`
* `gfx.linalg.utils.transform_aabb(aabb, matrix)`
* `gfx.linalg.utils.transform_sphere(sphere, matrix)`
* `WorldObject.get_world_position()`
* `WorldObject.get_world_bounding_box(include_children=True)`
* `WorldObject.get_world_bounding_sphere(include_children=True)`
* `Camera.show_object(target)`
* `gfx.show(object)`

New examples:
* `examples\show_util.py`
  image
* `examples\show_stl.py`
  image


Korijn:
I'm intending to merge this after I address that to do item, but I am curious to hear how we could make this even simpler. Like, what is the shortest path we can create from an STL file to interactive vis?

almarklein:
We're close!

examples/show_util.py
```diff
@@ -0,0 +1,12 @@
+"""
+Demonstrates show utility
+"""
+
+import pygfx as gfx
+
+material = gfx.MeshPhongMaterial()
+geometry = gfx.box_geometry(100, 100, 100)
+cube = gfx.Mesh(geometry, material)
```
almarklein, line 9: I've grown to like this notation, and its a wee bit code ;)
```suggestion
cube = gfx.Mesh(
    gfx.box_geometry(100, 100, 100),
    gfx.MeshPhongMaterial(),
)
```

almarklein:
> what is the shortest path we can create from an STL file to interactive vis?

This is already really short. I suppose that any code leading up to the call to `show()` would mostly be loading or preparing the data. So if you want to enable shorter code, maybe that's where to look. E.g. passing a filename to show even.

That said, what would we `show()` to be? A function that you can throw almost anything at and it with a bit of luck it works? Or does it require a WorldObject? Both have their merits. The first option can be really useful, but can also become muddy, and I can imagine that it'd be worth its own lib (e.g. because it might get tons of dependencies to load all sorts of objects).

examples/show_util.py
almarklein: But this is personal, so feel free to ignore.

pygfx/utils/show.py
```diff
@@ -0,0 +1,47 @@
+"""
+Quickly visualize world objects with as
+little boilerplate as possible.
+"""
+
+from wgpu.gui.auto import WgpuCanvas, run
```
Korijn, line 6: Delay this import

pygfx/objects/_base.py
```diff
@@ -400,6 +402,32 @@ def look_at(self, target: Vector3):
...
+    def get_world_bounding_box(self):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        bbox = self._geometry.bounding_box()
+        transformed = np.array(
+            [
+                Vector3(*bbox[0]).apply_matrix4(self._matrix_world).to_array(),
+                Vector3(*bbox[1]).apply_matrix4(self._matrix_world).to_array(),
+            ]
+        )
+        bbox_world = np.array([transformed.min(axis=0), transformed.max(axis=0)])
```
Korijn, line 419: this needs to iterate over all 8 corners of the rotated box

pygfx/geometries/_base.py
```diff
@@ -88,3 +93,33 @@ def __init__(self, **kwargs):
 
             # Store
             setattr(self, name, resource)
+
+    def bounding_box(self):
+        if not hasattr(self, "positions"):
```
Korijn, line 98: to support image data; use self.grid

pygfx/objects/_base.py
```diff
@@ -400,6 +403,33 @@ def look_at(self, target: Vector3):
...
             self.rotation.premultiply(self._q.inverse())
 
+    def get_world_position(self):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        self._v.set_from_matrix_position(self._matrix_world)
+        return self._v.clone()
+
+    def get_world_bounding_box(self):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        bbox = self._geometry.bounding_box()
```
Korijn, line 413: take children into account

pygfx/objects/_base.py
```diff
@@ -400,6 +403,33 @@ def look_at(self, target: Vector3):
...
+                Vector3(*corner).apply_matrix4(self._matrix_world).to_array()
+                for corner in corners
+            ]
+        )
+        bbox_world = np.array([corners_world.min(axis=0), corners_world.max(axis=0)])
+        return bbox_world
+
+    def get_world_bounding_sphere(self):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        bsphere = self._geometry.bounding_sphere()
```
Korijn, line 426: take children into account

pygfx/geometries/_base.py
```diff
@@ -88,3 +93,33 @@ def __init__(self, **kwargs):
...
+
+        if self._aabb_rev == self.positions.rev:
+            return self._aabb
+        pos = self.positions.data
+        self._aabb = np.array([pos.min(axis=0), pos.max(axis=0)])
+        self._aabb_rev = self.positions.rev
+        return self._aabb
+
+    def bounding_sphere(self):
+        if not hasattr(self, "positions"):
```
Korijn, line 111: to support image data; use `self.grid`

pygfx/utils/show.py
```diff
@@ -0,0 +1,48 @@
...
+        scene.add(object)
+
+        background = Background(None, BackgroundMaterial((0, 1, 0, 1), (0, 1, 1, 1)))
+        scene.add(background)
+
+    camera = PerspectiveCamera(70, 16 / 9)
+
+    pos = object.get_world_position()
+    _, radius = object.get_world_bounding_sphere()
+    camera.position.copy(pos).add_scaled_vector(Vector3(1, 1, 1), radius * 1.25)
```
Korijn, line 34: should this be a util func on camera?

ivoflipse: Would that be the same as https://github.com/pygfx/pygfx/issues/194?

ivoflipse:
Would this close https://github.com/pygfx/pygfx/issues/107?

pygfx/utils/show.py
almarklein: Maybe `look_at_scene()` :)

Korijn: Yes it would!

Korijn: I ended up calling it `show_object`

Korijn:
Added an example to load an STL file:

image

Korijn:
I think this is just about ready!

The only minor nitpick I have left is setting the camera up vector for the teapot example. I couldn't figure it out.

Will rebase as soon as #227 is merged.

Korijn:
See #229

PR #229, opened 2021-12-23
Add show util

### LINKED ISSUE #107: Simple viewer shortcut

### LINKED ISSUE #194: Camera look_at utility

Korijn:
Closes #107
Closes #194

The idea is that you can visualize any world object immediately.

New API:
* `Geometry.bounding_box()`
* `Geometry.bounding_sphere()`
* `gfx.trimesh_geometry(mesh)`
* `Matrix4.to_ndarray()`
* `gfx.linalg.utils.transform(vectors, matrix)`
* `gfx.linalg.utils.aabb_to_sphere(aabb)`
* `gfx.linalg.utils.transform_aabb(aabb, matrix)`
* `WorldObject.get_world_position()`
* `WorldObject.get_world_bounding_box()`
* `WorldObject.get_world_bounding_sphere()`
* `Camera.show_object(target)`
* `gfx.show(object)`

New examples:
* `examples\show_util.py`
  image
* `examples\show_stl.py`
  image
* `examples\show_scene.py`
  image


pygfx/objects/_base.py
```diff
@@ -400,6 +403,26 @@ def look_at(self, target: Vector3):
...
+
+    def get_world_bounding_box(self, include_children=True):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        aabb = self._geometry.bounding_box()
+        aabb_world = transform_aabb(aabb, self._matrix_world.to_ndarray())
+        if self._children and include_children:
+            aabbs = np.stack(
+                [c.get_world_bounding_box() for c in self._children] + [aabb_world]
+            )
+            aabb_world = np.array([aabbs[:, 0].min(axis=0), aabbs[:, 1].max(axis=1)])
```
Korijn, line 419: Factor this out to a `merge_aabbs` linalg util

pygfx/objects/_base.py
```diff
@@ -400,6 +403,26 @@ def look_at(self, target: Vector3):
             self._q.set_from_rotation_matrix(self._m)
             self.rotation.premultiply(self._q.inverse())
 
+    def get_world_position(self):
+        self.update_matrix_world(update_parents=True, update_children=False)
+        self._v.set_from_matrix_position(self._matrix_world)
+        return self._v.clone()
+
+    def get_world_bounding_box(self, include_children=True):
+        self.update_matrix_world(update_parents=True, update_children=False)
```
Korijn, line 412: This causes a lot of unnecessary back and forth, can probably make it more efficient by only updating world matrices once

pygfx/objects/_base.py
```diff
@@ -400,6 +403,26 @@ def look_at(self, target: Vector3):
...
+        aabb = self._geometry.bounding_box()
```
Korijn, line 413: This needs to support world objects without geometry as well

pygfx/linalg/utils.py
```diff
@@ -1,7 +1,43 @@
...
+
+def transform_aabb(aabb, matrix):
+    corners = np.array(list(itertools.product(*aabb.T)))
+    corners_world = transform(corners, matrix)
+    return np.array([corners_world.min(axis=0), corners_world.max(axis=0)])
+
+
+def transform_sphere(sphere, matrix):
+    center = transform(sphere[:3], matrix)
+    radius = np.linalg.norm(transform(np.array([sphere[3], 0, 0]), matrix))
```
almarklein, line 42: I'm not sure I follow the radius transform. Should it not only apply the scale factor somehow?

pygfx/objects/_base.py
almarklein: Yes, e.g. we have a `Group` object (which we use nowhere yet I think).

pygfx/objects/_base.py
```diff
@@ -400,6 +403,26 @@ def look_at(self, target: Vector3):
...
+        if self._children and include_children:
+            aabbs = np.stack(
+                [c.get_world_bounding_box() for c in self._children] + [aabb_world]
+            )
+            aabb_world = np.array([aabbs[:, 0].min(axis=0), aabbs[:, 1].max(axis=1)])
+        return aabb_world
+
+    def get_world_bounding_sphere(self, include_children=True):
+        aabb = self.get_world_bounding_box(include_children=include_children)
+        return aabb_to_sphere(aabb)
```
almarklein, line 424: Wouldn't it be better to actually calculate with spheres here? How much harder would that be?

pygfx/utils/show.py
```diff
@@ -0,0 +1,43 @@
...
+    WgpuRenderer,
+    WorldObject,
+    Scene,
+    PerspectiveCamera,
+    OrbitControls,
+)
+
+
+def show(object: WorldObject):
+    from wgpu.gui.auto import WgpuCanvas, run
```
almarklein, line 18: Docstring ;)

pygfx/objects/_base.py
Korijn: It's non trivial to the point I would like to settle for this now and maybe revisit it later :)

pygfx/linalg/utils.py
Korijn: Yeah good point, it's wrong.

pygfx/geometries/_base.py
```diff
@@ -88,3 +94,50 @@ def __init__(self, **kwargs):
 
             # Store
             setattr(self, name, resource)
+
+    @classmethod
+    def from_trimesh(cls, mesh):
+        return cls(
+            positions=mesh.vertices.astype("f4"), indices=mesh.faces.astype("i4")
+        )
```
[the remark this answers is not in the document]
Korijn: Works for me, will adjust!

pygfx/objects/_base.py
Korijn: And Scene!

examples/show_stl.py
```diff
@@ -0,0 +1,27 @@
...
+
+teapot = trimesh.load(TEAPOT)
+
+mesh = gfx.Mesh(
+    gfx.Geometry.from_trimesh(teapot),
+    gfx.MeshPhongMaterial(),
+)
+
+# TODO: I'd rather change the camera up vector to be (0, 0, 1) for this model
+# but it's not clear to me how to do that
```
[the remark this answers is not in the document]
Korijn: I'll see if that does the trick

pygfx/linalg/utils.py
almarklein: Think a bit on how normals are scaled in the vertex shader, would this work?
* Create a center_offset (center offset by radius in any dimension).
* Transform that point
* The new radius is the distance between the `center_offset` and `center`.

pygfx/objects/_base.py
almarklein: Agreed :)

pygfx/geometries/_base.py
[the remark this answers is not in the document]
Korijn: I feel like that last one is already what the Geometry constructor offers. So I ended up making `gfx.trimesh_geometry(mesh)` which is pretty neat. I also expanded it to support vertex colors, texcoords and normals.

pygfx/objects/_base.py
Korijn: Decided against this as it ends up not shaving off much code at all

Korijn:
I think this is ready now!

pygfx/cameras/_base.py
```diff
@@ -38,6 +38,35 @@ def update_matrix_world(self, *args, **kwargs):
...
+            radius = bsphere[3]
+            distance = radius * distance_weight
+        else:
+            # whatever it is has no bounding sphere, so we just pick an
+            # arbitrary distance
+            distance = 100 * distance_weight
+        self.position.copy(pos).add_scaled_vector(
+            Vector3(*view_dir).normalize().negate(), distance
+        )
+        self.look_at(pos)
```
Korijn, line 68: This should be the center of the bounding volume

pygfx/cameras/_base.py
```diff
@@ -38,6 +38,35 @@ def update_matrix_world(self, *args, **kwargs):
...
+        if bsphere is not None:
+            radius = bsphere[3]
+            distance = radius * distance_weight
+        else:
+            # whatever it is has no bounding sphere, so we just pick an
+            # arbitrary distance
+            distance = 100 * distance_weight
+        self.position.copy(pos).add_scaled_vector(
+            Vector3(*view_dir).normalize().negate(), distance
+        )
```
Korijn, line 67: Should use the diagonal of the bounding box instead

Korijn:
There were quite a lot of gotchas here, but it's working nicely now.

I guess it only falls apart when the near/far planes are too small for the scene you're showing. That can be addressed in #231

PR #232, opened 2021-12-28
Add BoxHelper

### LINKED ISSUE #187: BoxHelper to show bounding box of an object
As ThreeJS says: Helper object to graphically show the world-axis-aligned bounding box around an object. 

This could be a relatively easy feature to support "selecting objects" before we get #180, and it also feels a bit like a precursor to #186.

Korijn:
Closes #187

* Add AxesHelper.set_colors
* Remove AxesHelper thickness kwarg and switch to thin lines
* Add BoxHelper
* Add BoxHelper.set_transform_by_object(object)
* Add BoxHelper.set_transform_by_aabb(aabb)
* Extend validate_helpers2 example
  image
* Add object_bounding_box example
  image

The two utility methods on BoxHelper allow you to configure the transform appropriately for world space or local space. See the example for usage.

### DIFF
--- a/examples/object_bounding_box.py
+++ b/examples/object_bounding_box.py
@@ -17,11 +17,18 @@ mesh = gfx.Mesh(
     gfx.trimesh_geometry(teapot),
     gfx.MeshPhongMaterial(),
 )
+mesh.rotation.set_from_euler(gfx.linalg.Euler(0.71, 0.91))
 scene.add(mesh)
 
-box = gfx.BoxHelper()
-box.set_object_world(mesh)
-scene.add(box)
+box_world = gfx.BoxHelper()
+box_world.set_object_world(mesh)
+scene.add(box_world)
+
+box_local = gfx.BoxHelper()
+box_local.set_object_local(mesh)
+mesh.add(box_local)
+
+box_local.material.color = (0, 1, 0, 1)
 
 
 if __name__ == "__main__":
--- a/examples/object_bounding_box.py
+++ b/examples/object_bounding_box.py
@@ -8,8 +8,7 @@ import trimesh
 import pygfx as gfx
 
 
-TEAPOT = Path(__file__).parent / "models" / "teapot.stl"
-teapot = trimesh.load(TEAPOT)
+teapot = trimesh.load(Path(__file__).parent / "models" / "teapot.stl")
 
 scene = gfx.Scene()
 
--- a/examples/object_bounding_box.py
+++ b/examples/object_bounding_box.py
@@ -1,0 +1,28 @@
+"""
+Demonstrates show utility with an STL file
+"""
+from pathlib import Path
+
+import trimesh
+
+import pygfx as gfx
+
+
+TEAPOT = Path(__file__).parent / "models" / "teapot.stl"
+teapot = trimesh.load(TEAPOT)
+
+scene = gfx.Scene()
+
+mesh = gfx.Mesh(
+    gfx.trimesh_geometry(teapot),
+    gfx.MeshPhongMaterial(),
+)
+scene.add(mesh)
+
+box = gfx.BoxHelper()
+box.set_object_world(mesh)
+scene.add(box)
+
+
+if __name__ == "__main__":
+    gfx.show(scene, up=gfx.linalg.Vector3(0, 0, 1))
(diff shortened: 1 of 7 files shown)


pygfx/helpers/_box.py
```diff
@@ -0,0 +1,69 @@
...
+    def set_object_local(self, object):
+        if object.geometry:
+            aabb = object.geometry.bounding_box()
+        else:
+            aabb = None
+        self.set_aabb(aabb)
+
+    def set_aabb(self, aabb):
+        if aabb is None:
+            raise ValueError("Object has no geometry")
```
almarklein, line 62: This error message is weird if used by the user - it seems to assume being called from the other methods.

almarklein: Also, docstrings :)

examples/object_bounding_box.py
```diff
@@ -0,0 +1,35 @@
...
+mesh.rotation.set_from_euler(gfx.linalg.Euler(0.71, 0.91))
+scene.add(mesh)
+
+box_world = gfx.BoxHelper()
+box_world.set_object_world(mesh)
+scene.add(box_world)
+
+box_local = gfx.BoxHelper()
+box_local.set_object_local(mesh)
+mesh.add(box_local)  # note that the parent is `mesh` here, not `scene`
```
almarklein, line 29: These two modes confuse me a bit, does the box *have* to be a child of the mesh in the second case? What happens if its not? What happens if we do the same for the first case?

An idea, without having thought this through: instead of these methods to set a target, what if you simply made the box the child of any world object, and the box would size itself to the bounds of that object?

examples/object_bounding_box.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+from pathlib import Path
+
+import trimesh
+
+import pygfx as gfx
+
+
+TEAPOT = Path(__file__).parent / "models" / "teapot.stl"
+teapot = trimesh.load(TEAPOT)
```
almarklein, line 12: Suggestion:
```suggestion
teapot = trimesh.load(Path(__file__).parent / "models" / "teapot.stl")
```

almarklein (commented):
Nice!

pygfx/helpers/_box.py
Korijn: Any suggestions for the error message?

examples/object_bounding_box.py
Korijn: The way I see it, a BoxHelper is just a box like an AxesHelper is just a set of axes. You can use it as you like; visualizing an object's bounding box is not its only purpose.

The methods are available just to make that one use case a little easier.

almarklein: That makes sense.

From my understanding `set_object_world` is more common/useful than `set_object_local`, right? The fact that in order for `set_object_local` to work correctly you must make it a child of the object seems a bit confusing to me. Is this use-case useful enough to justify a special method? If you think it is, we should probably document how to use it correctly.

pygfx/helpers/_box.py
almarklein: Maybe this should check that `aabb` is a bounding box, and the `None` check with the current error message should be in `set_object_world` and `set_object_local`.

examples/object_bounding_box.py
Korijn: Why is that confusing? :) I think it's very elegant use of the scene graph functionality. It's precisely what it's intended for.

An example use case of local space bounding box visualization is collision detection with oriented (=local) bounding boxes.

almarklein: > I think it's very elegant use of the scene graph functionality.

Ok, I see what you mean. Though I still find it a bit confusing that there are two methods that look very similar, but with one you have to also make sure to put the the object in the right position. Anyway, good docstrings come a long way :)

Korijn: How about I add a boolean kwarg to choose local/world space and merge the methods?

almarklein: Yeah, maybe that helps because then there's a default. But if you prefer the current state I'm also fine with that. Maybe a good criterium is to see what version leads to simpler docs.

Korijn:
I think I've managed to address all your feedback appropriately. :)

PR #234, opened 2022-01-14
Polyhedrons

Korijn:
This adds new built-in geometry generator functions:

* Tetrahedron
* Octahedron
* Icosahedron
* Dodecahedron

The functions include radius and subdivision parameters, everything is vectorized. Particularly the triangle subdivision feature was pretty tricky to get right!

I didn't expose the polyhedron generator itself via the public api (`gfx.polyhedron_geometry`) since it isn't very friendly (you need to provide seed vertex and index buffers) but users can still import it directly to use it if they want to.

Other changes:

* Since I ported the approach from three.js to vectorized numpy code I figured I should move our copy of their license to a more generic location
* Added unit tests for the polyhedron generator functions
* Promoted numerical numpy warnings to errors in the test suite

Screenshot of the example:

image

Alternative rendering with phong material and no texture mapping:

image


almarklein:
Nice! What do the subdivisions do? the same as for the cube?

There's this trick where you take the minimal polyhedron (the Tetrahedron?), then subdivide each face into two triangles, and push the new vertices to the unit sphere. You then repeat this process `n` times, and you get the mesh for a sphere that is uniformly sampled (as opposed to the very uneven size of the faces in a long-lat sphere). Texture mapping does not really work with such a sphere, but you need much less vertices to get a round-appearing mesh. Anyway, probably something for another day.

Korijn:
> Nice! What do the subdivisions do? the same as for the cube?

Sort of, yes! In the example screenshot you can see on the left an icosahedron that has been subdivided twice. 

> There's this trick where you take the minimal polyhedron (the Tetrahedron?), then subdivide each face into two triangles, and push the new vertices to the unit sphere. You then repeat this process `n` times, and you get the mesh for a sphere that is uniformly sampled (as opposed to the very uneven size of the faces in a long-lat sphere). Texture mapping does not really work with such a sphere, but you need much less vertices to get a round-appearing mesh. Anyway, probably something for another day.

I think that's equivalent to what is shown here!

Except everything is vectorized, no for loops and list pushing here :)

Korijn:
Face orientation fixed, looks good at subdivisions=0:

image

Korijn:
I added an example to show how subdivision works, it's equivalent to what I see here https://threejs.org/docs/?q=tetra#api/en/geometries/TetrahedronGeometry

It's clearly not equivalent to what you mentioned earlier, by the way.

So I'm satisfied with this PR for now.

almarklein:
> It's clearly not equivalent to what you mentioned earlier, by the way.

That's odd ... how else could one subdivide? Could it be that the subdivision results in an uneven distribution because in between the subdivisions the vertices are not normalized so that they're on the unit sphere? Maybe something to look at later.

Anyway, LGTM!

Korijn:
> > It's clearly not equivalent to what you mentioned earlier, by the way.
> 
> That's odd ... how else could one subdivide? Could it be that the subdivision results in an uneven distribution because in between the subdivisions the vertices are not normalized so that they're on the unit sphere? Maybe something to look at later.
> 
> Anyway, LGTM!

Yeah I think it could be fixed by using some other formula than linspace, probably something involved pi :) but let's save it for later

PR #237, opened 2022-01-21
Add Image WorldObject

### LINKED ISSUE #225: An Image WorldObject
Like Volume. We can now create a plane with a texture, but that does not position the data as you'd normally want. Plus it makes sense from the point of colormapping: In a Plane with a tex, that text is the map. In an Image with a tex, that tex is the colormap (of any dimension). 

Relates to #104

almarklein:
Closes #225.

Looks like a lot of code, but most is copied from the volume and adjusted for 2D.

### DIFF
--- a/pygfx/objects/_more.py
+++ b/pygfx/objects/_more.py
@@ -60,7 +60,7 @@ class Mesh(WorldObject):
 
 
 class Image(WorldObject):
-    """An object  representing a 2D image in space.
+    """An object representing a 2D image in space.
 
     The geometry for this object consists only of `geometry.grid`: a
     texture with the 2D data.
--- a/pygfx/objects/_more.py
+++ b/pygfx/objects/_more.py
@@ -59,10 +59,37 @@ class Mesh(WorldObject):
     """
 
 
+class Image(WorldObject):
+    """An object  representing a 2D image in space.
+
+    The geometry for this object consists only of `geometry.grid`: a
+    texture with the 2D data.
+
+    The picking info of an Image (the result of ``renderer.get_pick_info()``)
+    will for most materials include ``index`` (tuple of 2 int),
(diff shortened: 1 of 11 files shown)


pygfx/objects/_more.py
```diff
@@ -59,10 +59,37 @@ class Mesh(WorldObject):
     """
 
 
+class Image(WorldObject):
+    """An object  representing a 2D image in space.
```
Korijn, line 63:
```suggestion
    """An object representing a 2D image in space.
```

pygfx/renderers/wgpu/volumerender.py
```diff
@@ -462,7 +462,7 @@ def vertex_shader(self):
 
             // Store values for fragment shader
             varyings.position = vec4<f32>(ndc_pos);
-            varyings.world_pos = vec3<f32>(world_pos);
+            varyings.world_pos = vec3<f32>(world_pos.xzy);
```
Korijn, line 465: you sure about this? remains unexplained in the PR description

almarklein: world_pos is `vec4` but is cast to `vec3` here. I was not even aware that it was allowed to do that this way. Adding the `.xyz` makes it more explicit.

ivoflipse: Except that the code says xzy instead of xyz

almarklein: Oooh, shit.

almarklein: Good catch. My bad!

almarklein, line 465:
```suggestion
            varyings.world_pos = vec3<f32>(world_pos.xyz);
```

Korijn:
LGTM!

PR #239, opened 2022-01-26
Screenshot testing of examples on CI

Korijn:
This PR adds the framework for running examples on CI, just as in https://github.com/pygfx/wgpu-py/pull/238

Please help me identify which examples we want to run in the test suite!


Korijn:
I ran a simple script to see which examples I could import and access `example_module.canvas.draw` without getting exceptions, here's what I found:

# broken

(raise error on import)

- examples.geometry_polyhedron
- examples.integration_qt
- examples.post_processing1
- examples.post_processing2
- examples.text_with_qt

# other

(no `canvas` global)

- examples.clipping_planes
- examples.collection_line
- examples.colormap_channels
- examples.colormap_mesh
- examples.cube
- examples.custom_object
- examples.cylinder
- examples.flat_shaded_torus
- examples.geometry_cubes
- examples.geometry_klein_bottle
- examples.geometry_plane
- examples.geometry_torus_knot
- examples.image_plus_points
- examples.instancing_mesh
- examples.line_basic
- examples.line_performance
- examples.line_segments
- examples.line_thick
- examples.line_thin
- examples.manual_matrix_update
- examples.mesh_slice
- examples.multi_slice1
- examples.multi_slice2
- examples.offscreen
- examples.orbit_camera
- examples.panzoom_camera
- examples.picking_color
- examples.picking_mesh
- examples.picking_points
- examples.points_basic
- examples.scene_overlay
- examples.scene_side_by_side
- examples.skybox
- examples.spheres
- examples.transparency1
- examples.transparency2
- examples.validate_culling
- examples.validate_depth_clipping
- examples.validate_helpers1
- examples.validate_helpers2
- examples.validate_image1
- examples.validate_image2
- examples.validate_ndc
- examples.validate_volume
- examples.volume_render1
- examples.volume_render2
- examples.volume_slice1
- examples.volume_slice2
- examples.volume_slice3
- examples.volume_slice4
- examples.wireframe1
- examples.wireframe2

# compatible

(can be imported safely and exposes a `canvas` global)

- examples.clipping_planes
- examples.collection_line
- examples.colormap_channels
- examples.colormap_mesh
- examples.cube
- examples.custom_object
- examples.cylinder
- examples.flat_shaded_torus
- examples.geometry_cubes
- examples.geometry_klein_bottle
- examples.geometry_plane
- examples.geometry_torus_knot
- examples.image_plus_points
- examples.instancing_mesh
- examples.line_basic
- examples.line_performance
- examples.line_segments
- examples.line_thick
- examples.line_thin
- examples.manual_matrix_update
- examples.mesh_slice
- examples.multi_slice1
- examples.multi_slice2
- examples.offscreen
- examples.orbit_camera
- examples.panzoom_camera
- examples.picking_color
- examples.picking_mesh
- examples.picking_points
- examples.points_basic
- examples.scene_overlay
- examples.scene_side_by_side
- examples.skybox
- examples.spheres
- examples.transparency1
- examples.transparency2
- examples.validate_culling
- examples.validate_depth_clipping
- examples.validate_helpers1
- examples.validate_helpers2
- examples.validate_image1
- examples.validate_image2
- examples.validate_ndc
- examples.validate_volume
- examples.volume_render1
- examples.volume_render2
- examples.volume_slice1
- examples.volume_slice2
- examples.volume_slice3
- examples.volume_slice4
- examples.wireframe1
- examples.wireframe2

Korijn:
I've gone ahead and added at least all the validate examples.

As you can see I had to move the request_draw call out of the `if __name__ == "__main__":` part otherwise the draw function is not registered when the test calls `canvas.draw`.

Korijn:
For example in validate_culling I see the following difference between the screenshot on my machine vs CI:

test

The current `atol=1` doesn't allow for this difference, naturally. I will just copy the screenshots from CI into here for now, but let's think of a solution.

almarklein:
Let's stick to the validation examples for now. Maybe a simple blur is sufficient? We should look at what other tools (vispy, bokeh, skimage) use.

Korijn:
> Let's stick to the validation examples for now. Maybe a simple blur is sufficient? We should look at what other tools (vispy, bokeh, skimage) use.

Vispy uses similar metrics, but they do downsample the images before comparing. I will give that a try.

Korijn:
Tried it, didn't work, sadly. Still get diffs like this:

validate_culling-diff

Korijn:
Skimage latest version is not available on python 3.6 time to drop it?

almarklein:
> Skimage latest version is not available on python 3.6 time to drop it?

No objections.

I played a bit with a similarity function and I came up with this, which works for the validate_culling example (on my machine). Basically it checks what pixels are not close enough and for each of these it checks whether any of its direct neighbors is close enough. This should account for pixel-shifts. 

```py
def images_similar(im1, im2, *, atol):
    mask_rgba = np.isclose(im1, im2, atol=atol)
    mask = mask_rgba[:,:,0] & mask_rgba[:,:,1] & mask_rgba[:,:,2] & mask_rgba[:,:,3]

    neighbours_plus = [(-1, 0), (1, 0), (0, -1), (0, 1)]
    neighbouts_cross = [(-1, -1), (1, -1), (-1, 1), (1,1)]
    neighbours8 = neighbours_plus + neighbouts_cross

    yy, xx = np.where(mask==False)
    for x, y in zip(xx, yy):
        assert mask[y, x] == False  # make sure we didn't confuse x and y
        for dx, dy in neighbours8:
            rgba1 = im1[y+dy, x+dx]
            rgba2 = im2[y+dy, x+dx]
            if np.allclose(rgba1, rgba2, atol=atol):
                mask[y,x] = True
                break
        else:
            print(f"Nope, still failing at img[{y},{x}]")

    return np.all(mask)
```

almarklein:
I pushed my version, and only commented out the skimage stuff for now.

Korijn:
I rerendered the screenshots on my machine; as you can see they still aren't accepted on CI with this metric. :/

almarklein:
Oh, the tests also don't pass for me locally. I should have tested that. I'll have another stab.

> Oddly enough the diff seems blank, even though the test fails?

A white diff means that *all* pixels are different, right?

tests/test_examples.py
```diff
@@ -0,0 +1,136 @@
...
+        # assert similarity
+        is_similar = images_similar(stored_img, img, atol=1)
+        if not is_similar:
+            # if the test fails, write the new image
+            diff_path = diffs_dir / f"{module}.png"
+            imageio.imwrite(diff_path, img)
+            # and a diff image, so it can be inspected
+            diff_path = diffs_dir / f"{module}-diff.png"
+            diff_img = np.abs(stored_img - img)
+            imageio.imwrite(diff_path, diff_img)
```
almarklein, line 79: Should we also remove old diff images that now pass?

almarklein:
> A white diff means that all pixels are different, right?

Oh wait, they are rgba images, and the alpha channels likely match, so the alpha in the diff is zero. LOL

tests/test_examples.py
Korijn: If a test fails, it will write the new image and the diff image to the `screenshots/diffs` subfolder.

That way you can look at the original in the `screenshots` folder, and the new and the diff images in the `diffs` subfolder.

Is that not what you want?

Korijn:
> > A white diff means that all pixels are different, right?
> 
> Oh wait, they are rgba images, and the alpha channels likely match, so the alpha in the diff is zero. LOL

Right!! That makes a ton of sense :)

tests/test_examples.py
Korijn: Oh, now I get it. You mean, from previous test runs. You could add that, yeah.

Korijn:
We could generate two diffs, one for rgb and one for alpha?

tests/test_examples.py
almarklein: nah

almarklein:
I think it would be good if we could distinguish between examples that we only want to run-test, and examples that we want to compare the screenshot of. As it is currently, if you run with `--regenerate-screenshots`, it will make screenshots for all examples that are marked with  `# test_example = true`. Could be an extra flag, or maybe change the `test_example` var into an enum.

Korijn:
Summary of the status:

* Renders come out differently depending on the backend (vulkan, lavapipe, directx, metal).
* The differences are not just per-pixel, but bigger than that. E.g. lines come out thinner/thicker. This makes it non-trivial to compare screenshots when running locally (non-lavapipe) against those generated on CI (lavapipe)

I want to try a few things:

* How do specifically vulkan and lavapipe compare?
* Can we run lavapipe locally (maybe using docker)?

Korijn:
I added a script to be able to run with lavapipe locally. Install podman or docker on your system (or in WSL) and use it like this:

`python scripts\docker.py [--build] {...}`

If you pass `--build`, the `scripts\Dockerfile` will be built to produce the docker image needed to run.

Anything else you pass to the command will be executed in the container. Alternatively if you pass nothing, you will get an interactive bash session.

You can do pretty much anything you can do in development in there, and all important folders and files are mounted so you can edit them on your host/IDE as you are used to. Any screenshots generated also end up on your host, so you can commit them.

Now I'm going to see what happens if I use Vulkan on my host and compare.

Korijn:
I've compared D3D12, Vulkan and lavapipe locally, all combinations seem to pass now. Not sure what is going on. So I've commited my D2D12 generated screenshots now... let's see what happens.

Korijn:
@almarklein do you know what functions from wgpu I would need to call to determine if we are running against lavapipe?

almarklein:
```
adapter.properties["adapterType"]  # "DiscreteGPU", "IntegratedGPU", "CPU", of "Unknown"
```
En de adapter kun je in pygfx van de renderer krijgen:
```
renderer._shared.adapter
```

Korijn:
Here's what I get in the docker container when running test_examples:

`{'name': '', 'vendorID': 65541, 'deviceID': 0, 'driverDescription': '', 'adapterType': 'CPU', 'backendType': 'Vulkan'}`

And this is what I get on my host:

{'name': '', 'vendorID': 4318, 'deviceID': 9632, 'driverDescription': '', 'adapterType': 'DiscreteGPU', 'backendType': 'Vulkan'}

So that's clearly distinguishable, I think. Good!

almarklein:
Yeah, so the "DiscreteGPU" means you have dedicated hardware (e.g. an nvidea card), "IntegratedGPU" means e.g. intel graphics, and "CPU", means "emulated" GPU. I guess that the combination of adapterType being "CPU" with the backendType being Vulkan, is (at least for the time being) a good indicator that its LavaPipe.

Korijn:
OK, I'm happy with this PR for now. We can make it better in a next/third PR!

Korijn:
Just as an example, when I change a color in one of the tested examples, this is the error:

image

And two diffs are generated, the alpha diff is of course black (no difference) and the other one neatly shows the rgb diff:

validate_helpers2-rgb

scripts/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Set the working directory, and install development dependencies
+WORKDIR /app
+COPY dev-requirements.txt .
+RUN pip install -r dev-requirements.txt \
+    && pip cache purge --no-input
+
+# Install pygfx in editable mode
+# (we only copy the files that are actually needed for the install to succeed)
+COPY pygfx/__init__.py pygfx/__init__.py
+COPY setup.py README.md LICENSE .
```
Korijn, line 28: This could be one line/layer

Korijn:
Ready to merge!

almarklein:
```
$ git log
commit 91998bc98ef9d47006678ae271033af6df0716c2 (HEAD -> test-examples, origin/test-examples)
Author: Korijn van Golen <k.vangolen@clinicalgraphics.com>
Date:   Sun Feb 13 19:34:23 2022 +0100

    Add reference screenshot testing

$ python scripts/container.py --build --volumes pytest tests/test_examples.py
+ docker build --platform linux/amd64 -t pygfx -f scripts/Dockerfile .
[+] Building 1.3s (18/18) FINISHED
 => [internal] load build definition from Dockerfile                                                                 0.0s
 => => transferring dockerfile: 37B                                                                                  0.0s
 => [internal] load .dockerignore                                                                                    0.0s
 => => transferring context: 2B                                                                                      0.0s
 => [internal] load metadata for docker.io/library/ubuntu:20.04                                                      1.1s
 => [internal] load build context                                                                                    0.0s
 => => transferring context: 279.74kB                                                                                0.0s
 => [ 1/13] FROM docker.io/library/ubuntu:20.04@sha256:669e010b58baf5beb2836b253c1fd5768333f0d1dbcb834f7c07a4dc93f4  0.0s
 => CACHED [ 2/13] RUN apt-get update -y -qq --fix-missing     && apt-get install -y -qq --no-install-recommends cu  0.0s
 => CACHED [ 3/13] WORKDIR /app                                                                                      0.0s
 => CACHED [ 4/13] COPY dev-requirements.txt .                                                                       0.0s
 => CACHED [ 5/13] RUN pip install -r dev-requirements.txt     && pip cache purge --no-input                         0.0s
 => CACHED [ 6/13] COPY pygfx/__init__.py pygfx/__init__.py                                                          0.0s
 => CACHED [ 7/13] COPY setup.py README.md LICENSE .                                                                 0.0s
 => CACHED [ 8/13] RUN pip install -e .     && pip install --no-deps --install-option="--install-dir=/usr/local/lib  0.0s
 => CACHED [ 9/13] COPY pygfx/ pygfx/                                                                                0.0s
 => CACHED [10/13] COPY tests/ tests/                                                                                0.0s
 => [11/13] COPY examples/ examples/                                                                                 0.0s
 => [12/13] COPY docs/ docs/                                                                                         0.0s
 => [13/13] COPY conftest.py setup.cfg .                                                                             0.0s
 => exporting to image                                                                                               0.0s
 => => exporting layers                                                                                              0.0s
 => => writing image sha256:69f70fe5848aca45f0bbf42066cd1795d0465e1ce5d388eabde82be2b305b796                         0.0s
 => => naming to docker.io/library/pygfx                                                                             0.0s

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
+ docker run --rm -ti --platform linux/amd64 -v /Users/almar/dev/py/pygfx/pygfx:/app/pygfx -v /Users/almar/dev/py/pygfx/tests:/app/tests -v /Users/almar/dev/py/pygfx/examples:/app/examples -v /Users/almar/dev/py/pygfx/docs:/app/docs -v /Users/almar/dev/py/pygfx/conftest.py:/app/conftest.py -v /Users/almar/dev/py/pygfx/setup.cfg:/app/setup.cfg pygfx pytest tests/test_examples.py
================================================== test session starts ===================================================
platform linux -- Python 3.9.10, pytest-7.0.0, pluggy-1.0.0
rootdir: /app
collected 10 items

tests/test_examples.py F..F...F.F                                                                                  [100%]

======================================================== FAILURES ========================================================
____________________________________________ test_examples[validate_culling] _____________________________________________

module = 'validate_culling', pytestconfig = <_pytest.config.Config object at 0x4001660f10>

    @pytest.mark.parametrize("module", examples_to_test)
    def test_examples(module, pytestconfig):
        """Run every example marked for testing."""

        # render
        example = importlib.import_module(f"examples.{module}")
        img = example.canvas.draw()

        # check if _something_ was rendered
        assert img is not None and img.size > 0

        # we skip the rest of the test if you are not using lavapipe
        # images come out subtly differently when using different wgpu adapters
        # so for now we only compare screenshots generated with the same adapter (lavapipe)
        # a benefit of using pytest.skip is that you are still running
        # the first part of the test everywhere else; ensuring that examples
        # can at least import, run and render something
        if not is_lavapipe:
            pytest.skip("screenshot comparisons are only done when using lavapipe")

        # regenerate screenshot if requested
        screenshot_path = screenshots_dir / f"{module}.png"
        if pytestconfig.getoption("regenerate_screenshots"):
            imageio.imwrite(screenshot_path, img)

        # if a reference screenshot exists, assert it is equal
        if screenshot_path.exists():
            stored_img = imageio.imread(screenshot_path)
            # assert similarity
            is_similar = np.allclose(img, stored_img, atol=1)
            update_diffs(module, is_similar, img, stored_img)
>           assert is_similar, (
                f"rendered image for example {module} changed, see "
                f"the {diffs_dir.relative_to(ROOT).as_posix()} folder"
                " for visual diffs (you can download this folder from"
                " CI build artifacts as well)"
            )
E           AssertionError: rendered image for example validate_culling changed, see the examples/screenshots/diffs folder for visual diffs (you can download this folder from CI build artifacts as well)
E           assert False

tests/test_examples.py:102: AssertionError
-------------------------------------------------- Captured stderr call --------------------------------------------------
error: XDG_RUNTIME_DIR not set in the environment.
error: XDG_RUNTIME_DIR not set in the environment.
error: XDG_RUNTIME_DIR not set in the environment.
error: XDG_RUNTIME_DIR not set in the environment.
WARNING: lavapipe is not a conformant vulkan implementation, testing use only.
____________________________________________ test_examples[validate_helpers2] ____________________________________________
[...]
_________________________________________ test_examples[validate_mesh_colormap] __________________________________________
[...]
_____________________________________________ test_examples[validate_volume] _____________________________________________
[...]

tests/test_examples.py:102: AssertionError
================================================ short test summary info =================================================
FAILED tests/test_examples.py::test_examples[validate_culling] - AssertionError: rendered image for example validate_cu...
FAILED tests/test_examples.py::test_examples[validate_helpers2] - AssertionError: rendered image for example validate_h...
FAILED tests/test_examples.py::test_examples[validate_mesh_colormap] - AssertionError: rendered image for example valid...
FAILED tests/test_examples.py::test_examples[validate_volume] - AssertionError: rendered image for example validate_vol...
============================================== 4 failed, 6 passed in 55.28s ==============================================

**Comment by Korijn** [comment]
What? So.. can you share some diffs? What could be the source of the difference? Is there randomness involved somewhere? Or floating point differences? We've isolated the backend variable now, so that can't be it...

**Comment by almarklein** [comment]
This is indeed strange. So much for reproducibility with Docker :(  - I had to ran it a few times and double-check that I was on the right commit, before I believed it. Maybe the emulation spoils it?

E.g. culling:
validate_culling-alpha
validate_culling-rgb

Helpers2 is weird:
validate_helpers2-alpha
validate_helpers2-rgb

The others are just sporadic pixels being off.

**Comment by almarklein** [comment]
The weird marix-style error is probably due to the gradient that that example has.

**Comment by Korijn** [comment]
It does seem like emulation of floating point is hard to get right if I google that. Damn. Now I wonder if an arm build of the docker container could work.

**Comment by almarklein** [comment]
Except that we don't have wgpu binaries for Linux aarch64 :P

**Comment by Korijn** [comment]
> Except that we don't have wgpu binaries for Linux aarch64 :P

Doh.

**Comment by Korijn** [comment]
So I guess we have the following options to consider:

- Create linux aarch64 binaries for wgpu-native and wgpu-py, and see if that resolves the problem.
- Add a special CI job that you can trigger manually to regenerate screenshots and commit them to a branch. Not ideal, not having local development utilities means you have to resort to a trial and error style workflow (slow feedback loop).
- Try and find a similarity metric that accepts the differences between all possible wgpu backends.
- Don't support Apple Silicon, and rely on CI and devs with normal machines ;) to deal with screenshot verification.

**Comment by almarklein** [comment]
I would suggest that the set of examples that we generate and compare screenshots for, is a subset of the examples that we try to run like we do now. That way, we can run-test most examples, but keep screenshot-testing to e.g. only the validation examples. That should make it easier to manage, regardless of the direction we take.

----

I'm trying to think how the workflow will look like when a dev creates/updates a validation example. I think that in any case, the reference images should only be generated on CI. If the reference images would be generated locally, you'd get a workflow like this:

* Write code.
* Check visual output locally by eye.
* Generate images locally.
* Commit.
* If CI produces different images:
  * Update tolerances.
  * Go back to "Generate images".
* If another dev needs yet other tolerances, that dev might create a new PR, and this change affects how CI will compare the images.

So let's start by only doing screenshot comparisons on CI:

* Write code.
* Check visual output locally by eye.
* Make CI generate reference images (using a dedicated Actions workflow).
* Download the images to the local screenshots folder, and double-check by eye.
* Commit.

Then later, we can still support local screenshot comparisons by introducing tolerances. But these tolerances will *not* be used on CI - only locally.

**Comment by berendkleinhaneveld** [comment]
I think I would also go with just screenshot generation on CI.

* Add new pipeline that:
    * generates (new) screenshots
    * optionally commit those screenshots, dependent on arguments/environment
* Create user script that has two functions:
    * (wait and) download images for current pushed commit, so the user can inspect / diff them locally
    * trigger the pipeline to generate and commit new screenshots

We could leverage this [workflow example](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#accessing-and-using-event-properties) to prevent people from making manual changes to the screenshots and only allow for CI to actually update the screenshots.

**Comment by Korijn** [comment]
>   * (wait and) download images for current pushed commit, so the user can inspect / diff them locally
>   * trigger the pipeline to generate and commit new screenshots

I realized later that this part is potentially very hard, because it requires authentication etc.

> We could leverage this [workflow example](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#accessing-and-using-event-properties) to prevent people from making manual changes to the screenshots and only allow for CI to actually update the screenshots.

Well, not directly; how would we detect that the images were generated and committed by CI?

**Comment by berendkleinhaneveld** [comment]
Triggering workflows/pipelines can be done through GitHub CLI tool, and authentication is done through the tool itself (https://cli.github.com/manual/gh_auth).

Downloading artifacts probably doesn't need auth as it is a public project.

And I think you're right about that example: that wouldn't do it. Actually, it does not really matter who made/committed the images: as long as they will be the same as what CI generates, it is fine! Still would be handy to let an action generate and commit the new images in the context of a pull request.

**Comment by almarklein** [comment]
> Triggering workflows/pipelines can be done through GitHub CLI tool, 

And with a button on the GH action page, if the yml has the right flag.

**Comment by Korijn** [comment]
Decided on the following plan:

* Remove the docker emulation thing, sadly didn't solve the issue
* Introduce separate comment to toggle screenshot verification independently from just verifying an example can run without error
* Skip screenshot verification tests if not running on CI + Linux + (Software-renderer) Vulkan
* Introduce separate github action workflow to generate screenshots, should be easy to trigger manually, and download screenshots so you can commit them to your branch

All other improvements/enhancements are left for future work.

**Comment by Korijn** [comment]
Since I force pushed earlier some code was lost that I want to keep around for future reference, so I dug it up from the reflog, here it is:

```python
def images_similar(im1, im2, *, atol):
    """Comparing pixels entry-wise with an absolute tolerance
    threshold. If there are mismatching pixels, scan the neighbouring
    pixels of each of the mismatches, to see if any of those match
    within the threshold.

    This allows for minor pixel-shift differences
    that have been observed between the lavapipe Vulkan software
    renderer implementation used in CI, and the other wgpu rendering
    backends used outside of CI."""
    mask_rgba = np.isclose(im1, im2, atol=atol)
    mask = (
        mask_rgba[:, :, 0]
        & mask_rgba[:, :, 1]
        & mask_rgba[:, :, 2]
        & mask_rgba[:, :, 3]
    )

    neighbours_plus = [(-1, 0), (1, 0), (0, -1), (0, 1)]
    neighbouts_cross = [(-1, -1), (1, -1), (-1, 1), (1, 1)]
    neighbours8 = neighbours_plus + neighbouts_cross

    yy, xx = np.where(~mask)
    for x, y in zip(xx, yy):
        assert not mask[y, x]  # make sure we didn't confuse x and y
        for dx, dy in neighbours8:
            rgba1 = im1[y + dy, x + dx]
            rgba2 = im2[y + dy, x + dx]
            if np.allclose(rgba1, rgba2, atol=atol):
                mask[y, x] = True
                break
        else:
            print(f"Nope, still failing at img[{y},{x}]")

    return np.all(mask)
```

Korijn:
> * Remove the docker emulation thing, sadly didn't solve the issue
> * Introduce separate comment to toggle screenshot verification independently from just verifying an example can run without error
> * Skip screenshot verification tests if not running on CI + Linux + (Software-renderer) Vulkan

This part is done.

> * Introduce separate github action workflow to generate screenshots, should be easy to trigger manually, and download screenshots so you can commit them to your branch

Still need to do this.

Korijn:
Alright, the workflow to regenerate screenshots was added. I've configured it to also run for any pull request by default, so you don't need to wait as much. :)

README.md
```diff
@@ -90,14 +90,38 @@ pip install -U https://github.com/pygfx/pygfx/archive/main.zip
...
+
+When running the test suite with `--slow` (enabled on CI), pytest will run every example in a subprocess,
+to see if it can run and exit cleanly. You can opt out of this mechanism by including the comment
+`# run_example = false` in the module.
+
+### Type 2: Checking if examples output an image
+
+You can also (independently) opt-in to output testing for examples, by including the comment
+`# test_example = true` in the module. Output testing means the test suite will
+attempt to import the `canvas` instance global from your example, and call `draw()` on it
```
almarklein, line 107: Mmm, we're also hiding the canvas behind the renderer since #275. Can we use the global renderer instead?

README.md
```diff
@@ -90,14 +90,38 @@ pip install -U https://github.com/pygfx/pygfx/archive/main.zip
...
+to see if an image is produced.
+
+To support this type of testing, ensure the following requirements are met:
 
 * The `WgpuCanvas` class is imported from the `wgpu.gui.auto` module.
 * The `canvas` instance is exposed as a global in the module.
+* A rendering callback has been registered with `canvas.request_draw(fn)`.
+
+Reference screenshot are available in the `examples/screenshots` folder,
```
almarklein, line 116:
```suggestion
Reference screenshots are stored in the `examples/screenshots` folder,
```

README.md
```diff
@@ -90,14 +90,38 @@ pip install -U https://github.com/pygfx/pygfx/archive/main.zip
...
+the test suite will compare the rendered image with the reference.
+
+Note: this step will be skipped when not on running CI. Since images will have subtle differences
+depending on the system on which it was rendered, that would make the tests unreliable.
```
almarklein, line 120:
```suggestion
Note: this step will be skipped when not running on CI. Since images will have subtle differences
depending on the system on which they are rendered, that would make the tests unreliable.
```

tests/examples/test_examples.py
```diff
@@ -0,0 +1,148 @@
...
+        f"the {diffs_dir.relative_to(ROOT).as_posix()} folder"
+        " for visual diffs (you can download this folder from"
+        " CI build artifacts as well)"
+    )
+
+
+def update_diffs(module, is_similar, img, stored_img):
+    diffs_dir.mkdir(exist_ok=True)
+    diffs = {
+        diffs_dir / f"{module}-rgb.png": np.abs(stored_img[..., :3] - img[..., :3]),
```
almarklein, line 133: I suspect these should be `.astype(np.int16)`'d before subtracting them, to avoid overflow errors. Also wondering whether you want to amplify the error. If the red channel differs by 5, that'd be hard to see.

almarklein:
Nice!

It looks like now the image test is run for every Ubuntu build. If these tests take significantly longer, we can also create a separate single build for them, similar to the Linting step.

I think you accidentally added `offscreen.png`.

Korijn:
> Nice!
> 
> It looks like now the image test is run for every Ubuntu build. If these tests take significantly longer, we can also create a separate single build for them, similar to the Linting step.

They don't take very long at all!

> I think you accidentally added `offscreen.png`.

Will remove

README.md
Korijn: Well, I'm specifically relying on the OffscreenCanvas-only `draw()` method, which is not available on the Renderer. So, I'm not sure, what do you think?

tests/examples/test_examples.py
Korijn: Done!

README.md
almarklein:
```py
im = example.renderer.target.draw()
```

Korijn:
Done! You like it? :)

.github/workflows/ci.yml
```diff
@@ -58,16 +58,20 @@ jobs:
...
-          - name: Test Linux py38
+          - name: Test py38
             os: ubuntu-latest
             pyversion: '3.8'
-          - name: Test Linux py39
+          - name: Test py39
             os: ubuntu-latest
             pyversion: '3.9'
-          - name: Test Linux py310
+          - name: Test examples py39
```
almarklein, line 70: If we make this a separate job (like the lint-build above), it would get rid of most the `if`'s. I think the `Install llvmpipe` step is also only needed for this build. We could also do this in a later PR to get the merges quicker.

README.md
Korijn: Done!

.github/workflows/ci.yml
Korijn: Done!

README.md
```diff
@@ -90,14 +90,38 @@ pip install -U https://github.com/pygfx/pygfx/archive/main.zip
...
-## Example testing
+## Testing examples
 
-Include the comment `# test_example = true` in an example to have pytest run it as part of the test suite.
+There are two types of tests for examples included with pygfx:
 
-To support testing an example, ensure the following requirements are met:
+### Type 1: Checking if examples can run
+
+When running the test suite with `--slow` (enabled on CI), pytest will run every example in a subprocess,
```
almarklein, line 99: The `--slow` is no more, right?

almarklein:
LGTM!

PR #241, opened 2022-01-28
Run examples on CI

Korijn:
Simpler variant of #239 that we can merge more quickly. Only runs examples, doesn't do screenshot verification.

I'd like to merge this and put #239 in draft for a while so we can think about it some more.


almarklein:
Good call.

PR #238, opened 2022-01-25
Improved colormaps support

almarklein:
Closes #104.

* [x] Proper support for colormaps that are gray, gray+alpha, rgb, rgba.
* [x] Per-vertex colors can be gray, gray+alpha, rgb, rgba.
* For Image:
  * [x] proper support for gray, gray+alpha, rgb, rgba.
  * [x] Colormaps.
* For Volume and VolumeSlice:
  * [x] proper support for gray, gray+alpha, rgb, rgba.
  * [x] Colormaps
* For Mesh:
  * [x] Refactor colormaps.
  * [x] Remove clim property. 
  * [x] Per-vertex colors.
* For MeshSlice:
   * [x] Added note about not implementing colormapping or per-vertex colors for now.
* For Line (and ThinLine):
  * [x] Colormaps.
  * [x] Per-vertex colors (was already implemented).
* For points:
  * [x] Colormaps.
  * [x] Per-vertex colors (was already implemented).
* Do I forget objects, maybe background?
* [ ] A few standard colormaps.
* [ ] A complete lib of builtin colormaps.
* [x] Docs.
* [x] Self review / check todos.
* [x] Rebase and update validation example to be tested on CI.

I'm building this list up for completeness, not everything will go in this PR.

----
image

image

image

image

PR #242, opened 2022-01-31
More colormaps

almarklein:
Sort of an addendum to #238 

* [x] Add a few colormaps from MPL.
* [x] Use these in two examples.
* [x] Add a validation test for colormaps.

Builtin colormaps in use:
image
image

and the new validation example:

image


Korijn:
LGTM!

PR #245, opened 2022-02-11
Camera fixes

almarklein:
I was playing with a very simple 2D points visualization, and I ran into some issues. This fixes these.

* [x] The panzoom contro's default initialization is odd for 2D data. I think these numbers were taken from the orbit controls. Let's assume the data is 2D.
* [x] With a perspective camera, points further away from the camera are smaller. We may add an option for this later, but for now this should not be.

### DIFF
--- a/pygfx/controls/_panzoom.py
+++ b/pygfx/controls/_panzoom.py
@@ -17,9 +17,9 @@ class PanZoomControls:
     ) -> None:
         self.rotation = Quaternion()
         if eye is None:
-            eye = Vector3(50.0, 50.0, 50.0)
+            eye = Vector3(0, 0, 0)
         if target is None:
-            target = Vector3()
+            target = Vector3(eye.x, eye.y, eye.z - 100)
         if up is None:
             up = Vector3(0.0, 1.0, 0.0)
         self.zoom_value = zoom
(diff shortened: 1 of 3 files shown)


pygfx/controls/_panzoom.py
```diff
@@ -17,9 +17,9 @@ def __init__(
     ) -> None:
         self.rotation = Quaternion()
         if eye is None:
-            eye = Vector3(50.0, 50.0, 50.0)
+            eye = Vector3(0, 0, 0)
         if target is None:
-            target = Vector3()
+            target = Vector3(eye.x, eye.y, eye.z - 100)
```
Korijn, line 22: So by default you're looking towards negative z?

Korijn:
> With a perspective camera, points further away from the camera are smaller. We may add an option for this later, but for now this should not be.

So after this change, they'll have the same size regardless of camera distance under perspective projection?

Is that consistent with lines?

pygfx/controls/_panzoom.py
almarklein: Exactly!

almarklein:
> So after this change, points will have the same size regardless of camera distance under perspective projection?
> Is that consistent with lines?

Yes, and yes. And later we'll probably add a material prop to make it behave like it did before this PR 😆   but at least for now its consistent.

Korijn:
OK then, LGTM!

PR #247, opened 2022-02-13
Fix box geometry

### LINKED ISSUE #246: Bug in box_geometry()
When you create a box that is not a perfect cube, the sides don't fit. So if in e.g. the `cube.py`example, you change to `gfx.box_geometry(100, 200, 200)`, you get this:

image

I looked into it briefly. It seems like the segments data is corect for the data, so my impression that the matrix might be wrong. @Korijn you're more familiar with this code, any ideas?

https://github.com/pygfx/pygfx/blob/b20b4570ee6455b1c860926c4aca9a55a40b2b5a/pygfx/geometries/_box.py#L59-L81

Korijn:
Closes #246 

The left/right/top/bottom planes needed to be rotated 90 degrees about their normal to account for the manner in which they are generated here. This PR fixes that and improves variable names a little.

Added a (tested!) validate example for a box with anisotropic dimensions.

image

### DIFF
--- a/examples/validate_box.py
+++ b/examples/validate_box.py
@@ -1,0 +1,31 @@
+"""
+Example showing the box geometry.
+"""
+# test_example = true
+
+from wgpu.gui.auto import WgpuCanvas, run
+import pygfx as gfx
+
+canvas = WgpuCanvas()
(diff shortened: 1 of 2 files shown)


almarklein:
Thanks!

PR #248, opened 2022-02-16
Better lines and improved helpers

### LINKED ISSUE #235: Helpers barely visible on hidpi screens
Because they make use of thin lines, which are 1 physical pixel wide, they become rather hard to see.

almarklein:
Lines:
* [x] Lines compensate for thickness-loss due to aa (our lines were too thin).
* [x] Add an `aa` prop to the line material, so the additional anti-aliasing can be disabled.
* [x] Adjust `render_mask` accordingly (non-antialised lines don't need to render in the transparency pass).

Helpers:
* [x] All helpers use the normal lines instead of the thin ones. Closes #235.
* [x] The helpers have a `thickness` arg.
* [x] Renamed Axes' `length` arg to `size` for consistency with the other helpers.

Examples:
* [x] Adjusted examples where needed.
* [x] Made validation examples use extra thick lines, so we can be more forgiving at pixel mismatches at the edges.
* [x] Remove background in validation examples to avoid pixel mismatches.

image

### DIFF
--- a/examples/multi_slice1.py
+++ b/examples/multi_slice1.py
@@ -19,7 +19,7 @@ scene = gfx.Scene()
 background = gfx.Background(None, gfx.BackgroundMaterial((0, 1, 0, 1), (0, 1, 1, 1)))
 scene.add(background)
 
-scene.add(gfx.AxesHelper(length=50))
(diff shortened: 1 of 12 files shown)


Korijn:
LGTM!

PR #251, opened 2022-02-21
Fix that lines flip into infinity under certain perspective

### LINKED ISSUE #250: (Thick) Line material displaces vertices to infinity under certain conditions

almarklein:
Fixes #250


Korijn:
Can you explain what went wrong here?

Glad to see a fix though, I played around a bit looking for a simple fix but clearly that wouldn't have worked out :)

almarklein:
Previously we divided the whole ndc by its `w` value, to make it easier to convert to screen coords. But you cannot do that. I cannot exactly explain why that caused what we were seeing, but I do know (now) that you should leave that  `w` value intact :)

PR #252, opened 2022-03-01
Fix for strong reference cycle in WorldObject

berendkleinhaneveld:
In order for world objects to be properly garbage collected when removed from the tree structure, they can't have a strong reference to their parent.


Korijn:
LGTM! How'd you identify this one?

berendkleinhaneveld:
Found it by thinking about how objects would be deleted if elements were removed from the 'dom' in pygui :)

PR #253, opened 2022-03-01
Add 'before' argument to `WorldObject.add` method

berendkleinhaneveld:
This allows for inserting objects at specific positions in the children array. This can be useful in cases where the scene tree is dynamically rendered with a reactive library/framework (such as [pygui](https://github.com/fork-tongue/pygui) 😉).


Korijn:
LGTM

PR #258, opened 2022-03-15
Event system

### LINKED ISSUE #255: Event system
During work on [pygui](https://github.com/fork-tongue/pygui) I needed WorldObjects/Meshes to be able to handle events, so I made a subclass of Mesh to be able to handle events such as 'onClick'. Would be great to have built-in support for events instead. 
We might want to think about what kind of events to support. And which kind of objects can fire/handle events (WorldObject, Renderer?, Canvas?).

berendkleinhaneveld:
Closes #255 .

Added the following concepts:

* `Event` (and some subclasses): capture event and metadata
* `EventTarget` mixin: make objects be able to handle events
* `RootEventHandler` mixin: special `EventTarget` that manages event propagation and such

The `WgpuRenderer` has the `RootEventHandler` mixed in, so that it can serve as root event handler. The `WgpuRenderer` also gains two methods:

* `enable_events`
* `disable_events`

which adds handlers to the canvas in order to intercept all the events that a canvas might emit. These then get translated into `Event` objects (and also get picking info, if applicable) and handled by the `RootEventHandler`'s `dispatch_event` method.

In the added [example](examples/events.py) you can find how this fits together.

Further actions:
- [x] Make it possible to register with the root handler (for camera, controls, etc)
- [x] Create/update example for `set_pointer_capture`
- [x] Improve readability and naming of things
- [x] Update other examples
  - [x] Update `add_default_event_handler` calls
  - [x] Update `handle_event` usages
- [x] Update pan/zoom controls
- [x] Fix tests
- [x] Add some tests for `dispatch_event`
- [x] Update description of PR


Korijn:
I'll review this now and also add the notes that I took yesterday during our pair programming session.

Korijn:
To start, I kicked off a discussion on our observations wrt jupyter-rfb events here: https://github.com/vispy/jupyter_rfb/issues/56

examples/events.py
```diff
@@ -0,0 +1,123 @@
...
+
+geometry = gfx.box_geometry(40, 40, 40)
+material = gfx.MeshPhongMaterial()
+selected_material = gfx.MeshPhongMaterial(color="#FF0000")
+hovered_material = gfx.MeshPhongMaterial(color="#FFAA00")
+
+state = {
+    "selected": None,
+    "hovered": None,
+}
```
Korijn, line 35: I think you can make the example more readable if you just have two variables `selected_obj` and `hovered_obj`

examples/events.py
```diff
@@ -0,0 +1,123 @@
...
+camera = gfx.PerspectiveCamera(70, 16 / 9)
+camera.position.z = 400
+
+controls = gfx.OrbitControls(camera.position.clone())
+controls.add_default_event_handlers(renderer, canvas, camera)
+
+scene = gfx.Scene()
+
+geometry = gfx.box_geometry(40, 40, 40)
+material = gfx.MeshPhongMaterial()
```
Korijn, line 28: renaming this to `default_material` would make some parts of this example more easy to read and understand

pygfx/controls/_orbit.py
```diff
@@ -176,9 +176,9 @@ def update_camera(self, camera: "Camera") -> "OrbitControls":
         camera.zoom = zoom
         return self
 
-    def add_default_event_handlers(self, canvas, camera):
+    def add_default_event_handlers(self, event_root, canvas, camera):
         """Apply the default interaction mechanism to a wgpu autogui canvas."""
-        canvas.add_event_handler(
+        event_root.add_event_handler(
```
Korijn, line 181: don't forget to rename this to `renderer` still

pygfx/objects/_events.py
```diff
@@ -0,0 +1,310 @@
...
+    def current_target(self) -> "EventTarget":
+        return self._current_target
+
+    def stop_propagation(self):
+        """Stops propagation of events further along in the scene tree."""
+        self._bubbles = False
+
+    def cancel(self):
+        self._cancelled = True
+        self._bubbles = False
```
Korijn, line 89: you might want to call `self.stop_propagation()` instead

pygfx/objects/_events.py
```diff
@@ -0,0 +1,310 @@
...
+            return
+
+        target = event.target
+        while target and target is not self:
+            # Update the private current target field
+            event._current_target = target
+            target.handle_event(event)
+            if pointer_id is not None and pointer_id in EventTarget.pointer_captures:
+                # Prevent people from shooting in their foot by calling set_pointer_capture
+                # on POINTER_UP events
```
Korijn, line 297: I'm having second thoughts about this here, but I do think it's good that we prevent any and all dangling pointer captures

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -692,3 +706,61 @@ def snapshot(self):
...
+        )
+
+    def convert_event(self, event: dict):
+        """Converts Wgpu event (jupyter event dict) to Pygfx Event object,
+        adds picking info and then dispatches the event in the Pygfx event system.
+        """
+        event_type = event["event_type"]
+        target = None
+        info = None
+        if "x" in event or "y" in event:
```
Korijn, line 746: shouldn't it be `and`? you're using both in the next line

pygfx/objects/_events.py
```diff
@@ -0,0 +1,310 @@
...
+        self._cancelled = True
+        self._bubbles = False
+
+    @property
+    def is_cancelled(self):
+        return self._cancelled
+
+    def __getitem__(self, key):
+        """Make ``Event`` work as a dict as well to be compatible with the jupyter_rfb
+        event spec."""
```
Korijn, line 97: as you already mentioned yourself, we want to create a hard line between the event systems in pygfx from that in wgpu, so better not provide this compatibility

pygfx/objects/_events.py
```diff
@@ -0,0 +1,310 @@
...
+
+    @property
+    def type(self) -> str:
+        """A string representing the name of the event."""
+        return self._type
+
+    @property
+    def time_stamp(self) -> float:
+        """The time at which the event was created (in seconds). Not actual timestamp
+        so please only use for relative time measurements."""
```
Korijn, line 65: I'm not sure about glfw, but at least qt and the browser (jupyter-rfb) provide a timestamp in their events already, so that could be passed along maybe, in order for it to be the true timestamp. I'll add it here: https://github.com/vispy/jupyter_rfb/issues/56

pygfx/objects/_events.py
```diff
@@ -0,0 +1,314 @@
...
+        return self._target
+
+    @property
+    def current_target(self) -> "EventTarget":
+        """The object that is currently handling the event. During event bubbling
+        this property will be updated whenever an event bubbles up in the hierarchy."""
+        return self._current_target
+
+    @property
+    def is_cancelled(self) -> bool:
```
Korijn, line 86: personally, I would be ok leaving off `is_` just so it's shorter, but it's your call :)

pygfx/objects/_events.py
```diff
@@ -0,0 +1,314 @@
...
+        middle = ", ".join(attrs)
+        suffix = ">"
+        return "".join([prefix, middle, suffix])
+
+
+class KeyboardEvent(Event):
+    def __init__(self, *args, key, modifiers=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.key = key
+        self.modifiers = modifiers or []
```
Korijn, line 130: might want to default to a tuple `(,)` instead so it's immutable

pygfx/objects/_events.py
berendkleinhaneveld: Applied. Still have to go through the rest of the code to make it compatible with the new event objects.

pygfx/objects/_events.py
berendkleinhaneveld: I've added the `time_stamp` arg to the Event init function such that it can be passed along.

berendkleinhaneveld: I just read the web spec [here](https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp) and there the `timeStamp` is in milliseconds. Shall I change it back to be like that? Instead of seconds?

pygfx/objects/_events.py
berendkleinhaneveld: It is also asymmetrical with `bubbles`, so yeah: I've removed the `is_`.

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -692,3 +706,61 @@ def snapshot(self):
         )
 
         return np.frombuffer(data, np.uint8).reshape(size[1], size[0], 4)
+
+    def enable_events(self):
+        # TODO: check not a texture
```
Korijn, line 711: don't forget to add this check :)

pygfx/objects/_events.py
```diff
@@ -0,0 +1,310 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def dispatch_event(self, event: dict):
+        # Check for captured pointer events
+        pointer_id = getattr(event, "pointer_id", None)
+        if pointer_id is not None and pointer_id in EventTarget.pointer_captures:
+            captured_target = EventTarget.pointer_captures[pointer_id]
+            # Set the target to be the captured target
+            event._target = captured_target
```
Korijn, line 282: I feel like this needs to be done "more neatly", maybe create an `event._retarget(new_target)` method?

pygfx/objects/_events.py
berendkleinhaneveld: Done!

pygfx/renderers/wgpu/_renderer.py
berendkleinhaneveld: I've added a check for `hasattr`. Should be enough, right?

Korijn:
If you want to implement "click" events, I think here's what you would need to do:

* Define a double click interval constant
  * On windows the default is 500ms, we can make this fancy and try to query it from the OS at a later point
* Maintain a dict of pointers (similar to the pointer capture dict)
* On pointer down:
  * If the pointer is already in the dict and the time delta is within the interval:
    * Increment click count by 1
  * Otherwise initialize click count to 1
  * In the pointer dict, register the target, timestamp and click count
* On pointer up:
  * Look up the pointer in the register and get the target
  * Find the current target and the old target's common ancestor
  * Dispatch a click event on that ancestor with the click count

Korijn:
We should check if we need to use weak references to targets (throughout this event system)

pygfx/objects/_events.py
```diff
@@ -0,0 +1,351 @@
...
+                and tracked_click["target"] is event.target
+                and event.time_stamp - tracked_click["time_stamp"] < CLICK_DEBOUNCE
+            ):
+                tracked_click["count"] += 1
+                tracked_click["time_stamp"] = event.time_stamp
+            else:
+                RootEventHandler.click_tracker[pointer_id] = {
+                    "count": 1,
+                    "time_stamp": event.time_stamp,
+                    "target": event.target,
```
Korijn, line 340: this probably needs to be a weakref

pygfx/objects/_events.py
```diff
@@ -0,0 +1,351 @@
...
+            modifiers=self.modifiers,
+            ntouches=self.ntouches,
+            touches=self.touches,
+            clicks=clicks,
+            pointer_id=self.pointer_id,
+            bubbles=self.bubbles,
+            target=self.target,
+            cancelled=self.cancelled,
+            time_stamp=self.time_stamp,
+            # **self._data,
```
Korijn, line 167: lets make sure we test `copy()`

Korijn:
The implementation is really clean. Amazing job, this event system is on a level that I don't think any other graphics lib can stand up to.

pygfx/objects/_events.py
berendkleinhaneveld: I've added a simple test. If you have something fancier in mind, please let me know!

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+        Subclasses can overload this method. Events include widget
+        resize, mouse/touch interaction, key events, and more. An event
+        is a dict with at least the key event_type. For details, see
+        https://jupyter-rfb.readthedocs.io/en/latest/events.html
+        """
+        event_type = event.type
+        for callback in self._event_handlers[event_type]:
+            if event.cancelled:
+                break
+            callback(event)
```
almarklein, line 271:
```suggestion
            try:
                callback(event)
            except Exception:
                log_exception(f"Error during handling {event_type} event", err)
```

And you'd need this, taken from wgpu-py.
```py
import logging

logger = logging.getLogger("pygfx")

err_hashes = {}

def log_exception(kind, err):
    """Log the given exception instance, but only log a one-liner for
    subsequent occurances of the same error to avoid spamming (which
    can happen easily with errors in the event handlers).
    """
    msg = str(err)
    msgh = hash(msg)
    if msgh not in err_hashes:
        # Provide the exception, so the default logger prints a stacktrace.
        # IDE's can get the exception from the root logger for PM debugging.
        err_hashes[msgh] = 1
        logger.error(kind, exc_info=err)
    else:
        # We've seen this message before, return a one-liner instead.
        err_hashes[msgh] = count = err_hashes[msgh] + 1
        msg = kind + ": " + msg.split("\n")[0].strip()
        msg = msg if len(msg) <= 70 else msg[:69] + "…"
        logger.error(msg + f" ({count})")
```

almarklein: mmm ... with a minor change in wgpu-py, we might re-use its `log_exception` function ...

I could pick this up in a later PR.

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+        event is encountered.
+        """
+        EventTarget.pointer_captures[pointer_id] = self
+
+    def release_pointer_capture(self, pointer_id):
+        """Release the pointer capture for the object that was registered
+        to the given pointer_id.
+        """
+        if pointer_id in EventTarget.pointer_captures:
+            del EventTarget.pointer_captures[pointer_id]
```
almarklein, line 285:
```suggestion
       EventTarget.pointer_captures.pop(pointer_id, None)
```

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+
+class RootEventHandler(EventTarget):
+    """Root event handler for the Pygfx event system."""
+
+    click_tracker = {}
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def dispatch_event(self, event: dict):
```
almarklein, line 297: docstring

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+        pointer_id = getattr(event, "pointer_id", None)
```
almarklein, line 298: If no id is present, can we assume that it's just the mouse and give it id 0 or -1 or something?

almarklein (commented):
Wow! This is looking great! Some of my comments may be because it was not actually ready for review yet :)

examples/clipping_planes.py
```diff
@@ -39,7 +39,7 @@ def create_scene(clipping_planes, clipping_mode):
 camera.position.z = 250
 
 controls = gfx.OrbitControls(camera.position.clone())
-controls.add_default_event_handlers(canvas, camera)
+controls.add_default_event_handlers(renderer, canvas, camera)
```
almarklein, line 42: I wonder whether we can remove the canvas from this list, since the renderer already has a reference to it. Handlers need it for the size and for requesting a new draw. Maybe we can replicate these methods on the renderer? We could leave this for after this PR - its big enough as it is.

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+    """
+
+    def __init__(
+        self,
+        type: Union[str, EventType],
+        *,
+        bubbles=True,
+        target: "EventTarget" = None,
+        time_stamp: float = None,
+        cancelled: bool = False,
```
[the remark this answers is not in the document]
Korijn: Yes, for copy.

pygfx/objects/_events.py
Korijn: Mouses, pens and touches are all pointers and they all have a pointer id.

Korijn:
We need a new wgpu release before we can merge this, it depends on both of these:

* https://github.com/pygfx/wgpu-py/pull/243
* https://github.com/pygfx/wgpu-py/pull/245

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+        **kwargs,
+    ):
+        self._type = type
+        # Using perf_counter_ns instead of perf_counter
+        # should give us a bit more accuracy
+        self._time_stamp = time_stamp or perf_counter_ns() / 1000000
```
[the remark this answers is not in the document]
almarklein: I wouldn't have guessed that it mattered for precision.

> Note that the standard on the web is an integer in milliseconds.

Yes, but this is Python :)  or do you mean that since this system is so similar to web events, that it might make sense to follow the web here too? (which is a valid argument too)

> I don't mind which choice we make ultimately.

Indeed, it is not really clear which is more natural, and it also does not matter so much. 🤷

pygfx/objects/_events.py
almarklein: Right, this is the generic dispatcher.

Shouldn't this be `if isinstance(event, PointerEvent)` then?

Korijn: This is shorter! :)

pygfx/objects/_events.py
```diff
@@ -0,0 +1,358 @@
...
+        self._target = target
+        self._update_current_target(target)
+
+    def _update_current_target(self, target):
+        self._current_target = target
+
+    def __getitem__(self, key):
+        """Make the extra kwargs available directly on the event object through
+        bracket syntax."""
+        return self._data[key]
```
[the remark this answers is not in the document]
Korijn: This is a tough one, but we decided against it. The renderer converts wgpu events to pygfx events, and that's the entry point. From there on, everything in pygfx uses pygfx events.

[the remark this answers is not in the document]
almarklein: Ok. Related, what is the use-case for `_data`? Do we use it in pygfx itself?

pygfx/objects/_events.py
berendkleinhaneveld: Great suggestion, I'll leave it up to you! I'll resolve this comment for now then.

berendkleinhaneveld: Actually, I'll put this in place already, and then you can update / rework this part later.

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -692,3 +706,63 @@ def snapshot(self):
...
+            self._target.add_event_handler(
+                self.convert_event,
+                "key_down",
+                "key_up",
+                "pointer_down",
+                "pointer_move",
+                "pointer_up",
+                "wheel",
+                "close",
+                "resize",
```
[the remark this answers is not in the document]
berendkleinhaneveld: Yes, great suggestion. I've also added some doc strings while I was at it. So not many lines have been saved ;)

pygfx/objects/_events.py
berendkleinhaneveld: I've updated the docstring, could you check if it is sufficient?

almarklein: LGTM!

pygfx/objects/_events.py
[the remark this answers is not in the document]
berendkleinhaneveld: Hmm, it is currently used for adding picking info, but I guess we could make an actual property for that instead.

pygfx/objects/_events.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def __init__(
+        self,
+        type: Union[str, EventType],
+        *,
+        bubbles=True,
+        target: "EventTarget" = None,
+        time_stamp: float = None,
+        cancelled: bool = False,
+        **kwargs,
```
almarklein, line 73: Kwargs come in now, but are simply ignored.
```suggestion
```

berendkleinhaneveld: I do use this functionality to convert from `wgpu-py` events to these events, so I would like to keep this in here.

pygfx/helpers/_gizmo.py
```diff
@@ -542,15 +547,15 @@ def handle_event(self, event, renderer, camera):
             # De-highlight the object
             self._highlight()
             self._canvas.request_draw()
-            event["stop_propagation"] = True
+            self.release_pointer_capture(event.pointer_id)
```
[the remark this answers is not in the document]
berendkleinhaneveld: Yes, indeed. `release_pointer_capture` is mostly needed for when you want to release the capture _before_ a `pointer_up` event. Do you want me to remove this line?

pygfx/objects/_events.py
```diff
@@ -0,0 +1,411 @@
...
+    ):
+        super().__init__(*args, **kwargs)
+        self.x = x
+        self.y = y
+        self.button = button
+        self.buttons = buttons or ()
+        self.modifiers = modifiers or ()
+        self.ntouches = ntouches
+        self.touches = touches or {}
+        self.pick_info = (pick_info and pick_info.copy()) or {}
```
Korijn, line 165: why does pick_info need to be copied?

pygfx/objects/_events.py
```diff
@@ -0,0 +1,411 @@
...
+        )
+        return result
+
+
+class WheelEvent(PointerEvent):
+    def __init__(self, *args, dx, dy, modifiers=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.dx = dx
+        self.dy = dy
+        self.modifiers = modifiers or ()
```
Korijn, line 195: PointerEvent already has modifiers

pygfx/objects/_events.py
berendkleinhaneveld: Hmmm, yeah, actually no need :/

pygfx/helpers/_gizmo.py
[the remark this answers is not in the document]
Korijn: I'd say so, yes?

pygfx/objects/_events.py
berendkleinhaneveld: Fixed by adding just the `event_type` argument that will be ignored instead of every kwarg.

PR #265, opened 2022-03-23
Add triangle example

almarklein:
Compare this with https://github.com/pygfx/wgpu-py/blob/main/examples/triangle.py :)


almarklein:
Arg, I accidentally undid a necessary change.

almarklein:
Fixed.

Korijn:
Now I'm wondering how much smaller still this could be if a plotting lib was built on top of this. :) Anyway, awesome!

PR #268, opened 2022-03-25
WIP: fix subplots  v2

### LINKED ISSUE #266: Events and subplots

I'm working on a POC.

almarklein:
This is another attempt to fix #266. This time, the renderer sets the viewport on the camera, and other components can use that. Note that the renderer already set the viewport size for the camera (so the camera could adjust to the aspect ratio). Now it can also adjust for the offset.

Todo:
* [ ] Also update orbit controls.
* [ ] Revert examples that I changed for testing.
* [ ] Give camera a simple API so that controller et al. don't have to use the private `_viewport` prop.
* [ ] Give renderer a `request_draw` method.
* [ ] Remove `canvas` from the args of the methods that register default events (controllers and gizmo).

### DIFF
--- a/examples/helpers_gizmo.py
+++ b/examples/helpers_gizmo.py
@@ -32,8 +32,10 @@ def animate():
     # We render the scene, and then the gizmo on top,
     # as an overlay, so that it's always on top.
     controls.update_camera(camera)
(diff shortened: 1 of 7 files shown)


almarklein:
Superseded by #275

PR #273, opened 2022-03-30
WIP: fix subplots v3

almarklein:
This is another attempt to fix https://github.com/pygfx/pygfx/issues/266. This is my personal favorite. It introduces a new `View` class. This is basically a small helper object to hold together the renderer, scene, camera, and viewport. It helps targeting events better, and enables writing simpler code.

This PR also puts some some stuff on the renderer, so that user code will hardly ever need the canvas object. This simplifies example code some more, because there is one less object to deal with.

Todo if we go further:
* [ ] Update any camera's and controllers that I did not yet update.
* [ ] Update examples.
* [ ] Maybe also include `controls` in the view, so it can auto-update the camera?
* [ ] Update docs.

### DIFF
--- a/pygfx/renderers/wgpu/_renderer.py
+++ b/pygfx/renderers/wgpu/_renderer.py
@@ -259,6 +259,26 @@ class WgpuRenderer(RootEventHandler, Renderer):
             )
 
     @property
+    def logical_size(self):
+        """The size of the render target in logical pixels."""
+        target_psize, target_lsize = get_size_from_render_target(self._target)
+        return target_lsize
+
+    @property
+    def physical_size(self):
+        """The physical size of the internal render texture."""
+        target_psize, target_lsize = get_size_from_render_target(self._target)
+        # Determine the physical size of the render texture
+        if self._pixel_ratio:
+            pixel_ratio = self._pixel_ratio
+        else:
+            pixel_ratio = target_psize[0] / target_lsize[0]
+            if pixel_ratio <= 1:
+                pixel_ratio = 2.0  # use 2 on non-hidpi displays
+        # Determine the physical size of the internal render textures
+        return tuple(max(1, int(pixel_ratio * x)) for x in target_lsize)
+
+    @property
     def blend_mode(self):
         """The method for handling transparency:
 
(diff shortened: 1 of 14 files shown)


pygfx/renderers/wgpu/_renderer.py
```diff
@@ -354,6 +374,12 @@ def _set_wobject_pipelines(self):
             self.blend_mode, weakref.WeakKeyDictionary()
         )
 
+    def render_views(self, *views):
+        """Comvenience function to render the given views and then flush."""
+        for view in views:
+            self.render(view.scene, view.camera, viewport=view.viewport, flush=False)
```
Korijn, line 380: or... `view.renderer`? feels like this method belongs on view, not on renderer

Korijn:
* Do you still get 1 big implicit view if you don't create any View objects?
* If you don't want to use pygfx input handling and events, can you still render scenes to subregions of the screen manually, without Views?

pygfx/renderers/wgpu/_renderer.py
Korijn: Hm... at the very least I think there should be a check that `view.renderer is self`

almarklein:
> [...] we should be careful to not provide APIs that take flexibility/freedom away from users. If users want to handle subplot interaction (or interaction in general) differently, they should be able to!

I agree. And the modular (not-entangled) components are a great part of that.

However, to have all the info in the right place, you sometimes need to pass multiple objects to a function, e.g. `render()` and `add_default_handler()`. I think that the core problem of #266 is that viewport is yet another piece that needs to be passed around in order for things to work. We *can* do that the verbatim way, but things start to become repetitive ...

The `View` collects some of these together so it can be passed around as a whole. I think it also helps semantically. At least personally thinking of a view as one "thing" being rendered feels simpler than the whole `(renderer, camera, scene, viewport)` tuple :)

> Do you still get 1 big implicit view if you don't create any View objects?

Yes!

> If you don't want to use pygfx input handling and events, can you still render scenes to subregions of the screen manually, without using Views?

Yes, the only places where a view is required is `renderer.render_views()` (which is optional), and `add_default_event_handlers()`. It should be possible to support both `xxxx.add_default_event_handlers(view)` and `xxxx.add_default_event_handlers(renderer, camera)`.

Korijn:
I can accept the design given your clarifications, but for some reason it still feels like its purely a convenience thing.

Having to pass a class instance instead of using method arguments and keyword arguments doesn't feel like an improvement (?) and it doesn't seem pythonic. On the other hand, I have no alternative suggestions other than just live without a View class. :)

almarklein:
I've proposed a new approach, by just providing a way to make the controller aware of the viewport: #274  I also updated the description of #266.

My current take is to use that approach, and then add the `View` as a convenience object.

almarklein:
Superseded by #275

PR #274, opened 2022-04-01
WIP: fix subplots v0

### LINKED ISSUE #266: Events and subplots

almarklein:
Fixes #266 by making it possible to supply the viewport to the controller.

Todo:
* [x] Give renderer some functionality so it can replace the canvas.
* [x] Make panzoom controller aware of viewport.
* [x] Create new example.
* [ ] Make orbit controller aware of viewport.
* [ ] Update some more examples.


examples/scene_subplots1.py
```diff
@@ -0,0 +1,92 @@
...
+
+    w, h = renderer.logical_size
+    w2, h2 = (w - 30) / 2, (h - 30) / 2
+    viewport1 = 10, 10, w2, h2
+    viewport2 = w / 2 + 5, 10, w2, h2
+    viewport3 = 10, h / 2 + 5, w2, h2
+
+    controls1.update_camera(camera1, viewport1)
+    controls2.update_camera(camera2, viewport2)
+    controls3.update_camera(camera3, viewport3)
```
almarklein, line 81: This is how I've now supplied the controller with the viewport. An alternative could be to use a separate method.

pygfx/linalgtest.py
```diff
@@ -0,0 +1,294 @@
+"""
```
Korijn, line 1: I guess this module doesn't belong in this PR

examples/scene_subplots1.py
Korijn: I think a separate method is nicer because I can see adjustments to the viewport occurring on a "different schedule":

* When the window is resized
* When the viewports are being resized interactively (if this is implemented at application level)
* When the viewports are being reconfigured/animated (if this is implemented at application level)

Whereas the camera only really needs to be updated just before rendering.

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -713,6 +726,14 @@ def snapshot(self):
 
         return np.frombuffer(data, np.uint8).reshape(size[1], size[0], 4)
 
+    def request_draw(self, draw_function=None):
+        """Request a new draw at the target canvas. If the renderer's
```
Korijn, line 730:
```suggestion
        """Forwards a request_draw call to the target canvas. If the renderer's
```

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -713,6 +726,14 @@ def snapshot(self):
...
+        target is not a canvas (e.g. a texture) this function does
+        nothing.
+        """
+        if hasattr(self.target, "request_draw"):
+            self.target.request_draw(draw_function)
```
Korijn, line 735:
```suggestion
        request_draw = getattr(self.target, "request_draw", None)
        if request_draw:
            request_draw(draw_function)
```

Korijn:
Yeah, I like this approach the best.

pygfx/linalgtest.py
almarklein: Arg, I shouldn't put stuff like this in the repo dir :/

examples/scene_subplots1.py
```diff
@@ -0,0 +1,96 @@
...
+    viewport2 = w / 2 + 5, 10, w2, h2
+    viewport3 = 10, h / 2 + 5, w2, h2
+
+    controls1.update_camera(camera1)
+    controls2.update_camera(camera2)
+    controls3.update_camera(camera3)
+
+    controls1.update_viewport(viewport1)
+    controls2.update_viewport(viewport2)
+    controls3.update_viewport(viewport3)
```
Korijn, line 85: Just wondering if we can out this code in a resize event handler instead of the animation loop?

pygfx/controls/_base.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @property
+    def viewport(self):
+        """The viewport that this controller applies to. Can be None to indicate
+        the full canvas. This is used to correct the positions of events.
+        """
+        return self._viewport
+
+    @viewport.setter
+    def viewport(self, viewport: Tuple[float, float, float, float]):
```
Korijn, line 20: x, y, width, height? from top left to bottom right? maybe add a docstring?

almarklein:
This seemed to progress pretty well, but I really hate the `viewport` property on the controller. It only needs it when handling events, so that's where it should be passed, in the event handler. But then we cannot do `add_default_event_handlers`, because we need to update the viewport from the application code.

So what if the viewport was an object that can be changed in-place. We can pass that to `add_default_event_handlers` and also change the value from the application code. I'm working on *yet another*  POC, that sits a bit in between this one and the `View` version.

almarklein:
To clarify, this is where it's going if we take this approach the clean way, I think:

```py
@renderer.add_event_handler("pointer_down", "pointer_move", "pointer_up", "wheel")
def on_event(self, event):
    w, h = renderer.logical_size
    w2, h2 = (w - 30) / 2, (h - 30) / 2

    viewport1 = 10, 10, w2, h2
    viewport2 = w / 2 + 5, 10, w2, h2
    viewport3 = 10, h / 2 + 5, w2, h2
    viewport4 = w / 2 + 5, h / 2 + 5, w2, h2

    controls1.handle_event(event, renderer, camera, viewport=viewport1)
    controls2.handle_event(event, renderer, camera, viewport=viewport2)
    controls3.handle_event(event, renderer, camera, viewport=viewport3)
    controls4.handle_event(event, renderer, camera, viewport=viewport4)

...

def animate():

    controls1.update_camera(camera1)
    controls2.update_camera(camera2)
    controls3.update_camera(camera3)
    controls4.update_camera(camera4)
   
    w, h = renderer.logical_size
    w2, h2 = (w - 30) / 2, (h - 30) / 2

    viewport1 = 10, 10, w2, h2
    viewport2 = w / 2 + 5, 10, w2, h2
    viewport3 = 10, h / 2 + 5, w2, h2
    viewport4 = w / 2 + 5, h / 2 + 5, w2, h2

    renderer.render(scene0, camera0, flush=False)
    renderer.render(scene, camera1, viewport=viewport1, flush=False)
    renderer.render(scene, camera2, viewport=viewport2, flush=False)
    renderer.render(scene, camera3, viewport=viewport3, flush=False)
    renderer.render(scene, camera4, viewport=viewport4, flush=False)
    renderer.flush()

```

almarklein:
Superseded by #275

PR #275, opened 2022-04-07
 WIP: fix subplots v4

### LINKED ISSUE #266: Events and subplots

almarklein:
~3~ ~4~ 5 time's a charm! Fixes #266 by introducing a Viewport class.

* [x] Update all examples that use `add_default_event_handlers` to drop the canvas arg.

### DIFF
--- a/pygfx/controls/_base.py
+++ b/pygfx/controls/_base.py
@@ -1,7 +1,8 @@
-from typing import Tuple
+from typing import Tuple, Union
 
 from ..linalg import Vector3
 from ..utils.viewport import Viewport
+from ..renderers import Renderer
 
 
 class Controller:
@@ -15,8 +16,12 @@ class Controller:
         camera.zoom = zoom
         return self
 
-    def add_default_event_handlers(self, viewport, camera):
-        """Apply the default interaction mechanism to a wgpu autogui canvas."""
+    def add_default_event_handlers(
+        self, viewport: Union[Viewport, Renderer], camera: "Camera"
+    ):
+        """Apply the default interaction mechanism to a wgpu autogui canvas.
+        Needs either a viewport or renderer, pus the camera.
+        """
         viewport = Viewport.from_viewport_or_renderer(viewport)
         viewport.renderer.add_event_handler(
             lambda event: self.handle_event(event, viewport, camera),


pygfx/renderers/wgpu/_renderer.py
```diff
@@ -404,48 +433,41 @@ def render(
...
-
-        # Determine the physical size of the first and last render pass
-        framebuffer_size = tuple(max(1, int(pixel_ratio * x)) for x in logical_size)
+        pixel_ratio = physical_size[1] / logical_size[1]
 
         # Update the render targets
-        self._blender.ensure_target_size(device, framebuffer_size)
+        self._blender.ensure_target_size(device, physical_size)
+
+        # The viewpoer could be a Viewport object
```
ivoflipse, line 446: viewpoer -> viewport

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -404,48 +433,41 @@ def render(
...
+        if hasattr(viewport, "rect"):
+            viewport = viewport.rect
```
Korijn, line 448: propose to rename the kwarg `viewport` to `rect`, and always require a tuple to be passed (so no support for Viewport object to be passed in)

pygfx/utils/viewport.py
```diff
@@ -0,0 +1,68 @@
...
+            return (0, 0) + self.renderer.logical_size
+        else:
+            return self._rect
+
+    def is_inside(self, x, y):
+        """Get whether the given position is inside the viewport rect."""
+        vp = self.rect
+        return vp[0] <= x <= vp[0] + vp[2] and vp[1] <= y <= vp[1] + vp[3]
+
+    def render(self, scene, camera):
```
Korijn, line 64: maybe add `**kwargs` to allow pass through of other arguments?

examples/scene_subplots1.py
```diff
@@ -0,0 +1,110 @@
...
+def layout(event=None):
+    w, h = renderer.logical_size
+    w2, h2 = (w - 30) / 2, (h - 30) / 2
+    viewport1.rect = 10, 10, w2, h2
+    viewport2.rect = w / 2 + 5, 10, w2, h2
+    viewport3.rect = 10, h / 2 + 5, w2, h2
+    viewport4.rect = w / 2 + 5, h / 2 + 5, w2, h2
+
+
+layout()
```
Korijn, line 90: Can you add a comment to explain why this call is here? Think it will be useful for readers of this example to understand better

pygfx/controls/_base.py
```diff
@@ -0,0 +1,50 @@
+from typing import Tuple
+
+from ..linalg import Vector3
+from ..utils.viewport import Viewport
+
+
+# todo: I hate that these classes are plural, maybe rename to Controller?
```
Korijn, line 7: I don't particularly mind if you rename :)

pygfx/utils/viewport.py
```diff
@@ -0,0 +1,68 @@
...
+    """Defines a rectangular area on a renderer."""
+
+    def __init__(self, renderer, rect=None):
+        if not isinstance(renderer, Renderer):
+            raise TypeError("Viewport first arg must be a Renderer.")
+        self._renderer = renderer
+        self.rect = rect
+
+    @classmethod
+    def from_viewport_or_renderer(cls, x):
```
Korijn, line 14: so, why not just implement this in the constructor?

pygfx/helpers/_gizmo.py
```diff
@@ -493,10 +493,10 @@ def _update_visibility(self):
 
     # %% Event handling
 
-    def add_default_event_handlers(self, renderer, camera):
+    def add_default_event_handlers(self, viewport, camera):
         # Store objects that we need outside the event handling.
-        self._renderer = renderer
-        self._canvas = renderer.target
+        viewport = Viewport.from_viewport_or_renderer(viewport)
+        self._viewport = viewport
```
Korijn, line 499: this is a slight discrepancy between the gizmo and the controls, why does one need to store a reference on self but not the other?

pygfx/controls/_base.py
```diff
@@ -0,0 +1,50 @@
...
+        viewport.renderer.add_event_handler(
+            lambda event: self.handle_event(event, viewport, camera),
+            "pointer_down",
+            "pointer_move",
+            "pointer_up",
+            "wheel",
+        )
+
+    def handle_event(self, event, viewport, camera):
+        pass
```
Korijn, line 34: why are the empty methods (`__init__` and `handle_event`) needed?

pygfx/controls/_base.py
almarklein: I might do that in another PR :)

pygfx/helpers/_gizmo.py
almarklein: Because the gizmo also does a bunch of things at draw time, where it needs the viewport and rect. Will add a comment to explain.

pygfx/controls/_base.py
almarklein: Good point. In an earlier version the `__init__` wasn't empty, I will remove it, and raise notimplemented in the placeholder `handle_event`.

pygfx/utils/viewport.py
almarklein: I'll add the only arg that's left: `clear_color`.

almarklein: On second thought, that `clear_color` arg is not needed unless you're doing really special stuff (because the default behavior almost always does what you want). If you have no objections, I'd rather keep this signature nice and lean :)

examples/scene_subplots1.py
almarklein: I looked into auto-calling a callback when registering for a "resize" event, but this is too much of a rabbit-hole. I just added a comment :)

almarklein:
Ready from my end! 🚀

pygfx/controls/_base.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def update_camera(self, camera: "Camera") -> "Controller":
+        """Update the transform of the camera with the internal transform."""
+        rot, pos, zoom = self.get_view()
+        camera.rotation.copy(rot)
+        camera.position.copy(pos)
+        camera.zoom = zoom
+        return self
+
+    def add_default_event_handlers(self, viewport, camera):
```
Korijn, line 18: can you type hint viewport to be `Union[Viewport, Renderer]` or something like that, to clearly signal the duck typing here?

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -359,6 +392,7 @@ def render(
         scene: WorldObject,
         camera: Camera,
         *,
+        rect=None,
         viewport=None,
```
Korijn, line 396: I guess you forgot to remove this line

almarklein:
Done! That second-to-last commit was me renaming an example to get `scene_subplot1` (the old way) and `scene_subplot2` (the viewport way).

PR #278, opened 2022-04-09
Synchronized video example

kushalkolar:
Hi, I looked at the image and events examples to create an example of playing multiple videos at high framerate. If there are better approaches let me know! This runs at ~320Hz, I have an AMD RX 570 GPU.

Correction: I was measuring fps wrong, it gives at least 60 Hz, not sure how to see beyond that?

This also works in jupyter notebooks using: https://github.com/vispy/jupyter_rfb

By the way, I was also able to integrate this with a slider using ipywidgets to have multiple videos that can be random-access scrollable at very high speed! This could be of very high interest to the neuroscience community.

### DIFF
--- a/examples/synced_video.py
+++ b/examples/synced_video.py
@@ -20,10 +20,10 @@ colormap1 = gfx.cm.plasma
 
 # Just create a 512x512 random image
 def create_random_image():
+    rand_img = np.random.rand(512, 512).astype(np.float32) * 255
+
     return gfx.Image(
-        gfx.Geometry(
-            grid=gfx.Texture(np.random.rand(512, 512).astype(np.float32) * 255, dim=2)
-        ),
+        gfx.Geometry(grid=gfx.Texture(rand_img, dim=2)),
         gfx.ImageBasicMaterial(clim=(0, 255), map=colormap1),
     )
 
--- a/examples/synced_video.py
+++ b/examples/synced_video.py
@@ -1,0 +1,54 @@
+"""
+Example demonstrating synced video rendering
+"""
+
+import numpy as np
+from wgpu.gui.auto import WgpuCanvas, run
+import pygfx as gfx
+
+
+canvas = WgpuCanvas(size=(512 * 3 + 100, 400))
+renderer = gfx.renderers.WgpuRenderer(canvas)
+scene = gfx.Scene()
+camera = gfx.OrthographicCamera(1800, 550)
+camera.position.y = 256
+camera.scale.y = -1
+camera.position.x = 1736 / 2
+
+colormap1 = gfx.cm.plasma
+
+
+# Just create a 512x512 random image
+def create_random_image():
+    return gfx.Image(
+        gfx.Geometry(grid=gfx.Texture(np.random.rand(512, 512).astype(np.float32) * 255, dim=2)),
+        gfx.ImageBasicMaterial(clim=(0, 255), map=colormap1),
+    )
+
+
+images = list()
+img_pos = 0
+
+
+for i in range(3):
+    images.append(create_random_image())
+    images[-1].position.x = img_pos
+    img_pos += 50 + 512
+    scene.add(images[-1])
+
+
+# update with new random image
+def update_img(obj):
+    for img in images:
+        img.geometry.grid = gfx.Texture(np.random.rand(512, 512).astype(np.float32) * 255, dim=2)
+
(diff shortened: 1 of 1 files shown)


Korijn:
Very cool! Thank you for the contribution.

Have you seen [the example `geometry_plane.py`](https://github.com/pygfx/pygfx/blob/main/examples/geometry_plane.py)? It renders an actual .mp4 file to a plane, so you might find it interesting.

examples/synced_video.py
```diff
@@ -0,0 +1,54 @@
...
+    images.append(create_random_image())
+    images[-1].position.x = img_pos
+    img_pos += 50 + 512
+    scene.add(images[-1])
+
+
+# update with new random image
+def update_img(obj):
+    for img in images:
+        img.geometry.grid = gfx.Texture(np.random.rand(512, 512).astype(np.float32) * 255, dim=2)
```
Korijn, line 43: Instead of creating a new Texture, you can also just update the data. I think that would improve performance.

kushalkolar: Thanks for the suggestion!

I tried this:
`img.geometry.grid.data[:] = np.random.rand(512, 512).astype(np.float32) * 255`

This causes the image freeze, no updating

Korijn: Ah, yes! You'd need to add this in order to let the renderer know the buffer needs to be re-uploaded to the GPu on the next frame:

`img.geometry.grid.update_range((0, 0, 0), img.geometry.grid.size)`

kushalkolar:
> Have you seen [the example geometry_plane.py](https://github.com/pygfx/pygfx/blob/main/examples/geometry_plane.py)? It renders an actual .mp4 video file to a plane, so you might find it interesting.

Thanks, I didn't notice that!

I notice it uses `gfx.plane_geometry` and `gfx.MeshBasicMaterial`, are these better than `gfx.Image` and `gfx.ImageBasicMaterial` for performance?

My goal is to random-access scroll through tens of gigabytes of 4-6 videos stored as memmaped numpy arrays, along with thousands of lines. So far this library seems to have that level of performance: https://github.com/kushalkolar/wgpu-viz-prototype/blob/master/scrolling_calcium_vids.ipynb 

Here's an example of scrolling through 40GB of data with 4 videos and 1,500 lines that encircle neurons in the video.

https://user-images.githubusercontent.com/9403332/162635312-ddb19f52-c6e6-4365-adf1-9c327f548b1b.mp4

Korijn:
> I notice it uses `gfx.plane_geometry` and `gfx.MeshBasicMaterial`, are these better than `gfx.Image` and `gfx.ImageBasicMaterial` for performance?

No, there shouldn't be a performance difference. Just the APIs are different, I think an Image is better suited to your use case. :+1:

> My goal is to random-access scroll through tens of gigabytes of 4-6 videos stored as memmaped numpy arrays, along with thousands of lines. So far this library seems to have that level of performance: https://github.com/kushalkolar/wgpu-viz-prototype/blob/master/scrolling_calcium_vids.ipynb

Awesome! That should definitely be possible with pygfx. I think the code you wrote in this example should already do the trick, but it can be improved further of course. :)

If you identify any particular bottlenecks we'd love to hear, so we can improve pygfx!

examples/synced_video.py
kushalkolar: Thanks! This works: `img.geometry.grid.update_range((0, 0, 0), img.geometry.grid.size)`

Performance is still the same with 3 vids, but with 6 recreating the texture gives ~10Hz, and just updating the data gives ~20Hz!

kushalkolar:
Thanks! Is there a way to see how far the renderer can be pushed, it's not displaying over the refresh rate of my monitor (60Hz) even though `max_fps=900` is passed to `WgpuCanvas`.

Korijn:
> Thanks! Is there a way to see how far the renderer can be pushed, it's not displaying over the refresh rate of my monitor (60Hz) even though `max_fps=900` is passed to `WgpuCanvas`.

We saw this recently in another issue as well, there seems to be something else imposing a limit. I've created #279 to track this problem, we'll try to look into it soon!

examples/synced_video.py
Korijn: Awesome!

almarklein (approved):
Very clear example to demonstrate image updates!

PR #281, opened 2022-04-11
More custom object examples

### LINKED ISSUE #280: how to pass uniform parameter to shader?
The custom_object.py example shows how to create a custom object with the user supplied shader code. In that example,  it passes uniform buffer to shader at bindings0 with the pygfx-builtin  render_info.stdinfo_uniform. What's the template to define a custom uniform buffer and then pass it to shader? Does it need the device object in wgpu to create that buffer?

Thanks for the excellent work!

almarklein:
Closes #280

The examples build up to increasingly complete/realistic world objects:
1. Minimal example (what we had)
2. Add uniforms and honor camera. 
3. Add geometry and use more shader features.

This requires some stuff from #275

### DIFF
--- a/examples/custom_object1.py
+++ b/examples/custom_object1.py
@@ -1,5 +1,15 @@
 """
-Example that implements a custom object and renders it.
+Example that implements a minimal custom object and renders it.
+
+This example simply draws a triangle at the bottomleft of the screen.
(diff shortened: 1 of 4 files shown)


Korijn:
LGTM

PR #283, opened 2022-04-12
Improve performance of synced video

almarklein:
I had not spotted this in the earlier review, but the updating was done multiple times per draw. The traverse method is intended to do something with each object in the scene graph.

It should not matter *that* much, because the uploading of the data should still happen only once, but the generation of data etc. does seem to affect the FPS.

cc @kushalkolar


Korijn:
Good point! LGTM.

Issue #148, opened 2021-10-01
Improve API of Controls (Orbit) to better support saving and restoring state


berendkleinhaneveld:
OrbitControls currently takes `eye`, `target` and `up` as init parameters, but not all those values are actually updated. So saving and restoring state of a camera and controls is not as straight-forward as it could be.

Would be nice to try and make the API for the parameters a bit more symmetric: use the same parameters as input and output.

Korijn:
I propose we just mimic the API from three.js: https://threejs.org/docs/index.html?q=Controls#examples/en/controls/OrbitControls

* saveState() stores a snapshot of the camera/controller's current state
  * a snapshot is comprised of the target, position and zoom
* resetState() restores the last snapshot

almarklein:
I think this should work?
* `savestate` stores the internal state and also returns an opaque object that represents the state.
* `restorestate` restores the given state, or the internal state if none is given.

Korijn:
Exactly

PR #287, opened 2022-04-14
example with video or image updates in subplots with pan and zoom

kushalkolar:
This is a more complex example that kind of combines `synced_video.py` with subplots. Each view can be panned and zoom independently. The pan & zoom can be reset.

If there's not a method to reset `PanZoomController` I think I can easily add one? It would make this example much easier.

Demo in a notebook with `jupyter_rfb`:

https://user-images.githubusercontent.com/9403332/163316431-92b95380-47b6-4fae-8390-d16444579da3.mp4

### DIFF
--- a/examples/scene_subplots_video.py
+++ b/examples/scene_subplots_video.py
@@ -1,5 +1,6 @@
 """
-An example combining `synced_video.py` with subplots
+An example combining `synced_video.py` with subplots.
+Double click to re-center the images.
 """
 
 from wgpu.gui.auto import WgpuCanvas, run
@@ -72,35 +73,15 @@ def layout(event=None):
     viewports[3].rect = w / 2 + 5, h / 2 + 5, w2, h2
 
 
-reset_cameras = False
-
-
 def animate():
     for img in images:
         # create new image data
         img.geometry.grid.data[:] = np.random.rand(*dims).astype(np.float32) * 255
         img.geometry.grid.update_range((0, 0, 0), img.geometry.grid.size)
 
-    global reset_cameras
-
-    # reset the cameras if `reset_camera` is set to True
-    if reset_cameras:
-        # this should work in the future
-        # for camera, image in zip(cameras, images):
-        #     camera.show_object(image)
-
-        # until a way to center the controller in the scene is implemented
-        for camera, controller, cntrl_default in zip(
-            cameras, controllers, cntl_defaults
-        ):
-            controller.load_state(cntrl_default)
-            controller.update_camera(camera)
-
-        reset_cameras = False
-    else:
-        for camera, controller in zip(cameras, controllers):
-            # if not reset, update with the pan & zoom params
-            controller.update_camera(camera)
+    for camera, controller in zip(cameras, controllers):
+        # update with the pan & zoom params
+        controller.update_camera(camera)
 
     # render the viewports
     for viewport, s, c in zip(viewports, scenes, cameras):
@@ -108,6 +89,15 @@ def animate():
 
     renderer.flush()
     canvas.request_draw()
+
+
+def center_objects(ev):
+    for con, cam, img in zip(controllers, cameras, images):
+        con.show_object(cam, img)
+        con.zoom(1.3)
+
+
+renderer.add_event_handler(center_objects, "double_click")
 
 
 layout()
--- a/examples/scene_subplots_video.py
+++ b/examples/scene_subplots_video.py
@@ -62,18 +62,18 @@ for i in range(4):
     cntl_default["target"] = controller.target.clone()
     cntl_defaults.append(cntl_default)
 
-w_div = 2
-h_div = 2
 
-
-# 2x2 arrangement for the viewports
-def produce_rect(w, h):
-    return [
-        (0, 0, w / w_div, h / h_div),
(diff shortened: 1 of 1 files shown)


kushalkolar:
Might be useful to add the version with the reset button to the `jupyter_rfb` repo?

Korijn:
I guess the reset view code is somrwhat related to #148 

I'll try and define a better API for Controllers onthat issue and take resetting into account

examples/scene_subplots_video.py
```diff
@@ -0,0 +1,144 @@
...
+def animate():
+    for img in images:
+        # create new image data
+        img.geometry.grid.data[:] = np.random.rand(*dims).astype(np.float32) * 255
+        img.geometry.grid.update_range((0, 0, 0), img.geometry.grid.size)
+        # img.geometry.grid = gfx.Texture(np.random.rand(*dims).astype(np.float32) * 255, dim=2)
+
+    w, h = canvas.get_logical_size()
+
+    rects = produce_rect(w, h)
```
Korijn, line 91: I would recommend updating the viewport rects in a separate function using the canvas resize event, for an example of how to achieve this see:

https://github.com/pygfx/pygfx/blob/main/examples/scene_subplots2.py#L79-L86

examples/scene_subplots_video.py
```diff
@@ -0,0 +1,144 @@
...
+            controller.pan(pan_delta)  # pan to initial state
+
+            # set zoom and distance to initial state
+            controller.zoom_value = cntl_default["zoom_value"]
+            controller.distance = cntl_default["distance"]
+
+            # update camera with the new params
+            controller.update_camera(camera)
+
+        reset_cameras = False
```
Korijn, line 112: You could simplify the animation loop a lot by putting this code in a separate function.

I can see why you decided not to hook it up to Qt/jupyter though, for the sake of this example relying on auto-gui of course.

Korijn:
@almarklein the CI build failure is pretty bizarre, it looks like the python interpreter is failing to shut down at the end of the test suite, with a stack trace pointing to `imageio_ffmpeg`.

almarklein:
> I'm also confused we even have this ffmpeg plugin installed, it's not in our dev-requirements.txt.

Actually, it is :)

That example is run as part of our test_examples. So indeed imageio-ffmpeg is used at some point ...

I did do a release of imageio-ffmpeg yesterday ... it looks like it introduced a new bug! I will have a look, maybe we can disable testing that example for now.

almarklein:
Reported: https://github.com/imageio/imageio-ffmpeg/issues/69

Korijn:
> I did do a release of imageio-ffmpeg yesterday ... it looks like it introduced a new bug! I will have a look, maybe we can disable testing that example for now.

I didn't feel like disabling testing of an example, it works fine with the previous version of imageio-ffmpeg, so I pinned that in this PR:  https://github.com/pygfx/pygfx/pull/288

Korijn:
> I guess the reset view code is somewhat related to #148. I'll try and define a better API for Controllers on that issue and take resetting into account.

FYI this is being worked on now in #297

examples/scene_subplots_video.py
kushalkolar: Thanks, is there anything particularly not-recommended about updating the viewports like this? In `fastplotlib` I'm using this method for updating the viewports for an arbitrary `i x j` grid:

```python
def _produce_rect(j, i, ncols, nrows, w, h):
    return [
        ((w / ncols) + ((i - 1) * (w / ncols))),
        ((h / nrows) + ((j - 1) * (h / nrows))),
        (w / ncols),
        (h / nrows)
    ]
```

https://github.com/kushalkolar/fastplotlib/blob/796f291086703cb752836d69d1ef2b52cfdf337d/fastplotlib/layouts.py#L9

Korijn: Just performance optimization. It's code that technically doesn't need to run on every frame. But you're free to choose to do it anyway.

kushalkolar: ah makes sense, didn't realize that `renderer.add_event_handler` is like a callback signal (sorry I'm coming from years of working in Qt). I think I can generalize it with the event handler for an `i x j` grid.


anyways I'll update this PR soon with the even handler!

Korijn:
@kushalkolar Are you still interested in completing this work?

kushalkolar:
Yea I'm away right now (in Netherlands actually!), will get back to it soon

kushalkolar:
@Korijn updated to use the event handler for the canvas resize events, also uses `Controller.save_state()` and `load_state()` instead of calculating the deltas manually. The reset works in a notebook with `ipywidgets`, but I guess I have to wait for this to have a more elegant way to reset the panzoom:
https://github.com/pygfx/pygfx/issues/301#issuecomment-1125791272
> To be clear, the `camera.show_object` function now supports any object. However, the `Controller` classes don't have a `look_at` or `show_object`-like utility method just yet. That would be the remaining work in order to close this issue.

kushalkolar:
Might be best to wait for #301 or I can remove the reset stuff and modify when #301 is done

kushalkolar:
this is ready to go now! double click to re-center the objects

Korijn (approved):

almarklein:
Nice!

PR #289, opened 2022-04-14
bump imageio-ffmpeg

almarklein:
The issue has been fixed.

Also refactor how we define our requirements, so we can be more specific.


Korijn:
Can you also update the README?

PR #291, opened 2022-04-15
Make one example fail

almarklein:
Just to see if CI picks it up. Should be fixed with https://github.com/pygfx/wgpu-py/pull/259/ (i.e. CI should detect the error)


Korijn:
That error is pretty specific! Awesome to see this improvement in naga.

almarklein:
Yes, it's getting better with each update :)

Ok, this confirms that we do detect the failure in the shader. So we can close this!

PR #292, opened 2022-04-15
Adjust to new wgpu-py

almarklein:
* [x] Bump to latest wgpu-py (yet to be released).
* [x] Make necessary API and WGSL changes.
* [x] Update buffer bindings to use a direct array (not via a struct).
* [x] Disabled run-testing the examples that don't use auto-gui.


.gitignore
```diff
@@ -6,8 +6,8 @@ pyproject.toml
 poetry.lock
 .DS_Store
 examples/screenshots/diffs
-
-# Special for this repo
+examples/offscreen.png
+ Special for this repo
```
Korijn, line 10: Think you lost a `#` there

almarklein: Oooh nice catch. I suspect this happened when I edited this file in VI; I'm still struggling how to go into insert mode without an insert key on Mac 😆

Korijn (approved):

PR #294, opened 2022-04-16
Pointer_enter and pointer_leave events v1

### LINKED ISSUE #262: High level pointer events
As the use cases come up, we can implement the remaining pointer events.

* Over
* Enter
* Leave
* Out
* Got capture
* Lost capture
* Cancel

Reference docs: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events

I think maybe over/out are most immediately valuable in pygfx to support material swapping on hover.

Korijn:
Closes #262

* Make event.copy more generic 
* Add support for pointer_enter and pointer_leave events by tracking event targets
* Refactor events example
  * Leverage pointer_enter and pointer_leave events for global-less hover materials
  * Add option to click background to deselect
  * Showcase the power of the event system a little more, and add some code comments to explain

### DIFF
--- a/pygfx/objects/_events.py
+++ b/pygfx/objects/_events.py
@@ -347,12 +347,12 @@ class RootEventHandler(EventTarget):
             previous_target = (previous_target_ref and previous_target_ref()) or None
             # Get the current target for this pointer (if any)
             new_target = event.target
-            # Update the current target for this pointer
-            RootEventHandler.target_tracker[pointer_id] = (
-                new_target and ref(new_target)
-            ) or None
             # Check if the target has changed since the previous move event
             if previous_target is not new_target or first_move:
+                # Update the current target for this pointer
+                RootEventHandler.target_tracker[pointer_id] = (
+                    new_target and ref(new_target)
+                ) or None
                 if not first_move:
                     # Dispatch a `pointer_leave` event for the previous target
                     ev = event.copy(type="pointer_leave")
--- a/pygfx/objects/_events.py
+++ b/pygfx/objects/_events.py
@@ -18,6 +18,8 @@ class EventType(str, Enum):
     POINTER_DOWN = "pointer_down"
     POINTER_MOVE = "pointer_move"
     POINTER_UP = "pointer_up"
+    POINTER_ENTER = "pointer_enter"
+    POINTER_LEAVE = "pointer_leave"
     CLICK = "click"
     DOUBLE_CLICK = "double_click"
     # Wheel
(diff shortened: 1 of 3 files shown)


pygfx/objects/_events.py
```diff
@@ -333,6 +338,32 @@ def dispatch_event(self, event: Event):
...
+            first_move = pointer_id not in RootEventHandler.target_tracker
+            # Get the previous target for this pointer (if any)
+            previous_target_ref = RootEventHandler.target_tracker.get(pointer_id)
+            previous_target = (previous_target_ref and previous_target_ref()) or None
+            # Get the current target for this pointer (if any)
+            new_target = event.target
+            # Update the current target for this pointer
+            RootEventHandler.target_tracker[pointer_id] = (
+                new_target and ref(new_target)
+            ) or None
```
Korijn, line 353: Only update if the target has changed

pygfx/objects/_events.py
```diff
@@ -333,6 +338,32 @@ def dispatch_event(self, event: Event):
...
+            if previous_target is not new_target or first_move:
+                # Update the current target for this pointer
+                RootEventHandler.target_tracker[pointer_id] = (
+                    new_target and ref(new_target)
+                ) or None
+                if not first_move:
+                    # Dispatch a `pointer_leave` event for the previous target
+                    ev = event.copy(type="pointer_leave")
+                    # Retarget to the previous target
+                    ev._retarget(previous_target)
```
Korijn, line 360: I could just pass the target to the copy call

pygfx/objects/_events.py
```diff
@@ -333,6 +338,32 @@ def dispatch_event(self, event: Event):
...
+            previous_target = (previous_target_ref and previous_target_ref()) or None
+            # Get the current target for this pointer (if any)
+            new_target = event.target
+            # Check if the target has changed since the previous move event
+            if previous_target is not new_target or first_move:
+                # Update the current target for this pointer
+                RootEventHandler.target_tracker[pointer_id] = (
+                    new_target and ref(new_target)
+                ) or None
+                if not first_move:
```
almarklein, line 356: IIUC both `previous_target`  and `new_target` are either an object or `None`. Can't we simply do something like this?
```py
if new_target is not previous_target:
     ... store new target in tracker
    if previous_target:
        ... dispatch LEAVE
    if new_target:
      ... dispatch ENTER
```

Korijn: The complication here is that `None` represents the event root (the renderer), and so, on the very first move event, the root needs to receive a pointer_enter event even though both previous and new target are `None`.

Korijn: See #296 for an alternative implementation where the root is also tracked

PR #296, opened 2022-04-18
Pointer_enter and pointer_leave events v2

### LINKED ISSUE #262: High level pointer events

Korijn:
Closes #262

* Same as #294 except here the root event handler is also tracked in the `RootEventHandler.target_tracker` which simplifies the code a little bit

I have to say, I like this version better!

### DIFF
--- a/examples/events.py
+++ b/examples/events.py
@@ -7,6 +7,7 @@ the items from that group (because the group has a double-click
 event handler).
 """
 
+from functools import partial
 from random import randint, random
 
 from wgpu.gui.auto import WgpuCanvas, run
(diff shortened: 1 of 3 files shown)


almarklein (approved):
I like this too. LGTM!

PR #297, opened 2022-04-18
Controllers save and load state 

### LINKED ISSUE #148: Improve API of Controls (Orbit) to better support saving and restoring state

Korijn:
Closes #148

* Adds `Controller.save_state()` and `Controller.load_state(state=None)` to all controller implementations
* Controllers create an initial snapshot upon creation
* Extends orbit and panzoom controller examples to demonstrate new features

orbit_saveload

panzoom_saveload

### DIFF
--- a/examples/orbit_camera.py
+++ b/examples/orbit_camera.py
@@ -1,5 +1,8 @@
 """
 Example showing orbit camera controller.
+
+Press 's' to save the state, and
+press 'l' to load the last saved state.
 """
 
 import imageio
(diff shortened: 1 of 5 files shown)


almarklein:
LGTM 🚀

Issue #301, opened 2022-04-26
controller method to center camera in a scene


kushalkolar:
It would be useful to have a controller method that pans the camera such that the edges of the set camera position form a bounding box for all `Geometry` objects. I notice that there's a method to get the bounding box for `Geometry` objects, but not for the scene.

I'm wondering if there's a better way to possibly implement this:

```python
x1, y1, x2, y2 = (0, 0, 0, 0)
for obj in scene.children():
    if hasattr(obj, 'geometry')
    bb = obj.geometry.bounding_box()
    x1 = np.min(x1, bb[0][0])  
    y1 = np.min(y1, bb[0][1])
    x2 = np.max(x2, bb[1][0])
    y2 = np.max(x1, bb[1][1])

center = Vector3(
    (x2 - x1) / 2,
    (y2 - y1) / 2,
    0.0
)

pos = camera.position
delta = pos.clone().sub(center)
controller.pan(delta)
controller.update_camera(camera)
```

Korijn:
Try `camera.show_object`? Does that do the trick for you?

kushalkolar:
Nope this doesn't work (at least for Image objects), for example modifying the `geometry_image.py` example throws an exception:

```python
"""
Show an image.
"""

import imageio
from wgpu.gui.auto import WgpuCanvas, run
import pygfx as gfx


canvas = WgpuCanvas()
renderer = gfx.renderers.WgpuRenderer(canvas)
scene = gfx.Scene()

im = imageio.imread("imageio:astronaut.png")[:, :, 1]

image = gfx.Image(
    gfx.Geometry(grid=gfx.Texture(im, dim=2)),
    gfx.ImageBasicMaterial(clim=(0, 255)),
)
scene.add(image)

camera = gfx.OrthographicCamera(512, 512)
camera.position.set(0, 0, 0)
camera.scale.y = -1

camera.show_object(image)

canvas.request_draw(lambda: renderer.render(scene, camera))
canvas
```

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [1], in <cell line: 26>()
     23 camera.position.set(0, 0, 0)
     24 camera.scale.y = -1
---> 26 camera.show_object(image)
     28 canvas.request_draw(lambda: renderer.render(scene, camera))
     29 canvas

[...]

PR #303, opened 2022-04-27
Fix Geometry.bounding_box for Image world object type and add docstrings

Korijn:
Related to #301

### DIFF
--- a/pygfx/geometries/_base.py
+++ b/pygfx/geometries/_base.py
@@ -117,10 +117,16 @@ class Geometry(ResourceContainer):
         if hasattr(self, "grid"):
             if self._aabb_rev == self.grid.rev:
                 return self._aabb
-            self._aabb = (
-                np.array([np.zeros_like(self.grid.data.shape), self.grid.data.shape])
-                - 0.5
-            )
+            grid_shape = list(self.grid.data.shape)
+            # ensure coordinates are 3D
+            if len(grid_shape) == 2:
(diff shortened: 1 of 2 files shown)


pygfx/geometries/_base.py
```diff
@@ -117,10 +117,23 @@ def bounding_box(self):
...
-            self._aabb = (
-                np.array([np.zeros_like(self.grid.data.shape), self.grid.data.shape])
-                - 0.5
-            )
+            grid_shape = list(self.grid.data.shape)
+            # if grid is multichannel, e.g. 3xf4
+            format = self.grid.format
+            if len(format) == 4 and format[1] == "x" and format[0] != "1":
+                # then strip last dimension from shape
+                grid_shape.pop(-1)
```
Korijn, line 125: I guess this could still go wrong if you supply e.g. a grayscale/single-channel 2D image but it still has a shape like `(320, 240, 1)`. Not sure how to account for that.

ivoflipse: Should we determine the number of dimensions somewhere, so you can use that to determine the appropriate shape downstream?

Korijn: I just noticed there is a `dim` property so I can use that

almarklein: > I just noticed there is a dim property so I can use that

Exactly

PR #304, opened 2022-04-27
Change colors of axes helper to rgb and add arrow heads

Korijn:
I always find myself looking at the axes helper, unable to remember which color maps to which direction. Changing the colors so that XYZ maps to RGB makes that a little easier.

I also added arrow heads to make it look more as you would expect:

Much better! ✨ 

Obviously when we get text rendering we should just add text nodes.


pygfx/helpers/_axes.py
```diff
@@ -9,10 +17,12 @@ class AxesHelper(Line):
 
     Parameters:
         size (float): The length of the lines (default 1).
-        thickness (float): the thickness of the lines (default 2 px).
+        thickness (float): The thickness of the lines (default 2 px).
+        arrow_start (float): Percentage where the arrow head should start (default 0.80).
```
Korijn, line 21: Actually this shouldn't be a parameter, the arrow head should just maintain an equilateral triangle when viewed from the side, so I can compute where it should start using arrow_radius. That will give the axes helper a consistent look.

Korijn, line 21: Actually this shouldn't be a parameter, the arrow head should just maintain an equilateral triangle when viewed from the side, so I can compute where it should start using arrow_radius. That will give the axes helper a consistent look.

Korijn:
I didn't realize that thickness is in screen-size pixels. Doh.

Korijn:
Well. I'd like to merge this and think about something better later.

almarklein:
So now the arrowhead's size (in world coords) is determined by the line's thickness (in screen coords)?

PR #305, opened 2022-05-03
Text rendering

### LINKED ISSUE #20: Text rendering
# Text rendering

An overview of our options. Latest update 16-11-2021.

## Goals

I think our goal should be:

* Aim for high quality text rendering (though not necessarily perfect).
* Make sure that there is a default font that looks good and is complete and always available.
* Provide support (one way or another) for all languages, including CJK.
* Preferably support user-provided fonts, so graphs in papers can be publication-worthy quality with a uniform style.
* Preferably an approach where we can start simple and improve on with further steps.

### Leveraging a GUI toolkit

We could leverage the text support of e.g. Qt to create text overlays. We have an example for this in pygfx.

### Bitmaps

Historically, fonts are rendered using glyph bitmaps that match the screen resolution. All glyphs in use are put into an atlas: a large image/texture that has all glyphs packed together.

A downside is that the bitmaps should match the screen pixels, so the bitmaps are font-size specific, and rendering glyphs at an angle is not possible without degrading quality.

### SDF

In 2007 Chris Green from Valve [published](https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf) an approach that uses scalar distance fields (SDF) to render glyphs. This approach makes it possible to create a special kind if bitmap (which encodes distances to the edge instead of a mask), that can then be used to render that glyph at any scale and rotation. This was a big step in games, but also for generic visualization.

There are variations on the way to calculate an SDF. The current industy standard seems to be [anti-aliased euclidian distance transform](https://weber.itn.liu.se/~stegu/aadist/edtaa_preprint.pdf).

### MSDF

Around 2015, Viktor Chlumsky [published](https://github.com/Chlumsky/msdfgen/files/3050967/thesis.pdf) his Master thesis, in which he proposes a method for a Multi-channel Signed Distance Field (MSDF). The key contribution is that directional information is encoded too, making is possible to produce sharp corners. It produces better results than normal SDF, at a smaller scale. The fragment shader is changed only slightly, causing just a minor performance penalty.

### Glyphy

The [Glyphy](https://github.com/behdad/glyphy) C library converts bezier curves to arcs and then calculates the signed distance to those arcs in the shader, producing better results than normal SDF. It's not well documented, but is by the author of Harfbuzz ...

### Vector rendering

It's also possible to render the Bezier curves from the glyphs directly on the GPU. See e.g. http://wdobbie.com/. This approach seems to require some preprocessing to divide the bezier curves into cells. If this can be changed to create an atlas (i.e. each cell is one glyph) then this may be a feasible approach for a dynamic context like ours. Also see [Slug](https://sluglibrary.com/), which looks pretty neat, but is not open source and covered by a patent.

Also see this fairly simple approach: https://medium.com/@evanwallace/easy-scalable-text-rendering-on-the-gpu-c3f4d782c5ac It uses a prepass to calculate the winding number for the rough shape, then composes the final shape and applies aa in a fragment shader. A disadvantage (for us) is that it needs multiple passes, making it harder to fit into our renderer, I think. 

### How visvis does text

https://github.com/almarklein/visvis/tree/master/text

Visvis produces glyphs on the fly using FreeType. Freetype is available by default on Linux, on Windows it must be installed. If FreeType is not available, visvis falls back on system that uses pre-rendered fonts. Visvis also ships with FreeSans, so there is always a good sans font available. Further, it can make use of other fonts installed on the system.

The rendered glyps are on a fixed resolution and put in an atlas texture. When rendered, the glyph is sampled using a smoothing kernel, giving reasonable results at multiple scales.

Further, visvis includes a mini-SDL to create bold, italic, and math symbols using escaping with backslash.

### How vispy does text

https://github.com/vispy/vispy/tree/main/vispy/util/fonts and https://github.com/vispy/vispy/tree/main/vispy/visuals/text

Vispy produces SDF glyphs on the fly. It first gets a glyph bitmap using FreeType (via Rougier's FreeType wrapper) on Windows and Linux, and Quartz on MacOS. The bitmap is then converted to an SDF using either the GPU or Cython. It can use most fonts installed on the system. The code to list and load fonts is platform specific.

The SDF glyphs are packed into an atlas texture, and rendered using regular SDF rendering. 

### How Matplotlib does text

Unless I'm mistaken, they use the GUI backend to render text.

## Discussion

SDF rendering already provides pretty solid results and is widely used. MSDF seems like a very interesting extension that may make it possible to create better results with smaller SDF sizes. Whether this is true depends also on the complexity of the glyphs - it has been reported that it does not provide much advantage for e.g. Chinese glyphs. We'd need tests to visually judge what glyph sizes are needed in both approaches.

In terms of dependencies, to create SDF we can get away with just FreeType to create the bitmaps, plus steal some code from vispy to generate the SDF. For MSDF there is that one project, but its not that actively maintained, so at this point it may be a liability. It may be worth investigating if its possible to generate an MSDF from a bitmap.  

Vector rendering seems to be popular lately. It produces better results, but is more complex and less performant than SDFs. Further you may need a bitmap fallback for high scales. There is also the question of how to generate the vector-data from the font files.

If you want to do render text perfectly, you should take into account kerning, ligatures, combinatory glyphs, left-to-right languages, etc. We can do kerning, but the rest is way more complex. Tools that handle these correctly are e.g. Slug and [Harfbuzz](https://harfbuzz.github.io).  The latter might be feasible to use - there are multiple Python wrappers.

## More details obtaining a glyph

I imagine that somewhere in our code we have a method `create_glyph(font, char)` that returns an SDF/MSDF/bezier-curve-set and some metadata. We can place this info in an atlas texture together with other used glyphs. Now, how does this function work? 

*The below more or less assumes SDF / MSDF, but similar questions apply to the Bezier-based rendering.*

#### Using prerendered glyph maps

We could produce prerendered glyph atlas, so `create_glyph()` samples one glyph from that (offline) atlas and to put it in our GPU atlas. That way, the dependencies for rendering the glyphs are only needed by devs, and not the users.
See e.g. https://github.com/Chlumsky/msdf-atlas-gen A disadvantages is that the font and available characters are limited by what atlasses we chose to ship. If we'd ship CJK fonts, that would grow the package size a lot, so maybe we'd need a `pygfx-cjk-fonts` or something.

Some measurements for rendering all chars from a font:

* OpenSans   ->  1011 glyphs -> 1.5MB @ 32   0.5MB @ 16
* FreeSans   ->  4372 glyphs -> 7MB @ 32     2.3MB @ 16
* NotoSans   ->  2300 glyps  -> 3.5MB @ 32   1.1MB @ 16
* NotoSansJP -> 16735 glyps  ->   ??          15.1MB @ 16

This shows that with 16x16 MSDF glyphs, the atlas is about 3x the size of the .ttf file. Or 9x for 32x32. Not bad, but I don't now yet whether 16x16 is enough, especially for the more complex glyphs in CJK fonts.

### Using a tool to render glyph on the fly

If we can produce the SDF glyph from the fontfile at the user, then support for all languages, and custom fonts is obtained. The cost is that we need a dependency for this.  

One option is FreeType. There is Rougiers https://github.com/rougier/freetype-py which provides binaries for multiple platforms (but not MacOS M1 or Linux aarch64 yet). IIUC freetype can render a glyph bitmap at any scale, but to create an SDF we need to do some extra work ourselves. I am not sure if a MSDF is feasible, because you'd miss the vectors, or can we do edge detection on a high-rez glyph to find the orientation?

Another option is https://github.com/Chlumsky/msdfgen. We'd need to build and distribute the code for all platforms that we support. Since this code is not *that* actively maintained, this feels a bit tricky.

## Typesetting

Typesetting is the process of positioning the quads that contain the glyphs. This includes kerning (some combination of glyhs have a custom distance). Also inclded are justifcation, alignment, text wrapping etc.

I think its not too hard to do this in pure Python, though existing solutions may make things easier. Examples are [Slug](https://sluglibrary.com) and [Harfbuzz](https://harfbuzz.github.io), but these are not feasible dependencies. I don't know of any Python libraries for typesetting.

For the kerning we would need the kerning maps. FreeType can provide this info. I *think* that msdfgen also provides it in the CSV/JSON metadata.

## Formatting

We should consider a formatting approach so users can create bold/italic text etc. Visvis uses escaping, but maybe an micro-HTML-subset would be nicer. Or something using braces. Or maybe its out of scope. This is definitely something that we can do *later*.

## Fonts of interest

* OpenSans is a pretty nice and open font.
* FreeSans is a GNU project that aims to cover almost all Unicode, except CJK.
* Noto is a family of over 100 fonts that aims to have full Unicode coverage. It basically has a font for each script. The plain NotoSans has over 3k glyphs and with support for Latin, Cyrillic and Greek scripts it covers 855 languages.

### LINKED ISSUE #271: Text API
A proposal for the text API.

Interesting links (for inspiration):
* https://www.npmjs.com/package/troika-three-text
* https://github.com/almarklein/visvis/wiki/Text
* https://vispy.org/api/vispy.visuals.text.text.html

### General discussion

Which props belong on the world-object, and which on the material? ThreeJS' troika has everything at the Object3D and uses a modified mesh material. Initially I liked the idea of putting most props on the material, since they affect how the text is rendered. But most of these props affects the positioning of the text, so they feel quite "structural".

### Alternative 1

A `Text` wobject and corresponding material. Not sure how to to bold/italic unless it is applied to the whole text.

class `Text(WorldObject)`, with properties:

* text
* font: the name of a builtin or system font, or a path to a font file
* font_size
* color
* anchor_x
* anchor_y
* angle
* max_width: if text goes beyond this, it will wrap
* line_height
* text_align

class `TextMaterial(Material)`, with properties:

* color

Maybe this could be a MeshMaterial ...

### Alternative 2

To have better support for formatting, we could introduce an object to hold a piece of text plus its formatting. We can then provide ways to generate these easily from e.g. Markdown, HTML or Latex.

class `TextPiece`, with properties:

* text: the Unicode text
* font
* font_size
* weight: bold, normal, etc.
* italic

We could have e.g. a `MathTextPiece` that has its own specific font and shaper, based on Latex input. Or a `MusicalNotesTextPiece`, if someone wants to go crazy. Bottom line is that with this we solve the "itemization" step, and that the TextPiece is responsible for the shaping. In comes Unicode, out come font indices and (relative) vertex positions.

class `Text(WorldObject)`, with properties:

* pieces: a list of `TextPiece` objects.
* text: shortcut to create a single text piece.
* markdown: shortcut to create simple formatted text.
* html: shortcut to create somewhat more advanced formatted text (e.g. `<small>` etc.)

... rest the same as alternative 1.

### Alternative 3

The geometry consists of an array of TextPiece objects. We have some functions to generate it.
```py
text = Text(
    gfx.text_geometry(markdown="hello **world**"),
    gfx.TextMaterial(...)
)
```

All props go on the material (except the specifics for the `TextPiece`). The buffers with positions and atlas indices are private.

### Alternative 4

The geometry consists of the actual positions and atlas indices, and we have functions to generate it. The code sample is the same as above, except now TextPieces are an internal construct, and the `text_geometry` function has arguments for all "properties" that affect positioning (e.g. `text_align`). In effect the material is "clean" and you can *always* put a different material on the same geometry without incurring unexpected CPU processing.

This is actually quite clean. The only downside that I see now, is that users may be a bit confused why some aspects are set as arguments in `text_geometry` and some as material properties.

almarklein:
Closes #271 and closes #20 

Todo:

* General
  * [x] High level version of the API.
  * [x] Document the text rendering process.
  * [x] Be specific about what font properties affect what, and where the user can set them.
  * [x] Text can be a child of another object.
  * [x] Make it such that some steps can be overloaded for custom behavior.
  * [x] Make it such that the builtin steps can (hopefully) be replaced without affecting the API.
  * [x] Flesh out the API.
  * [x] Make things work for RTL fonts.
  * [x] Make basics work for  vertical scripts.
  * [x] Fix bug that when text is replaced, parts of the old text seems to linger sometimes.
  * [x] Implement `geometry.bounding_box()` (or gfx.show() wont work).
* Discuss
  * [x] Place to put properties that affect positioning.
  * [x] Screen-space vs world-space, how to express in API.
  * [x] What builtin fonts to include (Noto sans!).
  * [x] What should happen with the text object's transform if it's set to render in screen space.
* Itemization
  * [x] Plain text
  * [x] Very basic markdown.
  * [x] Tests.
* Font selection
  * [x] Basic font selection (stub).
  * [x] Include builtin fonts, a selection of noto fonts.
  * [x] Decide what to do about font management. #379
  * [x] Clean up the font manager module.
  * [x] Include / delete builtin fonts.
  * [x] Tests.
* Shaping
  * [x] Stub shaping that just about works.
  * [x] Basic shaping with Freetype.
  * [x] Advanced shaping with Harfbuzz.
  * [x] Tests.
* Glyph generation and atlas
  * [x] Glyph generation with Freetype. 
  * [x] Atlas that can pack variable sized glyphs.
  * [x] Dynamic atlas sizing (grow and shrink). 
  * [x] Tests.
* Layout (minimal stuff)
  * [x] Put chars in a straight line.
  * [x] Text size.
  * [x] Alignment: `TextGeometry.anchor`
  * [x] Support custom layout.
  * [x] Rotate text at an angle, also in screen space.
  * [x] Tests.
* Rendering
  * [x] Render rectangles at the glyph positions.
  * [x] Render the texture content.
  * [x] Actual SDF rendering. 
  * [x] Approximate weight.
  * [x] Slanting.
  * [x] Outlines. 
  * [x]  Picking.
  * [x] Validation examples.
* Examples
  * [x] Minimal text example.
  * [x] Show white on black, and black on white and make sure they feel equally thick.
  * [x] A waterfall of glyphs, to strain the atlas.
  * [x] Use new `gfx.show()`
* Final stuff
  * [x] Update new validation example screenshots.
  * [x] Docs.
  * [x] More tests?
  * [x] Fix todos & self-review.

Tasks for later are listed in #391.

----

Status: can render text, but each character is just a little rectangle. Note: it says "Hello world" :)
image


pygfx/geometries/_text.py
```diff
@@ -0,0 +1,167 @@
...
+            assert part.positions.dtype == np.float32
+            assert part.positions.shape == (part.indices.size, 3)
+            part.offset = glyph_count
+            glyph_count += part.indices.size
+            indices_arrays.append(part.indices)
+            positions_arrays.append(part.positions)
+
+        # Store
+        self.indices = Buffer(np.concatenate(indices_arrays, 0))
+        self.positions = Buffer(np.concatenate(positions_arrays, 0))
```
Korijn, line 109: Is there a specific reason why these are not being passed to the super constructor?

pygfx/geometries/_text.py
```diff
@@ -0,0 +1,167 @@
...
+        positions_arrays = []
+        glyph_count = 0
+        for part in self._text_parts:
+            assert part.indices.dtype == np.uint32
+            assert part.positions.dtype == np.float32
+            assert part.positions.shape == (part.indices.size, 3)
+            part.offset = glyph_count
+            glyph_count += part.indices.size
+            indices_arrays.append(part.indices)
+            positions_arrays.append(part.positions)
```
Korijn, line 105: This is a common thing (concatenating buffers and updating indices) in the geometry subpackage; seems like a good candidate for a reusable function. Doesn't need to be a part of this PR!

pygfx/geometries/_text.py
```diff
@@ -0,0 +1,167 @@
...
+        elif isinstance(align, int):
+            align = {-1: "left", 0: "center", 1: "right"}[align]
+        elif not isinstance(align, str):
+            raise TypeError("Align must be a string.")
+        align = align.lower()
+        if align not in alignments:
+            raise ValueError(f"Align must be one of {alignments}")
+        self._text_align = align
+
+    def _position(self):
```
Korijn, line 154: I'm imagining you intend to expose these helper functions the geometry to update positions as needed when certain properties are changed, right? Makes sense!

pygfx/geometries/_text.py
```diff
@@ -0,0 +1,167 @@
...
+            indices_arrays.append(part.indices)
+            positions_arrays.append(part.positions)
+
+        # Store
+        self.indices = Buffer(np.concatenate(indices_arrays, 0))
+        self.positions = Buffer(np.concatenate(positions_arrays, 0))
+
+    @property
+    def max_width(self):
+        """The maximum width of the text (in model coords). Text will
```
Korijn, line 113: I'm not sure what model space means... :) Font space?

pygfx/materials/_text.py
```diff
@@ -0,0 +1,50 @@
...
+        self.color = color
+
+    def _wgpu_get_pick_info(self, pick_value):
+        # This should match with the shader
+        _ = unpack_bitfield(pick_value, wobject_id=20, index=26, x=9, y=9)
+        return {}
+
+    @property
+    def screen_space(self):
+        """Whether the text is applied in screen space (in contrast to model space)."""
```
Korijn, line 26: Is this equivalent to using an ndc camera?

pygfx/renderers/wgpu/textrender.py
```diff
@@ -0,0 +1,158 @@
...
+            $$ endif
+
+            var varyings: Varyings;
+            varyings.position = vec4<f32>(ndc_pos.xy + delta_ndc * ndc_pos.w, ndc_pos.zw);
+            varyings.world_pos = vec3<f32>(world_pos.xyz / world_pos.w);
+            varyings.pointcoord = vec2<f32>(point_coord);
+            varyings.size = f32(size);
+
+            // Picking
+            varyings.pick_idx = u32(i0);
```
Korijn, line 121: Is this going to support individual glyph picking? I didn't expect that sort of awesomeness 🤠

Korijn:
Awesome start, I'm super stoked to see this. I just had some questions which I added as comments.

pygfx/geometries/_text.py
almarklein: Not really. Might look a bit cleaner indeed.

pygfx/geometries/_text.py
almarklein: Good question. The docstring is wrong - it would depend on the `screen_space` prop of the material. This is worth some more thought and discussion ...

pygfx/geometries/_text.py
almarklein: Yes indeed. All values that affect positioning can be applied relatively easily *after* the shaping and the whole make-sure-the-sdf-glyph-is-available stuff is done. So it feels like a good idea to make them available as properties. The Geometry and WorldObject are both free of object-specific props so far, so it looks like we need to break that tradition here, and the geometry felt like the best place to do that.

pygfx/materials/_text.py
almarklein: It means that the text is flat on the screen, no matter the orientation of the text-object and camera. Similar to how Points are represented with a circle in screen space. This is good for labeling stuff.

If `False`, the text behaves more like other objects, so you can e.g. put it on a surface in a 3D scene.

I had not previously realized that we'd have both use-cases. I think I like the idea of creating separate materials for these, because their use is profoundly different.

pygfx/renderers/wgpu/textrender.py
almarklein: Yes. Even the location within the glyph :D   But note that glyph != character. Multiple characters can become a single glyph in the shaping step, so we'd need to account for that.

almarklein:
> I'm also glad we already thought about the API in advance

Me too. That's what made it possible to just implement the whole thing with stub rectangles. Now it's "just" a matter of filling in the blanks.

pygfx/materials/_text.py
Korijn: Ah I see! Very useful:)

pygfx/renderers/wgpu/textrender.py
Korijn: 💡 got it!

pygfx/geometries/_text.py
Korijn: See #306

pygfx/geometries/_text.py
Korijn: I'm fine with putting them on the geometry class for now, it makes most sense.

(Three.js also provides such buffer affecting functions directly on geometry)

pygfx/geometries/_text.py
Korijn: I would expect max width to be expressed in some font relative unit, so that line wrapping is not going to have different results based on how the text is rendered (billboard-style or 3d).

pygfx/materials/_text.py
Korijn: Thinking of the easiest way to communicate this in docstrings, I'm reminded of the graphics term "billboard" which is used to describe objects that are always rendered facing the camera.

pygfx/geometries/_text.py
almarklein: BTW, I'm not sure if the logic here is the same. E.g. the indices here are references to glyphs in the atlas. Maybe I should give them a different name to make that clear.

pygfx/geometries/_text.py
almarklein: Yeah exactly. It operates in the same space as the font size and other positions.

pygfx/utils/text/_atlas.py
```diff
@@ -0,0 +1,160 @@
+import threading
+
+import numpy as np
+
+from ...resources import Texture
+
+
+# todo: how to invalidate the pipeline of objects that use the atlas when the texture is changed?
```
Korijn, line 8: I've been thinking a lot about this, since the lighting system also needs it... I'm imagining some sort of dependency tracking mechanic using weakrefs...

almarklein:
Progress on atlas and reading the atlas texture in the shader. The positioning needs some work :)

image

pygfx/utils/text/_atlas.py
almarklein: Yeah, a more generic approach to solve these use-cases sounds like a good idea.

almarklein:
Working on SDF.

Plain bitmap (64x64):
image

SDF (64x64):
image

almarklein:
Worked mostly on getting the sizing correct. Once you know what units you're talking about in the shader, you can also properly implement outlineing (hooray SDF).

image

There's now also a validation example for the text size. Now we can compare to e.g. what the browser does (top) with our result (bottom):
image

The shaping seems good. The font looks a wee bit thinner though, and the aa-smoothing looks a bit much when zoomed in like this.

Korijn:
Seems like the difference in crispness could be related to a difference in DPI values?

almarklein:
> Seems like the difference in crispness could be related to a difference in DPI values?

Was partly #359 and partly too much aa in the text shader.

almarklein:
Basic font selection mechanism. Will create a new issue to discuss how to do this properly, but this demonstrates it working: This was one string containing codepoints from 3 separate Noto font files.
image

almarklein:
Using SDF allows us to do italics and bold in the shader. Top shows an actual italic/bold font, bottom (gray) shows a regular font with fake italics and sdf-based boldness.
image

image

The result is not perfect, but it does mean we can produce these font effects when an actual font for that effect is not available. Plus we can scale to many weight levels (also thinner text). The bold text is very rounded in SDF, an effect that can be expected. This is something that may be improved (in the future) by using a more advanced SDF approach.

almarklein:
Last week I realized that also with SDF the glyphs vary considerably in size. Using a fixed size for each glyph in the atlas is thus wastefull. I therefore replaced the atlas with one that supports glyphs of varying widths.

The magic sauce (the rectangle packing, which is a research topic on itself) I took from vispy. I then added support for resizing, and also to free glyphs. In some cases the atlas can then be only repacked instead of resized.

I also implemented an example that stresses the atlas. During which I discovered several edge cases and bugs :) 

Screen Recording 2022-11-17 at 16 14 52

Korijn:
I was wondering how transforms are applied to text objects.

For example, [in the famous valve paper](https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf), you can see the words "NO TRESPASSING" on a wall.

I would expect this to be represented in a scene graph like this:

* Scene
  * Mesh (the wall)
    * Text (the text)

That way, I can position and scale the text relatively to the wall, e.g. to center it on the wall and make it about 20% of the width of the wall. Then when I move on to position the wall in my scene (using translation, rotation or scaling) the text will just stay where it should be.

Is that supported now?

almarklein:
> Is that supported now?

Yes, this works, see the hello_text.py example. It's (currently) not like the text will fill 100% of the object - the text simply has a certain size depending on the number of chars and the font size, but that's also how other objects work, so I think this is ok.

This reminds me, we should also think what should happen with the text object's transform if it's set to render in screen space 🤔

almarklein:
> we should also think what should happen with the text object's transform if it's set to render in screen space

I think I know the answer. This text should just display in screen coords, also when attached to a tree of nested objects. So the transform is only used to position the text, nothing else. No scaling, not rotation.

Korijn:
> > we should also think what should happen with the text object's transform if it's set to render in screen space
> 
> I think I know the answer. This text should just display in screen coords, also when attached to a tree of nested objects. So the transform is only used to position the text, nothing else. No scaling, not rotation.

Sounds right to me. It should be consistent with other screen space things we (could) offer, like points and lines.

Korijn:
Notes from our meeting:

# For later (move to #379)

* Make things work for vertical fonts.
* Implement stretch font prop (stretch all text, stretch only whitespace, stretch only characters).

# Complete now (as part of this PR):

* Alignment.
* Rotate text at an angle, also in screen space.
* Decide what to do about font management. Text: font management #379
* Clean up the font manager module.

almarklein:
Japenese, Chinese and Korean render correctly (though horizontally) by default:

image

Turned out that supporting vertical text was not so hard: with `TextGeometry(..., direction='ttb')`:

image

almarklein:
Update on putting text at an angle: I initially felt against using the world-object transform for that, because the `world_transform` contains rotations and scaling from parents. However, in screen-space-mode, the rotation and scale from the local transform don't affect the final position, so we can still use them to rotate and scale the text. I added an extra transform to the `Text(WorldObject)` containing just the local rotation and scale, so that these can be applied in screen space.

Korijn:
As a response to your bounding box commit, I think it will come back as a discussion item when we get to line breaking support later.

In photoshop-like applications, there are usually two modes of operation: point text and text box.

* With point text mode, the user sets the anchor point and has to perform line breaking manually (by inserting newlines). This is very similar to what you've implemented now (although I'm not sure even manual line breaks are supported yet?)
* With text box mode, the user draws a bounding box, and lines are wrapped automatically to fit.

almarklein:
I don't think the line breaking matters for the bounding box. The two modes you describe would be equivalent to setting the `max-width` property. I both cases (wrapping or no wrapping) the geometry knows the bounding box.

The problem right now is related to screen-space mode. In world-space mode, the text's bounding box is well defined and known. But in sceen-space mode, it should be considered a point (a bounding box with zero volume). The problem is that the geometry does not know the mode.

This could be an argument to move the `screen_space` prop from the material to the geometry.

almarklein:
Ready for review!

Two open points:
* Ignore the font manager for now. We still need to decide what to do with font manager, also short-term to get this merged. See #379 
* I propose to move the `screen_space`  prop to the `TextGeometry`. This way the bounding box can be set to the space that the text occupies when rendering in world space. If you consider that the option determines how the text occupies the world, is does seem fitting to place it at the geometry, I guess ...

Korijn:
At geometry level, the bounding box function should return the _local_ bounds, not world bounds

For screen space things, it's gonna be weird. I doubt there is anything sensible to be done with that scenario in terms of frustum culling or the show_object functions we have. Screen space by definition bypasses the world, view and projection transforms, so what use is there to having that bounding box...

I think you are right, and the screen_space prop should be moved to geometry.

almarklein:
> At geometry level, the bounding box function should return the local bounds, not world bounds

Yeah, world-space isn't the correct term. Shall we call it model-space or object-space, or ... ?

> For screen space things, it's gonna be weird.

Well there's still a *point* in the scene that represents where the text is "attached to". This point could still participate in e.g. `show_object`. An alternative would be to give it a null bounding box, but IFAIK this is currently not possible.

Anyway, I'll move `screen_space`  to the geometry. Then if False, the bounding box will encompas the text. If True it will be a point.

pygfx/utils/text/_fontfinder.py
```diff
@@ -0,0 +1,260 @@
...
+    dirs = []
+    dirs.extend(WinFontDirs)
+
+    # Get win32 font directory from the registry
+    ms_folders = r"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders"
+    try:
+        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, ms_folders) as user:
+            dirs.append(winreg.QueryValueEx(user, "Fonts")[0])
+    except OSError:
+        dirs.append(os.path.join(os.environ["WINDIR"], "Fonts"))
```
Korijn, line 206: why not add both of these dirs?

pygfx/utils/text/_fontfinder.py
```diff
@@ -0,0 +1,260 @@
...
+
+try:
+    HOME = os.path.expanduser("~")
+except Exception:  # Exceptions thrown by home() are not specified...
+    HOME = "/home"  # Just an arbitrary path
+
+WinFontDirs = [
+    os.path.join(HOME, "AppData/Local/Microsoft/Windows/Fonts"),
+    os.path.join(HOME, "AppData/Roaming/Microsoft/Windows/Fonts"),
+]
```
Korijn, line 235: I would recommend extending this with the common user folder, and not relying on HOME expansion:

```suggestion
WinFontDirs = [
    os.path.join(os.environ["LOCALAPPDATA"], "Microsoft/Windows/Fonts"),
    os.path.join(os.environ["APPDATA"], "Microsoft/Windows/Fonts"),
    os.path.join(os.environ["ALLUSERSPROFILE"], "AppData/Local/Microsoft/Windows/Fonts"),
    os.path.join(os.environ["ALLUSERSPROFILE"], "AppData/Roaming/Microsoft/Windows/Fonts"),
]
```

pygfx/utils/text/_fontfinder.py
almarklein: Good point. Took this from MPL, but I see no reason not to add both.

pygfx/utils/text/_fontfinder.py
```diff
@@ -0,0 +1,260 @@
...
+
+    # Get win32 font directory from the registry
+    ms_folders = r"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders"
+    try:
+        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, ms_folders) as user:
+            dirs.append(winreg.QueryValueEx(user, "Fonts")[0])
+    except OSError:
+        dirs.append(os.path.join(os.environ["WINDIR"], "Fonts"))
+
+    return [os.path.abspath(d) for d in dirs if os.path.isdir(d)]
```
Korijn, line 208:
```suggestion
    dirs.append(os.path.join(os.environ["WINDIR"], "Fonts"))

    # Get win32 font directory from the registry
    ms_folders = r"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders"
    try:
        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, ms_folders) as user:
            dirs.append(winreg.QueryValueEx(user, "Fonts")[0])
    except OSError:
        pass

    return list({os.path.abspath(d) for d in dirs if os.path.isdir(d)})
```

pygfx/utils/text/_fontfinder.py
```diff
@@ -0,0 +1,265 @@
...
+
+def get_builtin_fonts():
+    """Get a list of fonts that is shipped with PyGfx."""
+    dir_paths, file_paths = find_fonts_from_dir(get_resources_dir(), False)
+    return file_paths
+
+
+def get_entrypoints_fonts():
+    """Get a list of fonts defined via the setuptools entrypoint. TODO someday."""
+    raise NotImplementedError()
```
Korijn, line 47: I think we need to add one more directory fonts can be found in, and use an environment variable to configure it's location. That will allow users to ship font files with applications. Should be trivial.

pygfx/utils/text/_fontfinder.py
```diff
@@ -9,37 +9,98 @@
...
+        self._family_name = family_name
+        self._style_name = style_name
+        self._name = None
+        self._codepoints = codepoints
+
+    def __repr__(self):
+        return f"<FontFile {self.name} at 0x{hex(id(self))}>"
+
+    def __hash__(self):
+        return hash(self.name)
```
Korijn, line 36: What if two different versions of the same font are found?

almarklein: Then one will be used. This'd be either the first or the last added. Would be good to make this behavior explicit and document it.

pygfx/utils/text/_fontfinder.py
almarklein: We have `font_manager.add_font_file()`.

pygfx/utils/text/_fontfinder.py
almarklein: This will be `logger.debug()`'ed.

almarklein:
Font manager is done too. I have nothing to add 🚀

Korijn (approved):
Let's merge!

I do think we should each go over all the discussion here one more time to ensure everything that's still open is captured in #391

Issue #308, opened 2022-05-05
Event system breaks down with multiple canvases


almarklein:
E.g. with the `two_canvases.py`  example, I get:
```
Running script: "/Users/almar/dev/py/pygfx/examples/two_canvases.py"
Error during handling pointer_move event
Traceback (most recent call last):
  File "/Users/almar/dev/py/wgpu-py/wgpu/gui/base.py", line 22, in log_exception
    yield
  File "/Users/almar/dev/py/wgpu-py/wgpu/gui/base.py", line 267, in _dispatch_event
    callback(event)
  File "/Users/almar/dev/py/pygfx/pygfx/renderers/wgpu/_renderer.py", line 789, in convert_event
    self.dispatch_event(ev)
  File "/Users/almar/dev/py/pygfx/pygfx/objects/_events.py", line 359, in dispatch_event
    self.dispatch_event(ev)
  File "/Users/almar/dev/py/pygfx/pygfx/objects/_events.py", line 376, in dispatch_event
    target = target.parent or self
AttributeError: 'WgpuRenderer' object has no attribute 'parent'
```

This happens here:
https://github.com/pygfx/pygfx/blob/728a4b7434879d3606f32836eda3971ed73905f8/pygfx/objects/_events.py#L365-L376

The problem is that target is *another* renderer than self. I strongly suspect that this is because of the class attributes:
https://github.com/pygfx/pygfx/blob/728a4b7434879d3606f32836eda3971ed73905f8/pygfx/objects/_events.py#L193-L198
and
https://github.com/pygfx/pygfx/blob/728a4b7434879d3606f32836eda3971ed73905f8/pygfx/objects/_events.py#L296-L302

I think these should all be (normal) attributes on the `RootEventHandler`. For the `pointer_captures ` this would require some refactoring to access it via `event.root`, I think.

PR #310, opened 2022-05-13
Fix event handling with multiple canvases

berendkleinhaneveld:
Make `click_tracker` and `target_tracker` instance attributes on `RootEventHandler` in order to have proper support for multiple canvases (and thus multiple root event handlers).
Also make sure that when a pointer is captured, pointer events (for the captured pointer_id) on other canvases are not handled.

Bonus: fixed a type annotation

### DIFF
--- a/pygfx/objects/_events.py
+++ b/pygfx/objects/_events.py
@@ -303,9 +303,9 @@ class RootEventHandler(EventTarget):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # Dictionary to track clicks, keyed on pointer_id
-        self.click_tracker = {}
+        self._click_tracker = {}
         # Dictionary to track targets, keyed on pointer_id
-        self.target_tracker = {}
+        self._target_tracker = {}
 
     def dispatch_event(self, event: Event):
         """Dispatch the given event.
@@ -360,12 +360,12 @@ class RootEventHandler(EventTarget):
         # Update the target tracker on all `pointer_move` events
         if event.type == EventType.POINTER_MOVE:
             # Get the previous target for this pointer (if any)
-            previous_target_ref = self.target_tracker.get(pointer_id)
+            previous_target_ref = self._target_tracker.get(pointer_id)
             previous_target = (previous_target_ref and previous_target_ref()) or None
             # Check if the target has changed since the previous move event
             if previous_target is not target:
                 # Update the current target for this pointer
-                self.target_tracker[pointer_id] = (target and ref(target)) or None
+                self._target_tracker[pointer_id] = (target and ref(target)) or None
                 if previous_target is not None:
                     # Dispatch a `pointer_leave` event for the previous target
                     ev = event.copy(type="pointer_leave", target=previous_target)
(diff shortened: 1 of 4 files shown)


pygfx/objects/_events.py
```diff
@@ -373,7 +373,7 @@ def dispatch_event(self, event: Event):
                     captured_target.release_pointer_capture(pointer_id)
             if not event.bubbles or event.cancelled or target is self:
                 break
-            target = target.parent or self
+            target = getattr(target, "parent", None) or self
```
Korijn, line 376:
```suggestion
target = getattr(target, "parent", self)
```

berendkleinhaneveld: I've thought about this before committing, but it ain't the same ;)

almarklein:
>  when target doesn't have a parent attribute

When would this happen?

Other then this question, this looks fine!

Korijn:
> > when target doesn't have a parent attribute
> 
> When would this happen?

The Scene, usually :)

berendkleinhaneveld:
> When would this happen?

I wondered myself as well, but Friday afternoon I just wanted it to work ;)
I think it's what Korijn said: the last iteration it is the WgpuRenderer. This subtlety is missing from the tests because it is hard to instantiate a renderer there. I'll figure out a way to trigger this in a test so that we don't get bitten by this again.

berendkleinhaneveld:
Haha, ok: it is exactly this problem (#308 ) which you apparently already discovered ;)
I'll see if I can replicate it in a test and then try to fix it with your suggestions from #308 .

berendkleinhaneveld:
> do we want the other canvas to process events, or should the pointer capture prevent that?

I've committed a solution where a pointer capture prevents the pointer events even on the other canvas. If we don't need that, I suggest to put the `set_pointer_capture` and `release_pointer_capture` on the `RootEventHandler` instead, so that the `pointer_captures` can be part of a root event handler instance, like the `click_tracker` and `target_tracker` now.

almarklein:
> do we want the other canvas to process events, or should the pointer capture prevent that?

I think the one canvas should capture, preventing events in other canvases - which is (I think) what you implemented.

pygfx/objects/_events.py
```diff
@@ -296,13 +300,12 @@ def release_pointer_capture(self, pointer_id):
...
-    click_tracker = {}
-    # Dictionary to track targets, keyed on pointer_id
-    target_tracker = {}
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        # Dictionary to track clicks, keyed on pointer_id
+        self.click_tracker = {}
+        # Dictionary to track targets, keyed on pointer_id
+        self.target_tracker = {}
```
almarklein, line 308: Should these be private variables?

pygfx/objects/_events.py
```diff
@@ -331,12 +334,24 @@ def dispatch_event(self, event: Event):
...
-            event.stop_propagation()
+            captured_target_ref, event_root_ref = EventTarget.pointer_captures[
+                pointer_id
+            ]
+            captured_target, event_root = (
+                captured_target_ref(),
+                event_root_ref and event_root_ref(),
+            )
+            # If the pointer was captured in the context of another root event
+            # handler, then let's not handle this event
```
Korijn, line 345:
```suggestion
            # If the pointer was captured in the context of another root event
            # handler, then let's not handle this event. It will be handled by
            # the appropriate RootEventHandler
```

Korijn (approved):

PR #312, opened 2022-05-15
Fix problems in "colormap_mesh.py" and "transparency1/2.py"

### LINKED ISSUE #295: colormap issue
Tried to run the colormap_mesh.py example, and it works fine. However, when replacing the orthographic camera by a perspective one:

camera = gfx.PerspectiveCamera(70, 1)
camera.position.z = 10

the ob4 (color for each vertex)  gives a strange result as shown in the attached image:

image

Don't know why the colormap for ob4 depends on the camera.

panxinmiao:
Fixes #295

and others.


pygfx/renderers/wgpu/_blender.py
```diff
@@ -132,7 +132,7 @@ def get_shader_code(self, blender):
             @location(1) pick: vec4<u32>,
         };
         fn get_fragment_output(depth: f32, color: vec4<f32>) -> FragmentOutput {
-            if (color.a < 1.0) { discard; }
+            // if (color.a < 1.0) { discard; }
```
panxinmiao, line 135: I guess the color interpolation accuracy in the fragment shader will lead to alpha not exactly equal to 1.0.
I tried to change the threshold to 0.99, and the result was also correct. 

But I think maybe they should not be discarded directly here, just treat them as opaque.
Or we may need a threshold attribute like "alpha_test" to control the discarding here.

pygfx/renderers/wgpu/_blender.py
```diff
@@ -316,14 +316,14 @@ def get_color_attachments(self, blender, clear_color):
...
-                "load_op": wgpu.logger.clear,
+                "load_op": wgpu.LoadOp.clear,
                 "store_op": wgpu.StoreOp.store,
             },
             {
                 "view": blender.reveal_view,
                 "resolve_target": None,
                 "clear_value": reveal_clear_value,
-                "load_op": wgpu.logger.clear,
+                "load_op": wgpu.LoadOp.clear,
```
Korijn, line 326: These two "spelling errors" should be merged into master asap, the other change (alpha) is probably going to need to some input from @almarklein

pygfx/renderers/wgpu/_blender.py
almarklein: It's important that (semi)transparent fragments are discarded here, otherwise the depth buffer will be set, and the blending will be incorrect. We could lower the threshold, but I want to be really sure that is necessary.

Is this what fixes #295? Do you know why? It's not obvious to me at first glance.

panxinmiao: > Is this what fixes #295? Do you know why? It's not obvious to me at first glance.

Yes, It is. 

When using "vertex color", the pixel color in the fragment shader will be calculated according to the interpolation of three vertices of the triangle. I guess there will be a floating-point precision problem.  Although the alpha value of all the three vertices color is 1.0, it may not be exactly equal to 1.0 after interpolation,  there will be precision error. At this time, some pixels may meet the alpha < 1.0 and be discarded.
Looks like that we need to keep this line of code and slightly lower the threshold.

pygfx/renderers/wgpu/_blender.py
Korijn: If you can split the PR we can merge this part immediately:)

Korijn: See #314

Korijn:
Can you rebase this PR?

pygfx/renderers/wgpu/_blender.py
Korijn: Can we define a constant for this?

pygfx/renderers/wgpu/_blender.py
```diff
@@ -25,6 +25,8 @@ class BasePass:
 
     render_mask = 3  # opaque end transparent
 
+    EPSILON = 1e-6
```
almarklein, line 28: There is also [WGSL constants](https://www.w3.org/TR/WGSL/#creation-time-consts) but the logic to fiddle one in at the module level is probably more code than what we have now :)

almarklein:
I double-checked that these are indeed all occurrences in this module where we compare against alpha. LGTM!

Korijn:
Wouldn't it make more sense to provide the constant via jinja instead of using f-strings? I worry it will get messy pretty quick if we mix the two approaches

almarklein:
@Korijn good point. We could define an `alpha_compare_epsilon` in the base `Shader` in `shadercomposer.py`. Though we already use simple string replace in some places in this module, so that would not be out of place either, and that way the value of epsilon stays local the module.

Korijn:
OK. @panxinmiao can you try that alternative approach and see which ends up with cleaner code? :) Then we can choose the cleanest solution and merge this.

panxinmiao:
> Wouldn't it make more sense to provide the constant via jinja instead of using f-strings? I worry it will get messy pretty quick if we mix the two approaches

I was going to do this at first, but it didn't work.

https://github.com/pygfx/pygfx/blob/8913416f37683013516e549a6b46fb8d47b99e7e/pygfx/renderers/wgpu/_shadercomposer.py#L647-L651

Here, the whole `blending_code` is passed into the Jinja template as a parameter. It seems that the internal parameters will no longer be resolved.

> There is also [WGSL constants](https://www.w3.org/TR/WGSL/#creation-time-consts) but the logic to fiddle one in at the module level is probably more code than what we have now :)

I tried to add a const in shader code here like this

```
        blending_code = """
        let alpha_compare_epsilon : f32 = 1e-6;   // Using 'const' keyword will cause wgsl parsing error.
        {{ blending_code }}
        """
```
But `const` keyword cause an error, `let` works well.  So do we use this as the final solution ?

Korijn (approved):

Korijn:
We should probably open a bug report for the WGSL error at... maybe the [Naga repo](https://github.com/gfx-rs/naga)?

almarklein:
Thanks! I really like the end result!

PR #313, opened 2022-05-16
Allow vertex and index buffers

almarklein:
Different render functions may need to either use a buffer as an index or vertex buffer, while others use storage buffers. This PR makes that both such objects can use the same buffer objects (and thus use the same geometry). 

This *should* fix an issue in #309, but I have not tested it :D


panxinmiao:
> This _should_ fix an issue in #309, but I have not tested it :D

I tested it，It works! :)

Korijn (approved):

Issue #311, opened 2022-05-14
Manage dependencies of render functions


almarklein:
The render functions produce (the input for) the pipeline object for a world object. Certain inputs can be changed without affecting the pipeline, e.g. a color. Other inputs need to trigger a pipeline rebuild when they change, because they're build in.

Things that can change:
* all things that is in the pipeline as part of a uniform, like colors etc.
* partial updates to buffers (e.g. moving mesh vertices).
* partial updates to textures (e.g. updating new glyphs to the text atlas).

Things that require a pipeline rebuild:
* a buffer gets replaced by another buffer object (e.g. new geometry is attached).
* a texture object gets replaced (e.g. a `material.map`, or the glyph atlas gets resized).
* The render textures change (related to `renderer.blend_mode` and the object's opaqueness).

This is currently handled with a `.rev` attribute that gets bumped. The renderer compares the object's `rev` with its last seen `rev` and builds a new pipeline if it has changed.

There are two limitations with this approach:

* It's not the renderer code that bumps the rev, but logic in the object/geometry/materials. This is an abstraction leak (see #272) but also makes it less precise (e.g. replacing any buffer in a geometry will trigger a rebuild even if that buffer is not used).
* The pipeline is not only affected by its own stuff, but also by external things. This used to only be the `renderer.blend_mode`, for which we had a solution, but now that lights and a glyph atlas are coming soon, we need a more generic solution. In particular, we have:
  * `renderer.blend_mode`.
  * the lights in the scene.
  * atlas texture.
  * it's likely that there will be more when we think of fog, shadows, etc.

These are two separate (though related) problems. It me or may not be easier to tackle both at once.

----

I have some ideas on paper, but they need a bit more thought before I write them down here.

A brief summary of my thoughs on handling the external dependencies:
* The renderer function lets the renderer know what external factors it depends on. The renderer uses this to put the pipeline object in a specific "context". This context has a hash, so it can be invalidated (and with it all the pipeline objects stored within) when any of its "factors" change.

almarklein:
I just realized that some of the limitations that we have are due to our design, not wgpu-limitations. In particular, it should be possible to replace a buffer or texture without having to rebuild the pipeline, as long as it has (some of) the same properties. That is what binding-layouts are for.

This means that (if we made some changes) we won't need a recompile when the glyph atlas changed. And maybe a uniform could contain a dynamic number of lights.

One possibility to make more effective use of this, is to make our Buffer and Texture object's resizable.

[...]

PR #315, opened 2022-05-25
Refactor pipeline building

### LINKED ISSUE #192: Developer docs
In pygfx we have some internal API which is not intended for users, but for developers. We should document this under a separate chapter in the docs.

Topics:
* How uniforms work.
* How the shader templating works.
* Explain what WGSL is autogenerated, and what is not.
* List the common functions that are made available.

### LINKED ISSUE #249: Render mask constants
Instead of using integer literals as bitmasks, we could define constants and binary-or them, to improve readability of code a little bit.

So this is how it is now:

```python
render_mask = 1  # opaque
render_mask = 2  # transparent
render_mask = 3  # opaque and transparent
```

This is the suggestion:

```python
render_mask = RENDER_MASK_OPAQUE
render_mask = RENDER_MASK_TRANSPARENT
render_mask = RENDER_MASK_OPAQUE | RENDER_MASK_TRANSPARENT

# or
RENDER_MASK_BOTH = RENDER_MASK_OPAQUE | RENDER_MASK_TRANSPARENT
render_mask = RENDER_MASK_BOTH
```

Although the variable names are a bit long. :)

almarklein:
This is an attempt at #311. 

Somewhat high-level idea of what changed:

* Implements `Trackable` as a replacement for the `bump_rev`:  a system that can record the usage of properties, and then publish a certain change-level when any of these properties changes.
* Writing (custom) shaders has changed:
  * The role of the shader functions is now part of the Shader class.
  * A few methods must be implemented to provide render details in steps.
  * Tracking is enabled in these steps, so that they can be re-done when needed.
  * Things should now be more clear/consistent/documented.
* When stuff changes, the minimal amount of work is done to apply the change. Were previously the whole pipeline was rebuild (incl. recompilation), now it only re-does what's needed.
* A notable result is that colormaps and even materials can be swapped cheaply as long as their formats match.
* Three objects play a role when the shader "does it's thing": the wobject, the environment, and the shared object.
* There is a new environment class that holds information on the environment. Now only blend-mode. Later also light-counts. This object is re-used, wherever it is a match, and it is used as a hash to store environment-specific shaders and pipeline objects.
* The `Shared` object now has its own module. It contains the uniform buffer that is shared between all objects, and later the glyph atlas will go there.
* The flake8 rule F821 is enabled again (detecting undefined variables).

Other details / todo:
* [x] There is an enum for `RenderMode`. Closes #249. 
* [x] I wrote docs on how to write shaders. I guess this closes #192.
* [x] Tests for all the new stuff.
* [x] Refactored most of the `pygfx.renderers.wgpu` package. 
* [x] Re-implemented all our shaders. 
* [x] Updated custom shader examples. 
* [x] Ran all the examples. 
* [x] Reviews.
* [x] Self-review and check for todos. 

Future work:
* Make the mesh use vertex buffers instead of storage buffers.
* Manage updates to buffers and textures more efficiently (without iterating over them).
* Caching of pipeline object and the like.


almarklein:
Status:
* The new system to record usage and track changes is implemented, and tested to some extend. (will add more tests when I am more confident that this is the way to go).
* I am working on the other steps, but the current code is a mess, so I have not committed that yet.

almarklein:
Status:
* Simple mesh working again.
* Can replace a colormap without invoking a shader re-compile or pipeline rebuild!
* Implemented reactivity for a few other props, e.g. `.wireframe` does a recompile and `.side` only a pipeline compose.
* Still many lose ends though.

almarklein:
Status:

* Trackables now use an internal `._store` - stuff stored on it is automatically tracked. This makes for an easier API: less work in our Materials etc.  Plus the internal code is also simpler.
* Trackables can now deal with external stuff!
* That ... made things more complex. I refactored the whole module twice (or three times?) to get it in a state where it felt clean/understandable, without references all over the place.
* Much more tests!
* I explain the module and some design choices in the docstring.

Now up to updating the rest of the code ...

almarklein:
Status update:

This is starting to take shape. But found that the current situation does not work with multiple renderers 😆. I'm now thinking how to make this work with renderer state (blend mode) and other environment stuff (lights, shadows, etc.). It looks like this involves a lot of weaksets, weakkeydicts and all that. Trying to figure out a what that feels reasonably elegant ...

almarklein:
I think I've found a good solution. Converging and cleaning up again.

almarklein:
This is about ready 🚀  It is also massive. Probably easiest to read the new doc page and the updated custom object examples.

panxinmiao:
A great work!

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+
+This document explains how to write shaders for PyGfx. This may be useful
+if you want to improve the existing shaders, add new shaders to PyGfx, or
+if you want to implement custom shaders in your own project.
+
+
+The shader class
+----------------
+
+A shader object derives from ``WorldObjectShader``. Its purposes is to
```
Korijn, line 12:
```suggestion
A shader object derives from ``WorldObjectShader``. Its purpose is to
```

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+    )
+
+    @register_wgpu_render_function(SomeWorldObject, SomeMaterial)
+    class SomeShader(WorldObjectShader):
+
+        type = "render"  # must be "render" or "compute"
+
+        def get_resources(self, wobject, shared):
+            bindings = {
+                0: Binding("u_stdinfo", "buffer/uniform", shared.uniform_buffer),
```
Korijn, line 37: Is `u_stdinfo` a mandatory binding?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+
+The shader must implement a few methods. A typical shader is shown below:
+
+.. code-block:: python
+
+    from pygfx.renderers.wgpu import (
+        register_wgpu_render_function, WorldObjectShader, Binding, RenderMask
+    )
+
+    @register_wgpu_render_function(SomeWorldObject, SomeMaterial)
```
Korijn, line 30: I get the "some..." analogy here, but it may be more illustrative to use code that actually works and does something?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+                "primitive_topology": wgpu.PrimitiveTopology.triangle_list,
+                "cull_mode": wgpu.CullMode.none,
+            }
+
+        def get_render_info(self, wobject, shared):
+            n_vertices = ...
+            n_instances = 1
+            render_mask = wobject.render_mask
+            if not render_mask:
+                render_mask = RenderMask.all
```
Korijn, line 60: does the renderer default to all if you omit the key, or is it a required key to return?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+                "render_mask": render_mask,
+            }
+
+        def get_code(self):
+            return (
+                self.code_definitions()
+                + self.code_common()
+                + self.code_vertex()
+                + self.code_fragment()
+            )
```
Korijn, line 72: is this a required format?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+WGSL code and templating
+------------------------
+
+The shader code is written in `WGSL <https://www.w3.org/TR/WGSL/>`_. We use `jinja2-templating <https://jinja.palletsprojects.com/>`_
+to allow flexible code generation. Here's an example:
+
+.. code-block:: python
+
+        def get_resources(self, wobject, shared):
+            self["scale"] = 1.2
```
Korijn, line 114: is there an important reason this uses brackets instead of dots to access the attribute?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
+Writing PyGfx shaders
+=====================
+
+This document explains how to write shaders for PyGfx. This may be useful
```
Korijn, line 4:
```suggestion
This document explains how to write shaders for PyGfx for the WgpuRenderer. 
This may be useful if you want to improve the existing shaders, add 
new shaders to PyGfx, or if you want to implement custom shaders in 
your own project.
```

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+                ...
+                var out = get_fragment_output(varyings.position.z, color);
+                return out;
+            }
+            """
+
+Picking
+-------
+
+The output struct in the previous section also has a ``pick`` field that can
```
Korijn, line 191:
```suggestion
The `output` struct of the fragment shader also has a ``pick`` field that can
```

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+.. code-block:: python
+
+        def code_fragment(self):
+            return """
+            @stage(fragment)
+            fn fs_main(varyings: Varyings) -> FragmentOutput {
+                ...
+                var out = get_fragment_output(varyings.position.z, color);
+
+                // The write_pick templating variable is builtin too
```
Korijn, line 206: when is it true?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+                ...
+                return out;
+            }
+            """
+
+
+Colormapping
+------------
+
+Many materials in PyGfx support colormapping. We distinguis between colormaps
```
Korijn, line 250:
```suggestion
Many materials in PyGfx support colormapping. We distinguish between colormaps
```

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+with image input data, and vertex input data (texture coordinates). The number of
+channels of the input data must match the dimension of the colormap (1D, 2D or 3D).
```
Korijn, line 252:
```suggestion
channels of the input data must match the dimensionality of the colormap (1D, 2D or 3D).
```

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+The base shader class has two corresponding helper functions, and there
+is a wgsl helper function.
+
+For images / volumes:
+
+.. code-block:: python
+
+        def get_resources(self, wobjwect, shared):
+            ...
+            extra_bindings = self.define_img_colormap(material.map)
```
Korijn, line 263: how should these be returned?

docs/shaders.rst
```diff
@@ -0,0 +1,325 @@
...
+            ...
+
+        def code_fragment(self):
+            return """
+            @stage(fragment)
+            fn fs_main(varyings: Varyings) -> FragmentOutput {
+                ...
+                let color = sample_colormap(img_valu);
```
Korijn, line 271: where does `img_valu` come from? also, should it be `img_value`?

examples/cube.py
```diff
@@ -18,6 +18,19 @@
...
+m3 = gfx.MeshBasicMaterial(color="#11ff99")
+
+p1 = cube.geometry.positions
+p2 = gfx.Buffer(cube.geometry.positions.data * 0.7)
+
+g1 = cube.geometry
+g2 = gfx.Geometry()
+g2.positions = p2
+g2.normals = g1.normals
+g2.indices = g1.indices
```
Korijn, line 32: what is the point of these changes? they don't seem to be added into the scene at all?

examples/custom_object1.py
```diff
@@ -28,16 +28,51 @@ class TriangleMaterial(gfx.Material):
...
+        bindings = {
+            0: Binding("u_stdinfo", "buffer/uniform", shared.uniform_buffer),
+        }
+        self.define_bindings(0, bindings)
+        return {
+            "index_buffer": None,
+            "vertex_buffers": {},
+            "bindings": {
+                0: bindings,
+            },
```
Korijn, line 48: so this becomes...

```python
"bindings": {
    0: {
        0: Binding(....),
    }
}
```

Is that right?

examples/custom_object3.py
```diff
@@ -48,16 +48,71 @@ def color(self, color):
...
+        # Determine number of vertices
+        n = 3 * geometry.positions.nitems
+
+        # Define in what passes this object is drawn.
+        render_mask = wobject.render_mask
+        if not render_mask:  # i.e. set to auto
+            if material.is_transparent:
+                render_mask = RenderMask.transparent
+            else:
+                render_mask = RenderMask.opaque
```
Korijn, line 100: I think you should add a subsection to `docs/shaders.rst` to explain which render passes exist and how the rendermask relates to them.

examples/custom_object3.py
```diff
@@ -86,60 +141,17 @@ def vertex_shader(self):
...
 
-    def fragment_shader(self):
+    def code_fragment(self):
         return """
         @stage(fragment)
         fn fs_main(varyings: Varyings) -> FragmentOutput {
             var out: FragmentOutput;
-            let a = u_material.opacity * u_material.color.a;
-            out.color = vec4<f32>(u_material.color.rgb, 1.0);
+            out.color = vec4<f32>(u_material.color.rgb, u_material.opacity);
```
Korijn, line 149: the old code looks wrong, and the new code doesn't use the colormap's alpha attribute. is that intended?

pygfx/renderers/wgpu/__init__.py
```diff
@@ -1,11 +1,112 @@
...
+TODO: how we keep buffers and textures up-to-date
+
+## Caching
+
+Objects are stored for each specific environment type, by using the
+environment's hash. The environment includes a system to detect that
+it is no longer used to that all objects related to that environment
+can be cleaned up.
+
+We also want to re-use wgpu objects like pipelines and shadermodules.
```
Korijn, line 107: To be clear: this is not supported yet, right?

pygfx/renderers/wgpu/_environment.py
```diff
@@ -0,0 +1,167 @@
...
+        # self.uniform_buffer.data[xx] = yy
+
+    def register_pipeline_container(self, pipeline_container):
+        """Allow pipeline containers to register, so that their
+        env-specific wgpu objects can be removed.
+        """
+        self._pipeline_containers.add(pipeline_container)
+
+    def check_inactive(self, renderer, scene, renderer_state_hash, scene_state_hash):
+        """So some clean-up for the given renderer and scene,"""
```
Korijn, line 68:
```suggestion
        """Do some clean-up for the given renderer and scene."""
```

pygfx/renderers/wgpu/_environment.py
```diff
@@ -0,0 +1,167 @@
...
+        pass
+        # if isinstance(ob, Light):
+        #     lights.append(ob)
+
+    scene.traverse(visit)
+
+    state["lights"] = lights
+    scene_hash = (len(lights),)
+
+    return renderer_hash, scene_hash, state
```
Korijn, line 167: is this a temporary function?

pygfx/renderers/wgpu/_pipeline.py
```diff
@@ -0,0 +1,758 @@
...
+        # todo: also release shader modules that are no longer used
+        # todo: cache more objects, like pipelines once we figure out how to clean things up
+
+        assert isinstance(source, str)
+        key = source  # or hash(code)
+
+        m = self.get(key)
+        if m is None:
+            m = device.create_shader_module(code=source)
+            self.set(key, m)
```
Korijn, line 753: is it possible that a different device is used for the same source, and that that leads to problems?

docs/shaders.rst
almarklein: No, but it's common. I added the other 2 common uniforms to this doc (wobject's uniform buffer and material's uniform buffer).

docs/shaders.rst
almarklein: This is the way to set templating variables. Added a comment for clarification.

docs/shaders.rst
almarklein: From sampling the image texture. Added an extra line above to clarify this.

examples/cube.py
almarklein: I think this is me trying things out. Will clean this up.

examples/custom_object3.py
```diff
@@ -48,16 +48,71 @@ def color(self, color):
...
+@gfx.renderers.wgpu.register_wgpu_render_function(Triangle, TriangleMaterial)
 class TriangleShader(WorldObjectShader):
+
+    type = "render"
+
+    def get_resources(self, wobject, shared):
+        geometry = wobject.geometry
+
+        # Set templating variable "scale" to 0.2
+        self["scale"] = 0.2
```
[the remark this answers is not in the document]
almarklein: Will add a comment here too!

examples/custom_object3.py
almarklein: Good point, done!

examples/custom_object3.py
almarklein: mmm, I tried to simplify the render_mask logic by not using the color's alpha, but this is confusing in other ways. I made it more like we do in the real materials: multiply opacity with the color's alpha.

pygfx/renderers/wgpu/__init__.py
```diff
@@ -1,11 +1,112 @@
...
+                     │WgpuShaderModule│  │WgpuBindGroupLayouts│                   │
+                     └───────┬────────┘  └──────────┬─────────┘                   │
+                             │                      │                             │
+                             └──────────────────┐   │  ┌──────────────────────────┘
+                                                │   │  │
+                                            ┌───▼───▼──▼───┐         ┌─────────────┐
+                                            │ WgpuPipeline ◄─────────┤pipeline info│
+                                            └───────┬──────┘         └─────────────┘
+                                                    │
+                          Ubuffers                  │
```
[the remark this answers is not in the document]
almarklein: Fixed.

pygfx/renderers/wgpu/__init__.py
almarklein: Nope. Made that more clear.

pygfx/renderers/wgpu/_environment.py
almarklein: This code is ready to include scene-related things like lights, but does not actually do anything. I will comment out some lines here.

pygfx/renderers/wgpu/_pipeline.py
almarklein: No, because in pygfx we only use a single device.

docs/shaders.rst
almarklein: I updated the example in the docs to follow what most of our shaders do: collect bindings in a list and then convert the list to a dict.

almarklein:
💥

PR #319, opened 2022-07-02
Physically-Based Rendering system

panxinmiao:
Try to implement Physically-Based Rendering system.

Related & depend #309 

Accomplished:

- Add a PBR material `MeshStandardMaterial`, support:
  - metalness & metalness _map
  - roughness & roughness_map
  - emissive_map & emissive_intensity
  - normal_map & normal_scale
  - ao_map & ao_map_intensity 
  - env_map & env_map_intensity (for IBL)
  - ...
- Physically-Based Lighting
- IBL (Image-Based Lighting)
- Examples: `physically_based_rendering.py`, `physically_based_rendering2.py`

The first example shows the lighting rendering performance of materials with different metalness and roughness.

image

image

The second example shows a complete PBR rendering effect.

image

Run the examples to see the specific effects.

Todo:
- Refactor this under the new architecture (#315)
- Add support for some other PBR attributes (light_map, bump_map, etc.)
- Consider better APIs
- Code clean

Issue #323, opened 2022-07-05
Use vertex buffers for meshes


almarklein:
We use bindings (readonly storage buffers) about everywhere. For e.g. lines this is needed, but for meshes not really. Using vertex buffers instead could improve the performance, depending on your hardware.

The one issue holding this back is that when the wireframe property is set, the neighboring positions need to be accessed in order to get the picking coordinates.

Some options to move to vertex buffers:
* Make wireframe a material subclass with its own shader. Downside is that we must pick what lighting it supports.
* ~Reduce picking support for wireframe. Only pick vertex index, not sub-vertex barycentric coords.~
* Don't use vertex buffers but always use storage buffers. Does reduce the complexity of the codebase :)

Another option is to leave it as is. That way subclasses of mesh shaders do keep the freedom to implement more interesting/advanced stuff. Would be nice to have a benchmark to see how much it really matters performance wise. I suspect that for shaders with advanced lighting etc. it does not matter so much.

Korijn:
It's still baffling to me that we can't just set an option on the WGPU pipeline to only render the lines. That's how every other rendering engine does it, I'm pretty sure they don't do all this trickery just to render wireframes.

Korijn:
Also see https://github.com/m-schuetz/webgpu_wireframe_thicklines

almarklein:
> See here: https://github.com/gfx-rs/wgpu-native/commit/6d63d53044aae6bc958578d147dda7aca9e342d0

Yeah, it's biefly been there, but removed just a little while later: 
https://github.com/gfx-rs/wgpu-native/commit/c4c66465367c329fb5bc9455441ecf92dcb08700#diff-cc8e93e710bd9ac3eaa9ca09299e28ba2839191111c4157a43936d8e6e8b26a1

It seems like its (at least currently) not part of the webgpu spec. And if it would, it would likely be lines of 1 physical pixel thick.

panxinmiao:
I use vertex buffer (instead storage buffer) to transfer vertex related data not only for performance, I think it's also a standard practice. In my opinion, `VertexBuffer` exists for this purpose. My idea is that all geometry related attributes should be passed through vertex buffer, while material related attributes should be passed through uniform buffer. This can make the logic more unified.

After the last discussion (https://github.com/pygfx/pygfx/pull/309#issuecomment-1128904408), I searched some informations ( mainly the source code and implementation of threejs , :) ).  

[...]

PR #324, opened 2022-07-14
Lighting and Physically-Based Rendering system

panxinmiao:
This PR is the combination and refactoring of #309 and #319. 
It preliminarily realizes relatively complete lighting, shadows, and PBR system.

Main features:

- Lights & Shadows (and some helpers)
  - `AmbientLight`
  - `DirectionalLight` & `DirectionalLightShadow`
  - `PointLight` & `PointShadow`
  - `SpotLight` & `SpotLightShadow`

- New Phong Lighting model and Physically-Based Lighting model
- IBL support
- Two materials that are affected by light, `MeshPhongMaterial`, `MeshStandardMaterial`
  - The shader logic of `MeshPhongMaterial` has been greatly modified, so some old examples using `MeshPhongMaterial`  may have some compatibility problems.
  -  `MeshStandardMaterial` is a standard physically based material, using Metallic-Roughness workflow (Disney Principled)

I think for me, the process of realizing these functions is also the process of exploring some "wgpu" best practices.
There are still some issues that need to be discussed, some of which have been written in the code comments. Maybe I need to sort it out later, and we can discuss some specific issues in detail.


examples/light.py
```diff
@@ -0,0 +1,60 @@
+import time
```
almarklein, line 1: A short docstring to explain what these new examples demonstrate would be nice :)

examples/pbr.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+class SkyboxMaterial(gfx.Material):
+    def __init__(self, *, map, **kwargs):
+        super().__init__(**kwargs)
+        self.map = map
+
+
+@gfx.renderers.wgpu.register_wgpu_render_function(Skybox, SkyboxMaterial)
+class SkyboxShader(WorldObjectShader):
```
almarklein, line 24: The builtin Background material already supports skyboxes. If that was not sufficient, we could extend the functuinality or maybe introduce a builtin Skybox material. Could be done in a separate PR though.

almarklein (commented):
I'm on holiday so splitting this review in parts. I now looked at everything except for the stuff in renderers/wgpu.

This looks great already, and running the new examples feels unreal!

In the helmet example, how is the skybox reflected in the helmet, what component is responsible for this? It looks so nice!

I'll try to review the 2nd part soon.

Korijn:
Question: is the PBR model here compatible with blender PBR materials? It would be really cool if we can support the PBR material files you can download/purchase online for blender. :)

panxinmiao:
> In the helmet example, how is the skybox reflected in the helmet, what component is responsible for this? It looks so nice!
> 
This is IBL technology in PBR rendering. 
It treats the surrounding environment as one big light source, accomplished by manipulating a cubemap environment map. 
It treats each cubemap texel as a light emitter. According to the material roughness, sampling from prefiltered(mipmapped) environment texture with different levels of blur (different mipmap levels), as radiance.

panxinmiao:
> Question: is the PBR model here compatible with blender PBR materials? It would be really cool if we can support the PBR material files you can download/purchase online for blender. :)

At present, most PBR rendering systems should use [Disney principles](https://media.disneyanimation.com/uploads/production/publication_asset/48/asset/s2012_pbs_disney_brdf_notes_v3.pdf) (at least threejs).

I think blender should be the same, but I haven't verified it. :)

However, there may be some advanced PBR attributes that we have not yet implemented, such as "clearcoat", "physically based transparency", and so on.

docs/reference.rst
```diff
@@ -103,6 +127,21 @@ Helper objects
...
+
+.. autoclass:: pygfx.DirectionalLightHelper
+    :members:
+    :member-order: bysource
+
+.. autoclass:: pygfx.SpotLightHelper
+    :members:
+    :member-order: bysource
+
+.. autoclass:: pygfx.DirectionalLightShadowHelper
```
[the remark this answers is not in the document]
panxinmiao: Er, this class is provided because directional light shadows are a little special. This is also one of the reasons why `position` and `target` are used instead of `direction` in the `DirectionalLight`'s API.

Ideally, directional light affects the whole scene, has only direction and no position. However, we need a camera (from the perspective of light) to render shadow maps, so it must have a position, and only the part of scene within the shadow camera's view frustum can generate shadow maps. The resolution of shadow map is certain, and the view frustum of shadow camera is also limited. If the view frustum of shadow camera is too large, the shadow sampling accuracy of nearby objects in scene will be too low.

This tool class is designed to help users set reasonable parameters to generate appropriate directional light shadows. The specific effect can be seen in the `light_directional_shadow.py` example.

If we want to realize the position independent global directional light shadow, we need to automatically calculate the appropriate viewing cone of the shadow camera (for the orthogonal camera, the direction has been determined, that is, we need to calculate the appropriate camera position), so that it can exactly include the current display scene, and use the CSM(Cascaded Shadow Mapping) algorithm to generate shadow maps in different regions. Maybe we can add these advanced functions later.

For shadow generation, directional light is more troublesome, while spotlight is relatively simple. The shadow camera parameters can be completely determined by the parameters of spotlight itself.

Sorry for my less proficient English, I hope I can explain it clearly.  :sweat_smile:

pygfx/cameras/_orthographic.py
```diff
@@ -57,6 +57,8 @@ def set_view_size(self, width, height):
         self.visible_world_size = width, height
 
     def update_projection_matrix(self):
+        # update visible_world_size from width and height first
+        self.set_view_size(1, 1)
```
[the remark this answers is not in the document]
panxinmiao: In some cases, we hope to dynamically change the `width` and `height` parameters of the camera, and then recalculate the projection matrix. As in the `light_directional_shadow.py` example:
```
camera.width = 100
camera.height = 100
camera.update_projection_matrix()
```
the `update_projection_matrix` method compose projection matrix from `visible_world_size`, but when the width and height change, `visible_world_size` does not update.

pygfx/materials/_mesh.py
```diff
@@ -214,6 +240,20 @@ def shininess(self, value):
...
+        """
+        return bool(self.uniform_buffer.data["flat_shading"])
+
+    @flat_shading.setter
+    def flat_shading(self, value: bool):
+        self.uniform_buffer.data["flat_shading"] = value
+        self.uniform_buffer.update_range(0, 1)
+
 
 class MeshFlatMaterial(MeshPhongMaterial):
```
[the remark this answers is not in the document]
panxinmiao: Currently, only `MeshPhongMaterial` and `MeshStandardMaterial` support this attribute. So, I haven't defined it in base class for the time being. I think that when we unify the rendering logic of all mesh materials later, it should be a general attribute in the base class.

pygfx/objects/_base.py
```diff
@@ -246,6 +249,15 @@ def material(self):
     def material(self, material):
         self._store.material = material
 
+    @property
+    def receive_shadow(self):
+        """Whether this object receive shadows. Default False."""
+        return self._store.receive_shadow
+
+    @receive_shadow.setter
```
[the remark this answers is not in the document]
panxinmiao: Changing the value of `receive_shadow` will cause the shader of the wobject to be recompiled, but `cast_shadow` will not. `cast_shadow` only determines whether the wobject participates in rendering shadow maps.

But, `cast_shadow` of `Light` objects is an exception, which will affect shader logic, so for `Light` objects, `cast_shadow` is passed to the shader as uniform.

pygfx/objects/_lights.py
```diff
@@ -0,0 +1,356 @@
...
+_look_target = Vector3()
+_proj_screen_matrix = Matrix4()
+
+
+shadow_uniform_type = dict(light_view_proj_matrix="4x4xf4")
+
+
+class LightShadow:
+    def __init__(self, camera: Camera) -> None:
+        self.camera = camera
```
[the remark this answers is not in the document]
panxinmiao: Yes, exactly.

examples/pbr.py
panxinmiao: I tried, but there seems to be a problem. The background is upside down.  I see a line of code in the `skybox.py` example:
https://github.com/pygfx/pygfx/blob/8375cbeaf5afd1a0a6957973af6f26ab7b3d16f6/examples/skybox.py#L43
It is used to flip the background, but so the objects in the scene also flip.

The implementation of skybox in this example is relatively correct. The only problem is that when rendering the skybox, we don't need to care about the depth test or write the depth buffer. The skybox should always be rendered first,  and put behind all objects in the scene. But now pygfx doesn't expose this ability (to control depth test and depth write), so I have to render skybox in another scene alone.

The same is true for stencil test and stencil write. I think we need to expose this capability so that we can achieve effects such as "outline".

Korijn:
> > Question: is the PBR model here compatible with blender PBR materials? It would be really cool if we can support the PBR material files you can download/purchase online for blender. :)
> 
> At present, most PBR rendering systems should use [Disney principles](https://media.disneyanimation.com/uploads/production/publication_asset/48/asset/s2012_pbs_disney_brdf_notes_v3.pdf) (at least threejs).
> 
> I think blender should be the same, but I haven't verified it. :)
> 
> However, there may be some advanced PBR attributes that we have not yet implemented, such as "clearcoat", "physically based transparency", and so on.

Thanks for the information. When this PR is merged I will try and see if I can support gltf model import. That supports PBR materials. :)

pygfx/utils/color.py
```diff
@@ -255,6 +255,20 @@ def css(self):
         else:
             return f"rgba({int(255*r+0.5)},{int(255*g+0.5)},{int(255*b+0.5)},{a:0.3f})"
 
+    def multiply_scalar(self, factor):
+        """Multiply the color by a scalar. The alpha channel is ignored."""
+        self._val[0] *= factor
+        self._val[1] *= factor
+        self._val[2] *= factor
+        return self
```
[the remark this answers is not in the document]
panxinmiao: I updated the code.

But I want to explain a little. The reason why this function was added before is that in the constructor of Color, the rgba component is truncated to a maximum of 1.0.

However, in some cases, values greater than 1.0 are meaningful. In some lighting BSDF functions, the color of the light is divided by `PI` as irradiance. In order to get the artist friendly lighting (not physically correct, but more colorful and brighter), when passing the light color to the shader,  it will be multiplied by `PI` as compensation.

See: https://github.com/mrdoob/three.js/blob/085222f7ca0de96ab362a8b048235bf7187ebc08/src/renderers/webgl/WebGLLights.js#L220-L221

panxinmiao:
>   * The shader logic of `MeshPhongMaterial` has been greatly modified, so some old examples using `MeshPhongMaterial`  may have some compatibility problems.

The biggest difference is that the current `MeshPhongMaterial` does not have built-in lighting. When using this material, we must add lights to the scene to illuminate objects. So some old examples using `MeshPhongMaterial` need to be modified.

I thought three solutions:

- Keeping the original and current `MeshPhongMaterial` at the same time. The old material logic remains unchanged, and the new logic is used as a new material (we need to give a different name).

- Modify all the old examples and add lights manually in the scene which using MeshPhongMaterial.

- Provide a util scene class with default lights (DirectionalLight and AmbientLight), such as `LightScene` or `DefaultLightScene` and so on. Modify the old example and use this scene class.

I prefer the second or third.

examples/pbr.py
almarklein: Ah, good point. I guess we need to add a property to the builtin lightbox to set the orientation. Good for a later PR. Let's discuss/track in #325.

At some point we could also make the renderer aware of background classes and render these first.

pygfx/utils/color.py
```diff
@@ -255,6 +255,26 @@ def css(self):
         else:
             return f"rgba({int(255*r+0.5)},{int(255*g+0.5)},{int(255*b+0.5)},{a:0.3f})"
 
+    def multiply_scalar(self, factor):
```
almarklein, line 258: > The reason why this function was added before is that in the constructor of Color, the rgba component is truncated to a maximum of 1.0.

I'd be open to removing clipping of values here, and apply clipping in the shader instead.

pygfx/objects/_lights.py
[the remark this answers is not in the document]
almarklein: Let's document that, because it may be confusing :)

pygfx/objects/_base.py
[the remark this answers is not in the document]
almarklein: > `cast_shadow` only determines whether the wobject participates in rendering shadow maps.

Then this would affect the environment right? So we still want it to be trackable?

pygfx/cameras/_orthographic.py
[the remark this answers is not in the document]
almarklein: Ah, I see. But now you're also overriding the view aspect. I think we should pull these apart. The `set_view_size()` function should only set `self._view_aspect = width / height`. The remaining lines can be moved to `update_projection_matrix`.

almarklein (commented):
> When using this material, we must add lights to the scene to illuminate objects. [...] I thought three solutions:

I think we should make/keep it easy to make simple things, so some form of default lighting makes sense. I like `DefaultScene`.

Can we actually replicate the default light as it was, as if attached to the camera?

pygfx/objects/_base.py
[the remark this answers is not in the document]
panxinmiao: > Then this would affect the environment right? So we still want it to be trackable?

No, it isn't.

The shadow map is rendered every frame. When rendering, the objects in the scene which cast shadows are rendered one by one.  Other objects (which not cast shadows) will be skipped.

Just like rendering the scene itself, When rendering,  only the objects that need to be rendered in the scene are rendered one by one, and the invisible objects are directly skipped.

panxinmiao:
> Can we actually replicate the default light as it was, as if attached to the camera?

I think so. 
In this case, maybe we need something like `DefaultLightCamera` instead of `DefaultScene`?

pygfx/renderers/wgpu/_pipeline.py
```diff
@@ -607,27 +654,71 @@ def update_vertex_buffer_descriptors(self):
                 ],
             }
             vertex_buffer_descriptors.append(vbo_des)
+
+        instance_buffer = self.resources.get("instance_buffers", None)
+        if instance_buffer is not None:
```
almarklein, line 659: I'm not a fan of this approach, to be honest. Using a storage buffer for the instance matrices (as `main` currently does) can avoid quite a bit of logic and make the codebase simpler.

almarklein (commented):
I'm getting back from parental leave and finally found the time to do a thorough review. I left quite a few comments :) Also a few generic remarks:

* I think we'd have to decide on #323, because it affects the code in the PR quite a bit.
* You left some TODO's in the code. I commented on some, but there are some others (also in the shader code), of which I'm not sure whether it's something you plan to address in this PR.
* The examples introduce a few rather big files. These will be in Git history forever once we merge this. I know that ThreeJS does this too, but it scales badly if we get more such examples. I know that other solutions like auto-downloading the files also introduce complexities etc. though, so perhaps just adding them like ThreeJS does is the easiest solution? @Korijn thoughts?

almarklein:
> In this case, maybe we need something like DefaultLightCamera instead of DefaultScene?

A DefaultScene that has an ambient light and a directional light (or point light?) in the camera position/direction would be great 👍

almarklein:
I added some stuff to get the default lights back. To do so I added a DirectionalCameraLight that simply follows the direction of a given camera. Should probably also follow its position for shadows to work correctly, now that I think of it.

API-wise, such a light needs a reference to the camera, which make that this `DefaultScene` idea does not really work. So my suggestion now is that lights must be added to most examples. However, in the `show()` utility we can add lights by default 🥳

panxinmiao:
> * You left some TODO's in the code. I commented on some, but there are some others (also in the shader code), of which I'm not sure whether it's something you plan to address in this PR.

A few have been completed but the code or comments maybe have not been cleaned up. There are also some to-do items, I was not sure which way was better, so I want to listen to some suggestions from reviews. :sweat_smile:
I will sort out the code again.

panxinmiao:
> I added some stuff to get the default lights back. To do so I added a DirectionalCameraLight that simply follows the direction of a given camera. Should probably also follow its position for shadows to work correctly, now that I think of it.
> 
> API-wise, such a light needs a reference to the camera, which make that this `DefaultScene` idea does not really work. So my suggestion now is that lights must be added to most examples. However, in the `show()` utility we can add lights by default 🥳

I investigated some information. Many engines or 3D applications will use environment lighting when displaying objects by default, and provide a default environment (if the user does not set the environment).

see: https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/three-components/EnvironmentScene.ts

I think maybe i can add "env_map" support to all MeshMaterials, and provide a tool class to generate environment map from the scene (and a default scene like the above link). In this way, we can provide a standard default model display scenario to display various material models consistently.

almarklein:
> I think maybe i can add "env_map" support to all MeshMaterials, [..] provide a standard default model display scenario to display various material models consistently.

That also sounds like a good plan. If this is something that can be done in a later PR, perhaps that easier, because this PR already is quite big 😅. What I am aiming at with the default lighting is to have an easy way to make examples with phong materials work again.

almarklein:
Small change to the light that follows the camera. You can now just add a directional light to a camera. E.g. using `scene.add(camera.add(gfx.DirectionalLight()))`

pygfx/renderers/wgpu/_environment.py
```diff
@@ -57,6 +325,107 @@ def update(self, renderer, scene, blender=None, lights=None):
...
+
+        ambient_lights_buffer = self.ambient_lights_buffer
+        if not np.all(ambient_lights_buffer.data["color"][:3] == lights["ambient"]):
+            ambient_lights_buffer.data["color"].flat = lights["ambient"]
+            ambient_lights_buffer.update_range(0, 1)
+
+        # We only need to update buffers once before each draw
+        update_buffer(device, ambient_lights_buffer)
+
+        # TODO: Use light attributes directly to setup final lights_buffer?
```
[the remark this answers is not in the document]
panxinmiao: At present, the `uniform_buffer` of the light object does not actually map to the GpuBuffer. Instead, they only temporarily store the attribute data of the light and are uniformly packaged and mapped to the GpuBuffer in the update method of the environment. The `xx_lights_buffer` in the Environment actually undertakes the ability to communicate with the GPU to transfer light status. What I consider is whether the `xx_lights_buffer` in the Environment can be directly filled with data through the properties of the original light object. In this way, the light object does not need to maintain the `uniform_buffer` (actually, they do not need it at present).

pygfx/renderers/wgpu/_environment.py
```diff
@@ -57,6 +325,107 @@ def update(self, renderer, scene, blender=None, lights=None):
...
+            for i, light in enumerate(directional_lights):
+                direction = self._tmp_vector.sub_vectors(
+                    light.target.get_world_position(), light.get_world_position()
+                ).normalize()
+                light.uniform_buffer.data["direction"].flat = direction.to_array()
+
+                if light.cast_shadow:
+                    light.shadow.update_uniform_buffers(light)
+
+                    if light.shadow._map is None or light.shadow._map_index != i:
```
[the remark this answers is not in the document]
panxinmiao: `_map` and `_map_index` is used for internal shadow map rendering, and should not be used or accessd by user directly. 

Maybe should use reflection (`setattr` and `getattr`) to dynamically process this "attribute"?

pygfx/renderers/wgpu/_environment.py
```diff
@@ -36,9 +59,254 @@ def __init__(self, renderer_state_hash, scene_state_hash):
...
+                codes.append(code)
+
+        return "\n".join(codes)
+
+    def get_shader_kwargs(self, bind_group_index=1):
+        args = {
+            "num_dir_lights": self.dir_lights_num,
+            "num_spot_lights": self.spot_lights_num,
+            "num_point_lights": self.point_lights_num,
+            # "has_shadow": self.has_shadow,
```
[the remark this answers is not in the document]
panxinmiao: The uniform of light contains a boolean value of `cast_shadow` to control whether the corresponding light shadow rendering executed in the shader. I don't think it's meaningful to decide whether to recompile the shader by judging whether the whole scene contains shadows, so I removed this flag.

pygfx/renderers/wgpu/_environment.py
```diff
@@ -57,6 +325,107 @@ def update(self, renderer, scene, blender=None, lights=None):
...
+                            self.directional_lights_shadow_texture.create_view(
+                                base_array_layer=i
+                            )
+                        )
+                        light.shadow._map_index = i
+
+                if dir_lights_buffer.data[i] != light.uniform_buffer.data:
+                    dir_lights_buffer.data[i] = light.uniform_buffer.data
+                    dir_lights_buffer.update_range(i, 1)
+                    light.uniform_buffer._pending_uploads = []
```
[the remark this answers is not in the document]
panxinmiao: See the reply above: https://github.com/pygfx/pygfx/pull/324#discussion_r971504279

`light.uniform_buffer` does not really map and upload to the GPU. After the data to be uploaded is copied to the `dir_lights_buffer` in the Environment, it is considered that the data is "uploaded.".

pygfx/renderers/wgpu/_environment.py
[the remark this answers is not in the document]
almarklein: Aha. One thing to take into consideration is that if you have two scenes with different lights, they will share the same environment object if they have the same number of lights.

pygfx/renderers/wgpu/_environment.py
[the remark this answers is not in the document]
almarklein: What I mean is that in PyGfx there are three kind of attributes:
* Attributes only accessed by the object itself - these are private and should be prefixed with an underscore.
* Attributes accessed by the user. These are public. We'd want a `@property` for these so its documented, and enforce readonly when needed.
* Attributes accessed internally by other objects. This is the case here. If this is an attribute also used by the object, just prefix with underscore, but would be good to add a comment that the usage is intentional. If it's used only here to attach information to that object, its better to prefix it with e.g. `_env` to make that more clear.

pygfx/renderers/wgpu/_environment.py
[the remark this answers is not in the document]
almarklein: Ah, but the GPU representation for these objects is never created, so no need to invalidate the pending uploads (The `_pending_uploads` list contains at most one item for buffers, so its fine :))

pygfx/renderers/wgpu/_environment.py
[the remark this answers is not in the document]
panxinmiao: Ah, I see. The `_map` is not only accessed in `Environment`, it created in the `Environment`, and also used in shadow rendering (in `ShadowUtil`), so maybe the underscore should be removed here.

What I thought about earlier is that when users see these attributes (for example, from the code prompt and auto completion), they can realize that these attributes should not be accessed generally, so I define them as private attributes.

pygfx/renderers/wgpu/_mipmaputil.py
```diff
@@ -0,0 +1,191 @@
...
+
+            src_view = dst_view
+
+        self.device.queue.submit([command_encoder.finish()])
+
+
+_cache = WeakKeyDictionary()
+
+
+def get_mipmap_utils(device) -> MipmapUtils:
```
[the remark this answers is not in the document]
panxinmiao: I realized this, but I didn't find a suitable way to get this device. 😅

pygfx/renderers/wgpu/_environment.py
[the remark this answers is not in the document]
almarklein: > they can realize that these attributes should not be accessed generally, so I define them as private attributes.

Yeah, then its good to either add a prefix, e.g. `_wgpu_xx` or `_env_xx`, or to add a comment that accessing the private attribute is deliberate.

pygfx/renderers/wgpu/_mipmaputil.py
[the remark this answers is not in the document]
almarklein: That's fine, but at least the cache can be replaced with a global :)

almarklein:
@panxinmiao In #323 we've decided to use storage buffers everywhere. That means that the mesh shaders will need to be refactored. Preferably by subclassing the base `MeshShader`. The instance buffers must also be made storage buffers again. The use of VBO's in the shadow util is fine - this decision applies to the logic related to `Shader` classes.

We hope to get this PR merged soon, and I'm happy to help out - let me know if you prefer that I do the above refactorings! 👍

pygfx/renderers/wgpu/_mipmapsutil.py
```diff
@@ -172,16 +171,10 @@ def generate_mipmaps(
...
-
-    if _cache.get(device) is None:
-        utils = MipmapUtils(device)
-        _cache.setdefault(device, utils)
-
-    return _cache.get(device)
+def get_mipmaps_util(device):
+    if not hasattr(device, "mipmaps_util"):
+        device.mipmaps_util = MipmapsUtil(device)
+    return device.mipmaps_util
```
almarklein, line 177:
```suggestion
    if not hasattr(device, "_pygfx_mipmaps_util"):
        device._pygfx_mipmaps_util = MipmapsUtil(device)
    return device._pygfx_mipmaps_util
```

almarklein, line 177:
```suggestion
_the_mipmap_util = None


def get_mipmaps_util(device):
    global _the_mipmap_util
    if not _the_mipmap_util:
        _the_mipmap_util = MipmapsUtil(device)
    return _the_mipmap_util
```

almarklein:
Tweaked some examples, quite easy to make them much more visually attractive. Love the gradient coming from the spot lights :)

image

pygfx/renderers/wgpu/meshshader_phong.py
```diff
@@ -0,0 +1,287 @@
...
+            var i = 0;
+            $$ if num_point_lights > 0
+                loop {
+                    if (i >= {{ num_point_lights }}) { break; }
+                    let point_light = u_point_lights[i];
+                    var light = getPointLightInfo(point_light, geometry);
+                    if (! light.visible) { continue; }
+                    $$ if receive_shadow
+                    if (point_light.cast_shadow != 0){
+                        let shadow = get_cube_shadow(u_shadow_map_point_light, u_shadow_sampler, i, point_light.light_view_proj_matrix, geometry.position, light.direction, point_light.shadow_bias);
```
almarklein, line 215: I get a shader error here if I have more than one spot lights in the scene (that also cast shadows). I briefly tried to track the cause, but was not able to find it yet. Naga seems to indicate its the first argument. Is something somehow different between `n=1`  and `n>1` point lights? BTW, multiple directional lights and spot lights does work.

almarklein:
I did observe that the shadows from the spot and points lights look very different than that of the directional light. The latter seems much more blurred ...

panxinmiao:
> @panxinmiao In #323 we've decided to use storage buffers everywhere. That means that the mesh shaders will need to be refactored. Preferably by subclassing the base `MeshShader`. The instance buffers must also be made storage buffers again. The use of VBO's in the shadow util is fine - this decision applies to the logic related to `Shader` classes.
> 
Well, although I prefer VertexBuffer, it is more important to keep the architecture consistent, 😅 

> We hope to get this PR merged soon, and I'm happy to help out - let me know if you prefer that I do the above refactorings! 👍

Yes, that would be great. 😄

pygfx/renderers/wgpu/meshshader_phong.py
panxinmiao: Ah, I seem to have forgotten to test "multiple point light shadows". I will investigate this.

panxinmiao:
> I did observe that the shadows from the spot and points lights look very different than that of the directional light. The latter seems much more blurred ...

Yes, that's why I provided `DirectoryLightShadowHelper`, as I replied here: https://github.com/pygfx/pygfx/pull/324#discussion_r927225418.
In a large scene (especially when the viewport is not at the origin of world coordinates), you need to set reasonable shadow camera parameters.
We can use the CSM algorithm to improve it later.

pygfx/renderers/wgpu/meshshader_phong.py
panxinmiao: Bug have fixed.

panxinmiao:
I added a simple example of "wireframe_material.py": https://github.com/pygfx/pygfx/pull/324/commits/6eb3a907f791088e567252df7b29aba60783f69f
originally intended to provide a WireframeMaterial class (if we finally use VertexBuffer) to solve the wide wireframe problem. (Use barycentric coordinates)

Now we can use it as one example of a custom material. 😄

almarklein:
> Yes, that's why I provided DirectionalLightShadowHelper, as I replied here: https://github.com/pygfx/pygfx/pull/324#discussion_r927225418.

Ah, now I understand that comment better :)

> I added a simple example of "wireframe_material.py" [...] originally intended to provide a WireframeMaterial class (if we finally use VertexBuffer) to solve the wide wireframe problem. (Use barycentric coordinates)

Oh interesting. So my assumption that a wireframe material needs a storage buffer was false.

almarklein:
I refactored so that the mesh phong shader and pbr shader inherit from the base shader. These subclasses now only implement a function to apply the light. And this function can be re-used in other objects later. Some improvements that @panxinmiao applied to the new phong shader are applied as well.

So now we can have PBR shading with a wireframe

pygfx/renderers/wgpu/meshshader_phong.py
almarklein: Confirmed 👍

almarklein:
It seems to me like the light helpers could just be objects that you make a child of a light object. The positioning and orientation is then automatic. Why is this not the case and is the light object passed as an argument to the helper?

Korijn:
> It seems to me like the light helpers could just be objects that you make a child of a light object. The positioning and orientation is then automatic. Why is this not the case and is the light object passed as an argument to the helper?

Agree 100% that that would be better. in general we should strive to avoid any indirect relationships via non-parent/child props.

almarklein:
The one think making me not 100% sure is that ThreeJS also does it this way (the way its currently implemented, *not* using parent-child relationships). I could not find an explanation why Three does it this way.

Korijn:
> The one think making me not 100% sure is that ThreeJS also does it this way (the way its currently implemented, _not_ using parent-child relationships). I could not find an explanation why Three does it this way.

I can't think of any specific reason other than inconsistent historical design decisions, perhaps.

CONTRIBUTING.md
```diff
@@ -69,3 +69,13 @@ flake8 .
...
 detailed log messages. Can be an int or any of the standard level names.
+
+
+## Attributes
+
+We distinguish four kinds of attributes:
+
+* Public attibutes for the user: should in most cases be ``@property``'s.
+* Private attributes: prefixed with "_" as usual.
+* Attributes used by other parts of pygfx but not intended for the user: prefixed with "_gfx_".
```
Korijn, line 80: I don't understand how this is different from other private attributes.

CONTRIBUTING.md
```diff
@@ -69,3 +69,13 @@ flake8 .
...
+* Attributes to store wgpu-specific data on a WorldObject or Material (the objects themselves are unaware): prefixed with "_wgpu_".
```
Korijn, line 81:
```suggestion
* Attributes to store WGPURenderer-specific data on a WorldObject or Material (the objects themselves are unaware): prefixed with "_wgpu_".
```

pygfx/resources/_texture.py
```diff
@@ -33,6 +33,11 @@ def __init__(self, data=None, *, dim, size=None, format=None):
         self._data = None
         self._pending_uploads = []  # list of (offset, size) tuples
 
+        assert encoding in ("linear", "srgb")
```
Korijn, line 36:
```suggestion
        if encoding not in ("linear, "srgb"):
            raise ValueError("encoding must be one of 'linear' and 'srgb'")
```

pygfx/utils/show.py
```diff
@@ -34,6 +37,15 @@ def show(object: WorldObject, up=None):
 
     camera = PerspectiveCamera(70, 16 / 9)
     look_at = camera.show_object(object)
+    scene.add(camera)
+
+    # Add lights if there are none
+    for ob in scene.children:
```
Korijn, line 43: this should probably use traverse

CONTRIBUTING.md
```diff
@@ -69,3 +69,13 @@ flake8 .
...
 You can set the `PYGFX_LOG_LEVEL` environment variable to get more
 detailed log messages. Can be an int or any of the standard level names.
+
+
+## Attributes
+
+We distinguish four kinds of attributes:
+
+* Public attibutes for the user: should in most cases be ``@property``'s.
+* Private attributes: prefixed with "_" as usual.
```
Korijn, line 79: I also encountered `__update_buffer_color` which is not covered here

CONTRIBUTING.md
almarklein: Well normally private attributes are only used by the object itself. There are a few cases where an attribute or method of one object is used by another in a different module or even subpackage. The idea here is to prefix such cases, which is a way of saying, yes I use a private var of another object, but this is intentional. Or in other words, it's private to pygfx :)

CONTRIBUTING.md
almarklein: Good point, we never use double underscores except for one place where it added. Will change that into a single underscore.

pygfx/utils/show.py
almarklein: Yes!

pygfx/resources/_texture.py
almarklein: Nah, this will be replaced with the one from the gamma PR anyway.

pygfx/objects/_lights.py
```diff
@@ -0,0 +1,522 @@
...
+
+    uniform_type = dict(
+        color="4xf4",
+        cast_shadow="i4",
+        light_view_proj_matrix="4x4xf4",
+        shadow_bias="f4",
+    )
+
+    def __init__(
+        self, color="#ffffff", intensity=1, *, cast_shadow=False, position=(0, 0, 0)
```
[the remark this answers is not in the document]
almarklein: Yeah, that makes sense!

examples/pbr.py
```diff
@@ -13,39 +13,25 @@
 
 
 def load_gltf(path):
```
Korijn, line 15: We could extract this function from the example and put it alongside trimesh_mesh. It's a pretty useful thing to have a gltdf loader, even if it's not entirely feature complete just yet

almarklein: Good idea!

pygfx/renderers/wgpu/_shaderlib.py
```diff
@@ -0,0 +1,693 @@
...
+
+class Shaderlib:
+    def light_deps_basic(self):
+        return """
+        fn getDistanceAttenuation(light_distance: f32, cutoff_distance: f32, decay_exponent: f32) -> f32 {
+            if ( cutoff_distance > 0.0 && decay_exponent > 0.0 ) {
+                return pow( clamp( - light_distance / cutoff_distance + 1.0, 0.0, 1.0), decay_exponent );
+            }
+            return 1.0;
+        }
```
almarklein, line 25: @panxinmiao are these wgsl snippets translated from glsl in TheeJS? It looks like this snippet is from the non-physically-correct part: https://github.com/mrdoob/three.js/blob/a3a4c821229ac554f3ce53cc2b9460cfd205aaaa/src/renderers/shaders/ShaderChunk/lights_pars_begin.glsl.js

I think we'd want to implement things as if `PHYSICALLY_CORRECT_LIGHTS` is always True.

almarklein: Mmm, it looks like this is the only snippet where the `PHYSICALLY_CORRECT_LIGHTS` applies. So easy enough to fix :)

almarklein:
Also relevant, TheeJS is planning to improve their color management: https://github.com/mrdoob/three.js/issues/23614

Korijn:
> Also relevant, TheeJS is planning to improve their color management: [mrdoob/three.js#23614](https://github.com/mrdoob/three.js/issues/23614)

Seems like they are aligning with the choices we are making here, for the most part!

almarklein:
These PI factor are ... interesting ... there are long discussions on TheeJS on removing / adding back such factors, and I'm super confused.

almarklein:
I grepped the threejs source code for PI and applied it in our code in the same places.

From looking at discussion at ThreeJS some of these factors are there because most materials need them, but then in other places that factor is undone for physical rendering. It's messy, and our best guess is that the ThreeJS devs made sure it's correct or produces the same results as blender or something.

Going for physically correct lights means that the lights are a bit weak by default, so we need to tune them up in the examples.

almarklein:
Ok, so I replicated what ThreeJS does. Also made an example in ThreeJS to compare how things look with all the different lights, and everything matches up now. Some things we need to decide on.

What would be the default value of the `decay` prop of the point and spot lights? A value of 2 is physically realistic, but it makes the light falloff quick and the needed intensity depends a lot on how far away from objects it is. So a value of 0, meaning no falloff, might be a better default to get reasonable results in quick scripts, even if not physically correct.

What would be the default light intensities? With physical lights things are suddenly darker because one of these PI factors is gone. I think my proposal is to set ambient intensity relatively low, and all the other intensities to 3, so that an ambient light plus any other light look good by default. I will reconsider after a night sleep.

almarklein:
This is about ready from my end. Only thing left to do is update the examples screenshots.

panxinmiao:
I've just come back from vacation and travel. @almarklein  has done so many great work! :)

>@panxinmiao are these wgsl snippets translated from glsl in TheeJS? It looks like this snippet is from the non-physically-correct part: https://github.com/mrdoob/three.js/blob/a3a4c821229ac554f3ce53cc2b9460cfd205aaaa/src/renderers/shaders/ShaderChunk/lights_pars_begin.glsl.js

Yes, a lot of rendering logic comes from threejs, :)

>I think we'd want to implement things as if PHYSICALLY_CORRECT_LIGHTS is always True.

Actually, I didn't achieve "physically correct rendering". 
"Physical based rendering" is not exactly the same as "physically correct rendering", which is also what the “Disney Principle” says, their shading model is art directable and not necessarily physically correct.

To achieve a "physically correct lighting" scene, not only does the logic of the light object change (using lumen power), but also depends on whether the model itself has relevant resources. We may need pre baked physically correct lightmaps.
Or if we use physically correct scene based lighting, we need to generate a pre filtered radiance map from the environment map (use the formula and method here: "https://learnopengl.com/PBR/IBL/Diffuse-irradiance")，This is also the function I want to achieve later.

Korijn (approved):

Korijn:
I've merged, as this PR is now in an acceptable state, and so we can unblock all the other ongoing efforts.

Any further improvements or fixes will have to come in through new PRs. Thank you @panxinmiao for sticking with us, this was a long journey but it's really a major step forward for pygfx!

Next up: text rendering. I think that's the last missing piece of the puzzle before we can call ourselves "feature complete", and we can switch to "refinement" mode.

almarklein:
> "Physical based rendering" is not exactly the same as "physically correct rendering", which is also what the “Disney Principle” says, their shading model is art directable and not necessarily physically correct.

Interesting! This [issue](https://github.com/pmndrs/react-three-fiber/issues/2127), not from ThreeJS itself, but a sub-project in the community, made me think that we're better off making physically based rendering the default from the start. We could still consider enabling the "disney" approach, but let's make a new issue for that if someone want to dive in.

almarklein:
Here's the code I used to produce the ThreeJS examples for comparison. Saving for later: https://gist.github.com/almarklein/42b87985f97de2b13034612a5df95c6a

PR #327, opened 2022-09-01
Fix CI screenshot verification

berendkleinhaneveld:
Update validate_volume.png. Seems that the volume visualization has improved a little bit, but I don't know exactly why?


Korijn:
Probably the software rendering drivers have changed in the meantime.

PR #328, opened 2022-09-06
Remove unneeded request_draw calls

berendkleinhaneveld:
I noticed some unnecessary draws on mouse movement (without any buttons pressed), so I updated the `handle_event` code to only request new draws when there was an actual change.

I also would like to propose to add an `auto_update` argument to the controller classes to configure this behaviour. That would make it possible for the user to have more fine-grained control over exactly when to draw or not.

### DIFF
--- a/pygfx/controllers/_orbit.py
+++ b/pygfx/controllers/_orbit.py
@@ -216,14 +216,14 @@ class OrbitController(Controller):
                 self.rotate_stop()
             elif event.button == 2:
                 self.pan_stop()
-            viewport.renderer.request_draw()
         elif type == "pointer_move":
             xy = event.x, event.y
             if 1 in event.buttons:
                 self.rotate_move(xy),
+                viewport.renderer.request_draw()
             if 2 in event.buttons:
                 self.pan_move(xy),
-            viewport.renderer.request_draw()
+                viewport.renderer.request_draw()
         elif type == "wheel" and viewport.is_inside(event.x, event.y):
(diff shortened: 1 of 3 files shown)


berendkleinhaneveld:
I've added a commit which adds an `auto_update` argument to the Controllers.

Korijn:
> I noticed some unnecessary draws on mouse movement (without any buttons pressed), so I updated the handle_event code to only request new draws when there was an actual change.

I'd expect this to be necessary for on hover effects?

berendkleinhaneveld:
Hover effects can be done with objects themselves that listen to hover/mouse enter/move/leave events. So I think these calls shouldn't be necessary.

pygfx/controllers/_orbit.py
```diff
@@ -216,20 +218,23 @@ def handle_event(self, event, viewport, camera):
...
-            viewport.renderer.request_draw()
         elif type == "pointer_move":
             xy = event.x, event.y
             if 1 in event.buttons:
                 self.rotate_move(xy),
+                if self.auto_update:
+                    viewport.renderer.request_draw()
             if 2 in event.buttons:
                 self.pan_move(xy),
-            viewport.renderer.request_draw()
```
almarklein, line 226: IIUC this is the line that caused new redraws when the mouse was moved. I didn't understand that this was caused by the controller when I first read the PR. Just marking this to make it clear :)

almarklein (approved):
Nice!

PR #330, opened 2022-09-23
Colorspace and gamma correction

### LINKED ISSUE #329: rgb Image colors washed out
Howdie,  and thanks for this very cool looking library... just getting my toes wet.

forgive the basic question, and apologies if this is discussed elsewhere, but I'm seeing washed out colors in the astronaut image when I follow the examples:
```python
# using either:..

image = gfx.Image(
    gfx.Geometry(grid=tex),
    gfx.ImageBasicMaterial(clim=(0, 255)),
)

image = gfx.Mesh(
    gfx.plane_geometry(512, 512),
    gfx.MeshBasicMaterial(map=tex.get_view(filter="linear")),
)
```

compared to vispy:

Screen Shot 2022-09-20 at 3 04 45 PM

I'm on a mac (12.6), python 3.10, PyQt5 (though I also see it with jupyter-rfb in notebook).

curious if this would a lower OS-level driver-type issue? or something I can tackle within the pygfx code?

almarklein:
Closes #329.

It was observed in #329 that the colors are washed out compared to e.g. Matplotlib and Vispy. After some research it appears that this is due to the srgb colorspace texture of the final step. Most examples in wgpu upstream use srgb, but I guess this convention does not hold for our ecosystem? I'm not sure what's right here, but I agree that the colors look "washed out", and that without srgb things look more vibrant.

It was observed that the colors still don't match that of Matplotlib. In fact, Matplotlib applies *some* conversion, as e.g. an input "#33333" does not appear as such on screen. The color transform applied by Matplotlib is close to a gamma correction with `gamma=0.82`. Close but not exactly.

This PR turns off srgb by default, but keeps an option to enable it. It also adds a `gamma_correction`  property to the renderer. This way, things look more similar by default, and can be made pretty close to MPL if you want.

I will also have a look to see if colors can become the exact same value on screen. There appears to be a small offset or rounding or something.

Side note: the wgpu funtion `get_preferred_format` was implemented just recently in wgpu-native, and wgpu-py returns a hardcoded value at the moment. We'd need to review this pygfx code when we do the next wgpu-py update.

### DIFF
--- a/pygfx/renderers/wgpu/_renderer.py
+++ b/pygfx/renderers/wgpu/_renderer.py
@@ -91,6 +91,15 @@ class WgpuRenderer(RootEventHandler, Renderer):
             See the corresponding property for details.
         show_fps (bool): Whether to display the frames per second. Beware that
             depending on the GUI toolkit, the canvas may impose a frame rate limit.
+        blend_mode (str): The method for handling transparency. See the blend_mode
+            property for details.
+        sort_objects (bool): Whether to sort world objects before rendering. Default False.
+        enable_events (bool): Whether to enable user interaction events. Default True.
+        prefer_srgb (bool): Whether the srgb colorspace is preferred for the output texture.
+            Default False.
+        gamma_correction (float): The gamma correction to apply in the final render stage.
+            Typically a number between 0.0 and 2.0. Default 1.0 (no correction).
+
     """
 
     _shared = None
@@ -106,6 +115,8 @@ class WgpuRenderer(RootEventHandler, Renderer):
         blend_mode="default",
         sort_objects=False,
         enable_events=True,
+        prefer_srgb=False,
+        gamma_correction=1.0,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
@@ -132,9 +143,15 @@ class WgpuRenderer(RootEventHandler, Renderer):
         # Get target format
         if isinstance(target, wgpu.gui.WgpuCanvasBase):
             self._canvas_context = self._target.get_context()
-            self._target_tex_format = self._canvas_context.get_preferred_format(
-                self._shared.adapter
-            )
+            # Select output format. Here we assume that the srgb or non-srgb variant is also available.
+            fmt = self._canvas_context.get_preferred_format(self._shared.adapter)
+            if fmt.endswith("-srgb"):
+                if not prefer_srgb:
+                    fmt = fmt[:-5]
+            else:
+                if prefer_srgb:
+                    fmt = fmt + "-srgb"
+            self._target_tex_format = fmt
             # Also configure the canvas
             self._canvas_context.configure(
                 device=self._shared.device,
@@ -153,6 +170,7 @@ class WgpuRenderer(RootEventHandler, Renderer):
 
         # Prepare object that performs the final render step into a texture
         self._flusher = RenderFlusher(self._shared.device)
+        self.gamma_correction = gamma_correction
 
         # Initialize a small buffer to read pixel info into
         # Make it 256 bytes just in case (for bytes_per_row)
@@ -313,6 +331,17 @@ class WgpuRenderer(RootEventHandler, Renderer):
     @sort_objects.setter
     def sort_objects(self, value):
         self._sort_objects = bool(value)
+
+    @property
+    def gamma_correction(self):
+        """The gamma correction applied in the final composition step."""
+        return self._gamma_correction
+
+    @gamma_correction.setter
+    def gamma_correction(self, value):
+        self._gamma_correction = float(value)
+        if isinstance(self._target, wgpu.gui.WgpuCanvasBase):
+            self._target.request_draw()
 
     def _set_wobject_pipelines(self):
         # Each WorldObject has associated with it a wobject_pipeline:
@@ -511,6 +540,7 @@ class WgpuRenderer(RootEventHandler, Renderer):
             None,
             raw_texture_view,
             self._target_tex_format,
+            self.gamma_correction,
         )
         self.device.queue.submit(command_buffers)
 
--- a/examples/validate_color.py
+++ b/examples/validate_color.py
@@ -1,0 +1,55 @@
+"""
+This example draws squares of reference colors. These can be compared to
+similar output from e.g. Matplotlib.
+"""
+# test_example = true
+
+from wgpu.gui.auto import WgpuCanvas, run
+import pygfx as gfx
+
+colors1 = ["#ff0000", "#770000", "#00ff00", "#007700", "#0000ff", "#000077"]
+colors2 = ["#000000", "#333333", "#666666", "#999999", "#cccccc", "#ffffff"]
+
+canvas = WgpuCanvas()
+renderer = gfx.WgpuRenderer(canvas)
+camera = gfx.OrthographicCamera(6, 2)
+camera.position.set(2.5, 1, 0)
+scene = gfx.Scene()
+
+plane = gfx.plane_geometry()
+for i, color in enumerate(colors1):
+    m = gfx.Mesh(plane, gfx.MeshBasicMaterial(color=color))
+    m.position.x = i
+    m.position.y = 0
+    scene.add(m)
+for i, color in enumerate(colors2):
+    m = gfx.Mesh(plane, gfx.MeshBasicMaterial(color=color))
+    m.position.x = i
+    m.position.y = 1
+    scene.add(m)
+
+canvas.request_draw(lambda: renderer.render(scene, camera))
+
+
+def show_mpl():
+
+    import matplotlib.pyplot as plt
+    from matplotlib.patches import Rectangle
+
+    fig, ax = plt.subplots()
+    ax.set_facecolor("k")
+    plt.xlim([0, 6])
+    plt.ylim([-1, 3])
+
+    for i, color in enumerate(colors1):
+        ax.add_patch(Rectangle((i, 0), 1, 1, facecolor=color))
+    for i, color in enumerate(colors2):
+        ax.add_patch(Rectangle((i, 1), 1, 1, facecolor=color))
+
+    fig.show()
+
+
+if __name__ == "__main__":
+    print(__doc__)
+    show_mpl()
+    run()
--- a/pygfx/renderers/wgpu/_flusher.py
+++ b/pygfx/renderers/wgpu/_flusher.py
@@ -92,14 +92,22 @@ class RenderFlusher:
         self._device = device
         self._pipelines = {}
 
-        dtype = [("size", "float32", (2,)), ("sigma", "float32"), ("support", "int32")]
+        dtype = [
+            ("size", "float32", (2,)),
+            ("sigma", "float32"),
+            ("support", "int32"),
+            ("gamma", "float32"),
+            ("_padding", "float32"),
+        ]
         self._uniform_data = np.zeros((), dtype=dtype)
         self._uniform_buffer = self._device.create_buffer(
             size=self._uniform_data.nbytes,
             usage=wgpu.BufferUsage.UNIFORM | wgpu.BufferUsage.COPY_DST,
         )
 
-    def render(self, src_color_tex, src_depth_tex, dst_color_tex, dst_format):
+    def render(
+        self, src_color_tex, src_depth_tex, dst_color_tex, dst_format, gamma=1.0
+    ):
         """Render the (internal) result of the renderer into a texture."""
         # NOTE: cannot actually use src_depth_tex as a sample texture (BindingCollision)
         assert src_depth_tex is None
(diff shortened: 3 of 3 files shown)


Korijn:
> Most examples in wgpu upstream use srgb, but I guess this convention does not hold for our ecosystem?

This leaves me wondering if those wgpu upstream examples also have washed out colors/images?

Korijn:
This thread from UE4 seems to indicate the bit resolution of image textures also influences the outcome?

https://forums.unrealengine.com/t/problems-with-srgb-and-washed-out-textures/129683

Korijn:
There's also this article which suggests we would need to do something in our shaders: http://filmicworlds.com/blog/linear-space-lighting-i-e-gamma/

rougier:
The "ideal" gamma (for human perception) is supposed to be 2.2 and I think this the one used with sRGB on OpenGL. The output on #329 is really washed out. Depending on the format of the image, it may come with an embedded color profile that is applied by matplotlib/vispy but not pygfx. To test, you can make a new image out of the two displayed image (e.g. screnshot)and render them in pygfx.

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -314,6 +327,17 @@ def sort_objects(self):
...
+    @property
+    def gamma_correction(self):
+        """The gamma correction applied in the final composition step."""
+        return self._gamma_correction
+
+    @gamma_correction.setter
+    def gamma_correction(self, value):
+        self._gamma_correction = 1.0 if value is None else float(value)
+        if isinstance(self._target, wgpu.gui.WgpuCanvasBase):
+            self._target.request_draw()
```
Korijn, line 339: why does this setter trigger a draw?

pygfx/resources/_texture.py
```diff
@@ -23,7 +23,9 @@ class Texture(Resource):
             (e.g. from ``wgpu.TextureFormat``).
     """
 
-    def __init__(self, data=None, *, dim, size=None, format=None):
+    def __init__(
+        self, data=None, *, dim, size=None, format=None, gamma_correction=0.45
+    ):
```
almarklein, line 28: This is a leftover from the previous attemp. Will remove it - this module should remain unchanged.

almarklein:
This is ready for review. Some examples may need their image updated, and I spotted something that I need to clean up. Otherwise this is ready from my end. I updated the first post to reflect the current state. I also wrote up my current understanding on the matter: https://almarklein.org/gamma.html

almarklein:
One question for @panxinmiao, in the lights PR, the lights have both a color and an intensity. I presume that the color is in sRGB and the intensity is a scale factor to be applied in the physiscal color space (i.e. it represents the intensity in terms of number of photons). Would you agree?

Korijn:
You can read about three.js handling here: https://threejs.org/docs/index.html?q=color#manual/en/introduction/Color-management

pygfx/resources/_texture.py
```diff
@@ -36,6 +38,8 @@ def __init__(self, data=None, *, dim, size=None, format=None):
         # Backends-specific attributes for internal use
         self._wgpu_usage = 0
 
+        self.gamma_correction = gamma_correction
```
Korijn, line 41:
```suggestion
        self._gamma_correction = gamma_correction
```

Korijn (approved):
Reviewed and approved. Looks like you need to update _all_ the reference screenshots before CI will pass. :) It makes sense though!

I do wonder if we can really do without a flag on Texture. I guess that means we only support texture that provide sRGB values (not-physical) for now?

almarklein:
> You can read about three.js handling here: https://threejs.org/docs/index.html?q=color#manual/en/introduction/Color-management

Interesting. Looks like the user is responsible for providing a color in physical colorspace:

> Colors supplied to three.js — from color pickers, textures, 3D models, and other sources — each have an associated color space. Those not already in the Linear-sRGB working color space must be converted, and textures be given the correct texture.encoding assignment.

Similarly, colors on materials, lights etc. are in physical colorspace. They do a little trick though: values provided as hex and css are assumed srgb and converted automatically.

```
color.setHex( 0x808080 );
console.log( color.r ); // → 0.214041140
console.log( color.getHex() ); // → 0x808080
```

I'm not sure what to think of this yet. As with some other aspects of ThreeJS, I'm not sure whether this design was intentional or the result of a series of design decisions over the course of its history 🤔   

> I do wonder if we can really do without a flag on Texture. I guess that means we only support texture that provide sRGB values (not-physical) for now?

Good point. Technically one can always convert any colorspace to the colorspace that pygfx expects. Or use a flag on the texture to indicate how the content should be interpreted. 

Maybe we need to checkout some more how other render engines do this.

Korijn:
> Maybe we need to checkout some more how other render engines do this.

Three.js: [texture has a flag](https://threejs.org/docs/index.html?q=text#api/en/textures/Texture.encoding)
Unreal: [texture has a flag](https://forums.unrealengine.com/t/problems-with-srgb-and-washed-out-textures/129683)
Unity: [texture has a flag](https://docs.unity3d.com/Manual/LinearRendering-LinearTextures.html#DisablingsRGBSampling)
Godot: [texture has a flag](https://docs.godotengine.org/en/stable/classes/class_texture.html)

panxinmiao:
> One question for @panxinmiao, in the lights PR, the lights have both a color and an intensity. I presume that the color is in sRGB and the intensity is a scale factor to be applied in the physiscal color space (i.e. it represents the intensity in terms of number of photons). Would you agree?

Er, I'd like to talk about my point of view.

My idea is that in order to get physically correct calculations, all colors read by shader programs are assumed to be in linear space, including reading from srgb textures with automatic conversion, and reading directly from non-srgb textures.

Note that this is only related to the encoding of the texture itself, but not the original image, because the program cannot know whether the data of the original image is srgb space or linear space. For programs, there is no difference, only digitals (although most color maps should be in the srgb space, because designers and artists work in srgb space, through the display screen). When loading materials from images, you should know whether they are in linear space or srgb space, and make reasonable settings, rather than being determined by the engine. I checked some GLTF model loader codes. After loading image textures, they are set to srgb space by default, because most texture maps are indeed in srgb space

The same is true for the basic color in other cases. The shader program assumes that the color received is the color in the linear space. How this "color in the linear space" comes from is determined by the user. 
(In this sense, the current values in pygfx `Color` class are assumed to be in linear space.)

In fact, for basic rendering, as long as the input and output are in the same color space, the impact in linear space or srgb space is not significant, or even can be ignored. Only in physics based rendering, you need to pay attention to the problem of color space.

Back to the question of "light", our current light color value is really in linear space. Theoretically, it is not completely correct if users use the color value in the srgb space to set the light (very likely, because users also work through the display screen or take some colors from the web palette).
But I don't think it's critical, because the lighting effect of the final scene is not only determined by the light color, but also related to many other factors (light intensity, distance, attenuation, shader implementation logic, engine implementation logic, etc. In order to get artist friendly effects, our code even directly multiplies the light color by `PI` for correction). No one can directly set a very accurate light color value to get the desired scene lighting effect. They must make adjustments to the rendering results.

Therefore, I think it is more important for the engine to have a set of self consistent and complete logic, without worrying about whether the rendering result is completely consistent with that of some other software.

Of course, it would be better if we could extend the `Color` class to have more perfect color space management functions.
I think that all color space conversion should be completed in the user layer or the CPU, without the concern of the GPU or shader.

almarklein:
Let me explain how the current state of this PR works. In the shader, the color of the fragment, which can come from multiple sources, is first determined. This color is assumed to be sRGB and converted to linear space, then follow the light calculations. Some examples:
* A basic mesh with a uniform color: sRGB color comes in as a uniform of the material. 
* Phong Mesh with map: color is sampled from the texture using texcoords.
* Image: color is sampled from the texture (geometry.grid is interpreted as color).
* CT data: value is sampled from the texture (geometry.grid is interpreted as data), and put through a colormap.

My point is that in all cases we come to a point in the shader where we have the fragment's "base color". This is the point where (in this PR) we move to linear space. Also see how a texture can contain data in one case and color in another. But this works fine, no intervention needed. In fact, the user does not have to mark textures as linear or srgb. All goes automatically, as long as your colors are sRGB.

So I propose that in PyGfx we expect all colors to be sRGB. This results in an easier API (not having to mark color textures as srgb). As far as I know there are no disadvantages (except for when your color is physical/linear, I'll address below).

> I checked some GLTF model loader codes. After loading image textures, they are set to srgb space by default, because most texture maps are indeed in srgb space

Indeed, the glTF spec dictates that all colors are sRGB :)

Sometimes you may have colors in physical/linear space though. We need a solution for that. My current thought is a flag on a texture, which can be considered  "auto" by default, but can be set to prevent the gamma correction step. Similarly, a method on the `Color` class to convert linear colors to srgb would be nice (which you seem to suggest as well).

> In fact, for basic rendering, as long as the input and output are in the same color space, the impact in linear space or srgb space is not significant, or even can be ignored. Only in physics based rendering, you need to pay attention to the problem of color space.

I agree that physics based rendering is practically impossible if you don't use physical colors. However, blending and antialiasing also benefit from treating colors the right way. And the effects are also visible with Phong shading :)

> Back to the question of "light", our current light color value is really in linear space.

Well, we can say that the light color is sRGB and convert to linear in the shader, and have happy artists and correct results :)

> But I don't think it's critical, because the lighting effect of the final scene is not only determined by the light color [...]
 
That's a reassuring thought.

> our code even directly multiplies the light color by PI for correction

Any idea where that PI comes from? I wonder if its still necessary once we've set things up correctly all the way.

pygfx/resources/_texture.py
```diff
@@ -21,10 +21,18 @@ class Texture(Resource):
             set from the data. This must be a pygfx format specifier, e.g. "3xf4",
             but can also be a format specific to the render backend if necessary
             (e.g. from ``wgpu.TextureFormat``).
+        colorspace (str): If this data is used as color, it is interpreted to be
+            in this colormap. Can be "srgb" or "physical". Default "srgb".
```
Korijn, line 25:
```suggestion
            in this colorspace. Can be "srgb" or "physical". Default "srgb".
```

pygfx/resources/_texture.py
```diff
@@ -67,13 +76,11 @@ def rev(self):
...
     @property
-    def gamma_correction(self):
-        """ """
-        return self._gamma_correction
-
-    @gamma_correction.setter
-    def gamma_correction(self, value):
-        self._gamma_correction = float(value)
+    def colorspace(self):
+        """If this data is used as color, it is interpreted to be in this colormap.
```
Korijn, line 80:
```suggestion
        """If this data is used as color, it is interpreted to be in this colorspace.
```

panxinmiao:
I think the engine assumes that the input color space and output color space (rather than letting the user decide) may cause some more complex problems.

For example, some visualizations of "scientific calculation and data statistics" are related to data by color values.

Also, for ordinary users, although they may work in the srgb space, they may not be aware of this. For example, they plan to turn red (255, 0, 0) into "semi red". Many people will set the color to (128, 0, 0) - this is usually the correct practice in most use cases. In the srgb space, its display effect is exactly the "semi red" that users see, but in the linear space, this color is not half the brightness of the previous color(physically), and the resulting calculation may not be what users expect.

This is especially true for lights. In PBR rendering, users may expect to change the color of a light (255, 255, 255) to (128, 128, 128) to reduce its brightness by half. In the color palette on the display (in the srgb space), it does become half (in line with users' expectations), but our program processes it in the "so-called correct" way (conversion from srgb to linear space), and assumed that the user knows this (in fact, the user does not know it), which makes the results inconsistent with the user's expectations.

Of course, for experienced users who understand the principle behind this, all this is not a problem, but they will pay attention to the problem of color space and have the ability to set it reasonably.

>Indeed, the glTF spec dictates that all colors are sRGB :)

Some pre baked lightmaps or HDR maps may be in linear space.

>Any idea where that PI comes from? I wonder if its still necessary once we've set things up correctly all the way.

This is from three.js. I think the possible reason is related to this function in the shader:
```
    fn BRDF_Lambert(diffuse_color: vec3<f32>) -> vec3<f32> {
        return 0.3183098861837907 * diffuse_color;  //  1/pi = 0.3183098861837907
    }
```
As I replied earlier, different from the color of objects, the color of light is a "somewhat abstract" attribute, which is related to the physical properties of optics, but it cannot be fully expressed by simply using "color".

No one can tell clearly what the result is for a light with a specific color value (we choose from the palette) to illuminate the scene. This depends on many factors, including how you understand the "color value" of the light (which to some extent determines the rendering logic)
However, in general, users may tend to think that a light with half the color will also have half the effect of illuminating the scene

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -115,7 +113,6 @@ def __init__(
         blend_mode="default",
         sort_objects=False,
         enable_events=True,
-        prefer_srgb=False,
```
panxinmiao: IMHO,  The original way may be better (use `prefer_srgb` flag)

almarklein: The target textures are now always sRGB, there is no way to turn it off anymore.

almarklein: The `prefer_srgb` was an attempt to fix the issue before I understood well enough how all this works :)

Korijn (approved):

PR #333, opened 2022-10-11
add example of click events on image

kushalkolar:
Shows how to work with click events on images to utilize the image "data coordinates" of the click event.

it took me too long to find `event.pick_info` to get the data coordinates so I thought it'd be useful!


almarklein:
LGTM. I don't understand why the example screenshot tests all fail. Likely related to the recent PR related to srgb, but I don't see how it affects this PR?

Could you rebase this on the latest `main` just in case?

almarklein:
> It's already rebased on latest main. The problem is that we released wgpu 0.8.2 which changed offscreen rendering behavior (quote changelog):

Aah 🤦

almarklein:
Well we need to update several screenshot anyway in #324. We can either merge this as is or wait until that one is merged.

Korijn:
If you rebase this on latest main, the tests will pass again.

kushalkolar:
Should be good to go now!

almarklein:
Thanks!

Issue #334, opened 2022-10-11
Getting bounding box on Volume raises AttributeError


berendkleinhaneveld:
The following example (modified example/volume_slice4.py):
```python
import imageio
from wgpu.gui.auto import WgpuCanvas, run
import pygfx as gfx


canvas = WgpuCanvas()
renderer = gfx.renderers.WgpuRenderer(canvas)
scene = gfx.Scene()

voldata = imageio.volread("imageio:stent.npz").astype("float32")
nslices = voldata.shape[0]
index = nslices // 2

vol = gfx.Volume(
    gfx.Geometry(grid=voldata),
    gfx.VolumeSliceMaterial(clim=(0, 2000), plane=(0, 0, -1, index)),
)
scene.add(vol)

camera = gfx.OrthographicCamera(128, 128)
camera.position.set(64, 64, 128)
camera.scale.y = -1  # in this case we tweak the camera, not the plane

# the BoxHelper gets the bounding box of the volume which fails
box_world = gfx.BoxHelper()
box_world.set_transform_by_object(vol)
scene.add(box_world)


if __name__ == "__main__":
    canvas.request_draw(lambda: renderer.render(scene, camera))
    run()
```

crashes with the following exception:

```
Traceback (most recent call last):
  File "../pygfx/examples/volume_slice4.py", line 30, in <module>
    box_world.set_transform_by_object(vol)
  File "../pygfx/pygfx/helpers/_box.py", line 122, in set_transform_by_object
    aabb = object.get_world_bounding_box()
  File "../pygfx/pygfx/objects/_base.py", line 417, in get_world_bounding_box
    return self._get_world_bounding_box()
  File "../pygfx/pygfx/objects/_base.py", line 424, in _get_world_bounding_box
    aabb = self._store.geometry.bounding_box()
  File "../pygfx/pygfx/geometries/_base.py", line 138, in bounding_box
    grid_shape = self.grid.data.shape[: self.grid.dim]
AttributeError: 'TextureView' object has no attribute 'data'
```

PR #335, opened 2022-10-11
Fix getting bounding box on volume

berendkleinhaneveld:
Fix for #334.

It is possible to create the geometry for a volume in multiple ways:

```python
geometry = gfx.Geometry(grid=voldata)
geometry_with_texture = gfx.Geometry(grid=gfx.Texture(data=voldata, dim=3))
```
In the first case, the `voldata` will be a TextureView, which does not have a `data` attribute. That was causing the exception.


Korijn:
If you rebase this on latest main, the tests will pass again.

PR #336, opened 2022-10-12
Update triangle example

almarklein:
Update triangle example to match the new wgpu triangle example.


almarklein:
Also cleaned up the cube.py, which contained some test-code. 

And mad it so the light direction is `(0, 0, -1)` is the direction is ill-defined. This makes the light look down the neg z-axis by default, which is good for many basic examples.

Korijn (approved):

PR #337, opened 2022-10-12
Remove vertex_buffers and index_buffer

### LINKED ISSUE #323: Use vertex buffers for meshes

almarklein:
Closes #323 

I also looked at `index_buffer`. For the mesh we cannot use it, because we'd lose information on what face we're in. So I figured we could drop this one too (in favor of using a binding). This means that for all shaders, `Shader.get_resources` returns only a dict of dict of bindings, so I renamed that to `get_bindings`.


pygfx/renderers/wgpu/_pipeline.py
```diff
@@ -261,7 +261,7 @@ def __init__(self, shader):
 
         # The info that the shader generates
         self.shader_hash = b""
-        self.resources = None
+        self.bindingss = None  # dict of dict of bindins, so multi-plural
```
Korijn, line 264: Typo

Korijn (approved):
Less code is more better! ✨

pygfx/renderers/wgpu/_pipeline.py
almarklein: Will also rename "bindings" to something more sensible.

PR #339, opened 2022-10-13
Fix wrong mipmap sampling level in IBL

panxinmiao:
I found that IBL on the current main branch does not work correctly.

As shown below:

image

The back of the helmet is not smooth and should not project the surrounding environment obviously.

That is irradiance is not sampled correctly from the corresponding blur level according to the roughness.

This PR fixed this.

image


almarklein:
Ah, nice catch!

PR #348, opened 2022-10-14
Check that buffer/texture data is contiguous

### LINKED ISSUE #307: dealing with non-contiguous arrays
I got this issue when dealing when setting data for Geometry to create a line using output from sklearn's `PCA.transform()`. It can be fixed using `np.ascontiguousarray(data)` but I'm not sure if this could effect other use cases too. The exception I receive is:

```
ERROR:wgpu:Draw error: memoryview: underlying buffer is not C-contiguous
```

I'll try to get a minimal code example that produces this, I was using some of my own data.

almarklein:
Closes #307.

### DIFF
--- a/pygfx/resources/_buffer.py
+++ b/pygfx/resources/_buffer.py
@@ -50,6 +50,10 @@ class Buffer(Resource):
         # Get nbytes
         if data is not None:
             mem = memoryview(data)
+            if not mem.c_contiguous:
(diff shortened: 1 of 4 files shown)


Korijn (approved):

almarklein:
Buffer and texture data does not actually have to be contiguous. The issue in #307 is more subtle than I originally thought.

Issue #85, opened 2020-11-02
IO (e.g. glTF)


almarklein:

Korijn:
I just stumbled on this link, sharing it here so we can reference it later:

https://github.com/KhronosGroup/glTF-Sample-Models

In particular it's got the Sponza scene which is a pretty great testbed for any engine:

![image](https://user-images.githubusercontent.com/1882046/158356088-f4cea0c3-7a4a-4202-9607-1ad69636fcb5.png)

almarklein:
Also look at [Pyrender](https://github.com/mmatl/pyrender) for gltf support (it uses trimesh extensively).

Korijn:
I propose to close this one as #349 improved supported formats greatly.

Issue #322, opened 2022-07-05
Caching of shader modules and pipeline objects


almarklein:
Right now, each world object gets its own shader modules and pipeline objects (plural because of multiple render passes). However, sometimes they could be re-used. Same holds for e.g. bind groups and bind group layouts.

In cases when there are a lot of objects in a scene, it's likely that most have the same material type, so this form of caching could save significant gpu memory and startup-speed for scenes with many similar objects.

The most tricky part is likely destructing the objects when nothing uses it anymore. -> A solution is a cache that uses weakrefs and also attaching the cached objects on the world object.

Some notes:

* A reusable cache as an "internal service" that is used throughout pygfx for all features that require caching, with a global toggle would be a very useful debugging feature. It can also provide statistics such as cache hit/miss ratios and usage stats.
* Could make more use of caching in the renderer. There are some todo's indicating possible locations.
* ~~We could re-use the command queue instead of recording it on each draw.~~ Could use render bundles (tracking in  #518)

Locations where caching is/can be used:

* [x] In the text subpackage, we apply some caching of font objects and obviously glyphs in the atlas, but that's all really specific to text, so it does not seem right to bring that into the same system.
* [x] `_mipmapsutil.py` #516 
* [x] `_shadowutil.py`. #530
* [x] `_pipeline.py` #525
* [x] `_renderer.py`
* [x] `_flusher.py` #533
* [x] `_blender.py` #533 
* [x] May just have to scan over the whole wgpu subpackage to identify more cases.
* [x] `_environment.py` -> a lot of texture and view creations, do some checks!
* [x] `_mipmapsutil.py` does not cache texture views and bind_groups. Do a bit of benchmarking to see whether that matters. If it does, add caching. If it does not, maybe remove some caching of those objects in other places ...

Korijn:
Some notes:

* Caching as an "internal service" that can be toggled off would be a useful debugging feature.

almarklein:
Updated the top post to list locations of interest for caching. Also spun the render bundle idea into a separate issue.

kushalkolar:
This is exciting! Naive question, does it make sense to store the cache on disk too to reuse it between pygfx runs? I've seen the "Building Vulkan shader cache" thing that pops up on steam the first time you launch a game or after updating gpu drivers.

almarklein:
>  does it make sense to store the cache on disk too to reuse it between pygfx runs?

[...]

PR #349, opened 2022-10-14
Make gltf loader a lot more generic

Korijn:
Relates to #85

* Adds a validation example for lighting and shadows
* Adds an example that renders the Sponza scene with custom tweaked lighting
* Revamps the gltf loading utility, I've broken it up into sub utilities that are now all available generally:
  * `gfx.load_scene(path)` which can load any file trimesh can handle (there's a lot) and return a list of Meshes (with potentially submeshes, etc)
  * `gfx.trimesh_material(material)` which will convert a Trimesh Material to a pygfx Material
  * `gfx.trimesh_geometry(mesh)` which will convert a Trimesh Geometry to a pygfx Geometry (we already had this)
  * `gfx.pillow_image(image)` which will convert a Pillow image to a pygfx Texture
* When comparing test screenshot output the diffs are now computed lazily, which makes executing this tests a bit faster

Fixes:

* I've moved the uv coordinate conversion from load gltf to `gfx.trimesh_geometry` so that it is applied generally
* I noticed that the AxesHelper with its default configuration is unusable. I've updated it a little bit to more sensibly size the arrow heads. Really the problem here is that the line thickness is in screen space and every other parameter is in world space. We can address that in the future.

image

image


almarklein:
Wow!

> Note: I tried enabling cast and receive shadow but the WGSL compilation process just kept blowing up.

Maybe #322 will help?

pygfx/utils/gltf.py
```diff
@@ -28,6 +28,7 @@ def load_gltf(path):
 
 def _parse_mesh(mesh):
     visual = mesh.visual
+    # todo: why?
     visual.uv = visual.uv * np.array([1, -1]) + np.array([0, 1])  # uv.y = 1 - uv.y
```
almarklein, line 32: @panxinmiao any idea about the reason for this texcoord flip?

pygfx/utils/gltf.py
```diff
@@ -41,22 +42,34 @@ def _parse_geometry(mesh):
...
+    if pbrmaterial.roughnessFactor is not None:
+        material.roughness = pbrmaterial.roughnessFactor
+
+    if pbrmaterial.metallicFactor is not None:
+        material.metalness = pbrmaterial.metallicFactor
+
+    if pbrmaterial.normalTexture is not None:
+        material.normal_map = _parse_texture(pbrmaterial.normalTexture)
+        # todo: why is this different from the default for MeshStandardMaterial?
+        material.normal_scale = (1, -1)
```
almarklein, line 69: My guess is differeny conventions or something. Maybe @panxinmiao has an idea?

almarklein (approved):

pygfx/utils/gltf.py
panxinmiao: This is because the uv coordinate systems of wgpu and opengl are different. The wgpu uses the D3D and Metal coordinate systems. The coordinate origin is in the upper left corner, while the opengl coordinate origin is in the lower left corner. trimesh loads textures according to the opengl coordinate system.

pygfx/utils/gltf.py
panxinmiao: To be honest, I'm not sure it's right. I guess it may also be because of the inversion of the UV coordinate system.

When I was writing the helmet example, I used threejs to load the helmet model, and tracked the `normalScale` attribute, it is indeed set to (1, -1), so I set it to (1, -1) here too, but it seems that the final rendering effect is almost the same as the default value (1,1).

panxinmiao: It took me a bit of time to figure it out. We should probably use the default (1, 1).

The root cause here is still the problem of the UV coordinate system.

For the standard GLTF model, there may be `tangent` vertex attributes, corresponding to the normalMap provided by it. At this time, we usually do not care about the coordinate system, because they are consistent.

However, some models do not contain the `tangent` vertex attribute. At this time, we use the screen-space derivatives to automatically calculate it. threejs's uv coordinate system and GLTF specification are y-flipped, so they compensate it by setting `normalScale.y = - 1`. The wgpu does not have this problem.

In addition, at present, we have not resolved the vertex attribute `tangent`, and all tangents are automatically calculated.
Although the [Sponza scene](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Sponza) example does have the `tangent` attribute.

I think it should be, but perhaps we need some more obvious examples to test to ensure that this is correct. 😅 

More details can be found here:
https://github.com/mrdoob/three.js/issues/11438
https://github.com/KhronosGroup/glTF-Sample-Models/pull/336

panxinmiao:
> Note: I tried enabling cast and receive shadow but the WGSL compilation process just kept blowing up.

I fixed it in #351, 😅 

BTW, when we render a large scene with multiple objects (especially when the rendering logic is complex), the performance may be poor. There are many objects outside the field of view, which do not need to be rendered. Before rendering, We should exclude them according to the camera's view frustum.
This part of the logic should be completed in the CPU, I think it should be in the `render()` method of `_render.py`.

pygfx/utils/gltf.py
almarklein: Let's put this in a comment.

Korijn:
Before I merge main I will add a test/example to prevent regressions.

Korijn:
I added an example that combines MeshPhongMaterial and MeshStandardMaterial with PointLight, AmbientLight, SpotLight and DirectionalLight to check that all combinations are working properly. I confirmed that it failed before merging main.

Korijn:
I had a ton of fun positioning and configuring the lighting. I'm really pleased with this result, doesn't it look amazing?

image

panxinmiao:
It's amazing!

PR #351, opened 2022-10-17
Fix shadows in MeshStandardMaterial

panxinmiao:
In my previous code, I redefined the structure of shadows in the shader, but only modified the rendering logic of MeshPhongMaterial, and forgot to modify MeshStandardMaterial. 😅 

This should solves the problem of shadows in #349.


Korijn (approved):

almarklein (approved):

almarklein:
Thanks for the after-care 😄  👍

Issue #332, opened 2022-10-10
Background API


almarklein:
Currently our background is implemented by creating a Background object. We have materials for colors (uniform colors and gradients), images, and cube maps.

An alternative (as I found was listed as a todo in the background module) is to set a background property on the renderer. Something to consider and decide about.

Korijn:
We should keep in mind that users should be able to compose "layers" to render themselves, and that backgrounds should not throw a wrench in that party. :)

almarklein:
Copying @panxinmiao's comment from #253:

I plan to add two attributes "background" and "environment" to the "Scene".

By setting the "background" attribute, we can distinguish the background object from other ordinary wobjects in the scene, so that the background can be rendered before all other objects. In addition, we can also consider making the "background" parameter accept a "Color" or "Texture", so that we can automatically generate it internally without exposing the "Background" or "SkyBox" objects in API.

By setting the "environment" attribute (should be a cubemap), we can affect all PBR materials (those without "env_map" set separately) in the scene. This saves the code in common use cases that we need to set the same environment map to multiple PBR material objects in a scene.

almarklein:
My first inclination is that I like how currently backgrounds are world objects just like everything else. However, backgrounds are a bit different than other objects ... and being able to do this would be pretty awesome:
```py
scene.background = scene.environment = cube_map_texture
```

I'd be interested to know what @Korijn thinks about this.

PR #353, opened 2022-10-20
SkyBox and enviroment map

panxinmiao:
The current SkyBox rendering logic is incorrect, and some maps are reversed. So in the previous example, when the same cubet exture is used as the scene background and pbr env_map, the rendering result is incorrect.

I made and added a standard cubemap ("examples/textures/cubemap.jpg") for testing.

In the `env_map_test.py` example, I compared the rendering results of the new SkyBox and the current SkyBox, The environment map of the ball uses the same cube texture as SkyBox. Observe the environment reflection of the ball and the text direction. The left one is correct, and the right one is the current wrong rendering.

image

In another example `env_map_cube_camera_test.py`, it is shown that using the cube map for the (correct) background SkyBox, and then using `CubeCamera` can generate the exact same cube map from the scene.

image


almarklein:
This PR is based off your other PR, making review a bit harder, especially when the other PR changes ... Could you make this PR only the new changes? E.g. by cherry picking the last few commits?  Let us know if you need some help with the Git commands!

panxinmiao:
> This PR is based off your other PR, making review a bit harder, especially when the other PR changes ... Could you make this PR only the new changes? E.g. by cherry picking the last few commits? Let us know if you need some help with the Git commands!

Ah, mainly because I used CubeCamera to test in the `env_map_cube_camera_test.py` example to verify the correctness of the new SkyBox rendering,

I will modify this example temporarily and update it after `CubeCamera` merges.

panxinmiao:
In addition, If you agree, I plan to add two attributes "background" and "environment" to the "Scene".

By setting the "background" attribute, we can distinguish the background object from other ordinary wobjects in the scene, so that the background can be rendered before all other objects. In addition, we can also consider making the "background" parameter accept a "Color" or "Texture", so that we can automatically generate it internally without exposing the "Background" or "SkyBox" objects in API.

By setting the "environment" attribute (should be a cubemap), we can affect all PBR materials (those without "env_map" set separately) in the scene. This saves the code in common use cases that we need to set the same environment map to multiple PBR material objects in a scene.

almarklein:
> In addition, If you agree, I plan to add two attributes "background" and "environment" to the "Scene".

That is interesting, but also a change in API. Let's discuss in #332!

My suggestion for now would be to create a new PR with just the skybox fix, so that we can merge that quickly.

panxinmiao:
> You might want to consider making `env_map_cube_camera_test.py` an example that is executed on CI with screenshot verification.

I'm not sure if we will eventually keep these two examples. At present, this PR is not final. I just want to use these two examples to reveal the current problem of sky box rendering.

At least when the wrong sky box rendering logic is deleted, we won't have the first example for comparison testing. 😅

panxinmiao:
I updated this PR and corrected the rendering logic of `Skybox`.

The implementation of new logic mainly depends on the following changes:

- Add "depth_write" and "depth_test" identifiers in "render_info", so that we can more flexibly control the "depth write" and "depth test" of the target.

- In the render() method, handle the `Background` object separately to ensure that it is always rendered first.

panxinmiao:
Screenshot verification failed on "validate_volume". 
It seems that it is due to other changes, which should not be caused by this PR, 😅

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -464,13 +465,28 @@ def render(
...
+        background = None
+
+        def _get_wobjects(obj):
+            if isinstance(obj, Background):
+                nonlocal background
+                background = obj
+            else:
+                wobject_list.append(obj)
+
+        scene.traverse(_get_wobjects, True)
```
almarklein, line 479: I don't really see the point for this change. It should not matter when the background is rendered since the depth test will take care of it. And even if it would not, it's easy for the user to just put the background first in the scene.

pygfx/materials/_background.py
```diff
@@ -110,51 +109,18 @@ def map(self, map):
...
+        self.map = map
 
     @property
-    def up(self):
-        """A Vector3 defining what way is up. Can be set to e.g. the
-        controller's up vector. The given vector is "rounded" to the
-        closest vector that is fully in one dimension.
-        """
-        return self._up
-
```
almarklein, line 134: I think we still want this up property to be able to orient the skybox, e.g. if you want a skybox in a scene where z+ is up.

pygfx/renderers/wgpu/backgroundshader.py
```diff
@@ -161,3 +166,114 @@ def code_fragment(self):
             return out;
         }
         """
+
+
+@register_wgpu_render_function(Background, BackgroundSkyboxMaterial)
+class SkyboxShader(WorldObjectShader):
+
```
almarklein, line 173: Any particular reason why you implement a new shader instead of fixing the existing one?

almarklein (commented):
Thanks for looking into this!

pygfx/renderers/wgpu/backgroundshader.py
panxinmiao: Err, the main reason is that I think the current implementation is a little complicated...
In the current implementation, the Skybox does not have a box like "entity" as it literally means. Its implementation process only exists in shaders and is procedure oriented.

In the class design, Skybox, as a special world object, can be added to the scene. It has a box entity. The only difference from ordinary world objects is that its position is fixed relative to the camera (in other words, it always follows and surrounds the camera). That is,, although the Skybox cannot be translated, other transformations such as rotation and scaling can be applied to Skybox objects.

Typical scenarios include:
1. Adjust the rotation angle and orientation of skybox:
 For example: `background.rotation.multiply(rot)`
 (Er, I'm not sure if this is the problem that the `up` property was designed to solve , 😅 )

2. Background follows the overall transformation of the scene:
 For example:  when `scene.rotation.multiply(rot)`,  The background will rotate with the scene

I think it may also be helpful to add other Skybox texture mapping modes, (`EquirectangularMapping`, etc.), The current implementation maybe has some "CubeTexture specific".

Of course, if you prefer the current implementation method, I would be happy to fix it on the current implementation. 😄

pygfx/renderers/wgpu/_renderer.py
panxinmiao: When a cube box is used to represent the Skybox object (rather than just a plane), we can't simply set the depth of each vertex to 0.99999 in the shader. 
For this cube box, we need to ensure that it is drawed behind all objects without depth testing or writing its depth information (it won't obscure any object). 
So we need to make sure that it is drawn first. Users can do this easily (by putting the background first in the scene), but we can't guarantee that users will do this.

Korijn:
We intend to take the validation examples from this and merge them in a separate, new PR, and then close this PR.

almarklein:
Superseded by #466 and #507.

PR #354, opened 2022-10-20
Use a hash for the key of a shader source

almarklein:
I observed that when changing material properties that should trigger the shader source to be recompiled, that this does not happen in all cases. ~After some digging it turned out that the hash lookup is the problem.~ It turned out that when the shaders are compiled, a new pipeline build is not enforced. Not sure whether this bug has been always there, or fell victim in a refactoring. 

* [x] Fix the bug that a new pipeline build was not enforced after a shader rebuild.
* [x] Fixed that `LimeMaterial.aa` was not tracked.
* [x] Use a proper hash as a key for cached shader modules.

### DIFF
--- a/pygfx/renderers/wgpu/_pipeline.py
+++ b/pygfx/renderers/wgpu/_pipeline.py
@@ -5,6 +5,7 @@ actual dispatching / drawing.
 """
 
 import wgpu
+import hashlib
 
 from ...resources import Buffer, TextureView
 from ...utils import logger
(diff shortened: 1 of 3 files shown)


pygfx/renderers/wgpu/_pipeline.py
```diff
@@ -740,7 +741,7 @@ def get_shader_module(self, device, source):
         # todo: cache more objects, like pipelines once we figure out how to clean things up
 
         assert isinstance(source, str)
-        key = source  # or hash(code)
+        key = hashlib.sha1(source.encode()).hexdigest()
```
Korijn, line 744: Well, this is ridiculous and makes no sense. Is it possible that `source` is actually a subclass of string or something else manipulated by Jinja?

almarklein:
Got it. Updated the description.

Korijn (approved):

almarklein:
The CI error is a most probably a glitch. Merging.

PR #356, opened 2022-10-21
panzoom controller can adjust camera scale using right click and drag

kushalkolar:
This adds functionality to `PanZoomController`, right click + drag can scale the camera x & y, this is similar to functionality in `pyqtgraph`. Not sure if this is the best way to implement it.

I think there should at least be a kwarg that allows it to be disabled? And not sure if `Vector3(1, 1, 1)` is the best way to initialize it if `scale` is not cloned from the respective camera. 

https://user-images.githubusercontent.com/9403332/197129466-1199ffc5-4818-4069-8986-05c7f7ef916d.mp4


almarklein:
I'd say there'd also have to be an option to maintain the aspect ratio.

kushalkolar:
> I'd say there'd also have to be an option to maintain the aspect ratio.

For a 2D controller that would just have the same effect as zoom right?

Do you think this is fine as a part of the `PanZoomController` or should it be a separate controller entirely?

kushalkolar:
added `maintain_aspect` kwarg and attribute, and `scale_factor` as a kwarg, and `mode` as a property that allows `"maintain-apsect"` or `"freescale"`

pygfx/controllers/_panzoom.py
```diff
@@ -45,6 +59,26 @@ def __init__(
         # Save initial state
         self.save_state()
 
+    @property
+    def mode(self) -> str:
+        if self.maintain_aspect:
+            return "maintain-aspect"
+        else:
+            return "freescale"
```
almarklein, line 67: This can just be a bool. Something like:
```suggestion
    @property
    def maintain_aspect(self) -> bool:
        return self._maintain_aspect
```

pygfx/controllers/_panzoom.py
```diff
@@ -16,6 +17,9 @@ def __init__(
         up: Vector3 = None,
         zoom: float = 1.0,
         min_zoom: float = 0.0001,
+        scale: Vector3 = None,
```
almarklein, line 20: Now we have a scale and a zoom. I'd expect to have one vector that has the same values for each dimension when the aspect ratio should be maintained. Can we make the zoom a vector?

pygfx/controllers/_panzoom.py
```diff
@@ -72,41 +108,85 @@ def look_at(self, eye: Vector3, target: Vector3, up: Vector3) -> Controller:
...
-    def pan(self, vec3: Vector3) -> Controller:
-        """Pan in 3D world coordinates."""
+    def pan(self, vec3: Vector3, button: int) -> Controller:
+        """Pan or stretch in 3D world coordinates."""
         self.target.add(vec3)
+
+        if button == 1:
+            pass
+
+        elif button == 2:
```
almarklein, line 118: Instead of this triage, the code would be more readable (I think) if it had a separate `def stretch()` for the new functionality. Same for start and stop.

almarklein (changes requested):
Thanks! Sorry for the late reply.

pygfx/controllers/_panzoom.py
kushalkolar: ah, so we can have a zoom `(x, y z)` to replace scale

pygfx/controllers/_panzoom.py
kushalkolar: I agree

kushalkolar:
I'll wait for the refactor mentioned in #403 before making another attempt at this, the nesting got messy

PR #359, opened 2022-10-21
Tweak flusher smoothing

almarklein:
When working with text I observed that on a hirez display the output is smoothed quite a bit. This tunes this down a bit.

I know that the triage can be written in 3 lines or so, but each `if`  represents a different situation, for which we may later want a different approach, so I prefer to keep it explicit.

Also fixed that you could not set `renderer.pixel_ratio`.


almarklein:
I'm going to merge this now, because I need it to continue in #305.

Issue #306, opened 2022-05-04
Merge geometry buffers util


Korijn:
Create utility function to concatenate geometry buffers and update indices

Korijn:
This was added in #169

PR #362, opened 2022-10-26
Execute examples in the same process

Korijn:
This avoids many types of overhead, like process creation, python interpreter startup, module importing, and wgpu device acquisition.

We use the `runpy` module here (and not `exec`) since it facilitates the use case quite well but also because it executes the code in a module namespace.

With this change the full test suite runtime is improved as follows:

* Locally: `2:47s` to `53s`
* CI: `4:49s` to over 30 minutes and still running :/


Korijn:
I've converted this to a draft and will later add profiling code to identify the slow parts on CI.

Korijn:
I've isolated the issue to [#306](https://github.com/pygfx/wgpu-py/pull/306), so I will revert my changes here to just capture the changes needed to run examples in the main process, and keep this as draft until we fix https://github.com/pygfx/wgpu-py/pull/306

Korijn:
Turns out runpy.run_module actually avoids importing by loading the code in a temporary module object.

Korijn:
So actually, thanks to the Shared object, we are reusing the same device and adapter throughout.

So we are leaking other things in this case, I think.

almarklein:
It could be that the leak here, and the one you demonstrated in wgpu-py have different causes ...

Korijn:
Let's see what happens when we disable the event loop in terms of memory usage and runtime, as we saw in #384 it should at least make things much faster, but perhaps it will also resolve the gigantic memory leak.

This will require more thought though, because disabling the event loop like this essentially means _nothing is rendered_ when you run an example. :)

Korijn:
I have a theory. When you run an example, you schedule a call to `animate` as the draw function, and then execute all pending tasks on the event loop when you call `run()`. It stops when all pending tasks have been processed. However, `animate` schedules another call to itself. This is left in there until the next example runs. The next example then essentially runs two examples (itself and the previous one) when you call `run()`. And the next one will run three. And so on. That explains the climbing runtime!

almarklein:
> . However, animate schedules another call to itself. This is left in there until the next example runs. The next example then essentially runs two examples (itself and the previous one) when you call run(). And the next one will run three. And so on. 

That sounds very likely! Nice detective work!

Korijn:
Now we climb up to 746.30 MB, and runtime has hardly moved! Woohoo! Let's try enabling all examples again...

Korijn:
Alright, this is ready for review.

To be clear, on CI, there is a lot of variance in performance. The pytest portion of the test examples job varies between roughly 4:20 and 6:30 (m:ss) when looking the recent commits on the main branch.

However, locally the performance is very significantly improved, from multiple minutes to 25.33 _seconds_.

I still don't fully understand why this speedup isn't visible on CI. It seems that the runtime on CI is unchanged.

almarklein (approved):

examples/tests/test_examples.py
```diff
@@ -22,38 +24,56 @@
...
+        if allow_call_later:
+            orig_call_later(*args, **kwargs)
+
+    def wrapped_run(*args, **kwargs):
+        nonlocal allow_call_later
+        allow_call_later = False
+        orig_run(*args, **kwargs)
+
+    wgpu.gui.offscreen.call_later = wrapped_call_later
+    wgpu.gui.offscreen.run = wrapped_run
```
almarklein, line 66: This stuff can be removed once there is a new release of wgpu-py that has the upstream fix. I will make an issue to track that.

almarklein: #405

Issue #367, opened 2022-11-01
Allow image to be placed anywhere in the same scene


kushalkolar:
Currently instances of `Image` are placed according to the size of the image and it doesn't seem like it's possible to set the position of the image within the scene manually. This would also allow multiple images to be shown in the same scene and help with the implementation of option 1 here #360 

I notice that I can offset the `Image` position by adding values to the `geo.positions` in the shader code here, I could just feed it the information from `Geometry.positions` but I'm not sure if that's the best way (I've never written shaders before :smile: ) https://github.com/pygfx/pygfx/blob/main/pygfx/renderers/wgpu/imageshader.py#L71

Korijn:
Image has a position attribute, does it not work?

kushalkolar:
It seems like setting `positions` doesn't affect anything, I tried all combinations shown below. If I understand the shader code, it seems like it's just creating `pos1` and `pos2` from the texture dimensions? Unlike some of the other shaders where `positions` seems to be used.

```python
positions = [
    [150, 150, 0],
]

positions = [
    [150, 150, 0],
    [150, 150, 0],
]

positions = [
    [150, 150, 0],
    [150, 150, 0],
    [150, 150, 0],
]

positions = [
    [150, 150, 0],
    [150, 150, 0],
    [150, 150, 0],
    [150, 150, 0],
]
```

kushalkolar:
Looks like I should add a `Binding` for `geometry.positions`, will make a PR

almarklein:
Have you tried `image.position.set(150, 150, 0)`? (not the geometry, but the WorldObject)

kushalkolar:
ah, that works setting it on the `WorldObject`!

Korijn:
In general, all world object subclasses support position, rotation and scaling!

PR #369, opened 2022-11-02
Modify image shader to set image position

kushalkolar:
followup from #367 

without `positions` specified:

```python
image = gfx.Image(
    gfx.Geometry(grid=gfx.Texture(im, dim=2)),
    gfx.ImageBasicMaterial(clim=(0, 255)),
)
```
image

with `positions`, uses the positions to apply an offset:

```python
positions = [
    [100, 100, 0],
]

image = gfx.Image(
    gfx.Geometry(grid=gfx.Texture(im, dim=2), positions=positions),
    gfx.ImageBasicMaterial(clim=(0, 255)),
)
```

image

### DIFF
--- a/pygfx/renderers/wgpu/imageshader.py
+++ b/pygfx/renderers/wgpu/imageshader.py
@@ -4,7 +4,8 @@ from . import register_wgpu_render_function, WorldObjectShader, Binding, RenderM
 from ._utils import to_texture_format
 from ...objects import Image
 from ...materials import ImageBasicMaterial
-from ...resources import Texture, TextureView
+from ...resources import Texture, TextureView, Buffer
+import numpy as np
 
 
 vertex_and_fragment = wgpu.ShaderStage.VERTEX | wgpu.ShaderStage.FRAGMENT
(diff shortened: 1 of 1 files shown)


kushalkolar:
We could also allow `positions` to be a `4 x 3` array if a user want to manually assign the positions of all 4 corners, currently it requires a `1 x 3` array to just set the top left offset.

kushalkolar:
So after [this](https://github.com/pygfx/pygfx/issues/367#issuecomment-1301446101), could it still be useful to allow a `4 x 3` array in some use cases to display the image in a warped way, like as a trapezoid for example instead of a rectangle?

almarklein:
My inclination is that for such use-cases a mesh should be used. I can imagine support for deforming an image or volume, but then I think it would be another texture with per-pixel deformation.

Korijn:
Converting this to a draft to signal that it is not clear if we want to merge this or not.

kushalkolar:
> My inclination is that for such use-cases a mesh should be used. I can imagine support for deforming an image or volume, but then I think it would be another texture with per-pixel deformation.

makes sense. Perhaps we could raise exceptions if certain `Geometry` arguments aren't supported for specific WorldObjects? Setting `positions` for Image geometry doesn't raise any exception, it has no effect which is confusing for users.

almarklein:
> Perhaps we could raise exceptions if certain Geometry arguments aren't supported for specific WorldObjects? 

The problem with that is that we don't know how a geometry is used. It might be used by different world objects. E.g. the geometry for a mesh might also be used by a `Points`  object. In the example of `Image`, someone could in theory create a custom shader that does handle both `grid`  and `positions`.

Korijn:
My proposal:

* Close this PR
* Open a new issue if you want to further discuss the Geometry API
* Personally I would focus on providing high quality documentation and guides, rather than making the renderer perform validation on WorldObject/Geometry combinations. Performance is critical so we don't want to spend time every frame validating scene graphs.

PR #371, opened 2022-11-03
Update the example in the guide

FirefoxMetzger:
(I don't think this PR has an issue tracking the problem. If it does, please reference)

The current example in the guide is broken; `gfx.BoxGeometry` changed to `gfx.box_geometry` and the example doesn't add light sources, so you will only see a black screen.

This PR updates the guide to match what we are currently doing in the cube example. It also rewrites it a bit to reflect the addition of light sources and adds some (static) visualizations.

I also found a typo in `Contributing.md` that I'm fixing here.


docs/guide.rst
```diff
@@ -31,108 +31,131 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
-       ...  # we'll get to this
+    scene.add(cube)
 
+Objects are slightly more complicated than lights or cameras. They have a
+`geometry`, which controlls an object's form, and a `material`, which controls
+an object's appearance (color, reflectiveness, etc).
 
-    canvas.request_draw(animate)
+Now we have all the necessary ingredients and it is time to take a look. To do
+so we need to create a `canvas` to draw what we see (here a on-screen window)
```
almarklein, line 78:
```suggestion
so we need to create a `canvas` to draw what we see (here an on-screen window)
```

docs/guide.rst
```diff
@@ -31,108 +31,131 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+    def animate():
+        # rotate the cube a bit
+        rot = gfx.linalg.Quaternion().set_from_euler(gfx.linalg.Euler(0.005, 0.01))
+        cube.rotation.multiply(rot)
+        # update the canvas
+        renderer.render(scene, camera)
+        renderer.request_draw()
 
-    ...
+    # make the cube pop
```
almarklein, line 103: Pop may be confusion. Maybe something like:
```suggestion
    # Add a directional light to make the cube look better
```

docs/guide.rst
```diff
@@ -31,108 +31,131 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+        # redraw the scene
         renderer.render(scene, camera)
-        canvas.request_draw()
+        renderer.request_draw()
 
-    canvas.request_draw(animate)
-    app.exec_()
+
+    if __name__ == "__main__":
+        renderer.render(scene, camera)
```
almarklein, line 156:
```suggestion
```

docs/guide.rst
```diff
@@ -31,108 +31,131 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+        cube.rotation.multiply(rot)
+        # update the canvas
+        renderer.render(scene, camera)
+        renderer.request_draw()
 
-    ...
+    # make the cube pop
+    scene.add(gfx.DirectionalLight())
+    
+    # create a new window and new renderer
```
almarklein, line 106: I think this may be confusing. My suggestion would be to only show the new code here (also not the `run()`. People can see a few lines down how things are put together.

almarklein (commented):
Thanks! I like the somewhat playful style. Just a few small comments.

FirefoxMetzger:
@almarklein @Korijn It looks like we don't have RTD set up to generate temporary docs for PRs. I think it would be nice to have this to preview changes to the docs in a PR. Would one of you mind [enabling it on RTD](https://docs.readthedocs.io/en/stable/pull-requests.html)?

docs/guide.rst
FirefoxMetzger: 😅

docs/guide.rst
FirefoxMetzger: Fair. I was thinking of users that open an interpreter and copy past the bits into it while reading the guide. Those users will have to define a new canvas after closing it each time; hence the duplication.

Happy to update though, if you think this is more confusing than helpful.

FirefoxMetzger:
@almarklein A slightly tangental question, but I'm not sure if it is worth making an issue out of it: What is the logic behind calling `renderer.request_draw(animate)` and then later calling `renderer.request_draw()` again inside `animate`? 

I get the idea of a custom render loop where we modify the scene and then redraw, but the API calls for it seem odd, so I wonder if I am missing some context here.

> We already have one CI job doing the build. We might as well do that build on RTD and be able to view the docs too.

I would advocate for keeping both. The CI job building the docs is useful to debug if things break and will run much faster than the job on RTD. RTD on the other hand produces a rendered artifact that can be inspected, which is much better for checking styling and composition.

docs/guide.rst
almarklein: Yeah, I figured you had an interactive session in mind, but that's not really clear here.

Actually, in an interactive session, you could add the light, define the animate function, call `request_draw` on it, and the original canvas would now show the cube rotating!

almarklein:
> What is the logic behind calling renderer.request_draw(animate) and then later calling renderer.request_draw() again inside animate?

Some examples / use-cases only request a new draw when something has actually changed. Calling it from inside the animate function makes it draw continuously. (Calling `request_draw` without an argument re-uses the last used function).

almarklein:
I checked the pull-request-checkbox at RTD. It did trigger a build. Let's see if it also shows up here.

docs/guide.rst
FirefoxMetzger: @almarklein Oh, cool. In that case I think we should update to that and highlight it. Maybe we should even start with the renderer and a black screen and then make it come to life incrementally. What do you think?

FirefoxMetzger:
> Calling request_draw without an argument re-uses the last used function

Oh, this should definitely be documented somewhere. I noticed that calling it without any function from the start works, too. Is there a default placeholder?

> Calling it from inside the animate function makes it draw continuously.

That seems like an odd design choice. Won't we get an infinite recursion error at some point, because `animate` calls itself (through `request_draw`)? It may take a while to do so, because it only runs at 60 FPS, but it would seem like that.

FirefoxMetzger:
> edit: I think you may have to push a commit to trigger RTD.

Yep. It should get picked up on the next commit or for the next PR.

Korijn:
> That seems like an odd design choice. Won't we get an infinite recursion error at some point

It's not a direct call, it's just queueing the next render call in the event loop.

almarklein:
> It's not a direct call, it's just scheduling the next render call in the event loop.

Exactly. To add some background, the GUI backend is responsible for the scheduling, applying an FPS limit etc. Also good to know is that multiple calls to `request_draw` result in just one new draw.

almarklein:
>  I noticed that calling it without any function from the start works, too. Is there a default placeholder?

Yes, but that's a function that does nothing. You sure it wasn't set already? You can check the `canvas.draw_frame` attribute. That's what gets called.

FirefoxMetzger:
> You sure it wasn't set already? You can check the canvas.draw_frame attribute. That's what gets called.

`canvas.draw_frame` is indeed set to a bound function. However I didn't set this explicitly, hence me asking if there is a default value.

On my machine the following snippet produces the output below:

```python

import pygfx as gfx

# Populate a scene with a cube
scene = gfx.Scene()
geometry = gfx.box_geometry(200, 200, 200)
material = gfx.MeshPhongMaterial(color=(1, 1, 0, 1))
cube = gfx.Mesh(geometry, material)
scene.add(cube)

camera = gfx.PerspectiveCamera(70, 16 / 9)
camera.position.z = 400

scene.add(gfx.AmbientLight())
scene.add(gfx.DirectionalLight())

# Create a canvas and a renderer
renderer = gfx.renderers.WgpuRenderer(WgpuCanvas())
renderer.request_draw()
run()
```

image

docs/guide.rst
FirefoxMetzger: > Actually, in an interactive session, you could add the light, define the animate function, call request_draw on it, and the original canvas would now show the cube rotating!

I wanted to try this to see how to update the guide, but I didn't manage to make this work. `run()` is a blocking call and waits until the window is closed, so you can't modify the scene after calling it. 

At the same time, I don't get any render before calling `run()`. Instead, I get a light gray box and a friendly notification from Windows asking me if the program is stuck:

image

FirefoxMetzger: To resolve this in an elegant way, I moved the `animate` extension into the `putting it all together` section. So now there is a step-by-step hello world example followed by a full example that adds a custom draw function. How does this look?

almarklein:
> ```
> renderer.render(scene, camera)
> renderer.request_draw()
> ```

Aha, the first line renders the scene to the canvas. The second requests a new draw, but it does nothing, leaving the initial render intact. In about all cases we only call `renderer.render()` from inside an animate function. Now that I think of it, this might actually be required for buffer swapping etc.

docs/guide.rst
```diff
@@ -31,109 +31,116 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+and a `renderer` that will look at the scene (through the camera) and draw what
+it sees onto the canvas::
 
-.. code-block:: python
+    from wgpu.gui.auto import WgpuCanvas, run
 
-    # A renderer is associated with a canvas (or a texture) that it renders to
-    renderer = gfx.renderers.WgpuRenderer(canvas)
+    renderer = gfx.renderers.WgpuRenderer(WgpuCanvas())
+    renderer.render(scene, camera)
```
almarklein, line 85: The `renderer.render()` call should be done in the "draw event".
```suggestion
    renderer.request_draw(lambda: renderer.render(scene, camera))
```

Lambda's are a bit ugly. Maybe it makes sense to create an `animate()` function here with just that one line, and explain the idea. Then down below the animate function is extended.

docs/guide.rst
almarklein: If there is a window, but no event loop that keeps it alive, it looks unresponsive to the OS.

The `run()`  function starts the event loop and is blocking by default. If run from an IDE with interactivity set up (with the Qt event loop if using a Qt widget, or with asyncio when using glfw) then `run` does not block and things are truely interactive.

However, this event loop stuff is a bit of a mess in Python, so we cannot expect the user to have it set up ...

I like how you solved it.

docs/guide.rst
FirefoxMetzger: oh? Did I get that right that you aren't supposed to call `renderer.render` from outside a `draw_function`?

docs/guide.rst
FirefoxMetzger: > If run from an IDE with interactivity set up (with the Qt event loop if using a Qt widget, or with asyncio when using glfw) then run does not block and things are truely interactive.

Hm, I was running it in a normal python shell you'd get by calling the python executable. I guess you were referring to an ipython shell?

In any case, if you like the current solution then we can simply keep it like this and sort out the event loop when it becomes relevant the next time.

almarklein: > I guess you were referring to an ipython shell?

Yes, for example, or any IDE that can do some event loop integration. But this is another topic indeed. Let's forget about interactive use for now and go with the current approach 👍

docs/guide.rst
almarklein: Sorry for being a bit vague about this. I had to dig into the code a bit to get it straight :)  Let me try to write a piece of text that could be put in the guide (feel free to edit it to your style).

The `renderer.render(scene, camera)` call renders the scene, from the viewpoint of the camera, to the internal texture of the canvas. This does not put the image onto the screen yet; we need to request the canvas to perform a draw using `renderer.request_draw(animate)`. This makes the GUI schedule a draw event, in which it will call the given function, and then present the texture to the screen. Note that `renderer.request_draw()` can be called (without an argument) from hereon to schedule new draws.

FirefoxMetzger: I've updated the guide to reflect this. @almarklein Please double check if I got the logic right and this makes sense.

Korijn:
> I would advocate for keeping both. The CI job building the docs is useful to debug if things break and will run much faster than the job on RTD. RTD on the other hand produces a rendered artifact that can be inspected, which is much better for checking styling and composition.

I support this suggestion, but let's make sure we handle this in a separate PR. RTD can be quite finnicky to configure and it will generate some noise, most likely.

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+.. _full_example:
 
-    ...
+Animations
+----------
 
-    # The actual rendering
-    renderer.render(scene, camera)
+As promised in the previous section, here is a full example of how to use pygfx.
+It adds a little bit of flare to the hello world example by rotating the cube a
```
Korijn, line 112:
```suggestion
It adds a little bit of flair to the hello world example by rotating the cube a
```

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+    geometry = gfx.box_geometry(200, 200, 200)
+    material = gfx.MeshPhongMaterial(color=(1, 1, 0, 1))
+    cube = gfx.Mesh(geometry, material)
 
+    scene.add(cube)
 
-We can ask the canvas to schedule a draw event, and tell it what to do
-to perform a draw.
+Objects are slightly more complicated than lights or cameras. They have a
+`geometry`, which controlls an object's form, and a `material`, which controls
```
Korijn, line 74:
```suggestion
`geometry`, which controls an object's form, and a `material`, which controls
```

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
-    material = gfx.MeshPhongMaterial(color=(1, 1, 0, 01))
-    cube = gfx.Mesh(geometry, material)
+    import pygfx as gfx
 
-    scene.add(cube)
+    scene = gfx.Scene()
 
+Right now this scene is very desolate. There is no light, no objects, and
+nothing that can look at those objects. Let's change this by adding some
+`light` and a `camera`::
```
almarklein, line 56: The camera can be put as an object in the scene, but we don't do it here.
```suggestion
`light` and creating a `camera`::
```

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+`geometry`, which controlls an object's form, and a `material`, which controls
+an object's appearance (color, reflectiveness, etc).
 
-.. code-block:: python
+Now we have all the necessary ingredients and it is time to take a look. This is
+done by a `renderer` and happens in a two step process known as a draw call:
+First, the renderer looks through the camera we created earlier, paints an image
+of what it sees, and stores it in a intermediate buffer. Second, the renderer
+takes this buffer and hands it over to a `canvas` which is responsible for
+displaying the image (here inside an on-screen window)::
```
almarklein, line 82: Trying to make it a bit more precise:
```suggestion
Now we have all the necessary ingredients and it is time to take a look. This is
done by a `renderer`, which looks through the camera we created earlier, paints an image
of what it sees, and stores it in the internal buffer of the canvas. To display the result on
the screen, we need to request a draw. The canvas will schedule a good time to call our `draw_function`
and present the contents of its internal buffer to the screen when it returns.::
```

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
 
+    canvas = WgpuCanvas()
+    renderer = gfx.renderers.WgpuRenderer(canvas)
 
-    canvas.request_draw(animate)
+    def draw_function():
+        # update the internal buffer
+        renderer.render(scene, camera)
 
+        # Note: the handover to the canvas is implicit
```
almarklein, line 94: The internal buffer is already a responsibility of the canvas. Also at the low level (the canvas provides the texture to render to). So the handover is already inside `renderer.render()`.
```suggestion
```

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+        )
         cube.rotation.multiply(rot)
 
+        # update the internal buffer
         renderer.render(scene, camera)
-        canvas.request_draw()
 
-    canvas.request_draw(animate)
-    app.exec_()
+        # Note: the handover to the canvas is implicit
```
almarklein, line 150:
```suggestion
```

docs/guide.rst
```diff
@@ -31,109 +31,134 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+
+        # schedule the next draw call to show an animation
+        # Note: without arguments it will use the previous draw_function
+        renderer.request_draw()
```
almarklein, line 154: Worth noting (here, in the PR) that `renderer.request_draw()` is a wrapper for `canvas.request_draw()`. We use the renderer's version because then we don't need a reference to the canvas, leading to simpler code. So I don't think we need to mention it in the guide.

FirefoxMetzger: Hm, so as a user of pygfx would I be expected to not interact with the canvas, or would I need to configure it somehow?

My angle is that, if I am expected to interact with a canvas, then I would wonder what the value of the `renderer.request_draw` function is. However, if I am not expected to directly interact with a canvas, then I would wonder why we need to introduce it as one of the first things in the guide. 

Following that thought, I would also wonder if it would make sense to just have a default value for the canvas in `gfx.renderers.WgpuRenderer()` (my suggestion is the `auto` canvas) that a user can overwrite if they need to. That way we only need to bring in a canvas.

almarklein: Sorry if I only added confusion :)  I just wanted to mention it as "background". I don't think we should mention it in the guide, because in pygfx you typically don't need the canvas.

FirefoxMetzger: > because in pygfx you typically don't need the canvas.

In that case I would vote for leaving the canvas out of the guide completely. Perhaps we can combine this with a feature PR that makes the auto canvas the default canvas for a renderer?

almarklein: I would be against hiding it completely because it's still a useful concept. We could discuss this further, but let's do that in another issue/pr.

FirefoxMetzger: This is now resolved by #378 and #380.

FirefoxMetzger:
This PR waits until #380 is merged to use the new `gfx.show` in the guide.

FirefoxMetzger:
Alright, I rewrote the guide and based it on `gfx.show`. I think it is much more approachable now; however, we may wish to come up with more meaningful examples on how to use a `Canvas` and a `Renderer`.

I also didn't bring up the event loop explicitly, as we can now use the new `before_render` callback to make the animations work. If you think we should still introduce it, I would suggest that we come up with a small GUI example (maybe we can recycle an existing example) where we need an explicit event loop.

docs/guide.rst
```diff
@@ -31,109 +31,179 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
 
-.. code-block:: python
+If you run the above, you will notice that it looks exactly the same as our
+introductory example. Just as with a `Scene` above, `gfx.show` will create a
+`Canvas` for you unless you provide it explicitly, and by default this uses the
+``auto`` backend and hence we won't see any changes.
 
-    import pygfx as gfx
+Why then would we explicitly provide a `Canvas`? Well, beyond the `Canvas`
+itself a backend also provides an event loop. This event loop allows us to
```
almarklein, line 137: I think by far the most common use-case for providing a custom canvas is when someone builds a GUI, and wants gfx to render into a subwidget. If think should just say that here. It speaks more to the imagination, and now the user knows that PyGfx can indeed do that. The part about the event loop is true, but may be confusing to newcomers.

docs/guide.rst
```diff
@@ -31,109 +31,179 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+`Canvas` explicitly.
 
-    from PySide6 import QtWidgets
-    from wgpu.gui.qt import WgpuCanvas
+**Renderers**
 
+The third and final main building block is a `Renderer`. A `Renderer` is like an
+artist that brings all of the above together. It looks at the `Scene` through a
+`Camera` and draws what it sees onto the surface provided by the `Canvas`. Like
+any good artist, a `Renderer` is never seen without its `Canvas`, so to create a
```
almarklein, line 148: > Like any good artist, a `Renderer` is never seen without its `Canvas`

😆

almarklein (commented):
This makes the guide easier to read indeed. I think we can even mention the event loop less :)

docs/guide.rst
```diff
@@ -31,109 +31,179 @@ The `WGPU <https://github.com/pygfx/wgpu-py>`_ library provides the low level AP
...
+As you may expect by now, the output is the same as without the explicit
+reference because `gfx.show` will create a renderer if you don't provide it. For
+many applications this is perfectly fine; however, if we want to tackle more
+advanced problems (e.g., control the exact process on how objects appear to
+overlay each other) we may need to create it explicitly. For starters, it is
+enough to know that it exists and what it does, so that we can come back to it
+later when it becomes relevant.
+
+(@almarklein: Do we have examples that customize the renderer that we could
+reference here?)
```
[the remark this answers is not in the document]
FirefoxMetzger: @almarklein Hm, I was thinking about this as well but was wondering if this isn't too advanced a subject for a newcomer. The same for the blend mode. It's an excellent example, but I feel it would be better placed into a dedicated tutorial on blend modes and why they are important.

Re-reading the guide now, I think it could be fine to have the example as is. How do you think?

FirefoxMetzger:
Interesting. @almarklein @Korijn One of the validation examples fails after merging main into this PR. At the same time, this PR only touches the docs and doesn't contribute any code. Does one of you have an idea why this fails?

Korijn:
I'm personally suspicious of the upstream wgpu-native library and lavapipe software render not being entirely stable. I've been deep diving into some memory leak issues and I can reproduce them at the rust level. I think there's some race conditions and leaks outside our control.

This tech is all still very new. :)

FirefoxMetzger:
@Korijn Okay, if that isn't something we should fix in this PR, then this one is RTR from my side :)

almarklein:
That validation example also failed in another PR. My guess is that lavapipe had a new version that behaves slightly different for that shader.

docs/guide.rst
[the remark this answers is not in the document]
almarklein: Yeah, mentioning that the renderer allows advanced tuning of the rendering process, like you do now, is fine! 👍

Korijn (dismissed):

FirefoxMetzger:
@Korijn You are one hour early :D I had to still remove two comments I had to mark discussions with Almar. These are now gone, so now the PR is ready :)

Korijn (approved):

Issue #372, opened 2022-11-04
"geometry_plane.py" example doesn't render anything


FirefoxMetzger:
The example `geometry_plane.py` doesn't render anything. It produces a window, but then just hangs presumably because the rendering loop is never started.

Mainly creating this issue so that I can file a PR against it in a few minutes ... @almarklein @Korijn let me know if I should create issues for this and file PRs against issues, or if it is okay to just submit a PR directly.

almarklein:
For things that are simple fixes requiring little discussion, just a PR is fine IMO.

Korijn:
Yup, I agree with Almar!

PR #373, opened 2022-11-04
update geometry_plane.py

FirefoxMetzger:
Closes: #372 

This PR updates the example by adding a light source and calling the rendering loop in the same way as it is done in the guide. This avoids seeing a frozen window only when running the example as a stand-alone.

Note that the unit test for this example worked fine. Likely, because it doesn't run the `__main__` block which needed updating.

I also took the liberty to update the imageIO code in there to v3 :D


almarklein (approved):
Although we tend to use `request_draw()` in our examples and docs, just setting `.draw_frame` should be a valid way to set the animate function. This should work because backends schedule a draw event when initializing (and when resizing).

I found the problem, and will create a PR to fix it. Nevertheless, I think these changes are still good, because the make it more consistent with the other examples, and it cleans up how imageio is used.

Korijn (dismissed):
Agree with Almar's response!

FirefoxMetzger:
> Although we tend to use request_draw() in our examples and docs, just setting .draw_frame should be a valid way to set the animate function. This should work because backends schedule a draw event when initializing (and when resizing).

Do you think it would be useful to converge on one canonical way to do this?

Korijn (approved):

almarklein:
> Do you think it would be useful to converge on one canonical way to do this?

Well, it's the API of wgpu, which allows different ways of working in a few places (handling events being another). It gives some flexibility to apps built with it. I think I prefer leaving that as it is.

That said, in pygfx I'd focus on just one way to do things. I just checked, and this was the only use of `draw_frame` in pygfx :)

almarklein:
To elaborate a bit. The `draw_frame` attribute allows one to subclass a canvas and implement `draw_frame` as a method.

PR #375, opened 2022-11-04
Update examples to IIO v3

FirefoxMetzger:
This PR updates the examples to the ImageIO v3 API. For the most part, this is a simple renaming, except for `volread` calls, which are now `imread` calls, too. So those have changed.


almarklein (approved):
🚀

Korijn (approved):

Issue #378, opened 2022-11-08
allow customizing gfx.show


FirefoxMetzger:
In #376 we realized that `gfx.show` can massively simplify the examples and reduce boilerplate. Indeed, it reduces the current hello world example to:

```python
import pygfx as gfx

cube = gfx.Mesh(
    gfx.box_geometry(200, 200, 200),
    gfx.MeshPhongMaterial(color=(1, 1, 0, 1))
)

gfx.show(cube)
```

![image](https://user-images.githubusercontent.com/4402489/200676341-c53a23c4-a67a-4f5c-b045-6649f15d34ff.png)

Which is much more concise than the current example in the guide. (Although I must admit that this background doesn't feel very easy on the eyes.)

Thinking this a bit further, a nice addition here would be to be able to supply a custom `animate` function. The problem with that is that the `animate` function needs access to the `renderer` and potentially the `controller` to allow the scene to update correctly. To avoid this, I want to propose a `before_render` function that the default `animate` executes before rendering if present. With it, the rotating cube example could become:

```python
import pygfx as gfx

cube = gfx.Mesh(
    gfx.box_geometry(200, 200, 200),
    gfx.MeshPhongMaterial(color=(1, 1, 0, 1))
)

def animate():
    rot = gfx.linalg.Quaternion().set_from_euler(
        gfx.linalg.Euler(0.005, 0.01)
    )
    cube.rotation.multiply(rot)

gfx.show(cube, before_render=animate)
```

That said, we could go all the way and allow customizing all parts of the pipeline by wrapping `show` into a dataclass and then only creating those elements that haven't been set explicitly. Picking up the name from #376 (Display), we could do:

```python
import pygfx as gfx

cube = gfx.Mesh(
    gfx.box_geometry(200, 200, 200),
    gfx.MeshPhongMaterial(color=(1, 1, 0, 1))
)

def animate():
    rot = gfx.linalg.Quaternion().set_from_euler(
        gfx.linalg.Euler(0.005, 0.01)
    )
    cube.rotation.multiply(rot)

# run the application
display = gfx.Display()
display.on_before_render = animate
display.show(cube)
```

Where `display` could be customized further, e.g., by setting your own `display.canvas = wgpu.gui.offscreen.WgpuCanvas()`, etc. What do you think?

almarklein:
I agree that we need to make it easy to tweak show. Ideally it would be a smooth scale where you can use all defaults, or define all stuff yourself, and everything in between.

[...]

PR #380, opened 2022-11-09
make gfx.show customizable

FirefoxMetzger:
Closes: https://github.com/pygfx/pygfx/issues/378 and https://github.com/pygfx/pygfx/issues/383

This PR adds more flexibility to `gfx.show` by allowing the replacement of various parts of this default setup. In particular, it allows using a custom: `canvas`, `event_loop`, `renderer`, `controller`, `camera`, and `draw_function`. It also enhances the default draw function by introducing two callbacks: `before_render` and `after_render`, which allow customizing the draw call (eg. to add animations) without having to replace the entire thing.

I also changed the way lighting is treated by `gfx.show` and only add a light source when the user does not provide a `Scene` object. My angle here is that, if the user provides an entire scene, they probably have their reasons and don't want us to mess with it. I'm happy to revert this though if you think it would be better to add light.

I'm also not super happy with always having a controller (what if a person wants none), but I think this is a minor issue.

Finally, a point that we could consider is if we want to change the callbacks from kwargs on `show` to decorators. Ie., if we would like to enable something like:

```python
import pygfx as gfx

@gfx.show.before_render
def animate():
    ...

gfx.show()  # uses animate
```

---

Since this is my first PR that contributes actual code and I didn't find any active coverage tracking, I will ask explicitly: @almarklein  @Korijn What are your expectations towards tests and coverage?

### DIFF
--- a/pygfx/utils/show.py
+++ b/pygfx/utils/show.py
@@ -3,6 +3,8 @@ Quickly visualize world objects with as
 little boilerplate as possible.
 """
 
+import sys
+
 from .. import (
     Background,
     BackgroundMaterial,
@@ -21,7 +23,6 @@ def show(
     up=None,
     *,
     canvas=None,
-    event_loop=None,
     renderer=None,
     controller=None,
     camera=None,
@@ -29,11 +30,36 @@ def show(
     after_render=None,
     draw_function=None
 ):
-    """Visualize a given WorldObject in a new window with an interactive camera.
+    """Display a WorldObject
 
-    Parameters:
-        object (WorldObject): The object to show.
-        up (Vector3): Optional. Configure the up vector for the camera controller.
+    This function provides you with the basic scaffolding to visualize a given
+    WorldObject in a new window. While it does add scaffolding, it aims to be
+    fully customizable so that you can replace each piece as needed.
+
+    Parameters
+    ----------
+        object : gfx.WorldObject
+            The object to show.
+        up : gfx.Vector3
+            If set, set the default camera controller's up vector to this value.
+        canvas : WgpuCanvas
+            The canvas to use to display the object.
+        renderer : gfx.Renderer
+            The renderer to use while drawing the scene.
+        controller : gfx.Controller
+            The camera controller to use.
+        camera : gfx.Camera
+            The camera to use.
+        before_render : Callable
+            A callback that will be executed during each draw call before a new
+            render is made.
+        after_render : Callable
+            A callback that will be executed during each draw call after a new
+            render is made.
+        draw_function : Callable
+            Replaces the draw callback with a custom one. If set both
+            `before_render` and `after_render` will have no effect.
+        
     """
 
     if isinstance(object, Scene):
@@ -56,11 +82,8 @@ def show(
 
         canvas = WgpuCanvas()
         event_loop = run
-    elif event_loop is None:
-        # TODO: find the matching run function
-        raise ValueError(
-            "When providing a canvas you also need to provide the event loop."
-        )
+    else:
+        event_loop = sys.modules[renderer.target.__module__].run
 
     if renderer is None:
         renderer = WgpuRenderer(canvas)
--- a/pygfx/utils/show.py
+++ b/pygfx/utils/show.py
@@ -27,10 +27,10 @@ class Display:
 
     This class provides the basic scaffolding needed to visualize a given
     WorldObject. To do so the class chooses a sensible default for each part of
-    the full setup (@almarklein: what is the technical term for the full setup?)
-    unless the value is explicitly set by the user or exists as a child of the
-    ``object`` passed to ``show()``. For example, it will create a camera unless you explicitly set a
-    value for camera or if there is (at least) one camera in the scene.
+    the full setup unless the value is explicitly set by the user or exists as a
+    child of ``object``. For example, it will create a camera unless you
+    explicitly set a value for camera or if there is (at least) one camera in
+    the scene.
 
     Parameters
     ----------
@@ -205,12 +205,15 @@ def show(
     Parameters
     ----------
     object : gfx.WorldObject
-            The object to show. If it is not a :class:`gfx.Scene <pygfx.Scene>`
-            then Display will wrap it into a new scene containing lights and a
-            background.
+        The object to show. If it is not a :class:`gfx.Scene <pygfx.Scene>`
+        then Display will wrap it into a new scene containing lights and a
+        background.
     up : gfx.Vector3
         If set, and ``object`` does not contain a controller, set the camera
         controller's up vector to this value.
+        canvas : WgpuCanvas
+        The canvas used to display the object. If both ``renderer`` and
+        ``canvas`` are set, then the renderer needs to use the set canvas.
     canvas : WgpuCanvas
         The canvas used to display the object. If both ``renderer`` and
         ``canvas`` are set, then the renderer needs to use the set canvas.
@@ -218,12 +221,10 @@ def show(
         The renderer to use while drawing the scene. If both ``renderer``
         and ``canvas`` are set, then the renderer needs to use the set canvas.
     controller : gfx.Controller
-        The camera controller to use. If not set, Display will use the first
-        controller defined on ``object``. If there is none, Display will
-        create one.
+        The camera controller to use. 
     camera : gfx.Camera
         The camera to use. If not set, Display will use the first camera
-        defined on ``object``. If there is none, Display will create one.
+        in the scene graph. If there is none, Display will create one.
     before_render : Callable
         A callback that will be executed during each draw call before a new
         render is made.
@@ -231,8 +232,14 @@ def show(
         A callback that will be executed during each draw call after a new
         render is made.
     draw_function : Callable
-        Replaces the draw callback with a custom one. If set both
+        Replaces the draw callback with a custom one. If set, both
         `before_render` and `after_render` will have no effect.
+
+    Notes
+    -----
+    If you want to display multiple objects, use :class:`gfx.Group
+    <pygfx.Group>` instead of :class:`gfx.Scene <pygfx.Scene>` if you
+    want lights and background to be added.
 
     """
 
--- a/pygfx/utils/show.py
+++ b/pygfx/utils/show.py
@@ -18,14 +18,33 @@ from .. import (
     AmbientLight,
     DirectionalLight,
     Light,
+    Camera,
+    Controller,
 )
 
 
-class DecorSet:
-    """Set function attributes using decorators.
+class RenderCallback:
+    """A callback used during the draw call
 
-    This class implements a Descriptor that allows setting
-    it's
+    This class implements a Descriptor that manages the draw call callbacks of
+    Display. It's main purpose is to allow setting the default value of these
+    callbacks using a decorator-style syntax::
+
+        @gfx.Display.before_render
+        def animate():
+            ...
+
+        gfx.show()  # uses animate
+
+    Parameters
+    ----------
+    default : Callable
+        The default callback to use.
+
+    Notes
+    -----
+    Using the decorator feature above sets the class-level default for the
+    callback and affects all instances of Display that use the default value.
 
     """
 
@@ -57,42 +76,45 @@ class DecorSet:
 
 
 class Display:
-    """A Helper to Display Objects
+    """A Helper to display an object or scene
 
     This class provides the basic scaffolding needed to visualize a given
-    WorldObject in a new window. While it does add scaffolding, it aims to be
-    fully customizable so that you can replace each piece as needed.
+    WorldObject. To do so the class chooses a sensible default for each part of
+    the full setup (@almarklein: what is the technical term for the full setup?)
+    unless the value is explicitly set by the user or exists as a child of
+    ``object``. For example, it will create a camera unless you explicitly set a
+    value for camera or if there is (at least) one camera in the scene.
 
     Parameters
     ----------
-        object : gfx.WorldObject
-            The object to show.
-        up : gfx.Vector3
-            If set, set the default camera controller's up vector to this value.
-        canvas : WgpuCanvas
-            The canvas to use to display the object.
-        renderer : gfx.Renderer
-            The renderer to use while drawing the scene. If set, `canvas` is
-            ignored.
-        controller : gfx.Controller
-            The camera controller to use.
-        camera : gfx.Camera
-            The camera to use.
-        before_render : Callable
-            A callback that will be executed during each draw call before a new
-            render is made.
-        after_render : Callable
-            A callback that will be executed during each draw call after a new
-            render is made.
-        draw_function : Callable
-            Replaces the draw callback with a custom one. If set both
-            `before_render` and `after_render` will have no effect.
+    canvas : WgpuCanvas
+        The canvas used to display the object. If both ``renderer`` and
+        ``canvas`` are set, then the renderer needs to use the set canvas.
+    renderer : gfx.Renderer
+        The renderer to use while drawing the scene. If both ``renderer``
+        and ``canvas`` are set, then the renderer needs to use the set canvas.
+    controller : gfx.Controller
+        The camera controller to use. If not set, Display will use the first
+        controller defined on ``object``. If there is none, Display will
+        create one.
+    camera : gfx.Camera
+        The camera to use. If not set, Display will use the first camera
+        defined on ``object``. If there is none, Display will create one.
+    before_render : Callable
+        A callback that will be executed during each draw call before a new
+        render is made.
+    after_render : Callable
+        A callback that will be executed during each draw call after a new
+        render is made.
+    draw_function : Callable
+        Replaces the draw callback with a custom one. If set both
+        `before_render` and `after_render` will have no effect.
 
     """
 
-    before_render = DecorSet()
-    after_render = DecorSet()
-    draw_function = DecorSet()
+    before_render = RenderCallback()
+    after_render = RenderCallback()
+    draw_function = RenderCallback()
 
     def __init__(
         self,
@@ -142,37 +164,21 @@ class Display:
 
         Parameters
         ----------
-            object : gfx.WorldObject
-                The object to show.
-            up : gfx.Vector3
-                If set, set the default camera controller's up vector to this value.
-            canvas : WgpuCanvas
-                The canvas to use to display the object.
-            renderer : gfx.Renderer
-                The renderer to use while drawing the scene. If set, `canvas` is
-                ignored.
-            controller : gfx.Controller
-                The camera controller to use.
-            camera : gfx.Camera
-                The camera to use.
-            before_render : Callable
-                A callback that will be executed during each draw call before a new
-                render is made.
-            after_render : Callable
-                A callback that will be executed during each draw call after a new
-                render is made.
-            draw_function : Callable
-                Replaces the draw callback with a custom one. If set both
-                `before_render` and `after_render` will have no effect.
+        object : gfx.WorldObject
+            The object to show. If it is not a :class:`gfx.Scene <pygfx.Scene>`
+            then Display will wrap it into a new scene containing lights and a
+            background.
+        up : gfx.Vector3
+            If set, and ``object`` does not contain a controller, set the camera
+            controller's up vector to this value.
+
+        Notes
+        -----
+        If you want to display multiple objects prefer :class:`gfx.Group
+        <pygfx.Group>` over :class:`gfx.Scene <pygfx.Scene>` to avoid the pitfall of
+        forgetting to add lights to the scene.
 
         """
-
-        if (
-            self.canvas is not None
-            and self.renderer is not None
-            and self.canvas != self.renderer.target
-        ):
-            raise ValueError("Display's render target differs from it's canvas.")
 
         if isinstance(object, Scene):
             scene = object
@@ -193,22 +199,38 @@ class Display:
                 "Your scene does not contain any lights. Some objects may not be visible"
             )
 
-        if not self.camera:
+        existing_cameras = self.find_children(Camera)
+        if self.camera:
+            pass
+        elif len(existing_cameras) > 0:
+            self.camera = existing_cameras[0]
+        elif not self.camera:
             self.camera = PerspectiveCamera(70, 16 / 9)
             self.camera.add(DirectionalLight())
+            # do we really want to modify the scene and add
+            # a camera?
             self.scene.add(self.camera)
 
-        if self.renderer is not None:
-            self.canvas = self.renderer.target
-        elif self.canvas is None:
+        if self.renderer is None and self.canvas is None:
             from wgpu.gui.auto import WgpuCanvas
 
             self.canvas = WgpuCanvas()
             self.renderer = WgpuRenderer(self.canvas)
-        else:
+        elif self.renderer is not None:
+            self.canvas = self.renderer.target
+        elif self.canvas is not None:
             self.renderer = WgpuRenderer(self.canvas)
+        elif self.canvas != self.renderer.target:
+            raise ValueError("Display's render target differs from it's canvas.")
+        else:
+            pass
 
-        if self.controller is None:
+        controllers = self.find_children(Controller)
+        if self.controller is not None:
+            pass
+        elif len(controllers) > 0:
+            self.controller = controllers[0]
+        else:
             look_at = self.camera.show_object(object)
             self.controller = OrbitController(
                 self.camera.position.clone(), look_at, up=up
@@ -252,28 +274,35 @@ def show(
 
     Parameters
     ----------
-        object : gfx.WorldObject
-            The object to show.
-        up : gfx.Vector3
-            If set, set the default camera controller's up vector to this value.
-        canvas : WgpuCanvas
-            The canvas to use to display the object.
-        renderer : gfx.Renderer
-            The renderer to use while drawing the scene. If set, `canvas` is
-            ignored.
-        controller : gfx.Controller
-            The camera controller to use.
-        camera : gfx.Camera
-            The camera to use.
-        before_render : Callable
-            A callback that will be executed during each draw call before a new
-            render is made.
-        after_render : Callable
-            A callback that will be executed during each draw call after a new
-            render is made.
-        draw_function : Callable
-            Replaces the draw callback with a custom one. If set both
-            `before_render` and `after_render` will have no effect.
+    object : gfx.WorldObject
+            The object to show. If it is not a :class:`gfx.Scene <pygfx.Scene>`
+            then Display will wrap it into a new scene containing lights and a
+            background.
+    up : gfx.Vector3
+        If set, and ``object`` does not contain a controller, set the camera
+        controller's up vector to this value.
+    canvas : WgpuCanvas
+        The canvas used to display the object. If both ``renderer`` and
+        ``canvas`` are set, then the renderer needs to use the set canvas.
+    renderer : gfx.Renderer
+        The renderer to use while drawing the scene. If both ``renderer``
+        and ``canvas`` are set, then the renderer needs to use the set canvas.
+    controller : gfx.Controller
+        The camera controller to use. If not set, Display will use the first
+        controller defined on ``object``. If there is none, Display will
+        create one.
+    camera : gfx.Camera
+        The camera to use. If not set, Display will use the first camera
+        defined on ``object``. If there is none, Display will create one.
+    before_render : Callable
+        A callback that will be executed during each draw call before a new
+        render is made.
+    after_render : Callable
+        A callback that will be executed during each draw call after a new
+        render is made.
+    draw_function : Callable
+        Replaces the draw callback with a custom one. If set both
+        `before_render` and `after_render` will have no effect.
 
     """
 
--- a/pygfx/utils/show.py
+++ b/pygfx/utils/show.py
@@ -60,7 +60,7 @@ def show(
         draw_function : Callable
             Replaces the draw callback with a custom one. If set both
             `before_render` and `after_render` will have no effect.
-        
+
     """
 
     if isinstance(object, Scene):
@@ -81,13 +81,13 @@ def show(
     if renderer is not None:
         canvas = renderer.target
     elif canvas is None:
-        from wgpu.gui.auto import WgpuCanvas, run
+        from wgpu.gui.auto import WgpuCanvas
 
         canvas = WgpuCanvas()
         renderer = WgpuRenderer(canvas)
     else:
         renderer = WgpuRenderer(canvas)
-    
+
     if controller is None:
         look_at = camera.show_object(object)
         controller = OrbitController(camera.position.clone(), look_at, up=up)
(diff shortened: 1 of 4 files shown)


pygfx/utils/show.py
```diff
@@ -34,33 +44,47 @@ def show(object: WorldObject, up=None):
...
+        event_loop = run
+    elif event_loop is None:
+        # TODO: find the matching run function
+        raise ValueError(
+            "When providing a canvas you also need to provide the event loop."
+        )
 
-    scene.traverse(check_light, False)
-    if not light_count:
-        camera.add(DirectionalLight(0.8))
```
Korijn, line 52: I think this specific line has gone missing?

Korijn:
The improvements here are very nice and I think for now this will do.

I propose to keep track of your idea for a decorator and dataclass style implementation in a new ticket so we can follow up in the future.

Korijn (approved):

Korijn:
Hmm, looks like test_examples.py doesn't support gfx.show yet. I was a little concerned that would be the case... Should be doable though. I'll check it out tomorrow.

FirefoxMetzger:
I will also have to write a decent docstring for the new function and then there are unit tests ... 😅

pygfx/utils/show.py
```diff
@@ -34,33 +44,47 @@ def show(object: WorldObject, up=None):
...
+        from wgpu.gui.auto import WgpuCanvas, run
 
-    def check_light(ob):
-        nonlocal light_count
-        if isinstance(ob, Light):
-            light_count += 1
+        canvas = WgpuCanvas()
+        event_loop = run
+    elif event_loop is None:
+        # TODO: find the matching run function
```
almarklein, line 60: This should do the trick:
```py
sys.modules[renderer.target.__module__].run
```

almarklein: Do you think this would mean we can drop the `event_loop` argument (since it comes with the canvas)? If not I propose a different name for `event_loop`, because it sounds like an object, which may be confusing.

pygfx/utils/show.py
```diff
@@ -34,33 +44,47 @@ def show(object: WorldObject, up=None):
...
+    if controller is None:
+        look_at = camera.show_object(object)
+        controller = OrbitController(camera.position.clone(), look_at, up=up)
+        controller.add_default_event_handlers(renderer, camera)
+
+    if draw_function is None:
+
+        def animate():
+            if before_render is not None:
+                before_render()
```
almarklein, line 77: If we'd provided `renderer` as an argument to `before_render` and `after_render` we could e.g. request new draws in the user code.

pygfx/utils/show.py
```diff
@@ -34,33 +44,47 @@ def show(object: WorldObject, up=None):
 
         background = Background(None, BackgroundMaterial((0, 1, 0, 1), (0, 1, 1, 1)))
         scene.add(background)
+        scene.add(AmbientLight(0.2))
```
almarklein, line 47: I prefer the old version of determining whether to add lights. Imagine:
```py
cube = gfx.Mesh(...)
gfx.show(cube)
```
Works fine, now I modify my example to show two cubes:
```py
scene  = gfx.Scene()
scene.add(gfx.mesh(...))
scene.add(gfx.mesh(...))
gfx.show(scene)
```
and now it's all dark :)

pygfx/utils/show.py
```diff
@@ -34,33 +44,47 @@ def show(object: WorldObject, up=None):
...
+        raise ValueError(
+            "When providing a canvas you also need to provide the event loop."
+        )
 
-    scene.traverse(check_light, False)
-    if not light_count:
-        camera.add(DirectionalLight(0.8))
-        scene.add(AmbientLight(0.2))
+    if renderer is None:
+        renderer = WgpuRenderer(canvas)
```
almarklein, line 66: A custom canvas should be disallowed if a custom renderer is given. We don't really need the canvas here, we can do the `request_draw()` call below on the renderer instead.

pygfx/utils/show.py
```diff
@@ -34,33 +44,47 @@ def show(object: WorldObject, up=None):
...
+            renderer.request_draw()
 
-    def animate():
-        controller.update_camera(camera)
-        renderer.render(scene, camera)
+        draw_function = animate
 
-    canvas.request_draw(animate)
-    run()
+    canvas.request_draw(draw_function)
```
almarklein, line 89:
```suggestion
    renderer.request_draw(draw_function)
```

almarklein (changes requested):
This is becoming an all-round powertool!

pygfx/utils/show.py
Korijn: You could use a Group, but I agree that it's maybe surprising behavior

pygfx/utils/show.py
FirefoxMetzger: If we can find the `run` function for a given canvas then that would make it easier indeed. The question is if we want to allow people to experiment with their own event loop for a canvas, or if this is out of scope for now.

If it is out of scope, we can drop `event_loop` for now.

pygfx/utils/show.py
FirefoxMetzger: Would that be an advantage? The current default loop already requests a new draw at the end in any case, so I am inclined to say this would be the realm of using a custom `event_loop`, which could totally have a reference to the canvas and the renderer.

pygfx/utils/show.py
Korijn: Agree with dropping `event_loop` for now

pygfx/utils/show.py
FirefoxMetzger: The opposite holds true as well though "The scene works in `gfx.show`. Why don't I see anything in my actual setup!?".

It's meaningful to add lights if we only get an object and construct a default scene (users will likely place that object into their own scene later). However, when we get a full scene I think we should be mindful before modifying it; especially when we think of `gfx.show` as a debugging tool.

pygfx/utils/show.py
FirefoxMetzger: Sure.

pygfx/utils/show.py
FirefoxMetzger: I would just get the `canvas = renderer.target`. This makes the code overall cleaner :)

pygfx/utils/show.py
Korijn: How about just adding a kwarg `add_lights=True`

pygfx/utils/show.py
Korijn: I would recommend to see if the need comes up while you are rewriting more examples later on.

pygfx/utils/show.py
almarklein: > The opposite holds true as well though "The scene works in gfx.show. Why don't I see anything in my actual setup!?".

Good point. There is a chance of confusion either way.

What bothers me about `if isinstance(object, Scene)` is that it means the behavior of this function depends on whether a single object is passed, compared to multiple objects grouped in a scene. E.g. adding lights and background.

After some thought, I like @Korijn's suggestion about using `Group`: we can keep the behavior, but add a note in the docstring about the behavior when a `Scene` is given, and that a `gfx.Group()` should probably be used for multiple objects. And also use that in examples.

To take this idea further, if we create the camera, it should only be added to the scene (with a light attached) when we also create the scene here.

pygfx/utils/show.py
```diff
@@ -11,56 +14,104 @@
...
+    if not camera:
+        camera = PerspectiveCamera(70, 16 / 9)
+        camera.add(DirectionalLight(0.8))
+        scene.add(camera)
 
-    camera = PerspectiveCamera(70, 16 / 9)
-    look_at = camera.show_object(object)
-    scene.add(camera)
+    if renderer is not None:
+        canvas = renderer.target
```
almarklein, line 86:
```suggestion
        if not (canvas is None or canvas is renderer.target):
            raise ValueError("The given canvas does not match the given renderer.")
        canvas = renderer.target
```

pygfx/utils/show.py
```diff
@@ -11,56 +14,104 @@
...
 
-    def animate():
-        controller.update_camera(camera)
-        renderer.render(scene, camera)
+        draw_function = animate
 
-    canvas.request_draw(animate)
-    run()
+    canvas.request_draw(draw_function)
+    sys.modules[renderer.target.__module__].run()
```
almarklein, line 117: Use Canvas like in the line above.
```suggestion
    sys.modules[canvas.__module__].run()
```

pygfx/utils/show.py
```diff
@@ -11,56 +14,104 @@
...
+        light_gray = np.array((100, 100, 100, 255)) / 255
+
+        background = Background(None, BackgroundMaterial(light_gray, dark_gray))
         scene.add(background)
+        scene.add(AmbientLight(0.2))
+
+    if not camera:
+        camera = PerspectiveCamera(70, 16 / 9)
+        camera.add(DirectionalLight(0.8))
+        scene.add(camera)
```
almarklein, line 83:
```suggestion
        if scene is not object:
            camera.add(DirectionalLight(0.8))
            scene.add(camera)
```

pygfx/utils/show.py
almarklein: I added a suggestion for my last point. Could you update the docs to explain `Scene` vs `Group`?

Korijn: I really do think an `add_lights=True` kwarg would be better, honestly. It's explicit, and obvious.

almarklein: One reason I was leaning towards using the type of the given object to decide whether to add stuff, was that we also background and a camera to it. Would it make sense to generalize  `add_lights=True` to something like `modify_scene=True`?

Korijn: How about: `add=('lights', 'camera', 'background', 'controller')`

I wish I could put `action` in there :)

almarklein: mmm, what about `add_lights` and `add_background` args? (The camera being added is by virtue of adding the directional light.)

Korijn: They're all equally acceptable to me, I just think it's better to lean towards controlling the behavior of `gfx.show()` via explicit keyword arguments than letting it be derived from the scene graph in a rather opinionated manner.

FirefoxMetzger: Hm. At first, I was going to agree with `add_lights`, but when thinking about it a bit more I am not sure anymore if it solves the problem. 

Assuming `add_lights` has a default value like `add_lights=True` then chances are most people will use `gfx.show(scene)` and not know/care about most additional kwargs unless explicitly prompted to check. Especially if there are many. (In my experience, many devs only check the docs if something is broken and then only if SO doesn't have good suggestions.) So even though we can now turn this behavior off, it is still somewhat unintuitive and we are not exactly advertising it.
As such, the switch itself doesn't really solve the "Why does it work in `gfx.show`, but not in my application" problem.

We could, ofc, choose `add_lights=False` as a default but that would defeat the point of the kwarg. It's equal work to do `scene.add(gfx.SomeLight())` vs explicitly passing `add_lights=True`, so there is nothing gained. 

Also, in the case of us wrapping the input in a scene, we will likely have to do the lights and ignore this kwarg. This would make our kwargs quite complicated, because we already have `canvas`, `before_render`, and `after_render` that may be ignored in certain circumstances. That's quite a complex API for something that is meant to be a "quick scaffolding tool to debug your application or develop pygfx".

FirefoxMetzger: Thinking of `gfx.show` as a diagnostics/debugging tool, I like the following:

If we don't get a `Scene`, we create one and add a light since this scene typically contains the ambient light. If we do get a `Scene`, we check if there are light sources and if not, we show a warning along the lines of `"Your scene does not contain a light source and will appear black."`. 

This way, there are no surprises. We can visualize single objects as expected, we don't modify existing scenes, and we give useful feedback on "ad-hoc" scenes like the one from @almarklein 's description. Would this work for you @almarklein ?

pygfx/utils/show.py
FirefoxMetzger: How would this work for situations where we pass a scene to the renderer but don't pass a camera?

pygfx/utils/show.py
almarklein: Yeah, I think this could work 👍 . I'd say something like "... and some objects may not be visible".

pygfx/utils/show.py
almarklein: Then we create a camera, but don't add it to the scene. The camera does not have to be in the scene for it to work. The reason it is added to the scene in these lines, is to make the light source move along with it (as if you had a head-light).

Korijn:
So it seems to support gallery generation, we need to expose some kind of reference from each example module to the renderer and/or canvas.

That implies we need something to hold on to... like a `gfx.Display` class. :) I did not have this insight earlier, it's costing you time and rework, I'm sorry for that.

FirefoxMetzger:
> So it seems to support gallery generation, we need to expose some kind of reference from each example module to the renderer and/or canvas.

The exact requirement is that we need to decide if (and what) to draw based on the input that sphinx-gallery gives the scraper's callback. Those are:

- `block`, a 3-tuple with `(block_type, block_source, starting_lineno)` where `block_type="code"` and `block_source` is the raw source code of that block (for us the entire example)
- `block_vars` a dict where the relevant fields are `src_file` (path to file), `example_globals` (global variables), `fake_main` (variables defined in the main namespace). The variables are collected by running the example up to and including the current block.

I can make the gallery work as long as there is a way for me to inspect the above (or the global state of pygfx and wgpu) and check (1) if I should create an image (or GIF) and (2) how to generate the image.

Currently, this is done by checking for a global `gfx.Camera` and `gfx.Scene` and if both are defined I create a canvas and a renderer and take a snapshot. This is very hacky though, so I am hoping to find a better solution, especially in light of `gfx.show` hiding `gfx.Camera` in most examples.

> That implies we need something to hold on to... like a gfx.Display class. :) I did not have this insight earlier, it's costing you time and rework, I'm sorry for that.

@Korijn I didn't either. We started this discussion before I started writing a custom scraper for sphinx-gallery, so we could not have known of this requirement at that stage. What's I think more interesting/important is how we best proceed from here.

I think there are three ways forward:

- I change this PR, introduce `Display`, and use it to build the gallery
- We introduce something like [`plt.get_fignums()`](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.get_fignums.html) in pygfx to track canvases. (The docs are not very extensive. Essentially, it tracks currently open figures so that you can find and access them)
- We find a feature in `wgpu` that can check for currently existing canvases

>

Korijn:
I'm not a fan of tracking things globally.

> I change this PR, introduce Display, and use it to build the gallery

This seems most sensible.

almarklein:
Once we have `Display` as an alternative to `gfx.show()`, we can categorize all examples in three groups:
* Examples that use `Display`. In the gallery-builder we can look for the object and use it.
* Examples that do not use display, but do have a `renderer`. In the gallery-builder we can use that.
* Examples that do more custom stuff - we do not include these in the gallery.

FirefoxMetzger:
Alright. Let me get this branch up to data and then we see how it looks 🚀

FirefoxMetzger:
That turned out to take longer than I expected. I got a little sidetracked trying to find a good way to implement the getter idea from https://github.com/pygfx/pygfx/issues/383. With the current changes, the cube example simplifies to:

```python
import pygfx as gfx

cube = gfx.Mesh(
)


@gfx.Display.before_render
def animate():
    rot = gfx.linalg.Quaternion().set_from_euler(gfx.linalg.Euler(0.005, 0.01))
    cube.rotation.multiply(rot)


if __name__ == "__main__":
    gfx.show(cube)

```

(Although we probably want to use `disp = gfx.Display(); disp.show()` in the examples to allow the gallery to pick up the Display object and render it.)

I've done the following changes to the PR:

- I introduced `Display()` which now holds all the data for the setup and made `gfx.show` a thin wrapper around it.
- I introduced a Descriptor class that allows setting the default callback for Display using decorators (more below)
- I've overhauled all the docstrings in the class
- I went through all the open discussions and implemented the suggestions.

What I did not do is write additional tests, since the code will be tested by every example once I refactored them.

---

A note on the decorator-style callback: If you look closely, you will see that the decorator is called on the class-level (`@gfx.Display.before_render` vs `disp = gfx.Display() ... @gfx.disp.before_render`). This is because I couldn't figure out how to make `@gfx.disp.before_render` work directly. This decorator would call `gfx.disp.before_render.__get__` which will yield the currently set callback instead of the decorator function and I don't think we can work around this.

In leu of the preferred option, we can have the following three alternatives:

```python

#  Decorator on the class (<- I implemented this)
@gfx.Display.before_render
def animate():
    ...

#  Decorator on a different method
@disp.update.before_render
def animate():
    ...

#  Return a wrapper with a custom method
@disp.before_render.update
def animate():
    ...
```

The first is complementary to the other two (we can implement both in parallel), but I didn't do that, because the latter two look a little cryptic, which may not be super maintenance friendly. That said, if you like the API of those options I can implement these.

The downside of the class-level approach is that we don't have access to instance-level variables, so the best we can do is modify the default value of the callback. This is what the current implementation does and as such, there is a caveat in cases where we prepare multiple displays at the same time as they will all share this default value and multiple calls to `@gfx.Display.before_render` will overwrite each other. This should be rare though, since in most cases we will only have one display at the time, i.e., create a scene, display, close, next scene.

pygfx/utils/show.py
FirefoxMetzger: This is now implemented :)

pygfx/utils/show.py
FirefoxMetzger: I wasn't quite sure how to address this comment, especially after the reactor. Since I'm not sure what we were after here, I can't really say if this is resolved now or still something to improve. @almarklein could you have a look and let me know if we can resolve this?

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+        self.scene = None
+
+        if before_render is not None:
+            self.before_render = before_render
+
+        if after_render is not None:
+            self.after_render = after_render
+
+    def default_draw(self):
+        if self.before_render is not None:
```
Korijn, line 143: How can this ever be `None`? Doesn't it resolve to the class level RenderCallback always?

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+        # allow decorator-style setting of default
+        # callback
+        if instance is None:
+            return self
+
+        value = getattr(instance, self.private_name, None)
+        if value is not None:
+            return value
+
+        # persist callback to prevent it from changing
```
Korijn, line 67: Can you explain how this prevents it? I'm not super clear on the control flow here

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+
+            dark_gray = np.array((169, 167, 168, 255)) / 255
+            light_gray = np.array((100, 100, 100, 255)) / 255
+
+            background = Background(None, BackgroundMaterial(light_gray, dark_gray))
+            scene.add(background)
+            scene.add(AmbientLight(), DirectionalLight())
+        self.scene = scene
+
+        if len(self.find_children(Light)) == 0:
```
Korijn, line 197: Could you use `any()` here? That would enable the more efficient generator implementation you hinted at

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+        existing_cameras = self.find_children(Camera)
+        if self.camera:
+            pass
+        elif len(existing_cameras) > 0:
+            self.camera = existing_cameras[0]
+        elif not self.camera:
+            self.camera = PerspectiveCamera(70, 16 / 9)
+            self.camera.add(DirectionalLight())
+            # do we really want to modify the scene and add
+            # a camera?
```
Korijn, line 211: If you don't do this, the directional light is not in the scene either :)

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+        elif self.renderer is not None:
+            self.canvas = self.renderer.target
+        elif self.canvas is not None:
+            self.renderer = WgpuRenderer(self.canvas)
+        elif self.canvas != self.renderer.target:
+            raise ValueError("Display's render target differs from it's canvas.")
+        else:
+            pass
+
+        controllers = self.find_children(Controller)
```
Korijn, line 228: I don't think Controllers are a WorldObject subclass, so you will never find them in the scene

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+            self.controller = controllers[0]
+        else:
+            look_at = self.camera.show_object(object)
+            self.controller = OrbitController(
+                self.camera.position.clone(), look_at, up=up
+            )
+            self.controller.add_default_event_handlers(self.renderer, self.camera)
+
+        self.canvas.request_draw(self.draw_function)
+        sys.modules[self.canvas.__module__].run()
```
Korijn, line 241: I think we should open a ticket on wgpu-py to provide a simple reference, something like `WgpuCanvas.run`

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+
+    # this should probably live on WorldObject
```
Korijn, line 243: Agreed

pygfx/utils/show.py
```diff
@@ -16,9 +17,243 @@
...
+    def find_children(self, clazz):
+        """Return all children of the given type"""
+        objects = list()
+
+        # can we make traverse a generator?
+        # [x for x in self.scene.traverse(filter=lambda x: isinstance(x, clazz))
```
Korijn, line 249: `traverse` is more like `map`, it's purpose is to apply a function to all objects in a (sub)graph. I like your suggestion but it has a different purpose. You can just add a new function to `WorldObject`, I like the name `find_children`, could also be `query`, I don't mind.

Korijn:
Really nice work! 👍 

`@disp.before_render` if this can't be done, that too bad. I wouldn't worry about having a decorator feature

pygfx/utils/show.py
FirefoxMetzger: Right. Under the hood, the descriptor uses a `_<name>` attribute on the instance to manage the actual callback, i.e., if we do

```python
disp = Display()

disp.before_render = animate
```

then the descriptor creates `disp._before_render` to store the callback on `disp`. If that value doesn't exist or is `None` we will fall back to `gfx.Display.before_render.default_callback`. This is what the `if value is None: ...` does.

This works for most cases, except when we have an already running `disp` (e.g. inside an earlier notebook cell) that uses the default callback. If, in this case, we use the `@gfx.Display.before_render` decorator while setting up the next display object we may modify the running one. Since the decorator sets the class-level default (which the previous `disp` is using) we risk the rendering loop of the running renderer changing. To prevent this interaction I take the value `Display.before_render.default_callback` during the first run and persist it on the instance by setting `disp._before_render = Display.before_render.default_callback`. Now future changes of the class-level default can no longer impact the instance-level `disp` once the display starts running/showing.

Writing this out makes two enhancements come to mind: (1) We probably want to not use `default_callback` if the user has explicitly set `disp.before_render = None`. (2) We could add a `Display.finalize()` function that does all the setup that can be done without making a canvas pop up. I'm thinking in particular of setting all the callbacks and making them independent of the class-level defaults. The second one is not needed right now, but is quite easy to incorporate.

pygfx/utils/show.py
FirefoxMetzger: hm, as long as `bool(gfx.Light(...)) == True` this should be possible. Its a bit weird to think of the boolean-ness of a WorldObject, but I like how `if any(self.find_children(Light))` looks syntax-wise 🤩 I will change it.

pygfx/utils/show.py
FirefoxMetzger: Yes. my question is more from the angle of "do we want to modify the input scene?". It's similar to our discussion about adding lights by default: How worried are we about the side effects of permanently modifying the user's scene?

pygfx/utils/show.py
FirefoxMetzger: Oh, that's true. In that case I will drop this logic. @Korijn Do you know if there is another way to discover existing controllers, or will it be sufficient to only use one if passed to the Display?

pygfx/utils/show.py
Korijn: Yes, well, I think we had that discussion earlier, I think all proposals are still on the table. Let's just go with this approach for now and adjust it later if we think of something better.

pygfx/utils/show.py
Korijn: I don't think there is another way, no!

pygfx/utils/show.py
FirefoxMetzger: See: https://github.com/pygfx/wgpu-py/issues/307

pygfx/utils/show.py
FirefoxMetzger: @Korijn Should this happen in a separate PR where we move the function over, or do you want this to be part of this PR?

pygfx/utils/show.py
FirefoxMetzger: Okay. In that case, we can just add a `find_children` function. 

The prototype I had in mind was something like this:

```python
def traverse(self, filter:Callable[[WorldObject],bool], skip_invisible=False):
    if skip_invisible and not self.visible:
        return

    if not filter(self):
        return

    yield self

    for child in self._children:
        yield from child.traverse(callback, skip_invisible)
```

With the above you can also do modifications like:

```python
# set all light intensities to .8
for light in scene.traverse(filter=lambda x: isinstance(x, gfx.Light)):
    light.intensity = .8
```

So maybe the name `iter` or `iter_children` would make more sense?

FirefoxMetzger:
> @disp.before_render if this can't be done, that's too bad. I wouldn't worry about having a decorator feature too much, it's not really critical to making this work.

How do you think of `@gfx.Display.before_render`? If you think it's too much/weird I can also kick it out. In that case we can remove the entire descriptor class (`RenderCallback`) and just use standard assignments.

pygfx/utils/show.py
FirefoxMetzger: Then we go with explicitly passing one until a better option develops :)

pygfx/utils/show.py
Korijn: 👍 thanks for explaining

pygfx/utils/show.py
Korijn: Let's create a ticket and follow up later

pygfx/utils/show.py
Korijn: Nice, I like it! I would say just `iter` is a very nice fit.

pygfx/utils/show.py
FirefoxMetzger: No problem. Should I add a clarifying comment in the source for future reference somewhere? (I also did a refactor of this section just now, so it might be a little more intuitive now)

Korijn:
I am personally leaning towards kicking it out if the examples still look nice without them :)

pygfx/utils/show.py
FirefoxMetzger: See https://github.com/pygfx/pygfx/issues/385

FirefoxMetzger:
@Korijn 

With `@gfx.Display.before_render`:

```python
import pygfx as gfx

cube = gfx.Mesh(
)


@gfx.Display.before_render
def animate():
    rot = gfx.linalg.Quaternion().set_from_euler(gfx.linalg.Euler(0.005, 0.01))
    cube.rotation.multiply(rot)


if __name__ == "__main__":
    disp = Display()
    disp.show(cube)
```

Without `@gfx.Display.before_render`:

```python
import pygfx as gfx

cube = gfx.Mesh(
)


def animate():
    rot = gfx.linalg.Quaternion().set_from_euler(gfx.linalg.Euler(0.005, 0.01))
    cube.rotation.multiply(rot)


if __name__ == "__main__":
    disp = Display()
    disp.before_render = animate
    disp.show(cube)
```

Korijn:
If it simplifies the implementation a lot, I would say lose the decorator support for now.

FirefoxMetzger:
And the decorator is gone :) Now it's only @almarklein 's comment and my question to @almarklein in the docstring of `Display`.

@Korijn Thanks for reviewing this on a weekend, btw 🚀 At the same time, please don't feel pressured by me working on the weekend. My PhD has ruined my working habits, but that doesn't mean that there are any expectations towards getting things reviewed over the weekend.

pygfx/utils/show.py
```diff
@@ -11,56 +15,229 @@
...
-    controller.add_default_event_handlers(renderer, camera)
+        existing_cameras = self.find_children(Camera)
+        if self.camera:
+            pass
+        elif any(existing_cameras):
+            self.camera = existing_cameras[0]
+        elif not self.camera:
+            self.camera = PerspectiveCamera(70, 16 / 9)
+            self.camera.add(DirectionalLight())
+            self.scene.add(self.camera)
```
almarklein, line 149: I saw that the two of you already had a discussion about this, but I think the docs are pretty clear that if a `Scene` is provided, that we don't add lights, so let's be consistent. @Korijn ?
```suggestion
            if object is not scene:
                self.camera.add(DirectionalLight())
                self.scene.add(self.camera)
```

pygfx/utils/show.py
```diff
@@ -11,56 +15,229 @@
...
+    *,
+    canvas=None,
+    renderer=None,
+    controller=None,
+    camera=None,
+    before_render=None,
+    after_render=None,
+    draw_function=None,
+):
+    """Display a WorldObject
```
almarklein, line 201: This docstrings needs the "upstream" changes too :)

pygfx/utils/show.py
```diff
@@ -11,56 +15,229 @@
...
+        # can we make traverse a generator?
+        # [x for x in self.scene.traverse(filter=lambda x: isinstance(x, clazz))
+        self.scene.traverse(
+            lambda x: objects.append(x) if isinstance(x, clazz) else None
+        )
+
+        return objects
+
+
+def show(
```
almarklein, line 189: I wonder if whether we'd ever use this, since we've found that we need that `display` object to be able to pick it up in the gallery and tests ...

almarklein (commented):
This looks nice!

Not a fan of how that decorator looked, plus decorators may be intimidating for Python novices, so I'm glad you removed it.

My comments are mostly nitpicks in docstrings. Except for the one where the camera is added (or not added) to the scene.

pygfx/utils/show.py
almarklein: Still nice to have I guess, and does not look like a maintenance burden, so let's just keep it.

pygfx/utils/show.py
FirefoxMetzger: I will resolve https://github.com/pygfx/pygfx/pull/380#discussion_r1019023447 in favor of this discussion, since it is the same suggestion :)

Korijn: Yes, this is a good suggestion.

I would recommend to evaluate the condition in a single place though. For example you could put `add_lights = not isinstance(object, Scene)` near the top and re-use that condition in both places where lights are added.

FirefoxMetzger: Refactored.

FirefoxMetzger:
@Korijn @almarklein This resolves all the currently open conversations. I also have nothing else to add, so if the tests are happy this one is RTR from my side :)

almarklein (approved):

Korijn (approved):

Korijn:
Nice work! You can merge as far as I am concerned. For some reason CI was not triggered though. It may require another git push. :/

Issue #345, opened 2022-10-14
Generated example gallery in documentation


Korijn:
* we can probably run the following snippet in the pre_install hook for the RTD builds to have lavapipe available:

https://github.com/pygfx/pygfx/blob/b9f9e1ca43375be21097293320ac520cd6ea36a7/.github/workflows/ci.yml#L96-L99

Relates to #145

almarklein:
Discussed with @FirefoxMetzger some ideas on how to handle the images:
* Could render in GH actions, and put *somewhere*, like a bucker ot something, and load these in RTD.
  * Downside is managing the bucket and its access rights, reliance on another service, etc.
* Could render the screenshots on RTD, but sooner or later we'll likely hit the half hour (?) built time limit. 
  * Would the limit go away if we'd take a business account?
* We could render in GH actions and put the screenshots as CI artifacts, which we can then load in RTD. 
  * Is it even possible for an outside service to download artifacts?
  * Race condition: when pusing a new release, RTD starts downloading before our CI is done generating the images.
* Could we render in GH actions and then *push* the artifact to RTD?
* Might be simplest to use GH pages instead.
  * But would miss out on different doc versions, analytics, and other RTD niceties.

Korijn:
A couple of questions and suggestions:

* Would generating the gallery really take that much time? Unlike testing examples, we don't need strict process isolation between examples. That means you can retain the device instance and don't incur the initialization overhead again and again.
* Will rendering on RTD work? Depends on drivers, framebuffers, etc.
* You can publish build pipeline artifacts on GH but there is a max storage limit (sum of all builds artifacts) and a bandwidth limit and it is not very high.
* I would suggest sticking to RTD at first, and see how much we can squeeze out of the half hour time limit. How long does a build take right now? Can we shave off time? We can make installing dependencies (be very specific about which packages and versions to install and use wheels only) and cloning the repo very fast (shallow), and that is all relatively easy.

almarklein:
[Apparently](https://docs.readthedocs.io/en/stable/builds.html) the RTD build limit is 15 minutes by default and 30 minutes for a business account.

> we don't need strict process isolation between examples. That means you can retain the device instance and don't incur the initialization overhead again and again. 

True, but we also have leaks in our caching at the moment, so it will likely blow up if you run all examples after each other on the same device 😄 

A plan might be:
* Test that we can actually render on RTD (via lavapipe, installation commands in top post).
[...]

Issue #145, opened 2021-09-30
Organize examples


almarklein:
As our examples grow they becomes a bit harder to navigate. Might be good to think about organizing them better. Maybe a (max 1 layer deep) dir structure, or just more sensible names.

Related, I really like how the Threejs examples can be searched by keyword. Maybe something for our docs? (Technically a separate issue.)

almarklein:
I can think of three categories:

* "hello" examples that are minimal, and demonstrate a particular feature.
* "demo" more complete examples that are usually pretty.
* "validate" examples that produce a stable output that is tested on CI.

My suggestion would be to prefix all examples with either of the three terms. Some names can also be adjusted to include better keywords. Further some current examples should probably be either simplified for a hello-example, or made more pretty/complete for a demo example.

Issue #381, opened 2022-11-10
Creating video with pygfx


FirefoxMetzger:
I'm working on the sphinx gallery for the examples and I got the basic pieces working. At the moment we can generate a gallery and have sphinx-gallery use pygfx to render images and add them. Something like this:

![image](https://user-images.githubusercontent.com/4402489/201003035-3d9a7f76-6f12-4933-a5f0-d9982f794b43.png)

There are a few more things to iron out (like giving each example a title so that sphinx can link it automatically), but the basics are there.

One thing that is, perhaps, a bit of a luxury, but is still really useful for us is to show videos/animations in the gallery. This essentially works by creating a GIF that gets shown instead of the usual PNG. For this I have two discussion points:

- Is there an existing example of capturing a video with pygfx?
- The code for the video may have to be partly embedded in the example (to say "take a picture here, but do a video for X seconds there"). From a pedagogical perspective, this is suboptimal because we would ideally like to just show the example with the usual `run()` call so that they can see it nicely once they copy the code. Would an alternative be to say "we always show 5 sec of video for each render"?

I'm not 100% clear yet on how I want to solve this problem, but if you have ideas, I'm all ears :)

Korijn:
> * Is there an existing example of capturing a video with pygfx?

I don't believe so but I guess you could start with an offscreen rendering example and then manually perform the render loop for the desired number of frames. Essentially I imagine just a for loop where you call `animate()`, then `canvas.draw()` to get a frame as an ndarray, and at the end of the loop you write the ndarrays to a GIF or video with something like imageio.

I hope @almarklein has a clearer vision to share with you.

> * The code for the video may have to be partly embedded in the example (to say "take a picture here, but do a video for X seconds there"). From a pedagogical perspective, this is suboptimal because we would ideally like to just show the example with the usual `run()` call so that they can see it nicely once they copy the code. Would an alternative be to say "we always show 5 sec of video for each render"?

One suggestion: You could keep the main examples simple, ending with `gfx.show()` in the `if __name__ == "__main__":` statement, and have a separate folder with video examples that build on (by importing) the main examples by tacking on an alternative to the `gfx.show()` call. You can keep the code for generating the videos separate from the gallery that way, but still generate them.

FirefoxMetzger:
> I don't believe so but I guess you could start with an offscreen rendering example and then manually perform the render loop for the desired number of frames.

[...]

PR #384, opened 2022-11-12
Add sphinx gallery

FirefoxMetzger:
Closes: https://github.com/pygfx/pygfx/issues/345 #145

This PR contributes a sphinx gallery that converts examples into a nice RST file including output renders. It also reorganizes the example folder into `introductory`, `feature_demo`, and `validation` examples to track them better and to display examples by category in the gallery.

Note that this is a _large_ PR. Most of the changes are fairly light-weight, though, so a lot of the touching lines are things like adding a title in the docstring or renaming an example. The critical pieces are in `conf.py` (the sphinx config) and in `utils/gallery_scraper.py`.


FirefoxMetzger:
One current issue is that some examples (e.g. the PBR render of the helmet) use models from `examples/models`. Currently these are loaded by getting a path to the model via `__file__`; however, `__file__` is not defined when the example is run by sphinx, so those examples break. I'm not sure how we want to resolve this though. My intuitive approach would have been to think of a `gfx.sample_models.[helmet/teapot]` approach, but this means that we have to include this data in the package, which I am not a fan of. Then again, it might be useful because users that copy down the example from the docs will face the same pathing issues on their machines since a path based on `__file__` (pointing to the target of the copy/paste) will break too.

FirefoxMetzger:
Here is also a benchmark on how long the examples take to build on my machine. It doesn't look too bad:

<details>

```
computation time summary:
    - ..\examples\feature_demo\plot_collection_line.py:                    20.45 sec   0.0 MB
    - ..\examples\feature_demo\plot_pbr2.py:                               16.97 sec   0.0 MB
    - ..\examples\feature_demo\plot_multi_select.py:                        3.65 sec   0.0 MB
    - ..\examples\introductory\plot_hello_triangle.py:                      3.15 sec   0.0 MB
    - ..\examples\validation\plot_validate_color.py:                        2.03 sec   0.0 MB
    - ..\examples\validation\plot_validate_light_shadow.py:                 1.95 sec   0.0 MB
    - ..\examples\feature_demo\plot_scene_subplots_video.py:                1.92 sec   0.0 MB
    - ..\examples\validation\plot_validate_mesh_colormap.py:                1.87 sec   0.0 MB
    - ..\examples\feature_demo\plot_multi_slice2.py:                        1.83 sec   0.0 MB
    - ..\examples\feature_demo\plot_light_spotlight_shadows.py:             1.82 sec   0.0 MB
    - ..\examples\validation\plot_validate_volume.py:                       1.76 sec   0.0 MB
    - ..\examples\validation\plot_validate_image1.py:                       1.75 sec   0.0 MB
    - ..\examples\validation\plot_validate_depth_clipping.py:               1.72 sec   0.0 MB
    - ..\examples\feature_demo\plot_multi_slice1.py:                        1.70 sec   0.0 MB
    - ..\examples\validation\plot_validate_helpers2.py:                     1.70 sec   0.0 MB
    - ..\examples\feature_demo\plot_light_spotlights.py:                    1.65 sec   0.0 MB
    - ..\examples\validation\plot_validate_helpers1.py:                     1.63 sec   0.0 MB
    - ..\examples\validation\plot_validate_culling.py:                      1.63 sec   0.0 MB
    - ..\examples\feature_demo\plot_colormap_mesh.py:                       1.61 sec   0.0 MB
    - ..\examples\feature_demo\plot_wireframe_material.py:                  1.59 sec   0.0 MB
    - ..\examples\feature_demo\plot_mesh_depth_material.py:                 1.58 sec   0.0 MB
    - ..\examples\feature_demo\plot_cylinder.py:                            1.58 sec   0.0 MB
    - ..\examples\validation\plot_validate_image2.py:                       1.55 sec   0.0 MB
    - ..\examples\validation\plot_validate_box.py:                          1.53 sec   0.0 MB
    - ..\examples\validation\plot_validate_image_colormap.py:               1.52 sec   0.0 MB
    - ..\examples\feature_demo\plot_image_click_events.py:                  1.52 sec   0.0 MB
    - ..\examples\validation\plot_validate_ndc.py:                          1.51 sec   0.0 MB
    - ..\examples\feature_demo\plot_skybox.py:                              1.49 sec   0.0 MB
    - ..\examples\feature_demo\plot_light_shadow.py:                        1.48 sec   0.0 MB
    - ..\examples\feature_demo\plot_scene_in_a_scene.py:                    1.48 sec   0.0 MB
    - ..\examples\feature_demo\plot_scene_subplots2.py:                     1.47 sec   0.0 MB
    - ..\examples\feature_demo\plot_colormap_image.py:                      1.45 sec   0.0 MB
    - ..\examples\feature_demo\plot_custom_object3.py:                      1.44 sec   0.0 MB
    - ..\examples\feature_demo\plot_custom_object2.py:                      1.40 sec   0.0 MB
    - ..\examples\feature_demo\plot_colormap_channels.py:                   1.33 sec   0.0 MB
    - ..\examples\feature_demo\plot_picking_mesh.py:                        1.31 sec   0.0 MB
    - ..\examples\feature_demo\plot_custom_object1.py:                      1.21 sec   0.0 MB
    - ..\examples\feature_demo\plot_transparency2.py:                       1.19 sec   0.0 MB
    - ..\examples\feature_demo\plot_manual_matrix_update.py:                0.98 sec   0.0 MB
    - ..\examples\feature_demo\plot_panzoom_camera.py:                      0.97 sec   0.0 MB
    - ..\examples\feature_demo\plot_helpers_gizmo.py:                       0.96 sec   0.0 MB
    - ..\examples\feature_demo\plot_spheres.py:                             0.91 sec   0.0 MB
    - ..\examples\feature_demo\plot_light_directional_shadow.py:            0.88 sec   0.0 MB
    - ..\examples\introductory\plot_orbit_camera.py:                        0.85 sec   0.0 MB
    - ..\examples\feature_demo\plot_pbr.py:                                 0.82 sec   0.0 MB
    - ..\examples\introductory\plot_clipping_planes.py:                     0.82 sec   0.0 MB
    - ..\examples\feature_demo\plot_mesh_slice.py:                          0.81 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_cubes.py:                      0.78 sec   0.0 MB
    - ..\examples\feature_demo\plot_scene_subplots1.py:                     0.73 sec   0.0 MB
    - ..\examples\introductory\plot_object_bounding_box.py:                 0.70 sec   0.0 MB
    - ..\examples\feature_demo\plot_volume_render1.py:                      0.67 sec   0.0 MB
    - ..\examples\feature_demo\plot_two_canvases.py:                        0.63 sec   0.0 MB
    - ..\examples\feature_demo\plot_instancing_mesh.py:                     0.61 sec   0.0 MB
    - ..\examples\feature_demo\plot_show_scene.py:                          0.58 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_polyhedron.py:                 0.57 sec   0.0 MB
    - ..\examples\feature_demo\plot_scene_overlay.py:                       0.57 sec   0.0 MB
    - ..\examples\feature_demo\plot_physical_color.py:                      0.57 sec   0.0 MB
    - ..\examples\feature_demo\plot_image_plus_points.py:                   0.54 sec   0.0 MB
    - ..\examples\feature_demo\plot_transparency1.py:                       0.51 sec   0.0 MB
    - ..\examples\feature_demo\plot_synced_video.py:                        0.51 sec   0.0 MB
    - ..\examples\feature_demo\plot_show_image.py:                          0.50 sec   0.0 MB
    - ..\examples\feature_demo\plot_volume_render2.py:                      0.50 sec   0.0 MB
    - ..\examples\feature_demo\plot_world_bounding_box.py:                  0.49 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_plane.py:                      0.44 sec   0.0 MB
    - ..\examples\feature_demo\plot_picking_color.py:                       0.44 sec   0.0 MB
    - ..\examples\feature_demo\plot_volume_slice2.py:                       0.43 sec   0.0 MB
    - ..\examples\feature_demo\plot_volume_slice3.py:                       0.42 sec   0.0 MB
    - ..\examples\feature_demo\plot_wireframe2.py:                          0.37 sec   0.0 MB
    - ..\examples\feature_demo\plot_volume_slice4.py:                       0.36 sec   0.0 MB
    - ..\examples\feature_demo\plot_wireframe1.py:                          0.35 sec   0.0 MB
    - ..\examples\feature_demo\plot_line_basic.py:                          0.34 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_torus_knot.py:                 0.33 sec   0.0 MB
    - ..\examples\feature_demo\plot_line_thick.py:                          0.32 sec   0.0 MB
    - ..\examples\feature_demo\plot_volume_slice1.py:                       0.31 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_polyhedron_subdivisions.py:    0.29 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_klein_bottle.py:               0.29 sec   0.0 MB
    - ..\examples\feature_demo\plot_geometry_image.py:                      0.27 sec   0.0 MB
    - ..\examples\feature_demo\plot_picking_points.py:                      0.25 sec   0.0 MB
    - ..\examples\introductory\plot_cube.py:                                0.24 sec   0.0 MB
    - ..\examples\introductory\plot_light_basic.py:                         0.24 sec   0.0 MB
    - ..\examples\feature_demo\plot_line_segments.py:                       0.22 sec   0.0 MB
    - ..\examples\feature_demo\plot_flat_shaded_torus.py:                   0.22 sec   0.0 MB
    - ..\examples\feature_demo\plot_show_util.py:                           0.20 sec   0.0 MB
    - ..\examples\introductory\plot_offscreen.py:                           0.19 sec   0.0 MB
    - ..\examples\feature_demo\plot_line_performance.py:                    0.18 sec   0.0 MB
    - ..\examples\feature_demo\plot_post_processing2.py:                    0.00 sec   0.0 MB
    - ..\examples\feature_demo\plot_sponza_scene.py:                        0.00 sec   0.0 MB
    - ..\examples\feature_demo\plot_post_processing1.py:                    0.00 sec   0.0 MB
    - ..\examples\feature_demo\plot_show_stl.py:                            0.00 sec   0.0 MB
    - ..\examples\cube_qt.py:                                               0.00 sec   0.0 MB
    - ..\examples\cube_wx.py:                                               0.00 sec   0.0 MB
    - ..\examples\integration_qt.py:                                        0.00 sec   0.0 MB
    - ..\examples\introductory\points_basic.py:                             0.00 sec   0.0 MB
    - ..\examples\light_view.py:                                            0.00 sec   0.0 MB
    - ..\examples\text_with_qt.py:                                          0.00 sec   0.0 MB
```

</details>

examples/feature_demo/plot_pbr.py
```diff
@@ -40,7 +44,11 @@
 # Load meshes, and apply env map
 # Note that this lits the helmet already
 gltf_path = (
-    Path(__file__).parent / "models" / "DamagedHelmet" / "glTF" / "DamagedHelmet.gltf"
+    Path(__file__).parents[1]
```
almarklein, line 47: The [sphinx-gallery docs](https://sphinx-gallery.github.io/stable/faq.html#why-is-file-not-defined-what-can-i-use) say that one could use `os.getcwd()` instead of `__file__`. That would mean the examples *must* be run from the containing folder. So we could do:
```py
try:
    this_dir = os.path.dirname(__file__)
except NameError:
    this_dir = os.getcwd()
```

FirefoxMetzger:
Adding this link here for future reference.

It is possible to specify a custom static thumbnail image for a gallery example. This can be useful for gui-specific examples that we can't run on CI. For these, we can either run it offline and insert the thumbnail, or we could display a static image (eg the logo of the GUI backend used) for those examples.

Here is the link to the relevant docs: https://sphinx-gallery.github.io/stable/configuration.html#providing-an-image-for-the-thumbnail-image

examples/feature_demo/plot_pbr.py
FirefoxMetzger: That does the trick @almarklein. I've expanded this a bit to make it less magical when a user encounters it in the gallery, and - for now - I will add the following block to each example that uses `__file__`:

image

With this, the examples now render successfully :)

image

FirefoxMetzger:
I've reorganized this PR to make use of `Display` and introduced the ability to configure how gallery examples are built using comments of the form

```python
# sphinx_gallery_pygfx_<name> = <value>
```

This builds on top of sphinx-gallery's mechanism, which takes the form of `# sphinx_gallery_<name> = <value>` (which I have discovered today). This is nice because we can configure sphinx-gallery to remove these comments from the produced RST and IPYNB files, making the result look clean. To this, I've added a `pygfx` namespace to isolate the variables we care about.

With this, we can easily control when to render. To create a render below a given example's code block one has to specify

```python
# sphinx_gallery_pygfx_render = True
```

This will trigger our scraper to check the global scope for a variable named `"renderer"`. The variable to check can be controlled by specifying 

```python
# sphinx_gallery_pygfx_target_name = "<var_name>"
# eg.
# sphinx_gallery_pygfx_target_name = "disp"
```

inside the block. The resulting variable is assumed to be a Display, Renderer, or Canvas, and is used to render the image to display.

I've also added plumbing to procude animations/GIFs by (via `# sphinx_gallery_pygfx_animage = True`); however, this is not fully implemented yet, because I first need to figure out how to generate video using pygfx (https://github.com/pygfx/pygfx/issues/381#issuecomment-1328226839).

I've also fixed the `__file__` issue mentioned earlier.

What is missing now is to set up lavapipe on the CI runners that build the docs (GH Actions and RTD). However, it should already be possible to build the gallery locally as part of the normal doc-building procedure.

FirefoxMetzger:
Okay, now the tests are passing (with exception of the one that currently breaks globally). One downside is that the doc build now takes 74 minutes because we are rendering all the examples (twice). (For comparison, on my machine with GPU building the docs takes less than 5 minutes.)

Here is a summary of how long each example takes to build on GH Actions. We may wish to discuss which examples are worth building.

<details>

```
computation time summary:
    - ../examples/feature_demo/plot_pbr2.py:                               112.51 sec   0.0 MB
    - ../examples/validation/plot_validate_volume.py:                      106.87 sec   0.0 MB
    - ../examples/validation/plot_validate_light_shadow.py:                105.47 sec   0.0 MB
    - ../examples/validation/plot_validate_ndc.py:                         105.25 sec   0.0 MB
    - ../examples/validation/plot_validate_mesh_colormap.py:               105.07 sec   0.0 MB
    - ../examples/validation/plot_validate_color.py:                       103.77 sec   0.0 MB
    - ../examples/validation/plot_validate_culling.py:                     103.36 sec   0.0 MB
    - ../examples/validation/plot_validate_box.py:                         102.63 sec   0.0 MB
    - ../examples/feature_demo/plot_light_spotlight_shadows.py:            102.58 sec   0.0 MB
    - ../examples/feature_demo/plot_wireframe_material.py:                 102.55 sec   0.0 MB
    - ../examples/feature_demo/plot_custom_object3.py:                     102.46 sec   0.0 MB
    - ../examples/validation/plot_validate_helpers1.py:                    102.24 sec   0.0 MB
    - ../examples/validation/plot_validate_image_colormap.py:              102.19 sec   0.0 MB
    - ../examples/validation/plot_validate_image1.py:                      102.16 sec   0.0 MB
    - ../examples/validation/plot_validate_depth_clipping.py:              101.85 sec   0.0 MB
    - ../examples/validation/plot_validate_helpers2.py:                    101.81 sec   0.0 MB
    - ../examples/validation/plot_validate_image2.py:                      101.19 sec   0.0 MB
    - ../examples/feature_demo/plot_light_spotlights.py:                    98.47 sec   0.0 MB
    - ../examples/feature_demo/plot_multi_select.py:                        98.06 sec   0.0 MB
    - ../examples/feature_demo/plot_custom_object2.py:                      97.59 sec   0.0 MB
    - ../examples/feature_demo/plot_mesh_depth_material.py:                 97.28 sec   0.0 MB
    - ../examples/feature_demo/plot_colormap_mesh.py:                       95.81 sec   0.0 MB
    - ../examples/feature_demo/plot_scene_subplots2.py:                     95.68 sec   0.0 MB
    - ../examples/feature_demo/plot_scene_subplots_video.py:                95.11 sec   0.0 MB
    - ../examples/feature_demo/plot_collection_line.py:                     91.74 sec   0.0 MB
    - ../examples/feature_demo/plot_dynamic_env_map.py:                     90.46 sec   0.0 MB
    - ../examples/feature_demo/plot_colormap_image.py:                      90.44 sec   0.0 MB
    - ../examples/feature_demo/plot_custom_object1.py:                      89.49 sec   0.0 MB
    - ../examples/feature_demo/plot_image_click_events.py:                  83.58 sec   0.0 MB
    - ../examples/feature_demo/plot_multi_slice1.py:                        82.83 sec   0.0 MB
    - ../examples/feature_demo/plot_light_shadow.py:                        82.75 sec   0.0 MB
    - ../examples/feature_demo/plot_multi_slice2.py:                        80.21 sec   0.0 MB
    - ../examples/feature_demo/plot_picking_mesh.py:                        79.07 sec   0.0 MB
    - ../examples/feature_demo/plot_pbr.py:                                 75.18 sec   0.0 MB
    - ../examples/feature_demo/plot_cylinder.py:                            75.16 sec   0.0 MB
    - ../examples/feature_demo/plot_colormap_channels.py:                   72.98 sec   0.0 MB
    - ../examples/feature_demo/plot_transparency2.py:                       71.32 sec   0.0 MB
    - ../examples/feature_demo/plot_scene_in_a_scene.py:                    71.08 sec   0.0 MB
    - ../examples/feature_demo/plot_skybox.py:                              35.31 sec   0.0 MB
    - ../examples/feature_demo/plot_scene_subplots1.py:                     33.32 sec   0.0 MB
    - ../examples/feature_demo/plot_volume_render1.py:                      32.77 sec   0.0 MB
    - ../examples/feature_demo/plot_spheres.py:                             32.23 sec   0.0 MB
    - ../examples/feature_demo/plot_mesh_slice.py:                          31.76 sec   0.0 MB
    - ../examples/feature_demo/plot_light_directional_shadow.py:            30.83 sec   0.0 MB
    - ../examples/feature_demo/plot_manual_matrix_update.py:                29.99 sec   0.0 MB
    - ../examples/feature_demo/plot_panzoom_camera.py:                      26.79 sec   0.0 MB
    - ../examples/feature_demo/plot_scene_overlay.py:                       25.85 sec   0.0 MB
    - ../examples/feature_demo/plot_transparency1.py:                       25.64 sec   0.0 MB
    - ../examples/feature_demo/plot_volume_render2.py:                      25.24 sec   0.0 MB
    - ../examples/feature_demo/plot_instancing_mesh.py:                     23.48 sec   0.0 MB
    - ../examples/feature_demo/plot_physical_color.py:                      18.84 sec   0.0 MB
    - ../examples/feature_demo/plot_synced_video.py:                        18.82 sec   0.0 MB
    - ../examples/feature_demo/plot_line_basic.py:                          18.71 sec   0.0 MB
    - ../examples/feature_demo/plot_volume_slice4.py:                       18.07 sec   0.0 MB
    - ../examples/feature_demo/plot_picking_color.py:                       17.61 sec   0.0 MB
    - ../examples/feature_demo/plot_volume_slice2.py:                       17.34 sec   0.0 MB
    - ../examples/feature_demo/plot_volume_slice3.py:                       17.28 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_polyhedron.py:                 16.93 sec   0.0 MB
    - ../examples/feature_demo/plot_world_bounding_box.py:                  16.76 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_cubes.py:                      14.45 sec   0.0 MB
    - ../examples/feature_demo/plot_helpers_gizmo.py:                       14.44 sec   0.0 MB
    - ../examples/feature_demo/plot_wireframe2.py:                          13.71 sec   0.0 MB
    - ../examples/feature_demo/plot_picking_points.py:                      13.01 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_plane.py:                      12.78 sec   0.0 MB
    - ../examples/feature_demo/plot_line_thick.py:                          12.69 sec   0.0 MB
    - ../examples/feature_demo/plot_volume_slice1.py:                       12.29 sec   0.0 MB
    - ../examples/feature_demo/plot_wireframe1.py:                          11.98 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_torus_knot.py:                 11.19 sec   0.0 MB
    - ../examples/feature_demo/plot_image_plus_points.py:                   10.53 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_klein_bottle.py:               10.20 sec   0.0 MB
    - ../examples/feature_demo/plot_line_performance.py:                    10.14 sec   0.0 MB
    - ../examples/feature_demo/plot_flat_shaded_torus.py:                    9.63 sec   0.0 MB
    - ../examples/feature_demo/plot_line_segments.py:                        9.12 sec   0.0 MB
    - ../examples/feature_demo/plot_show_scene.py:                           9.11 sec   0.0 MB
    - ../examples/feature_demo/plot_show_image.py:                           8.06 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_polyhedron_subdivisions.py:     7.48 sec   0.0 MB
    - ../examples/feature_demo/plot_show_stl.py:                             6.30 sec   0.0 MB
    - ../examples/introductory/plot_orbit_camera.py:                         6.19 sec   0.0 MB
    - ../examples/feature_demo/plot_geometry_image.py:                       5.94 sec   0.0 MB
    - ../examples/introductory/plot_clipping_planes.py:                      5.26 sec   0.0 MB
    - ../examples/introductory/plot_light_basic.py:                          3.84 sec   0.0 MB
    - ../examples/introductory/plot_object_bounding_box.py:                  3.50 sec   0.0 MB
    - ../examples/introductory/plot_show_util.py:                            3.41 sec   0.0 MB
    - ../examples/introductory/plot_points_basic.py:                         2.76 sec   0.0 MB
    - ../examples/introductory/plot_hello_triangle.py:                       2.53 sec   0.0 MB
    - ../examples/introductory/plot_cube.py:                                 1.93 sec   0.0 MB
    - ../examples/introductory/plot_offscreen.py:                            0.47 sec   0.0 MB
    - ../examples/other/cube_qt.py:                                          0.00 sec   0.0 MB
    - ../examples/other/cube_wx.py:                                          0.00 sec   0.0 MB
    - ../examples/other/integration_qt.py:                                   0.00 sec   0.0 MB
    - ../examples/other/light_view.py:                                       0.00 sec   0.0 MB
    - ../examples/other/post_processing1.py:                                 0.00 sec   0.0 MB
    - ../examples/other/post_processing2.py:                                 0.00 sec   0.0 MB
    - ../examples/other/sponza_scene.py:                                     0.00 sec   0.0 MB
    - ../examples/other/text_with_qt.py:                                     0.00 sec   0.0 MB
    - ../examples/other/two_canvases.py:                                     0.00 sec   0.0 MB
```

</details>

Apart from that, what is now pending in this PR is:

- Get RTD to build the gallery (I don't have a good way to test this, but @almarklein offered to check if we can get PR builds working. This should help)
- Add video/GIF for some examples.

almarklein:
Wow, 100s for one frame ... 

> (twice).

Why are we rendering them twice?

almarklein:
>  takes 74 minutes

This is a problem for RTD, because the built limit is 15 min, 30 min for a business account.

almarklein:
>  @almarklein offered to check if we can get PR builds working. This should help)

No luck yet ...

FirefoxMetzger:
> Why are we rendering them twice?

Because sphinx-gallery will run the example incrementally to give us access to the variables. This involves calling either `run()` or `show()` at some point which triggers a render behind the scenes that gets discarded. We then render a second time when we explicitly call `img = canvas.draw()` inside the scraper to actually produce the image we save for the gallery.

> This is a problem for RTD, because the built limit is 15 min, 30 min for a business account.

Indeed. Though this assumes that we are using a similar instance/machine for both services.

We could also start considering building the gallery images offline and tracking them as a "static" asset that gets regenerated on regular intervals and then pulled in by RTD when building the docs. Once I am done reorganizing the examples, I don't think they will change frequently...

almarklein:
I think you mentioned that we can also simply specify the thumbnail to use. So could we run all examples on our own terms, store the screenshots, then run sphinx gallery with the custom screenshots that we just made? That way we can already reduce the build time by half. And much further once we can do #362.

FirefoxMetzger:
Yes, we could manually render all the examples and create a folder that contains the resulting images (either inside this repo, or externally). We would then pull in those images during CI builds and use them as output instead of rendering the images on the runner. This would save 50%-ish of the time.

We could also introduce a new environment variable similar to the one that forces the `auto` GUI to use the offscreen, maybe something like `WGPU_SPHINX_MODE` which would skip the rendering when `run()` is called. This would save the other 50%-ish of the time to build the gallery.

almarklein:
> Yes, we could manually render all the examples and create a folder that contains the resulting images (either inside this repo, or externally). We would then pull in those images during CI builds and use them as output instead of rendering the images on the runner. This would save 50%-ish of the time.

I meant in one CI build step:
* Generate screenshots and gifs for the gallery, store in a local folder.
* Build the gallery using the results in the local folder.

FirefoxMetzger:
> I meant in one CI build step:

Not sure I understood you correctly. If we render on GH Actions and build the gallery on RTD wouldn't we get a race condition?Both are triggered by pushes and I don't know a good way to ask RTD to wait for an action to finish before starting to build the docs.

almarklein:
Haha, communication is hard 😄   I meant to do it as part as the build itself, as a pre-processing for the Sphinx build. E.g. from the `conf.py` we could call something that renders the screenshots, and then Sphinx proceeds as usual, except the gallery logic will not run the examples, it will use the pre-rendered screenshots.

----

The long build time is a problem. Let's see what our options are now.
* If we apply my suggestion *and* use Korijns trick #362, we can probably run all examples in a short enough time, so it could run on RTD. Some obstacles though:
  * For this to work we need to update wgpu-py to the latest wgpu-native, to *hopefully* get rid of the memory leaks that we were seeing. I hope to get to that next week.
  * We don't know yet whether we can even render examples on RTD.
* Host docs on GH pages.
* Render screenshots at GH actions and somehow host docs into RTD. There are several options (see #345) but each is tricky in its own way.

The first option would definitely have my preference. My suggestion for moving forward would be to:
* Update this PR to only render screenshots for a few examples, so the build time is short enough for RTD.
* Try to get it to run on RTD.
* When we have a new wgpu-py, see if we can get #362 working.
* Update the gallery building to use that.

FirefoxMetzger:
> E.g. from the conf.py we could call something that renders the screenshots, and then Sphinx proceeds as usual, except the gallery logic will not run the examples, it will use the pre-rendered screenshots.

I think I'm still stuck not understanding because I still don't see how this would reduce build time. The problem is that CI uses CPU rendering, which is painfully slow and either approach (pre-processed render vs render during doc building) will suffer from that bottleneck.

Are you thinking of saving some overhead costs (#362)? We are already doing all the scraper-related rendering in the same thread, but I can check if we can shave off more time by reusing e.g., the canvas or the renderer where it makes sense (i.e., for examples that use a Display and don't define an explicit canvas).

Another issue with not running the examples is that it will not call the scraper on examples that are not run/rendered. As a result, we would have to hardcode the image paths into each example (we could, but it's a bit ugly). We would also lose any other output, e.g., if a block prints `FPS: 3` then the gallery captures that and will display it below the block when running the example.

> The long build time is a problem. Let's see what our options are now.

There is also the option to do something similar to what we do in wgpu when bindings are updated: We build the gallery locally (on a machine with GPU), upload the resulting images, skip any rendering on CI, and use the pre-rendered images instead.

This would keep CI fast and run on RTD without too much hassle (we can pull in external images as a pre-build step). The downside is that images don't update when examples update; however, this may be mitigated by examples not changing frequently.

Korijn:
We could also build in an option (environment variable, or a flag on renderer) to make calls to render/draw a noop to optimize the initial scraping of code.

FirefoxMetzger:
> We could also build in an option (environment variable, or a flag on renderer) to make calls to render/draw a noop to optimize the initial scraping of code.

Yes, I was thinking of this option as well something like `WGPU_SPHINX_MODE` (though I don't like environment variables) or simply a monkey-patch inside `conf.py` that makes `wgpu.gui.offscreen.run` a noop. This way, we'd save 50% of the runtime.

---

Also, different topic, we would ideally like to have PR builds on RTD set up to better test the gallery, both in terms of runtime and in terms of final result. 

I am now done with all my pending PRs except for this one, so I will pick up some new ones today or tomorrow. Should I look into RTD pr builds (and it registering as a check inside the PR)?

Korijn:
Sure, go ahead

> > We could also build in an option (environment variable, or a flag on renderer) to make calls to render/draw a noop to optimize the initial scraping of code.
> 
> Yes, I was thinking of this option as well something like `WGPU_SPHINX_MODE` (though I don't like environment variables) or simply a monkey-patch inside `conf.py` that makes `wgpu.gui.offscreen.run` a noop. This way, we'd save 50% of the runtime.

A monkey-patch in conf.py sounds good to me.

> Also, different topic, we would ideally like to have PR builds on RTD set up to better test the gallery, both in terms of runtime and in terms of final result.
> 
> I am now done with all my pending PRs except for this one, so I will pick up some new ones today or tomorrow. Should I look into RTD pr builds (and it registering as a check inside the PR)?

Go ahead!

almarklein:
> Are you thinking of saving some overhead costs (https://github.com/pygfx/pygfx/pull/362)? We are already doing all the scraper-related rendering in the same thread, but I can check if we can shave off more time by reusing e.g., the canvas or the renderer where it makes sense (i.e., for examples that use a Display and don't define an explicit canvas).

Yes, it'd avoid everything being rendered twice, and it would give us more room for tweaks to make it faster. But I completely overlooked the fact that we're already "doing" what #362 proposes 😄  Sorry for introducing confusion.

From what we saw in earlier experiments, re-using the device creates the biggest benefit, and this happens by default when the examples run in the same process.

This means that the building is slow for the same reason that #362 is slow: because of the memory leaks in wgpu-native. With some luck, the gallery building in this PR will become *much* faster when wgpu-py updates!

So, let's cut the build time in half by avoiding the initial draws. And then try things running on RTD.

FirefoxMetzger:
that's interesting. I added the monkeypatch that I suggested earlier this week. On my local system this does speed up the gallery building (didn't quantify explicitly, but the build logs reported about 1-2 sec less per example) while still producing the screenshots as expected.

On CI this seems to make a massive difference. Rendering the docs is now down to 3 minutes (!) from around 70-80 min. I'll look into RTD now to see if I can get a rendered output and see if everything is created correctly.

FirefoxMetzger:
@almarklein Some bad news on the rendering on RTD front. I've tried installing lavapipe on RTD, but I keep running into permission errors. I will have a think if I can come up with a way of installing it that doesn't involve `apt` and potentially reach out to the RTD devs for some feedback. Hopefully one of these approaches proves fruitful. If not, we may have to consider alternatives.

FirefoxMetzger:
@Korijn 

Here is the latest one that tried using `sudo` for permissions: https://readthedocs.org/projects/pygfx/builds/18800543/
Here is the one that tried to directly install via `apt-get`: https://readthedocs.org/projects/pygfx/builds/18800442/

almarklein:
Eek! Yeah, let's reach out to the RTD devs and see if something's possible ...

Korijn:
The only way to install apt packages is via `build.apt_packages`, and sadly PPAs are not supported. See here: https://github.com/readthedocs/readthedocs.org/blob/624b13d5a208f11c1a06a9422888ebc5fdb1114d/docs/user/config-file/v2.rst#build-apt-packages

FirefoxMetzger:
> Eek! Yeah, let's reach out to the RTD devs and see if something's possible ...

Yes. I've sent a message yesterday. Let's see what they say.

> The only way to install apt packages is via build.apt_packages, and sadly PPAs are not supported.

I am aware. Though I was hoping that there might be a way around this, i.e., that this is a limitation in the API of `apt_packages` rather than an explicitly chosen restriction.

Korijn:
There are ways to "install" .deb packages without root access, see here for example: https://askubuntu.com/a/350

It's not exactly trivial though..

FirefoxMetzger:
@Korijn That's exactly what I've been looking into just now ... you must be reading my mind 😅

My main doubt is if wgpu will pick up the local installation. Do you have any experience here? Or know where I can read up on what to configure to point wgpu to those drivers (maybe setting `PATH` is enough)?

Korijn:
On this page: https://launchpad.net/~oibaf/+archive/ubuntu/graphics-drivers
Read the section: `=== Using alternative drivers - Here be dragons ===`

I think this is what you want to do:

```
* zink + lavapipe - OpenGL emulated on a software Vulkan driver:
  Set VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/lvp_icd.x86_64.json MESA_LOADER_DRIVER_OVERRIDE=zink
* llvmpipe OpenGL software render:
  Set MESA_LOADER_DRIVER_OVERRIDE=llvmpipe
```

FirefoxMetzger:
._. I tried a fresh ubuntu 22.04 image in WSL and the gallery builds fine without having to rely on a custom PAA. However, doing the same on RTD results in a segfault. Could this be because WSL uses a different kernel than standard ubuntu?

Korijn:
> ._. I tried a fresh ubuntu 22.04 image in WSL and the gallery builds fine without having to rely on a custom PAA. However, doing the same on RTD results in a segfault. Could this be because WSL uses a different kernel than standard ubuntu?

Seems unlikely.

FirefoxMetzger:
Jackpot! 🚀 🤩  @Korijn @almarklein We now have a gallery on RTD: https://pygfx--384.org.readthedocs.build/en/384/_gallery/index.html 

Looks like the issue was just one example (`collection_line.py`), which I think we can afford to not build. Not 100% sure what the issue with it is (segfaults are hard to debug remotely; my guess is OOM), but it is also one of the heavy examples that we were discussing not to build on CI anyway. I am happy to exclude it from rendering for now.

FirefoxMetzger:
We are finally ready to give this PR a serious review :) 

There is likely some scaffolding left over from some of the experimentation that I've done to get this running, so if anything looks out of place feel free to question it. Otherwise, all the pieces should be in place.

Also note: In this PR; I have only done minimal changes to the examples to make them compliant with sphinx-gallery. Once this is merged, I will do a few follow-up PRs where I will refactor individual examples to prettify them or introduce `Display`.

Korijn (approved):

Korijn:
Looks good!

I'm surprised this doesn't suffer from the issues found in #362. Does sphinx-gallery sandbox the examples to subprocesses?

FirefoxMetzger:
> I'm surprised this doesn't suffer from the issues found in https://github.com/pygfx/pygfx/pull/362. Does sphinx-gallery sandbox the examples to subprocesses?

Not that I am aware of, no. Neither does CI/Docs which builds the gallery in GH Actions. It used to be slow, but ever since I monkey-patched `wgpu.gui.offscreen.run` and replaced it with a no-op, things have been running fast and smooth.

almarklein (changes requested):
Nice work! I'm glad that you managed to get it to work on RTD. Saves a lot of pain. And the result looks great already 🚀

pygfx/__init__.py
```diff
@@ -18,6 +18,8 @@
 from .utils.viewport import Viewport
 from .utils import cm, logger
 
+from .utils.gallery_scraper import _get_sg_image_scraper
```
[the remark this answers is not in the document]
FirefoxMetzger: I placed it here to make it more similar to a standard scraper and to allow downstream projects to use it. If you've built some application on top of pygfx you can set up your downstream `conf.py` using 

```python
sphinx_gallery_conf = {
    ...,
    "image_scrapers": ("pygfx",),
    "remove_config_comments": True,
}
```

and it will pick up our scraper and use it to build your gallery. More on this in the [sphinx-gallery docs](https://sphinx-gallery.github.io/stable/advanced.html#integrate-custom-scrapers-with-sphinx-gallery).

pygfx/utils/gallery_scraper.py
```diff
@@ -0,0 +1,100 @@
+import imageio.v3 as iio
```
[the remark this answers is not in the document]
FirefoxMetzger: Oh, right. then we probably want to conditionally import it here like we do sphinx-gallery.

FirefoxMetzger:
@almarklein I looked into WEBP as an alternative to GIF for animations. As you suspected, supported formats are hardcoded into the source ([here](https://github.com/sphinx-gallery/sphinx-gallery/blob/e2ecb35d9049d0df0a1db4f6d15be817745cd565/sphinx_gallery/scrapers.py#L339)). I've opened a ticket upstream to ask for support ([here](https://github.com/sphinx-gallery/sphinx-gallery/issues/1055)), but in the meantime I've implemented a workaround using the following hack:

```python
def _get_sg_image_scraper():
    # add webp as supported extension
    import sphinx_gallery.scrapers
    sphinx_gallery.scrapers._KNOWN_IMG_EXTS += ("webp",)

    return pygfx_scraper
```

This gives us WEBP support and makes our gradients look nice again.

pygfx/utils/gallery_scraper.py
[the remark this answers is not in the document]
FirefoxMetzger: Made this a conditional import. Will re-evaluate depending on our decision to move the scraper.

pygfx/utils/gallery_scraper.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if scraper_config["render"]:
+        path_generator = block_vars["image_path_iterator"]
+        img_path = next(path_generator)
+        iio.imwrite(img_path, canvas.draw())
+        images.append(img_path)
+
+    if scraper_config["animate"]:
+        frames = []
+        n_frames = scraper_config["duration"] * 30
```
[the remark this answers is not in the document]
FirefoxMetzger: I've added a comment. Also, since we can now write WEBP instead of GIF, we can actually get almost 30 FPS (30.303030 FPS effective), so the 25Hz vs 30Hz discussion resolves itself, too :)

FirefoxMetzger (commented):
I went over the files once more, and from my side, I am happy with the PR as is.

Once @almarklein is satisfied we can check this one off the list :)

Korijn:
> > but ever since I monkey-patched wgpu.gui.offscreen.run and replaced it with a no-op, things have been running fast and smooth.
> 
> @Korijn could be worth trying that in #362. This would mean it'd be related to asyncio 🤔

I can think of a number of ways that could be the problem indeed! I will investigate.

pygfx/__init__.py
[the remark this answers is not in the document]
almarklein: Ok, that makes sense. Then my proposal would be to not import `gallery_scraper` by default, but implement 
`_get_sg_image_scraper` here that imports it when called. It looks like Sphinx-Gallery defined `_get_sg_image_scraper` to be used as a factory function to support this exact pattern.

almarklein:
Nice to see smooth gradients in the animation ❤️   It looks like the thumbnail is static (as intended). Did you do anything to make that work, or does SphinxGallery accidentally make it static when it makes the thumbnail?

Korijn:
> Nice to see smooth gradients in the animation ❤️ It looks like the thumbnail is static (as intended). Did you do anything to make that work, or does SphinxGallery accidentally make it static when it makes the thumbnail?

Just FYI: the thumbnail was not static when it was still a .gif.

Also, I think the image quality is a little low, there seem to be a lot of compression artifacts. Can we raise the quality a little?

image

FirefoxMetzger:
> Did you do anything to make that work, or does SphinxGallery accidentally make it static when it makes the thumbnail?

No, this is "by accident". I expected the thumbnail generation to break when I create a WEBP, but instead, it is picked up and turned into a static WEBP.

At some point, I should look into how this process works and how to tell sphinx-gallery that it should either generate a PNG from the WEBP or do that manually and tell the gallery to use the image we generate as thumbnail instead. In my mind, the current approach is a good enough start and then we can revise from there in future PRs.

> Just FYI: the thumbnail was not static when it was still a .gif.

No, because the thumbnail has the same format as the original image. Since GIF is officially supported by sphinx-gallery and WEBP isn't there is likely something that doesn't behave 100% correct. Something we may want to investigate, but I'm not sure if we should do that here or in a follow-up.

That said, we want thumbnails to be static. If not, you will see a lot of moving things in the gallery overview, which will be a bit messy.

> Also, I think the image quality is a little low, there seem to be a lot of compression artifacts. Can we raise the quality a little?

Yes, because WEBP uses YUV420. I will have a look if we can control the compression quality 👍

FirefoxMetzger:
> Yes, because WEBP uses YUV420. I will have a look if we can control the compression quality 👍

That said, we already improve over the default quality that comes out of GIF (and its 256 colors) 🥲 

image

FirefoxMetzger:
@Korijn Now we are using lossless compression, so any compression artifacts should disappear. This comes at the cost of additional runtime though, so we may need to revisit this once our examples or animation needs have grown to the limitations of RTD build time.

pygfx/__init__.py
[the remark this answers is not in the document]
FirefoxMetzger: Sure, I'll move it to `__init__.py` :)

Korijn (dismissed):
Looks great!

almarklein (approved):

docs/conf.py
```diff
@@ -2,11 +2,27 @@
...
 ROOT_DIR = os.path.abspath(os.path.join(__file__, "..", ".."))
 sys.path.insert(0, ROOT_DIR)
 
+# -- Sphix Gallery Hackz -----------------------------------------------------
+# When building the gallery, render offscreen and don't process
+# the event loop while parsing the example
+
+
+def _ignore_offscreen_run():
+    wgpu.gui.offscreen.run = lambda: None
```
Korijn, line 18: After diving in (much) deeper as part of #362 I would say this is a safe choice when building the docs.

I see that in the scraper you call `canvas.draw()` which bypasses the event loop, which is why it works.

However, we should be mindful that if an example performs setup or initialization of the example in either the `animate`, `before_render` or `if __name__ == "__main__":` blocks, that code will not be executed (since it's pending on the event loop), and we'd need to refactor the example for it to appear correctly in the gallery.

No need to change anything. I just hope that makes sense and wanted to ensure we're all aware of this "gotcha".

FirefoxMetzger: I'm not sure I follow. Doesn't the `animate` function get executed as part of `canvas.draw`? Otherwise, why does `plot_cube.py` show a rotating cube in the animation when I make repeated calls to `canvas.draw`?

almarklein: > Doesn't the animate function get executed as part of canvas.draw?

Yes it should. Unless I'm mistaken, Korijn's comment would only apply to code that explicitly schedules an asyncio task (well and the `if __name__ == "__main__"`).

FirefoxMetzger: Inside sphinx-gallery, the `if __name__ == "__main__"` block will get executed. Examples are run as a "fake main" to replicate the behavior as-if it was run standalone.

almarklein: With https://github.com/pygfx/wgpu-py/pull/310, we should be all good (once we have a new wgpu-py release). I think we can resolve this and merge, right @Korijn ?

FirefoxMetzger:
@Korijn I didn't get a notification that you resolved the conversation (GH doesn't seem to do those), so I missed that we can merge this until now.

Master/Main has moved on, so I have re-synced the branch and resolved a merge conflict manually. Take a look and see if it still looks good :) From my side, we are still LGTM.

almarklein (dismissed):

examples/tests/test_examples.py
```diff
@@ -24,20 +24,34 @@
...
 
 # only test output of examples that opt-in
-examples_to_test = find_examples(query="# test_example = true", return_stems=True)
+examples_to_test_dict = {
+    path.stem: path for path in find_examples(query="# test_example = true")
+}
+examples_to_test = [x for x in examples_to_test_dict.keys()]
 
 
 @pytest.mark.parametrize("module", examples_to_run)
```
Korijn, line 39: you can just pass the full Path objects to `@pytest.mark.parametrize` and use `ids` keyword argument like this: `ids=lambda module: module.stem` to simplify

Korijn (dismissed):

FirefoxMetzger:
@Korijn the validation tests broke because of a pathing issue after the `ids` update. That's now fixed :)

Korijn (approved):

Korijn:
Awesome work!

FirefoxMetzger:
And with that the gallery is live! Nice!

PR #396, opened 2022-11-28
raise exception when showing closed canvas

### LINKED ISSUE #394: Repeated calls of `Display.show` freeze the program / get stuck
I was experimenting with ways to detect if the gallery has already rendered a particular display and noticed that it is currently impossible to make repeated calls to `Display.show`. E.g. the following snippet will freeze on the second `disp.show` call when using pyside:

```python
import pygfx as gfx

cube = gfx.Mesh(
)

if __name__ == "__main__":
    disp = gfx.Display()
    disp.show(cube)

    # stuck here forever
    disp.show(cube)
```

My guess is that this happens because the `canvas`/window was closed by the user and the second call to `show` waits forever for the previous canvas to re-appear; however, this never happens because it is already closed.

This can be both a bug or a feature. We could treat it as a bug and change this so that `show` checks the state of the canvas/renderer and creates a new one if needed. This would allow repeated calls to `show`. Alternatively, we can treat this as a feature, and we can document that this is supposed to be a one-time-use method and raise a descriptive exception on repeated calls. The latter could be used by the gallery to check if the example used `show` on the given display and if so the gallery could proceed with rendering the scene for the docs.

FirefoxMetzger:
Closes #394 

This PR adds a test to `Display.show` that checks if its canvas is closed or not. If it is, an exception is raised which prevents the application from freezing.

I've also added a unit test for this; however, it currently doesn't work because the offscreen canvas can not be closed. This is something we may wish to fix in wgpu-py, or if not, something we need to update here depending on how we decide.


tests/utils/test_display.py
```diff
@@ -0,0 +1,23 @@
...
+def test_repeated_show():
+    pytest.xfail("Offscreen canvas can currently not be closed.")
+
+    cube = gfx.Mesh(
+        gfx.box_geometry(200, 200, 200),
+        gfx.MeshPhongMaterial(color="#336699"),
+    )
+
+    # always use offscreen canvas
+    # @almarklein do we have a fixture for that?
```
Korijn, line 15: There's this but it only works if the environment variable is set before importing the canvas from the auto namespace.

https://github.com/pygfx/pygfx/blob/2b762a314ed67ac81f3ed7ce1bbb2bdee2b6e0f5/examples/tests/test_examples.py#L59-L66

Korijn (dismissed):

tests/utils/test_display.py
almarklein, line 15: We have the env var that makes the `auto` backend select the offscreen backend. But no special fixtures.

almarklein:
> This is something we may wish to fix in wgpu-py,

I think this is the way to go. Should be an easy fix.

tests/utils/test_display.py
Korijn: That also only works if it's set prior to importing, so it's pretty hard to use if you're not launching a subprocess.

almarklein: Good point.

FirefoxMetzger:
@Korijn I've created https://github.com/pygfx/wgpu-py/pull/309 to allow the offscreen canvas to be closed. How do you want to proceed with this PR considering that it's test requires upstream changes? 

At the moment, the test I've added will simply xfail until the canvas can be closed. We can either merge this as is, or we can wait for a new release of wgpu-py and then merge once CI can use the new version.

tests/utils/test_display.py
FirefoxMetzger: Hm, in that case I would say we leave this test as is. Would you like me to open a new issue for writing a fixture that forces to use an offscreen canvas during unit tests?

Korijn:
Can you make the xfail conditional on the version of wgpu-py being `<=` the current version?

tests/utils/test_display.py
Korijn: That would need to be an issue at wgpu-py. It would mean we need to make a function so we can resolve after import time.

FirefoxMetzger:
@Korijn Sure. I assume wgpu follows semver quite strictly, i.e., no `.dev` or `.3-rc1` or other strangeness in its version string? If so, I've added a conditional xfail on the wgpu version.

Korijn (approved):

Korijn:
> @Korijn Sure. I assume wgpu follows semver quite strictly, i.e., no `.dev` or `.3-rc1` or other strangeness in its version string? If so, I've added a conditional xfail on the wgpu version.

Correct! Thanks.

I like to use this strategy for sensitive assertions about library internals. Especially when you have to resort to workarounds or monkeypatching, it's really nice to set a "time bomb" that will go off once you finally get around to upgrading that dependency. It forces you to re-evaluate.

PR #398, opened 2022-11-29
Fix pbr metalness

panxinmiao:
To convert materials from trimesh, when "metalic_roughness_map" exists but "roughnessFactor" or "metalicFactor" does not exist, the corresponding default value should be 1.0.

In the current "pbr. py" example, "metalness_map" does not actually work:

image

Correct results:

image


Korijn (approved):
Nice!

PR #400, opened 2022-11-30
Small tweak to iter usage

Korijn:
Follow up to #399

I was reading the PR again and noticed we could avoid traversing the entire scene graph. Should be more performant for larger scenes.

### DIFF
--- a/pygfx/utils/show.py
+++ b/pygfx/utils/show.py
@@ -143,11 +143,11 @@ class Display:
                 "Your scene does not contain any lights. Some objects may not be visible"
             )
 
-        existing_cameras = [x for x in scene.iter(lambda x: isinstance(x, Camera))]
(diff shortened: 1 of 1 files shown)


pygfx/utils/show.py
```diff
@@ -143,11 +143,11 @@ def show(
                 "Your scene does not contain any lights. Some objects may not be visible"
             )
 
-        existing_cameras = [x for x in scene.iter(lambda x: isinstance(x, Camera))]
+        existing_camera = next(scene.iter(lambda x: isinstance(x, Camera)), None)
```
almarklein, line 146: Wow,  I have never thought about using iterators this way ...

almarklein (approved):

pygfx/utils/show.py
Korijn: :) It's their main selling point!

PR #401, opened 2022-12-02
Fix WorldObject.iter function

almarklein:
I got a warning that my scene did not have any lights, when in fact it did have lights 😄  Seems like a small oversight, unless it was intended to work this way, but then we cannot use `iter` the way we do. Also added a test to avoid regressions.

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -356,10 +356,8 @@ class WorldObject(EventTarget, RootTrackable):
         if skip_invisible and not self.visible:
             return
 
-        if filter_fn is not None and not filter_fn(self):
-            return
-
-        yield self
+        if not (filter_fn is not None and not filter_fn(self)):
+            yield self
 
         for child in self._children:
             yield from child.iter(filter_fn, skip_invisible)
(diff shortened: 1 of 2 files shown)


Korijn (approved):

pygfx/objects/_base.py
```diff
@@ -353,13 +353,15 @@ def traverse(self, callback, skip_invisible=False):
...
+        are included.
+        """
         if skip_invisible and not self.visible:
             return
 
-        if filter_fn is not None and not filter_fn(self):
-            return
-
-        yield self
+        if not (filter_fn is not None and not filter_fn(self)):
```
FirefoxMetzger, line 363:
```suggestion
        if filter_fn is None:
            yield self
        elif filter_fn(self):
            yield self
```

Could we write it like this? I tripped over this on my first pass, thinking that it would not yield when `filter_fn` is None. I think this makes it a little easier to read.

FirefoxMetzger (commented):
@almarklein Good catch! I left a small suggestion, but this is good as-is, too.

pygfx/objects/_base.py
almarklein: I agree that that `if`'s double negations are a bit confusing. PR's merged now though. Perhaps we can sneak your suggestion via another PR that touches related code 😃

FirefoxMetzger: Yes, my bad😅 I'll keep it in mind and will add it in a different PR.

PR #404, opened 2022-12-09
Fix and minor refactor "normal" in shader

panxinmiao:
I am adding more general environment mapping support for MeshMaterials, and find that the current "normal" flipping is incorrect.

In addition, I have slightly refactored the normal related code to make it more general.


panxinmiao:
I want to explain more about the example of "update_culling. py".

I believe the updated results are correct.

In this scene, there is a light source that follows the camera. When the z-scale of the camera is negative, it means looking backward. However, the actual position and direction of the camera have not changed (so the light source is the not changed too), so the objects in the scene will be “in their own shadows” when "looking from behind".

I also finnaly figured out why our camera needs the "flips_winding" attribute and why normal need flip according it in shader,  which I always wondered at the beginning (Because it seems that other 3D engines don't pay attention to this).

This is also related to another issue that I have always wanted to propose.

At present, our shaders works in the world coordinate system (world-space). That is to say, all positions, normal, direction, etc. used in our shaders are located in the world coordinate system. At this time, the "view direction" calculated in the world coordinate system may not be the real "view direction" (for example, when the projection matrix has some negative scaling).
So we need to pay extra attention to the "flips_winding" attribute of the camera in the shader.

It is better to make our shader work in "view-space". In this case, we need to apply the projection matrix to convert position, direction, etc. to the camera coordinate system before entering the shader. This is a big change. If so, there is a lot of work to be done.

But there are many other advantages to using "view-space". The main one is when we have a large world, doing lighting and other calculations in world space can cause issues as floating-point precision degrades when numbers get really large. View space keeps the camera at the origin meaning all calculations will be using smaller numbers. 

In addition, I think that using “view-space” as the workspace maybe also help simplify other algorithms, such as view frustum culling, automatic shadow camera parameter settings (especially the realization of global directional light shadow with position independence), etc.

panxinmiao:
> I want to explain more about the example of "update_culling. py".
> 
> ~I believe the updated results are correct.~
> 
> ~In this scene, there is a light source that follows the camera. When the z-scale of the camera is negative, it means looking backward. However, the actual direction of the light is the not changed, so the objects in the scene will be “in their own shadows” when "looking from behind".~

Er, I made a mistake. I checked the code, When the DirectionalLight is child of the camera, the direction will apply the transformation matrix of the camera.  So the new result is incorrect, Objects should also be illuminated when viewed from behind.

So, normal flipping should be independent of the "flips_winding" attribute of the camera?  To be honest, I don't quite understand what this attribute does here, 😅

panxinmiao:
I think I know the problem. There may be some problems with "view_direction" in the current shader lighting calculation code. I need to think it over again.

panxinmiao:
It took me some time to figure out this example. 

I've been thinking about what it means to scale a dimension of space negatively. In fact, when we scale a dimension of space negatively, it not only means that the position in that direction is reversed, but also means that the "inside and outside" of the object are reversed. Of course, this "inside and outside" refers to a direction in the mathematical sense, rather than what we understand in reality (it may be a bit confusing).

I think the "flips_winding" attribute and the relevant logic were originally used to deal with this problem. Through artificial logic modification to make it conform to the "results we imagine", although it destroys the mathematically correct logic and relationship.

I'm not sure whether we should retain the relevant logic of "flips_winding", but in any case, I think there is a problem with the current rendering results.

This is a little complicated. I'll try to describe it briefly. Let's illustrate this by rendering a frustum.

Imagine that this is a side of "frustum" located in world space (the world z-axis points to the right of the screen). The direction of light is fixed in world space (does not follow the camera transformation)

image

When the z-scale of the camera is -1, we convert it into view-space (This helps us to better understand this process. Although our actual lighting calculation is located in world space, it actually makes no difference in mathematics.).

 Let's not use the manually set "flips-windy" logic first.

It like this:

image

We can see that all positions and directions (including light, normals, etc.) are flipped on the Z-axis, and the "front-back" of the surfaces are also flipped. 

so, the triangle on the yellow surface will be culled by GPU, And the green surface will be rendered. This creates an effect similar to rendering the back-side.

Then consider the lighting results,  the normal direction of the green surface is to the right, because the GPU thinks it is the front-side (also mathematically), so the shader will not flip the normal. The normal faces the direction of light, so the surface is illuminated. At the same time, our line of sight also looks to the right (z+), so there is no specular effect, (This is a little strange. In fact, it cannot happen in reality. It only exists in mathematics）.

At this point, in this example, the rendering result is as follows. 
(Note that we have fixed the light direction in world space, so we need to modify the example to add DirectionalLight directly to the scene, not to the camera)

image

If, like the original example, the light is transformed with the camera (actually changing the direction of light in world space), the rendering result is as follows, The results are obvious and easy to understand.

image


When we use the "flips-windy" logic to manually modify, that is, we manually modify the (mathematical) "front-back" attribute of the surface, the rendering results will be easy to understand.

image

At this time, we artificially modified the (mathematical) "Front-Back" of the surface. so, the yellow surface will be rendered and the green will be culled.  At the same time, in the shader, in addition to ”front-back“, we also need to flip normals according to "flips-windy".  Therefore, the yellow surface will be illuminated at this time. Similarly, due to the view direction, there will be no specular effect.

So, now, when we fix the light direction in the scene, the rendering results like this:

image


When making the light direction follow the camera transformation, the rendering results like this:

image

panxinmiao:
I tested on three.js, and the results are consistent.

image

The test code:  https://gist.github.com/panxinmiao/5893323de44e1570593795f4e558560d

panxinmiao:
Therefore, I think whether we can delete the logic related to "flips_winding".  Although it tries to make the rendering result conform to our imagination, it actually destroys the logical correctness in mathematics, and may cause some confusion in rendering and lighting calculation.

In fact, negative dimension does not exist in reality, and we can't imagine what it looks like, sowhether it is better to render according to the actual logical results? Moreover, the rendering results according to logic can also help us to imagine the pictures that do not exist in reality to a certain extent.

What do you think? @almarklein @Korijn

almarklein:
I'll have to take some time to dive into this :)  I'll do this once I've rounded off the text PR.

almarklein:
Thanks for taking the time to dive into this and provide such detailed explanations! ❤️ 

I agree that in the current `main`, the idea to account for camera flipping is flaky, the implementation sloppy, and the result of the validation test is wrong 😄 

I do think that being able to flip dimensions (making a scale negative) has it's use-cases. 
* E.g. you load in a model, but want it mirrored. 
* Also sometimes you want to flip an axis of a complete world, because of how a coordinate frames is defined (e.g. in CT / MRI).
* Not sure about the camera, but people *can* flip it, so let's support it.

BTW: For some reason I took on the assumption that the camera's scale could be used to implement the 2nd use-case of flipping whole worlds. This was wrong. I'll get back to that in another issue.

I took the liberty to push new commits to this PR. The implementation is now much simpler. Basically, in the vertex shader we check the transformation matrices to see if they contain a flip. If they do, we swizzle the 3 vertices that make up the face, so that the GPU will correctly identify the face as either front or back. The fragment shader then no longer needs any trickery, except for flipping the normal for back faces.

With the new code, the validation example correctly has the bottom shapes unlit (because the light direction is opposite of the view direction):

image

I updated that example to include some more cases though (and made the directional light a child of the camera):

image

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -193,27 +193,31 @@ def code_vertex(self):
...
 
-            // If the camera flips a dimension, it flips the face winding.
-            // We can correct for this by adjusting the order (sub_index) here.
-            sub_index = select(sub_index, -1 * (sub_index - 1) + 1, u_stdinfo.flipped_winding > 0);
+            // If a transform has an uneven number of negative scales, the 3 vertices
+            // that make up the face are such that the GPU will mix up front and back
+            // faces, producing an incorrect is_front. We can detect this from the
+            // sign of the determinant, and reorder the faces to fix it.
+            let winding_world = sign(determinant(world_transform));
+            let winding_cam = sign(determinant(u_stdinfo.cam_transform));
```
almarklein, line 214: Does someone with better Math skills happen to know a more efficient way to calculate the sign of the determinant of a 4x4? Maybe @FirefoxMetzger ?

FirefoxMetzger: That depends on what info we have access to without doing any extra work. 

- If you have an arbitrary 4x4 matrix (`A`), `determinant(A)` is about as good as it gets. That is assuming that it is implemented using something like an LU decomposition internally or comes from a trusted source.
- If you have a transformation matrix in standard form (bottom row is `(0, 0, 0, 1)`) you can focus on `A[:3, :3]` and compute the determinant of that 3x3 matrix instead. This makes it slightly faster. There might even be a specialized algorithm available, but I would have to go and search for it.
- If you have the parts of the transform (translation, rotation, scale), then you should be able to simply do `winding_cam = prod(sign(scale))` and ignore translation and rotation (both always have `det(A)=1`). I.e., you can count the negative scale factors and if there is an uneven number you have `det(A)<0`.

Korijn: This is what Three does: "compute the determinant of that 3x3 matrix"

almarklein: Thanks for the suggestions. I did the 3x3 trick.

almarklein:
Here's another attempt to explain the situation in different words. Imagine the following scene, with the orange ball a directional light. The green and yellow two surfaces (mesh), and the black lines their normals.
image

When looking from the same direction as the light, you want to see both surfaces lit. For this to happen, the normal of the yellow surface must be flipped. We do this with this line:
```wgsl
normal = select(-normal, normal, is_front);
```
where `is_front` is a builtin variable that flags whether we're looking at the front or back of a face. The GPU determines it's value from the winding (i.e. order) of the vertices.

All is well ... until you start flipping axii (i.e. using negative scale values), because that messes up the winding, making the GPU produce wrong values for `is_front`. Mathematically you could say that flipping axii means turning the meshes inside-out.

In the current `main` I tried to correct for this, but only for the camera transform, and by doing additional flipping the normal in the fragment shader.

In this PR, this is changed to take the whole transform into account, and by swizzling the vertices of the face, so that the `is_front` is correct.

panxinmiao:
Thanks for the detailed explanation!
I like the current changes, so that we can use a unified `normal` in different lighting models and all subsequent calculations that require normals.

panxinmiao:
I added some changes, and there are still some problems when using "flat_shading" (the normal is calculated from screen-space derivatives).

I seem to understand the problem. Please do not merge this PR first.

panxinmiao:
I updated the "light_view.py" and "validate_culling.py" (you can press any key to switch the flat_shading mode) examples to test this.

panxinmiao:
I think it should be OK now. Now the code is ready for review. 😄

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -329,21 +346,34 @@ def code_fragment(self):
...
+                    $$ if use_normal_map is defined
+                        var normal_map = textureSample( t_normal_map, s_normal_map, varyings.texcoord );
+                        normal_map = normal_map * 2.0 - 1.0;
+                        let normal_map_scale = vec3<f32>( normal_map.xy * u_material.normal_scale, normal_map.z );
+                        normal = perturbNormal2Arb(view, normal, normal_map_scale, varyings.texcoord, is_front);
+                    $$ else
+                        var normal = vec3<f32>(varyings.normal);
+                    $$ endif
+                    // Flip the vertex normal if we're looking at the back face
+                    normal = select(-normal, normal, is_front);
```
almarklein, line 373: I refactored this to also do the flipping for backfaces when a normal map is supplied. Otherwise the back/inside of such meshes would be dark. @panxinmiao I think this makes sense, do you agree?

almarklein: Oh wait, this seems to be covered in `perturbNormal2Arb`. My bad. Will revert.

almarklein:
I'm also looking into the NormalMaterial and NormalLinesMaterial, to make them show the intended normals.

pygfx/renderers/wgpu/meshshader.py
panxinmiao: Yes, the flipped `normal` is a parameter of `perturbNormal2Arb`.

Here is a test model to confirm the correctness of the processing: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/NormalTangentTest

This is the test code: https://gist.github.com/panxinmiao/f4037a2a3bf037dfb476170a735bc76a

BTW, this also confirms a previous problem. When loading the pbr model from trimesh, the default value of `normal_scale` should be (1, - 1). I also corrected this problem in this PR.

almarklein:
> I'm also looking into the NormalMaterial and NormalLinesMaterial, to make them show the intended normals.

I did not change these much, because (I think) we want these to show the surface normal, which they already did. What I did do was:

* Make `flat_shading` work with the `NormalMaterial` by moving the calculation of the surface normal to the top of the shader.
* Add a `line_length` prop to `NormalLinesMaterial`.
* Make the lines of the `NormalLinesMaterial` stick out of the back face a tiny bit, so that it's at least visible there.
* Improve the documentation of these material classes.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -305,14 +322,27 @@ def code_fragment(self):
...
+            var surface_normal = vec3<f32>(varyings.normal);
+            $$ if flat_shading
+                let u = dpdx(varyings.world_pos);
+                let v = dpdy(varyings.world_pos);
+                surface_normal = normalize(cross(u, v));
+                // Compute the normal from the 'world_pos' partial derivatives with respect to window coordinates.
+                // Therefore no need to flip the normal by face orientation, but we need consider the sign of the camera dimension.
+                surface_normal = select(-surface_normal, surface_normal, varyings.winding_cam < 0.0);
+                // If this is a back face, we need to flip the normal to the front
+                surface_normal = select(-surface_normal, surface_normal, is_front);
```
panxinmiao, line 335: Uh, I think `$$ else` is missing here.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -329,21 +359,24 @@ def code_fragment(self):
...
-                normal = select(normal, -normal, (select(0, 1, is_front) + u_stdinfo.flipped_winding) == 1);
+                // Get normal used to calculate lighting
+                $$ if use_normal_map is defined
+                    // Get normal from the normalmap and use the PBR logic to deal with back faces
+                    let normal_map = textureSample( t_normal_map, s_normal_map, varyings.texcoord ) * 2.0 - 1.0;
+                    let normal_map_scale = vec3<f32>( normal_map.xy * u_material.normal_scale, normal_map.z );
+                    let normal = perturbNormal2Arb(view, surface_normal, normal_map_scale, varyings.texcoord, is_front);
+                $$ else
+                    // Flip the surface normal if we're looking at the back face
+                    let normal = select(-surface_normal, surface_normal, is_front);
```
panxinmiao, line 376: After adding `$$ else` above, there should be no need to flip the normals again.
just `let normal = surface_normal`

panxinmiao:
Using the test model, there are some problems when looking from the back.
image

Correct results:
image

I left some comments above.

almarklein:
@panxinmiao  I think I fixed the problem. Could you check again 😄 ?

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -539,26 +570,26 @@ def code_vertex(self):
...
+            let world_pos2 = u_wobject.world_transform * vec4<f32>(raw_pos + scaled_normal, 1.0);
+            let world_pos = world_pos1 * f32(r) + world_pos2 * (1.0 - f32(r));
             let ndc_pos = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos;
 
             var varyings: Varyings;
             varyings.world_pos = vec3<f32>(world_pos.xyz / world_pos.w);
             varyings.position = vec4<f32>(ndc_pos);
 
             // Stub varyings, because the mesh varyings are based on face index
+            varyings.normal = vec3<f32>(world_pos2.xyz - world_pos1.xyz);
```
panxinmiao, line 592: I think there are some problems in the normal calculation here. When the object has non-equal scaling, the calculation is incorrect.

panxinmiao:
> @panxinmiao I think I fixed the problem. Could you check again 😄 ?

The normalMap is OK now. But I think there are still some problems with normal calculation in `MeshNormalLineMaterial`

image

As shown in the picture above, when the x dimension is scaled by 0.5, it is incorrect to directly apply the model matrix to calculate the transformed normal (or calculate the endpoint position of the normal as in the code).

I have fixed it in other Mesh shaders, but not in `MeshNormalLineMaterial`

panxinmiao:
I submitted a commit to try to correct it.

almarklein:
All green! @panxinmiao Could you have another look?

Korijn (approved):

panxinmiao:
> All green! @panxinmiao Could you have another look?

LGTM! 😄

PR #408, opened 2022-12-14
Tame backgrounds

FirefoxMetzger:
This PR updates the examples that currently have a "lsd-background" (blue/green gradient) and replaces those backgrounds with the same background we use in `gfx.Display`. Where possible, I've also updated the examples to use `gfx.Display` instead of creating a renderer explicitly.


Korijn (approved):

almarklein (approved):
A like PR's that reduce code :)

PR #421, opened 2023-01-05
Fix line shader to avoid artifacts on straight lines

### LINKED ISSUE #419: Weird line rendering with a long line where relatively few points have a high alpha value
For example in this video the length of each of these lines (left to right) is 8410, and only a few points along the line have high alpha values. The artifact is that arrows appear at the edges of the where the colors fade out (could it be an interpolation thing?).

https://user-images.githubusercontent.com/9403332/210306005-19fa7d20-b237-416b-af0e-849ae57a6200.mp4

The lines are instantiated like this:

```python

pygfx.LineMaterial

pygfx.Line(
  geometry=pygfx.Geometry(positions=<[8410, 3]>, colors=<[8410, 4]),
  material=material(thickness=20, vertex_colors=True)
)
```

I tried adjusting the camera near and far planes but it doesn't make a difference. GPU is an nvidia RTX 3090, OS is Debian 11, `vulkaninfo` seems fine. 

Anyways the purpose is to use color intensity of a very long line that represents timepoints to indicate the magnitude of a quantitative variable, not sure if it's possible to do this with a simple Mesh instead if that's better than a Line?

image

almarklein:
Closes #419

Changes how the 5th vertex position is calculated, avoiding the generate case when the current point is exactly on the line between the two neighboring points (i.e. the line is straight). Also fixes the calculation of the angle so it can actually be used to decide on what side the "inside" is. 

I visually confirmed that the line examples still produce correct results, and that the artifacts in #419 are gone.

### DIFF
--- a/pygfx/renderers/wgpu/lineshader.py
+++ b/pygfx/renderers/wgpu/lineshader.py
@@ -305,7 +305,7 @@ class LineShader(WorldObjectShader):
             let npos2 = get_point_ndc(i);
             let npos3 = get_point_ndc(min(u_renderer.last_i, i + 1));
 
-            // Convert to logical screen coordinates, because that's were the lines work
+            // Convert to logical screen coordinates, because that's where the lines work
             let ppos1 = (npos1.xy / npos1.w + 1.0) * screen_factor;
             let ppos2 = (npos2.xy / npos2.w + 1.0) * screen_factor;
             let ppos3 = (npos3.xy / npos3.w + 1.0) * screen_factor;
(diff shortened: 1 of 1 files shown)


pygfx/renderers/wgpu/lineshader.py
```diff
@@ -344,19 +344,20 @@ def code_vertex_core(self):
...
 
-                // From the angle we can also calculate the intersection of the lines.
+                // From the angle we can also determine how long the ne vector should be.
                 // We express it in a vector magnifier, and limit it to a factor 2,
                 // since when the angle is ~pi, the intersection is near infinity.
                 // For a bevel join we can omit ne (or set vec_mag to 1.0).
                 // For a miter join we'd need an extra vertex to smoothly transition
                 // from a miter to a bevel when the angle is too small.
-                // Note that ne becomes inf if v1 == v2, but that's ok, because the
-                // triangles in which ne takes part are degenerate for this use-case.
```
almarklein, line 357: Short story is that this assumption did not hold :)

kushalkolar:
Wow you guys are amazing! This fix works :tada: 

Using the lines in these massive visualizations which were impossible with pyqtgraph :smile: 

https://user-images.githubusercontent.com/9403332/210887997-864b21ea-a228-44b5-a795-39d16286b8d0.mp4

Korijn (approved):

Issue #425, opened 2023-01-09
Gallery build breaks on windows


FirefoxMetzger:
In https://github.com/pygfx/pygfx/pull/422 we changed the behavior of building the gallery to only execute files that are in a path that is not `**/other/*`. This breaks on Windows - I guess because paths use `\` instead, which the current regex doesn't capture.

I'll submit a fix soon(TM).

FirefoxMetzger:
Another issue is that our test suite doesn't specify an encoding to use when testing the examples. This used to not be a problem, but the text rendering PR (https://github.com/pygfx/pygfx/pull/305) introduced some kyrillic in the example. Ofc, this breaks on Windows, which doesn't default to using UTF-8 🙄.

I'll leave the comment for this here and fix it in the corresponding PR, because they both small issues and seem related.

almarklein:
> This breaks on Windows - I guess because paths use \ instead, which the current regex doesn't capture.

Woops, my bad!

> Another issue is that our test suite doesn't specify an encoding to use when testing the examples.

I think Korijn fixed this in #424

FirefoxMetzger:
@almarklein Indeed. I will wait until #424 is merged before continuing this - otherwise, I think we will end up duplicating the work and may create an ugly merge conflict.

PR #424, opened 2023-01-06
asyncio fixture no longer needed with new wgpu-py

Korijn:
Closes #405

PR #426, opened 2023-01-09
Fix slash/backslash bug when building the gallery on Windows

FirefoxMetzger:
Closes: #425 

This PR adds `**\other\*` to the set of examples to exclude from execution when building the docs. This is necessary, because windows uses `\` as path separator instead of `/`.


Korijn (approved):

FirefoxMetzger:
`Fatal Python error: Aborted` doesn't sound good ... I didn't touch library code, so this could be a broken dependency?

@almarklein We recently released a new wgpu-py and seeing https://github.com/pygfx/pygfx/pull/424 I don't think pygfx compatible with it yet. Is the reason that we aren't pinning wgpu?

Korijn:
I think Almar assumed this update wouldn't break anything upstream as he incremented the patch version number. :)

FirefoxMetzger:
@Korijn RTR :) thanks for fixing the wgpu PR today.

Korijn (approved):

Issue #427, opened 2023-01-09
Add autosummary in the docs


FirefoxMetzger:
I am looking into https://github.com/pygfx/pygfx/issues/113 and I've noticed that our reference docs are currently a "dump" of all modules onto a single page. I think it would be cleaner if we had each module on a single page.

The downside of this would be that we can no longer Ctrl+F on a single page to find the function we want. The upside is that we don't serve a single gigantic page if users only want to check a single function/module. It is also advantageous if we ever want to introduce analytics, because we can get per-module view-counts.

almarklein:
> I think it would be cleaner if we had each module on a single page.

Maybe not per module, but per sub-package. E.g. geomeyty, materials, cameras, etc.

FirefoxMetzger:
On second thought, it might not be wise to split on the module/sub-package level. We make an effort to forward all classes and functions into the global `pygfx` namespace. Consequentially, there is no easy way for a user to know which sub-package to check when looking up a specific function. For example, how would I go from `gfx.Display` to searching in `gfx.utils.Display`?

From that perspective, I think it is easier if all classes and functions are documented on the same page using their user-facing name (e.g. `pygfx.Display` for `pygfx.utils.Display`).

almarklein:
We could have a page for each subpackage and still show `gfx.Display`, right?

FirefoxMetzger:
We could have a summary list for the main namespace that links to the docs in the sub-package. The downside to this is that we would have to manually maintain this list. 

Alternatively, we could duplicate the documentation and have the docs generated for both `gfx.Display` and `gfx.utils.Display`. I personally don't like this option very much, but am suggesting it here anyway for completeness.

almarklein:
> We could have a summary list for the main namespace that links to the docs in the sub-package. The downside to this is that we would have to manually maintain this list.

Ah, because then we cannot simply autodoc the whole module. I like the idea of documenting per subpackage, to have somewhat smaller chunks, and we can introduce things a bit in the subpackage's docstring. We can start with `automodule`, to make it easy, despite causing the objects to contain the subpackage name. @Korijn what do you think?

FirefoxMetzger:
I started putting something together in https://github.com/pygfx/pygfx/pull/430 .

Here is a link to the rendered docs on how I would do it (not finished, but you get an idea of what I am aiming at):
https://pygfx--430.org.readthedocs.build/en/430/reference.html

almarklein:
[...]

Issue #113, opened 2021-08-26
Make sure that the docs are complete


almarklein:
Complete as in all public classes, methods, and properties have a docstring that is complete.

Let's track per subpackage:
* [x] cameras
* [x] controls
* [x] geometries
* [x] helpers
* [ ] linalg
* [x] materials
* [x] objects
* [x] renderers -> also document the "common functions" that we make available internally.
* [x] resources
* [x] utils (I think we only have utils for private use)

almarklein:
I think we can leave this open to remind ourselves on the parts that are not yet documented.

almarklein:
Closing! The only remaining point is linalg, but that will get its own docs, I presume.

Korijn:
See https://github.com/pygfx/pylinalg/issues/3

PR #430, opened 2023-01-09
Complete the API reference docs

FirefoxMetzger:
Closes: https://github.com/pygfx/pygfx/issues/427

This PR adds missing API docs and begins to complete https://github.com/pygfx/pygfx/issues/113.


docs/reference.rst
```diff
@@ -1,13 +1,63 @@
...
 
-Resources
----------
+The primary way of accessing pygfx is by using the members of its top-level namespace.
+Currently this includes the following classes, which comprise the public API:
+
+.. autosummary::
+
+    pygfx.objects.WorldObject
+    pygfx.utils.color.Color
```
almarklein, line 12: Can we name these just `pygfx.WorldObject` etc?

almarklein: > It should be possible to create alias names or get autosummary to show a different name in the result

That would be great. That `pygfx.utils.color.Color` is rather ugly in hindsight 😄

FirefoxMetzger:
FYI @almarklein I began adding a mini gallery to the docstrings that I am editing. We briefly discussed this once during a meeting, but never dug deeper into it.

You can now see rendered examples of them here:
- `box_geometry`: https://pygfx--430.org.readthedocs.build/en/430/_autosummary/_autosummary/pygfx.geometries.box_geometry.html#pygfx.geometries.box_geometry
- `trimesh_geometry` https://pygfx--430.org.readthedocs.build/en/430/_autosummary/_autosummary/pygfx.geometries.trimesh_geometry.html#pygfx.geometries.trimesh_geometry

FirefoxMetzger:
@almarklein I just realized that I began to instinctively turn all the docstrings that I found into numpy/numpydoc style docstrings. However, I just realized that we currently use a mix of google-style and numpy style in that it mostly follows google's convention, but uses `Parameters` instead of `Args`. On the other hand, pylinalg imposes numpydoc style.

I didn't find any explicit statement on which kind of style to go for and was wondering if you have a strong opinion either way. (Since I am going over most of the codebase anyway to update docstrings we could combine this with a style change if desired or add an explicit comment on this somewhere.)

Korijn:
Personally I prefer numpydoc, but on the other hand I've never done anything this extensive with sphinx so feel free to opt for anything else you find more productive.

FirefoxMetzger:
I, too, lean towards the numpydoc style, due to the added flexibility provided by the increased number of sections when compared to google's style.

At the same time, @almarklein is writing a lot of the code for this project, so I think he should also weigh in on this.

almarklein:
In general I have a minor preference for the Google style because it feels lighter. But for this project I agree that numpydoc would probably be more suitable.

FirefoxMetzger:
@almarklein I'm done with the geometry submodule. If you like, feel free to review it.

FirefoxMetzger:
The helper module is done as well.

FirefoxMetzger:
@Korijn @almarklein what shall we do with the linalg module? It will (hopefully soon) be removed in favor of pylinalg, which makes it a bit redundant to document it extensively. I can, ofc do that if you think it has merit, but personally I would rather put a warning on the module and say "this module will disappear soon. check out pylinalg instead".

almarklein:
>  I would rather put a warning on the module and say "this module will disappear soon. check out pylinalg instead".

Yes, let's not waste effort on documenting the to-be-deprecated linalg.

pygfx/helpers/_box.py
```diff
@@ -4,14 +4,29 @@
...
 
-    Parameters:
-        size (float): The length of the box' edges (default 1).
-        thickness (float): the thickness of the lines (default 1 px).
-        color (Color): the color of the box.
     """
 
+    # @almarklein: Should we move this to geometry._box? It clearly has a
+    # usecase as a aabb helper, but seems to have merit outside of this context,
+    # too.
```
almarklein, line 28: I don't have a strong opinion on the module that this is in, but its still be a helper, so this module makes sense? 😃

pygfx/helpers/_box.py
```diff
@@ -46,27 +61,34 @@ def __init__(self, size=1.0, thickness=1, color="white"):
             dtype="f4",
         )
         positions -= 0.5
+
+        # @almarklein: does this mean we currently don't support non-isotropic scale?
+        # otherwise we may want to update this to set the tranform's scale instead.
```
almarklein, line 66: The geometry is always a cube, but the transform's scale can be used to stretch it. See `self.scale.set` below.

pygfx/helpers/_box.py
```diff
@@ -46,27 +61,34 @@ def __init__(self, size=1.0, thickness=1, color="white"):
...
         positions *= self._size
 
         geometry = Geometry(positions=positions)
         material = LineSegmentMaterial(thickness=thickness, color=color, aa=True)
 
         super().__init__(geometry, material)
 
+    # should this become a classmethod? (alternative constructor)
```
almarklein, line 74: I don't think so. With this method you can have a single helper object, and use it to "highlight" objects in the scene, e.g. as the user clicks them.

pygfx/helpers/_gizmo.py
```diff
@@ -297,10 +326,19 @@ def _highlight(self, object=None):
...
+        This method is overloaded from the base class to "attach" the gizmo to
+        the controlling object.
+
         """
+
+        # @almarklein: Assuming I interpreted above docstring correctly,
+        # wouldn't it be easier to make the gizmo a child of the controlled
+        # object and let the transform/linalg engine take care of the moving?
+        # (as its will move together with its parent, who's position the gizmo
+        # updates.)
```
almarklein, line 340: We tried that at the time, but decided against it. I can't recall the reasons though 🤔

setup.py
```diff
@@ -46,6 +46,12 @@
         "sphinx-gallery",
         "imageio",
         "matplotlib",
+        # duplicate example dependencies to avoid pyside6
```
almarklein, line 49: Maybe PySide6 should not be in the example deps. A few examples are qt-specific, but it can also be e.g. pyqt5. @Korijn ?

pygfx/geometries/_box.py
```diff
@@ -14,24 +14,43 @@ def box_geometry(
...
+    Creates a box (a rectangular cuboid) of the given size that is centered
+    around the local frame's origin. Faces may be subdivided by specifying the
+    number of segments along each axis. This will result in all faces parallel to
+    the given axis to be evenly devided into the requested number of segments.
+
+    .. @almarklein: What does it mean for corners to be square?
+
+    The normals for each side all point in the same direction; the corners are
+    actually square. Texture coordinates for each side run from 0 to 1.
+
```
almarklein, line 28: Meh, let's drop this comment. The thought was that if you'd share the corner vertices, they'd appear oddly rounded. But this comments feels out of place now.
```suggestion
```

pygfx/helpers/_gizmo.py
```diff
@@ -494,21 +532,25 @@ def _update_visibility(self):
...
         # some calculation at draw time (or right before a draw, to be
         # precise), and for these calculations it needs the viewport
-        # cand camera.
+        # and camera.
         viewport = Viewport.from_viewport_or_renderer(viewport)
         self._viewport = viewport
         self._camera = camera
 
+        # @almarklein: out of curiosity, what happens if we have more than one
+        # gizmo? Will events alias/collide?
```
almarklein, line 547: Unless I'm mistaken, the gizmo only gets events when its elements are clicked, so it *should* work.

pygfx/helpers/_lights.py
```diff
@@ -149,10 +182,28 @@ def _update(self):
...
+    --------
+    .. minigallery:: pygfx.SpotLightHelper
+
     """
 
+    # @almarklein: What is the expected shape of this indicator? Also, could we
+    # combine its code path with that for the spot light, but make the Mesh a
+    # cone that points into the opposite direction of the light source (ie. in
+    # such a way that the bottom of the cone has the light coming out, like a
+    # lamp)
```
almarklein, line 205: They look like this:
image

The idea of the helper to see quite precisely where it's pointed at, which would not be as clear with the "lamp". It could be worth improving the helper, but it'd require more discussion, so it'd be more suited in a separate issue/PR.

pygfx/geometries/_cylinder.py
```diff
@@ -143,7 +143,53 @@ def cylinder_geometry(
...
-    """Create geometry representing a cylinder."""
+    """Generate a cylinder or a cylinder segment.
+
+    .. @almarklein @korjin: isn't this inconsistend
+    .. with our convnetion that the z-axis is forward?
+
+    This function generates a cylinder or a cylinder segment. The cylinder's
+    axis runs along the local z-axis, and its midpoint is located at the local
+    origin. The cylinder's faces (top and bottom cap) are approximated by
+    regular N-sided polygons, with corners on a circle of the given radius.
```
almarklein, line 154: >  The cylinder's axis runs along the local z-axis

I don't think this is the case. I ran the `speres.py` example, and modify the geometry like this:
```py
geometry.positions.data[:,1] *= 0.5
```
Causing the spheres to be flat in the direction of their axis.
image

almarklein (commented):
Nice work so far!

docs/conf.py
```diff
@@ -48,7 +48,13 @@ def _check_objects_are_documented():
...
 
 
-_check_objects_are_documented()
+# TODO: @almarklein it is a bit odd to have a test like this in the config file
+# that will build the docs in particular because it runs before the docs are
+# built ... could we make it a "normal" unit test in something like `test_docs`?
+# Also: this breaks when we autogenerate the docs, because there will be no
+# explicit `.. autoclass::` line anymore. I could work out how to adapt this
+# function though, if you think it is worth keeping.
+# _check_objects_are_documented()
```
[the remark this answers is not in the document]
FirefoxMetzger: You still have to manually "mention" a new function in the autosummary list of the respective module. It's harder to miss, because you will see the module's docstring when you add the function to the module's `__init__.py`, but one could still miss it. (We could automate this, too, but it significantly reduces our flexibility in writing nice docs.)

That said, I'm not aware of any other repo that has such a check, so keeping/discarding depends on how "bad" we think it is if we miss a function in the docs. (We can always add it later and the docs update instantly.)

pygfx/helpers/_axes.py
```diff
@@ -13,11 +13,23 @@
 
 
 class AxesHelper(Line):
-    """An object indicating the axes directions.
+    """Generate an axis indicator.
```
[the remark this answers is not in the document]
FirefoxMetzger: This first sentence is what shows up in the summary table of the module, so I like to write them in a way that they are coherent when you see them all aggregated there. (ref: https://pygfx--430.org.readthedocs.build/en/430/_autosummary/pygfx.geometries.html)

As I understand it, [Numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes) recommends documenting the constructor's arguments in the classes docstring (under `Parameters`). Since we do that, I think it makes sense to match the rest of the style as if we are documenting the constructor.

pygfx/helpers/_axes.py
```diff
@@ -33,6 +45,7 @@ def __init__(self, size=1.0, thickness=2):
             dtype="f4",
         )
 
+        # TODO: maybe change to colors that are friendly for colorblind users.
```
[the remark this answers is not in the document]
FirefoxMetzger: Yes, I would not address this here, because the PR will get big enough as-is.

pygfx/helpers/_box.py
FirefoxMetzger: We could call it `WireframeBox` and then the name "Helper" disappears ;). I mainly think that a box wireframe can do more than just help draw a bounding box.

(also while on the topic of wireframes, shouldn't it be possible to write a shader that draws a wireframe instead of drawing faces, removing the need for explicit wireframe objects?)

pygfx/helpers/_box.py
FirefoxMetzger: In that case, why do we need `scale` _and_ `size`? Wouldn't it be enough to always construct a unit cube and use scale?

pygfx/helpers/_box.py
FirefoxMetzger: Ah, I see.

pygfx/helpers/_gizmo.py
FirefoxMetzger: So each gizmo has it's own event queue? It makes sense that only the gizmo that is clicked emits an event, but I don't know how it works on the receiving end if multiple gizmos are listening.

pygfx/helpers/_lights.py
FirefoxMetzger: That makes sense. In this case, lets take it up again once there is a specific need/use-case for an improvement.

pygfx/geometries/_cylinder.py
FirefoxMetzger: I got this from the source of the cylinder, which constructs the mantle positions using the following segment:

https://github.com/pygfx/pygfx/blob/eb085bd5010445ea7ddc807da0d4e7a012a6798c/pygfx/geometries/_cylinder.py#L36-L39

I didn't do any testing on this though, so it could be that there is a rotation somewhere that matches it with our convention.

docs/conf.py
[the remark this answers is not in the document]
almarklein: Anyway, let's drop the check here.

pygfx/helpers/_axes.py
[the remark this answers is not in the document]
almarklein: > This first sentence is what shows up in the summary table of the module, so I like to write them in a way that they are coherent when you see them all aggregated there.

Fair enough for those use-cases.

> Since we do that, I think it makes sense to match the rest of the style as if we are documenting the constructor.

From my experience it's more common to describe the object. The fact that you also describe the parameters to instantiate such an object does not conflict IMO. This is an example from the same numpydoc page:

image

pygfx/helpers/_axes.py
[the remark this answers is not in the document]
almarklein: #435

pygfx/helpers/_box.py
almarklein: > We could call it `WireframeBox` and then the name "Helper" disappears ;). I mainly think that a box wireframe can do more than just help draw a bounding box.

I'm not convinced yet. It still feels like a helper to me 😄. If you take your position to the extreme, you could argue that all objects in the helper submodule should be first class world objects. In the end it does not matter much where they live. Let's discuss in a new issue?

> (also while on the topic of wireframes, shouldn't it be possible to write a shader that draws a wireframe instead of drawing faces, removing the need for explicit wireframe objects?)

Yes, we have that! It's the `MeshMaterial.wireframe` property. But you'll see wires for all the edges of the triangle, so it serves a different purpose.

pygfx/helpers/_box.py
almarklein: Good point. I suspect that it has a `size` arg because many geometry-generating functions have it. And for these it makes sense, because `.scale` also sizes the children and you may not want to do that. But helper objects are unlikely to have children. I'd be ok with removing the `size` arg. @Korijn ?

pygfx/helpers/_gizmo.py
almarklein: Events are only dispatched to world objects if they are clicked on (or hovered over, etc.).

In more detail: the canvas generates the event. The renderer picks it up and checks if it applies to a certain world object, e.g. because the object is under the mouse, or when a world object is tracked (i.e. dragging). And the event propagates to parent world objects, similar to how it does in the browser dom.

pygfx/geometries/_cylinder.py
almarklein: Oh, but the cylinders are indeed not facing up:
image

I think this is something that we should change.

setup.py
Korijn: Can you explain why we want avoid pyside6?

pygfx/geometries/_cylinder.py
FirefoxMetzger: Let's discuss this in https://github.com/pygfx/pygfx/issues/436

pygfx/geometries/_text.py
```diff
@@ -201,14 +237,21 @@ def __init__(
...
-        coordinates at the object's point in the world. The object's
-        local rotation and scale can still be used to rotate and scale
-        the text. This mode is typically used for annotations.
+        """Text size unit (screen vs local).
+
+        Returns
+        -------
+        screen_space : bool
+            If False, text size uses the unit of the local frame (e.g. cm).
+            Otherwise it is uses the logical screen's units (e.g. px).
```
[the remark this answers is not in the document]
FirefoxMetzger: I will add this one manually because the text block will have to be reformatted.

pygfx/helpers/_gizmo.py
FirefoxMetzger: So we keep things as-is, or should we make an issue on attempting to change this (and shed some code in the process)?

setup.py
FirefoxMetzger: @Korijn because we don't need it when building the docs (keeps build times low), and because it started to break the build on RTD.

pygfx/helpers/_gizmo.py
almarklein: Let's keep it as-is. Tried to remember a bit harder. The thing is that if you make it a child of the object, you must undo the scaling of the world object, and depending on the mode, also the rest of the transform. This would make the calculations unnecessarily complex, defeating the purpose of using the parent-child relationship for selecting the object to control.

FirefoxMetzger:
🥵 okay. A little bigger than expected, but we are getting there. I went over the entire library and updated all the public docstrings to numpydoc format. I've also added the ones that were missing (which were a surprising few) and made sure that parameters (for functions or constructors) are documented.

I didn't update the docstrings for `pygfx.camera`, `pygfx.controllers`, and `pygfx.linalg`, because the former two are due for a larger refactor (iirc, @almarklein?) and the later is scheduled for deprecation.

What is missing now is to (1) find a way to get autosummary to generate a link of the form `pygfx.WorldObject` that actually links to `pygfx.objects.WorldObject`, (2) to create a long (ctrl+F -able) list of items in the public API, and (3) write a class template that hides attributes and methods from parent classes to declutter the docs a bit.

Another point (that I wasn't quite clear on) is if we should also document internal classes in the generated documentation. These would live as part of the respective sub-package docs, but won't be listed in the public API. If so, we may wish to do this in a follow-up PR though because this one is getting a bit large.

almarklein:
image

🤯 

> because the former two are due for a larger refactor (iirc, @almarklein?) 

Yes, don't waste your time on these :)

> Another point (that I wasn't quite clear on) is if we should also document internal classes in the generated documentation.

I think it's enough if code for internal use is documented simply by having docstring that can be read in the source code. At least for the time being.

docs/reference.rst
```diff
@@ -50,119 +72,9 @@ Camera objects
     :member-order: bysource
```
Korijn, line 72: I guess you will move the Camera objects to pygfx.cameras subpage?

pygfx/pkg_resources/__init__.py
```diff
@@ -0,0 +1 @@
+# @almarklein: why do we need this init file here?
```
Korijn, line 1: It relates to the packaging setup. This folder needs to be recognized as a python module. We load the binaries from this folder via package_resources, and it resolves paths by module names.

setup.py
Korijn: OK.

pygfx/helpers/_box.py
Korijn: The qualitative difference between geometry and helper is that helpers are WorldObjects and geometries are routines that produce geometry which can be attached to WorldObjects.

In this specific case, if anything, you could extract the parts of code that generate the box's geometry to a function in the geometry subpackage. But the box helper will remain.

pygfx/helpers/_box.py
Korijn: As mentioned in the other open conversation on _box.py, there is an inline geometry routine in the constructor here. That routine is parameterized so that users can generate whatever geometry they need. I guess it would make sense actually to also add subdivision parameters, since users may have UV mapping use cases for which they require more triangles on each side of the cube. But those are all concerns for a box geometry generation routine.

For the helper object, it doesn't matter too much, since indeed, you also have the standard scale/rotation/position transform options at your disposal. I ported these from three.js for the most part, and I assume the arguments are just there for one-liner visual debugging convenience. I would prefer the size arg to stay for this reason.

pygfx/helpers/_gizmo.py
Korijn: The event system is modelled after the browser dom and is pretty complex, but also very flexible and powerful :) it's one of pygfx' killer features, I would say. I am not familiar with anything this easy to use in other graphics libraries for python.

Korijn:
Nice work!!

> What is missing now is to (1) find a way to get autosummary to generate a link of the form `pygfx.WorldObject` that actually links to `pygfx.objects.WorldObject`, (2) to create a long (ctrl+F -able) list of items in the public API, and (3) write a class template that hides attributes and methods from parent classes to declutter the docs a bit.

In all three cases, if it ends up being costly (in terms of time and complexity), I am fine with the current output. Regarding (2):

There is a search bar on the documentation site and it works well. So it's not impossible to search for things. I don't think this is super important.

> Another point (that I wasn't quite clear on) is if we should also document internal classes in the generated documentation. These would live as part of the respective sub-package docs, but won't be listed in the public API. If so, we may wish to do this in a follow-up PR though because this one is getting a bit large.

I agree with Almar's reply to this.

docs/reference.rst
FirefoxMetzger: I looked into this over the weekend and there are two options:

1. We can hide the module in the autosummary table. This would turn `pygfx.utils.color.Color` into `Color`. Autosummary supports this out-of-the-box, and this is how I solved it for now.
2. We can add a small custom sphinx extension that extends autosummary by adding a custom prefix to the name in the table, i.e., it would turn `pygfx.utils.color.Color` into `pygfx.Color`.

In both cases the result would link to the same page that we use in the sub-package docs, so either solution works. Do you have any preference @almarklein ?

pygfx/helpers/_box.py
FirefoxMetzger: > Yes, we have that! It's the MeshMaterial.wireframe property. But you'll see wires for all the edges of the triangle, so it serves a different purpose.

Ah right, then it isn't equivalent to drawing a wireframe box as we do here.

> WorldObjects typically used for visual debugging, while the geometry package holds routines that produce geometry which can be attached to WorldObjects

Ah, that's the piece I was missing. It will be good to have this info in the docs of the helpers module, because it makes it clear why the box helper is relevant here. I'll go and update the docstring.

pygfx/helpers/_axes.py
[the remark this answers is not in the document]
FirefoxMetzger: Alright, then I will aim to write the docstrings from the classes/objects perspective.

pygfx/helpers/_box.py
FirefoxMetzger: @Korijn  Is it because of the name, or because of the ability to use `size` and `scale` independently? If the former, we could look into making `size` an alias of `scale` which should make our code a little less complex while maintaining the current behavior.

pygfx/helpers/_gizmo.py
FirefoxMetzger: > Events are only dispatched to world objects if they are clicked on (or hovered over, etc.).

This means the object's geometry right? Like, we cast a ray through the cursor's pixel position and the first geometry it hits is the one that generates events? In other words, objects without geometries won't receive events?

> And the event propagates to parent world objects, similar to how it does in the browser dom.

Oh, that's a really cool way to isolate events. 

Since it uses the parent/child relationship it does make me wonder though what this means for the new objects API and our plans of moving the parent/child relationship into `AffineTransform`. Our argument for this was that we only use the scene graph for transforms and that chaining should be testable in pylinalg. Since we are using it for events, too, does this change anything?

docs/reference.rst
FirefoxMetzger: @Korijn  Yes that is the plan. I didn't touch them yet because we said they are due for a refactor, but I will add the plumbing  to keep the current classes documented.

pygfx/pkg_resources/__init__.py
FirefoxMetzger: Ah right. There is some logic in `pygfx.utils._dirs` that relies on the folder being a module to access resources. Thanks for clarifying

FirefoxMetzger:
Okay, I resolved (2) and (3), and found a solution for (1) (see https://github.com/pygfx/pygfx/pull/430#issuecomment-1387739350  for details on what the individual points mean). Once we have resolved the current discussions this PR is RTR.

pygfx/helpers/_box.py
Korijn: I'm assuming there's no specific conscious design decision behind this.

docs/reference.rst
Korijn: I think that was about Controllers, not Cameras, actually. :)

pygfx/helpers/_gizmo.py
Korijn: > This means the object's geometry right? Like, we cast a ray through the cursor's pixel position and the first geometry it hits is the one that generates events? In other words, objects without geometries won't receive events?

Actually, no, all shaders render both a color pixel to a buffer, and picking info to a secondary buffer. We query that secondary buffer for picking information. It includes different things depending on the material used to render. For example, here is the picking implementation for Mesh:

* https://github.com/pygfx/pygfx/blob/4aafa96c1f7ac21d6d2b1a177a6a44dcb5779daa/pygfx/materials/_mesh.py#L39
* https://github.com/pygfx/pygfx/blob/4aafa96c1f7ac21d6d2b1a177a6a44dcb5779daa/pygfx/renderers/wgpu/meshshader.py#L396-L405

> Since it uses the parent/child relationship it does make me wonder though what this means for the new objects API and our plans of moving the parent/child relationship into `AffineTransform`. Our argument for this was that we only use the scene graph for transforms and that chaining should be testable in pylinalg. Since we are using it for events, too, does this change anything?

We'll need to be creative and find a way to satisfy all our design goals. Must haves for me are still the same:

* chaining should be testable in pylinalg
* parent/child relationships should be user friendly, traversable and queryable in pygfx

Korijn:
I'll be out on holiday between 26 jan and 7 feb, so I'll have limited access over my phone to approve PRs.

docs/reference.rst
almarklein: Let's use the builtin option and use `Color`.

docs/reference.rst
almarklein: Both, to some extend, probably ...

pygfx/resources/_texture.py
```diff
@@ -275,27 +285,45 @@ def format_from_memoryview(mem, size):
...
+        the underlying texture. This must be a pygfx format specifier, e.g.
+        "3xf4", but can also be a format specific to the render backend if
+        necessary (e.g. from ``wgpu.VertexFormat``).
+    view_dim : str
+        The dimensionality of the array (1, 2 or 3). If None, use the texture's
+        dimension. Or e.g. get a "2d" slice view from a 3d texture, or e.g.
+        "cube" or "2d-array".
+    aspect : str
+        Omit or pass None to use the default.
+        @almarklein: What does this do?
```
almarklein, line 320: It's a (rather low level) wgpu thing. I don't think we use it somewhere, but for now it's not in the way either...
```suggestion
    aspect : Enum
        The `wgpu.TextureAspect` for this view. Omit or pass None to use the default.
```

pygfx/utils/__init__.py
```diff
@@ -1,3 +1,21 @@
...
+    :toctree: utils/
+    :template: ../_templates/clean_class.rst
+
+    Color
+    load.load_scene
+    show.show
+    show.Display
+    viewport.Viewport
+    text.font_manager
+    cm
```
almarklein, line 14: In the docs there is also a `functions` section?  https://pygfx--430.org.readthedocs.build/en/430/_autosummary/pygfx.utils.html

almarklein: Also the docs on e.g. `cm` speak of an alias: https://pygfx--430.org.readthedocs.build/en/430/_autosummary/utils/pygfx.utils.cm.html#pygfx.utils.cm

pygfx/utils/color.py
```diff
@@ -33,19 +33,19 @@ def _float_from_css_value(v, i):
...
 
     * `Color(r, g, b, a)` providing rgba values.
     * `Color(r, g, b)` providing rgb, alpha is 1.
     * `Color(gray, a)` grayscale intensity and alpha.
     * `Color(gray)` grayscale intensity.
 
-    The above variations can also be supplied as a single tuple/list,
-    or anything that :
+    The above variations can also be supplied as a single tuple/list, or
+    anything that:
```
almarklein, line 48: The bullet lists in this docstring don't become bullets in the rendered docs. Must they be indented?

pygfx/materials/_line.py
```diff
@@ -101,26 +124,41 @@ def map(self, map):
...
+
+    A simple line, drawn with line_strip primitives that has a thickness
     of one physical pixel (the thickness property is ignored).
 
     You probably want to avoid this material, because its width is
     inconsistent and looks *very* thin on HiDPI monitors.
     """
 
+    # @almarklein: The docs advise to avoid this. Depricate before 1.0?
+
```
almarklein, line 137: It may still be useful because the thin lines are more performant, or for e.g. debugging, so I'm reluctant to remove them.

pygfx/materials/_line.py
```diff
@@ -101,26 +124,41 @@ def map(self, map):
...
 
 
 class LineArrowMaterial(LineSegmentMaterial):
-    """A material that renders line segments that look like little vectors."""
+    """Arrow (vector) line material.
+
+    A material that renders line segments that look like little vectors."""
+
+    # @almarklein: This is a thin (1:1) wrapper around LineMaterial. Out of
+    # curiosity, how does it do more than the base class?
```
almarklein, line 164: Because the shader associated with this material does something different.

pygfx/objects/_events.py
```diff
@@ -118,13 +137,63 @@ def _update_current_target(self, target):
...
+    ----------
+    args : Any
+        Positional arguments are forwarded to the :class:`base class
+        <pygfx.objects.Event>`.
+    x : int
+        The x position of the cursor or touch in screen space (px).
+    y : int
+        Thy y position of the cursor or touch in screen space (px).
+    button : int
+        @almarklein: where can I find the description of this attribute?
```
almarklein, line 176: This is the spec that we follow for values like these: https://jupyter-rfb.readthedocs.io/en/latest/events.html

pygfx/objects/_events.py
```diff
@@ -176,13 +245,49 @@ def copy(self, **kwargs):
...
+    ----------
+    args : Any
+        Positional arguments are forwarded to the :class:`base class
+        <pygfx.objects.Event>`.
+    width : int
+        The new width of the application window in screen space (px).
+    height : int
+        The new height of the application window in screen space (px).
+    pixel_ratio : float
+        The new pixel ratio. (@almarklein: is this aspect ratio or dpi or something else?)
```
almarklein, line 284:
```suggestion
        The new pixel ratio (ratio between logical pixels and physical pixels).
```

almarklein (commented):
I left a ton of comments. I think I got all the `@almarklein` refs 😄

pygfx/controllers/__init__.py
```diff
@@ -1,8 +1,17 @@
...
+
 Controllers define how cameras can be interacted with. They are
 independent of any GUI toolkits.
+
+.. autosummary::
+
+    Controller
+    PanZoomController
+    OrbitController
+    OrbitOrthoController
```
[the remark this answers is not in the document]
FirefoxMetzger: good catch!

pygfx/geometries/_base.py
```diff
@@ -6,36 +6,19 @@
 
 
 class Geometry(Trackable):
-    """A Geomerty object is a container for geometry data of a WorldObject.
+    """Base class for Geometry Subtypes.
```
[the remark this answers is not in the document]
FirefoxMetzger:
```suggestion
    """Generic class for Geometry Subtypes.
```

@almarklein What about this instead?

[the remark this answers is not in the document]
almarklein: Yes, that sounds good! 👍

pygfx/materials/_line.py
FirefoxMetzger: No problem. In that case, I'd say we just add that info to the docs so it is clear why it exists.

pygfx/materials/_line.py
FirefoxMetzger: Ah! so all the classes here that have an empty body have a shader in `pygfx.renderers.wgpu` that does something special when it encounters this class? I think this is something we should document in the narrative docs.

pygfx/objects/_more.py
```diff
@@ -4,21 +4,73 @@
...
+        The data defining the appearence of the object.
+    visible : bool
+        Whether the object is visible.
+    render_order : int
+        The render order (when applicable for the renderer's blend mode).
+    render_mask : str
+        Determines the render passes that the object is rendered in. It's
+        recommended to let the renderer decide, using "auto".
+    position : Vector
+        The position of the object in the world. Default (0, 0, 0).
```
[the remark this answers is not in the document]
FirefoxMetzger: Mmh. In this case, we might also want to have a stub constructor that hides the arguments we remove here. Otherwise, they will still show up in an IDE and be suggested for auto-completion which may be confusing to newcomers.

pygfx/objects/_more.py
```diff
@@ -28,51 +80,125 @@ def __init__(self, geometry=None, material=None, **kwargs):
...
-    ``renderer.get_pick_info()``) will for most materials include
-    ``vertex_index`` (int) and ``segment_coord`` (float, sub-segment coordinate).
     """
 
 
 class Points(WorldObject):
-    """An object consisting of points represented by vertices (3D positions).
+    """A point cloud.
+
+    An object consisting of points represented by vertices (3D positions).
```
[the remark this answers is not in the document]
FirefoxMetzger: I think this is true for any object that is rendered. Ultimately, it is just a set of vertices with some extra data defined on them x)

pygfx/resources/_buffer.py
```diff
@@ -6,27 +6,37 @@
 
 
 class Resource(Trackable):
+    """Resource base class."""
+
```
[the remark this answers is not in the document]
FirefoxMetzger: Yes, we can exclude the object from backreferencing, meaning that no objects will be shown. There might also be a way to only show a select number. I can investigate.

pygfx/resources/_buffer.py
```diff
@@ -6,27 +6,37 @@
...
-            (e.g. from ``wgpu.VertexFormat``).
+    """A contigous piece of GPU memory.
+
+    Buffers can be used as index buffer, vertex buffer, uniform buffer, or
+    storage buffer. You can provide (and update data for it), or use it as a
+    placeholder for a buffer with no representation on the CPU.
+
+    @almarklein: does this mean we differentiate between paged, pinned, and
+    device (GPU) memory?
+
```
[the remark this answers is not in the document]
FirefoxMetzger: Okay, so we don't differentiate explicitly in pygfx? I'm asking because this is a very big topic for deep learning and working with CUDA and I would have expected something similar for rendering.

If pygfx doesn't deal with this concept internally (or expose it in the Buffer class) then we can just leave it for now.

pygfx/utils/__init__.py
FirefoxMetzger: > In the docs there is also a functions section?

True, looks like the default template adds a block for functions defined on a module. I'll have a look and remove them.

> Also the docs on e.g. cm speak of an alias:

Good catch. this should be a neatly rendered page of a sub-module.

pygfx/geometries/__init__.py
```diff
@@ -1,3 +1,70 @@
...
+    :toctree: geometry/
+    :template: ../_templates/clean_class.rst
+
+    trimesh_geometry
+    octahedron_geometry
+    icosahedron_geometry
+    dodecahedron_geometry
+    tetrahedron_geometry
+    torus_knot_geometry
+    klein_bottle_geometry
```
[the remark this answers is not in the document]
FirefoxMetzger: Nice catch. I thought that autosummary would be smart enough to use the function template for functions and class template for classes. However, it looks like you always use the custom template if you specify a custom template.

I've updated it to treat functions correctly

pygfx/materials/__init__.py
```diff
@@ -1,3 +1,52 @@
...
+.. autosummary::
+    :toctree: materials/
+    :template: ../_templates/clean_class.rst
+
+    Material
+    pillow_image
+    trimesh_material
+
+    # @almarklein the descriptions for the materials are sparse, but I lack
+    # context to come up with something more precise. Could you make a suggestion?
```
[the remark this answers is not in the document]
FirefoxMetzger: applied.

pygfx/materials/_mesh.py
```diff
@@ -6,10 +6,53 @@
...
+    The direction of a face is determined using Counter-clockwise (CCW) winding;
+    i.e., if the fingers of your curled hand match the direction in which the
+    face's vertices are defined then your thumb points into the "FRONT"
+    direction of the face. If this is not the case for your mesh, adjust its
+    geometry (using e.g. ``np.fliplr()`` on ``geometry.indices``).
+
     """
 
+    # @almarklein: Could we be more precise on what "physical space" means for
+    # the map's color format?
```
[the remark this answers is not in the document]
FirefoxMetzger: I'll add a link in the docs.

pygfx/objects/_events.py
FirefoxMetzger: I'll add the link to the docstring.

pygfx/objects/_more.py
```diff
@@ -4,21 +4,73 @@
...
+    direct or indirect children/nested objects.
+
     """
 
 
 class Background(WorldObject):
-    """An object representing a scene background.
+    """The scene's background.
+
+    @almarklein: how come this is an independent object and not part of Scene?
```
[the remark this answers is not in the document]
FirefoxMetzger: okay. If this is already tracked then let's not make another issue for it.

pygfx/objects/_more.py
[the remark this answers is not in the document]
FirefoxMetzger:
```suggestion
    """An object representing a disconnected cluster of individual points.

```

Maybe like this?

pygfx/resources/_buffer.py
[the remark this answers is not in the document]
FirefoxMetzger: Okay, I disabled indirect backreferences for `pygfx.Resource`. Now, only examples that link to it during the narrative comments of the example will be included.

pygfx/utils/__init__.py
FirefoxMetzger: fixed.

pygfx/utils/color.py
FirefoxMetzger: Yep, they should be :)

pygfx/objects/_more.py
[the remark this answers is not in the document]
FirefoxMetzger: > Is there a guideline on how short/long the "title" should be?

Yes there is. They should be a one-liner and not use the function name ([ref](https://numpydoc.readthedocs.io/en/latest/format.html#short-summary)). The main thing is to have something snappy that looks good when looking through the table in the docs. Otherwise we can do whatever we think is meaningful.

> I think some are so short they barely add anything over the class name.

Yes. sometimes its hard to say something useful beyond a descriptive class name. (guess it means you chose good variable names :D). It's just odd to not have a short-description because then there is a table entry that is randomly empty, which looks like we forgot something.

pygfx/resources/_texture.py
```diff
@@ -4,27 +4,37 @@
...
+            The data is copied if it's not float32 or not contiguous.
+        dim : int
+            The dimensionality of the array (1, 2 or 3).
+        size : tuple, [3]
+            The extent ``(width, height, depth)`` of the array. If None, it is
+            derived from dim and the shape of the data. By creating a 2D array
+            with ``depth > 1``, a view can be created with format 'd2_array' or
+            'cube'.
+            @almarklein: I don't understand this behavior. Can you elaborate so
+            I can come up with a better docstring?
```
FirefoxMetzger, line 25: @almarklein This is the final discussion point. I'm not exactly sure what the second sentence (about the 2D and cube array means).

pygfx/objects/_more.py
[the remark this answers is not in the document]
Korijn: > Maybe like this?

I'd say the original was more understandable for a broader audience.

pygfx/objects/_more.py
```diff
@@ -4,21 +4,73 @@
...
+
     Can be e.g. a gradient, a static image or a skybox.
+
+    Parameters
+    ----------
+    geometry : Geometry
+        The geometry to use when rendering the background.
```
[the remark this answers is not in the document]
FirefoxMetzger: Seeing this in the code now, it looks like we allow `geometry` for backwards compatibility:

https://github.com/pygfx/pygfx/blob/774a983c0133f458e502b1eef34398f1e8174b93/pygfx/objects/_more.py#L24-L27

Is this safe to remove?

pygfx/objects/_more.py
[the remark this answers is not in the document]
FirefoxMetzger: @Korijn The original as in the version that is currently in the PR, or the version that Almar suggested?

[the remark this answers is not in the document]
Korijn: > the version that is currently in the PR

:)

pygfx/materials/_line.py
almarklein: This is how it works for all materials, yes - the shader is selected based on the class of the material, and a shader may also make choices based on the subclass and/or their properties.

pygfx/objects/_more.py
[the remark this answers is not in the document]
almarklein: 👍

pygfx/resources/_buffer.py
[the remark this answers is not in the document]
almarklein: There is no way to specify these things in wgpu. We can only select whether it's a uniform, vertex, read-only-storage, regular storage. And we can specify where/how the buffer is used.

pygfx/objects/_more.py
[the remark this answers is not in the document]
almarklein: I'm leaning towards forcing `Background(None, material)`, because all world objects specify geometry and material in that order. Let's not spend to much time on this though, since the background is a bit up in the air anyway.

docs/conf.py
```diff
@@ -85,6 +56,7 @@ def _check_objects_are_documented():
     "examples_dirs": "../examples",
     "gallery_dirs": "_gallery",
     "backreferences_dir": "_gallery/backreferences",
+    "exclude_implicit_doc": {r"Resource"},
```
almarklein, line 59: Exclude some more, e.g. base classes.
```suggestion
    "exclude_implicit_doc": {"WgpuRenderer", "Resource", "WorldObject", "Geometry", "Material", "Controller", "Camera", "show", "Display", "Group", "Scene", "Light"},
```

pygfx/objects/_more.py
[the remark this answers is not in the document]
FirefoxMetzger: okay. in this case I will rephrase the docstring instead of removing it.

pygfx/resources/_texture.py
FirefoxMetzger: We are down to one unresolved conversation. Once this one is cleared the PR is ready :)

almarklein, line 25:
```suggestion
        size : tuple, [3]
            The extent ``(width, height, depth)`` of the array. If None, it is
            derived from `dim` and the shape of the data. The texture can also
            represent a stack of images by setting `dim=2` and `depth > 1`.
            Any derived texture views must then have a `view_dim` of  e.g. 'd2_array' or 'cube'.
```

almarklein (commented):
Almost done!

pygfx/resources/_texture.py
FirefoxMetzger: 👍

FirefoxMetzger:
> Almost done!

@almarklein Done? 😅 

(This PR is getting too long for GH. My browser is becoming slow when I search for open discussions :D)

almarklein (approved):
Done!

Korijn (approved):

Korijn:
Amazing work!

PR #432, opened 2023-01-12
Test if CI works with mainline lavapipe

Korijn:
* Mostly to see if this could save significant amount of time spent in apt-get on CI builds
* Takes away a dependency on a 3rd party PPA
* Also switched PySide6 to PySide6-Essentials which excludes all the add-on modules, see here: https://doc.qt.io/qt-6/qtmodules.html

This seems to scrape off another 15-25 seconds in total. Personally I think it's worth it. If you agree @almarklein I will update the one screenshot and merge.


almarklein:
Great, I agree!

almarklein (approved):

PR #441, opened 2023-01-24
Fix pathing in sponza example

FirefoxMetzger:
We have an example rendering the sponza scene (nice 🚀). Unfortunately, though it currently breaks because trimesh isn't happy to receive a `pathlib.Path`. This PR fixes that :)

https://user-images.githubusercontent.com/4402489/214269724-2da96bb9-1339-41f1-88c6-4028171c021a.mp4

(How come our performance while panning breaks down so much?)

### DIFF
--- a/examples/other/sponza_scene.py
+++ b/examples/other/sponza_scene.py
@@ -15,7 +15,7 @@ import pygfx as gfx
 # run the following command in the root of this repo
 # git clone https://github.com/KhronosGroup/glTF-Sample-Models ../glTF-Sample-Models
 # so that the paths will be set up correctly
-repo_root = Path(__file__).parent.parent
+repo_root = Path(__file__).parents[1]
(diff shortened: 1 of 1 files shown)


almarklein (approved):

almarklein:
Curious that this wasn't caught earlier. Could it be that this depends on the version of trimesh? Anyway, giving it an "old school" path seems safer anyway.

FirefoxMetzger:
I think it's a windows/linux issue.

Korijn:
It worked on my machine :)

Anyway, performance is just poor for thus scene in general, not just while panning. It's visible when panning, but the FPS is just as low when you aren't panning.

The problem is that there are four shadow casting point lights in this scene, which means the scene is rendered 1+ 4*6 = 25 times for each frame. (A point light shadow map is a cube map, or six faces)

Korijn (approved):

PR #443, opened 2023-01-25
Update to latest wgpu

almarklein:
These are the changes necessary to update to the upcoming wgpu.

Still need to actually release wgpu-native and wgpy-py, and pin them in pygfx.


Korijn (approved):
Other than the reserved names in WGSL this doesn't seem too bad at all! :)

almarklein:
Indeed! Though it took me a while to figure out what was needed for the depth-stencil issue ...

PR #448, opened 2023-02-06
PyGfx -> pygfx

### LINKED ISSUE #437: Refer to PyGfx in a consistent way
We seem to be using "pygfx" and "PyGfx", and maybe even "PyGFX"? Let's use a single way of writing in our readme, docs and comments.

My Personal preference is "PyGfx".

almarklein:
Closes #437

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -22,7 +22,7 @@ or install the bleeding edge from Github:
 What is pygfx?
 --------------
 
-PyGfx is a render engine. It renders objects that are organized in a scene, and
+Pygfx is a render engine. It renders objects that are organized in a scene, and
 provides a way to define what the appearance of these objects should be.
 For the actual renderering, multiple render backends are available, but the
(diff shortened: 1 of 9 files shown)


docs/guide.rst
```diff
@@ -22,7 +22,7 @@ or install the bleeding edge from Github:
 What is pygfx?
 --------------
 
-PyGfx is a render engine. It renders objects that are organized in a scene, and
+Pygfx is a render engine. It renders objects that are organized in a scene, and
```
ivoflipse, line 25: I think we used `pygfx` at the start of the sentence, because its also the name of the package. You can't import `Pygfx`

I don't have a strong opinion, but it would be nice to be consistent

almarklein: To me, starting a sentence without a capital feels weirder than capitalizing "pygfx" in these cases.

Korijn (approved):

PR #449, opened 2023-02-13
Remove some internal objects from the global pygfx namespace

### LINKED ISSUE #438: Move some objects out of the root namespace
In #430 we can see some objects being documented that are really meant for internal usage. These should not be present in the root namespace.

* `id_provider`
* `RenderFunctionRegistry`

And then there are objects that power-users need when creating custom shaders. These objects should be in a sensible subpackage (probably `pygfx.renderers.wgpu`). Perhaps we can (at some point) have extended docs specific for these users.

* `register_wgpu_render_function`
* `register_svg_render_function`

Specific cases:
* `pillow_image()` (defined in `pygfx.materials`): It converts a PIL image to a Texture. I propose to rename to `texture_from_pillow_image` and move to the utils subpackage. I don't think it needs to be present in the global namespace.

almarklein:
Closes #438


Korijn (approved):

Korijn:
If you're going to rename pillow_image, then the trimesh_* functions should follow suit, I think they belong together (serving the purpose of providing compatibility with other libraries) and should be named consistently.

almarklein:
Good point. What I now did, is:
* rename trimesh_geometry -> geometry_from_trimesh
* rename trimesh_material -> material_from_trimesh
* rename pillow_image -> texture_from_pillow, but also made it local to the module that needs it (not a public function anymore).

BTW: I tried moving these stuff to a new  `utils.compat` module, but that caused import recursion problems. We'd need to clean up the utils subpackage first. But that's for another day.

Korijn (approved):

Issue #403, opened 2022-12-05
Improvements to camera, controller, and display


almarklein:
I was working on a text example with `show` and ran into several annoyances related to cameras. I figured I'd list these, along with existing related issues, and then do a proper camera/controller refactoring soon.

* When providing a camera to `Display`, it should use not update the view, because I already carefully set the initial view.
* The zoom for an `OrbitController` on a `OrthographicCamera` is broken, it should set the bounds when `Controller.zoom_changes_distance` is set.
* Make `PanzoomController` work with a perspective camera. 
* Use `@property`'s.
* #388. A start was made in [#356](https://github.com/pygfx/pygfx/pull/356), but it seems a bit stuck on some of the above issues.
* #113
* #231
* #300 (but also allow user to determine when to draw)
* Automatically update the camera ~(but also allow users to determine when it happens)~.

 Also a proposal: a controller is in most cases already directly associated with a specific camera because of `add_default_event_handlers`. My proposal is to make the wrapping explicit, and drop the need to call `update_camera` on each draw. I think this will simplify both internal and user code. Needs some though on the edge cases for controlling multiple camera's, but we'll figure that out.

----

Most work was done in #457. Here are the leftovers:

* [x] Trackball controller.
* [x] Fly controller.
* [x] #365
* [x] Make key-mouse-bindings configurable.
* [x] Better default bindings? (e.g. panning is not consistent).

Korijn:
As long as all the coupling happens one way, I'm OK with your proposed changes.

So that means, Camera should _not_ need to change to accomodate Controller requirements, but the other way around is fine.

Another way to put it: Controllers depend on the Camera API, but Cameras do not depend on the Controller API.

That's a crucial invariant to me for the architecture.

almarklein:
> Another way to put it: Controllers depend on the Camera API, but Cameras do not depend on the Controller API.

Completely agree; a camera knows nothing about a controller.

Korijn:
Some minor notes:

> * [Controller does not request draw on load_state #300](https://github.com/pygfx/pygfx/issues/300)

I'm fairly certain there will be users who will want to use controllers in their applications, _and_ maintain full control when a draw is requested. Of course it's the polar opposite of users who just want to throw a controller into a script and not worry about the details. But I think both use cases should be supported. Otherwise the first user group will end up having to reimplement controllers just so they can keep control over draw requests.

[...]

PR #457, opened 2023-02-15
Cameras and controller overhaul

### LINKED ISSUE #231: Update projection matrix in camera.show_object
In #229 a utility was added to snap a camera to a new position and orientation in order to guarantee a part of the scene graph is visible.

Improvements that can still be made:

- [ ] add option to reconfigure optimal projection matrix as well: fov, near and far planes (useful for gfx.show)

### LINKED ISSUE #300: Controller does not request draw on load_state
I overlooked that it would have been nice to add a request_draw call to Controller.load_state in #297

### LINKED ISSUE #388: Ability to scale camera-axii individually
It would be useful to have a way to automatically scale the camera to fill a world object on the canvas, for example if we plot a sine wave with small amplitudes it barely shows up unless the camera is scaled appropriately. Doing this automatically would be useful for dynamic scientific plots! I think I can add this functionality to `camera.show_object()` and `controller.show_object()`, but I couldn't figure out how to map from canvas coordinates to world coordinates w.r.t. the camera.

```python3
import numpy as np
import pygfx as gfx


canvas = WgpuCanvas()

scene = gfx.Scene()
xs = np.linspace(0, 30, 100)
ys = np.sin(xs) / 10
zs = np.ones(100)

positions = np.dstack([xs, ys, zs])[0].astype(np.float32)

line = gfx.Line(
    gfx.LineMaterial(thickness=2.0, color=(1, 1, 1, 1)),
)
scene.add(line)

controller = gfx.PanZoomController()
controller.add_default_event_handlers(renderer, camera)

controller.show_object(camera, line)

def animate():
    controller.update_camera(camera)


if __name__ == "__main__":
    canvas.request_draw(animate)
    run()
```

image

if you set `camear.scale.y = 0.1`:

image

almarklein:
Addresses #403

## Tasks

Contributions outside of cameras and controllers:

* [x] Introduced `WorldObject._look_at` which has a flag to indicate whether it is a light or camera.
* [x] Fixed what's forward for cameras, but **not** for lights. 
* [x] Fixed a bug in the calculation of the bounding box for grid (texture) geometry.
* [x] The `show()` func considers a given camera as already having a good initial view. Plus a few small tweaks in `show()`.

To do:

* [x] Gave camera's and controllers proper properties.
* [x] Make combination of controller and camera Just Work.
* [x] Make controllers stateless; they simply work with the state of the camera.
* [x] Controllers can control multiple cameras at the same time.
* [x] Proper support for plotting (no fixed aspect). Closes #388.
* [x] Closes #231, because we auto-set clip planes based on extent.
* [x] Also closes #300, but only because controller no longer has state :)
* [x] Make it possible to provide near and far (overloading the auto-selection).
* [x] Maybe pass `renderer` into controller's init?

Wrapping up:

* [x] Make sure the cameras are documented in the ref docs.
* [x] Write a piece in the guide on how to use the cameras.
* [x] Add validation example for camera `show` methods.
* [x] Update all examples.
* [x] Wait for #484

After this:

* Camera smoothing.
* More controllers.
* Make key-mouse-bindings configurable.
* Better default bindings? (e.g. panning is not consistent).

### Notion of extent

As I made the controllers work with both the ortho and perpective cameras, I introduced a notion of scale or scene size, which I called "extend". For the ortho camera this is simply the mean of `width` and `height`. For the perspective camera, I needed to introduce a new `extent` prop. It is set by `look_at`, `show_object` and the new `show_rect`. 

The extent is an indication of the size of the scene being viewed. It can be used to determine the distance to the scene (taking fov into account), and thus can be used by the orbit camera to know around what point to rotate. Further, it is used to automatically determine the near and far clipping plane to sensible values.

### Camera class hierarchy

Interestingly, the `extent` / `aspect` pair, and the `width` / `height` pair, are two ways to set the same thing! With that insight, the two cameras are kindof the same thing: an orthographic camera is simply a perspective camera with a fov of zero. 

So I made the `PerspectiveCamera` more generic by allowing `fov=0`, and the `OrthographicController` a thin wrapper that limits the fov to zero. With a perspective camera and an orbit controller, you can change fov by scrolling while holding alt/option. You can then lower it all the way to zero, smoothly transitioning from a perspective camera to an orthographic one 😄 

(At some point I introduced a `GenericCamera` from which both cameras inherited, but when I started to write the guide, that felt overly complex.)

### Controller class hierarchy

The orbit controller has the same method as the panzoom camera, and some more. And they work the same way. So I made the OrbitController inherit from the PanZoomController.

### Camera updates

This PR covers everything in #403, except that changes are applied to the camera automatically (and directly). That's a consequence of the fact that the controllers do not have state by themselves.


pygfx/controllers/_base.py
```diff
@@ -1,6 +1,7 @@
 from typing import Tuple, Union
 
-from ..linalg import Vector3
+import pylinalg.func as la
```
Korijn, line 3: You don't need the `.func` part :)

almarklein: Haha, thanks.

Related, are we going for `as la` or `as pla`? I started with the former, but the pylinalg tests use the latter.

Korijn: I guess `la` is preferable, just like it's `np` and not `npy`.

Korijn:
Is it a lot of work to also implement fly camera controls?

* W/S move forward/backward
* A/D strafe sideways
* Mouse movement controls viewing direction
* Optional:
  * Holding right mouse button and moving the mouse rolls camera about the viewing axis

It's much more convenient to navigate scenes like sponza with this kind of control scheme.

kushalkolar:
This is looking amazing!

What do you think about allowing the Controller to be "enabled" or "disabled". For the PanZoomController it seems like this might be possible by creating a property `enabled` which is checked in `handle_event()`, and then calling,  `pan_stop()`, `zoom_stop()` and `quickzoom_stop()`?

almarklein:
> What do you think about allowing the Controller to be "enabled" or "disabled"

I think that's a great idea (was planning for adding this already 😄 )

pygfx/controllers/_base.py
```diff
@@ -1,75 +1,115 @@
...
+    # Get vectors
+    screen_dist = 100
+    pos1 = la.vector_apply_matrix(
+        la.vector_apply_matrix((screen_dist, 0, center[2]), camera_projection_inverse),
+        camera_world,
+    )
+    pos2 = la.vector_apply_matrix(
+        la.vector_apply_matrix((0, screen_dist, center[2]), camera_projection_inverse),
+        camera_world,
+    )
```
Korijn, line 112: is `la.vector_unproject` not useful here?

pygfx/objects/_base.py
```diff
@@ -457,9 +471,23 @@ def update_matrix_world(
...
+        if isinstance(target, (tuple, list, np.ndarray)) and len(target) == 3:
+            target = Vector3(*target)
+        # Lower level look_at
         self.update_matrix_world(update_parents=True, update_children=False)
         self._v.set_from_matrix_position(self._matrix_world)
-        self._m.look_at(self._v, target, self.up)
+        if is_light_or_cam:
+            self._m.look_at(self._v, target, up)
+        else:
+            self._m.look_at(target, self._v, up)
```
almarklein, line 490: This should do the trick, right?

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+
+.. code-block:: python
+
+    controller = gfx.OrbitController(camera)
+    controller.add_default_event_handlers(renderer)
+
+
+There are currently two controllers: the :class:`~pygfx.controllers.PanZoomController`
+is for 2D content or in-plane visualization, and the :class:`~pygfx.controllers.OrbitController` is for 3D content.
+All controllers work with both perspective and orthographic cameras.
```
almarklein, line 306: This, and the fact that `show_object` etc show the exact same scene no matter the fov, is the main idea of this PR. Everything Just Works!

almarklein:
This is ready for review 🚀 

A few things remain:
* I want to add a validation example for show_object and show_rect, as these touch many of the fundamental mechanisms.
* I need to update almost *all* of the examples. Leaving that until the rest is settled.

Since we need to update most examples anyway ... a suggestion:
```py
# Instead of this
controller = gfx.OrbitController(camera)
controller.add_default_event_handlers(renderer)

# Allow this
controller = gfx.OrbitController(camera, renderer)
```

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
 data associated with these positions (normals, texcoords, colors, etc.).
 Multiple world objects may share a geometry.
 
 
-Materials
----------
+**Materials**
 
-Each world object also has a material. This material object defines the
+All visible world objects have a material. This material object defines the
```
Korijn, line 222: I would remove "visible", invisible objects can have materials too right?

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+    camera = gfx.PerspectiveCamera(50, 4/3)
+
+The first argument is the fov (field of view) in degrees. This is a value between 0 and 179,
+with typical values < 100.
+
+The second argument is the aspect ratio. A window on screen is usually a rectangle
+with 4/3 or 16/9 aspect. The aspect can be set so the contents better fit the window.
+
+When the fov is zero, the camera operates in orthographic mode, where there is
+no perspective. In this case it'd be good to also specify a 3d argument: the extent
```
Korijn, line 250:
```suggestion
no perspective. In this case it'd be good to also specify a 3rd argument: the extent
```

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+specifies the size of the thing being viewed. So if your scene is about 400 units in size:
+
+.. code-block:: python
+
+    camera = gfx.PerspectiveCamera(0, 1, 400)
```
Korijn, line 255: I don't understand. 400 is the size? In what dimension? Depth? Width? Diagonal?

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+
+
+**Orthographic camera**
+
+The second camera of interest is the :class:`~pygfx.cameras.OrthographicCamera`. It's similar
+to the perspective camera, except it's fov is always zero. It can be instantiated like this:
```
Korijn, line 261:
```suggestion
to the perspective camera, except its fov is always zero. It can be instantiated like this:
```

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+do not represent a physical dimension, e.g. for plotting data. The contents
+of the view are then stretched to fill the window.
+
+
+**Orienting cameras**
+
+Camera's can be oriented manually by setting their position, and then set their rotation
+to have them look in the correct direction. This last bit can be tricky, unless you use
+the :func:`.look_at()<pygfx.cameras.PerspectiveCamera.look_at>` method. This also sets
+the extend (based on what's visible at the target location), so you can now attach
```
Korijn, line 282:
```
the extent (based on what's visible at the target location), so you can now attach
```

I'm still not sure what "extent" means. Is it the near and far plane? Why does `look_at` handle that? The name implies it only affects orientation.

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+a controller and interact!
+
+Even easier is using :func:`.show_object()<pygfx.cameras.PerspectiveCamera.show_object>`.
+With this method you specify an object (e.g. the scene) and optionally a direction,
```
Korijn, line 286: It would make more sense for this method to adjust the near and far plane (assuming that's what "extent" means)

docs/guide.rst
```diff
@@ -196,31 +196,116 @@ the scene as desired::
...
+
+    camera = gfx.PerspectiveCamera(50, 4/3)
+
+The first argument is the fov (field of view) in degrees. This is a value between 0 and 179,
+with typical values < 100.
+
+The second argument is the aspect ratio. A window on screen is usually a rectangle
+with 4/3 or 16/9 aspect. The aspect can be set so the contents better fit the window.
+
+When the fov is zero, the camera operates in orthographic mode, where there is
```
Korijn, line 249: What is the idea behind supporting zero fov on a perspective camera to effectively change it to an orthographic camera? Is it more convenient than swapping out camera instances?

docs/guide.rst
almarklein: I meant visible as in having an apperance, so excluding lights, cameras, groups, etc. I think I'll change it to say "Most".

docs/guide.rst
almarklein: I tried to explain the reasoning in the top post. I should've mentioned that I added explanations there, sorry!

docs/guide.rst
almarklein: Width and height, depending on the aspect.

docs/guide.rst
almarklein: This extent thing clearly needs a better explanation 😄 

It's basically the equivalent of `width` and `height` (it's the mean of these). So it's a measure for the size of the scene that the camera is to cover.

Based on this value, and the fov, the controller knows how far away from the scene the camera is, and thus e.g. around what point it should rotate. 

Since it's a measure for the size of the scene, it's a nice basis for calculating the near and far clip planes. But it *does not* directly represent the clip planes in any way.

Back to `look_at`, it works exactly as usual, except it assumes that you made the camera look at the center of the scene, so now it knows the distance between camera and scene, and from that it can calculate the extent, so that controllers can just pick up the camera and work with it (and clip planes can be calculated that likely make sense for the current scene).

Korijn:
> ```python
> controller = gfx.OrbitController(camera, renderer)
> ```

I think it's okay and convenient, but it may also give the impression that the controller will keep a reference (which I think it shouldn't). I would say the current `add_default_event_handlers` method should also remain available.

---

I think the guide should also briefly explain how the near and far plane settings work, would you agree?

docs/guide.rst
Korijn: Ok, I get it now. That's pretty cool!

docs/guide.rst
Korijn: Can you give an example? How would I determine a proper value for extent, as a user?

docs/guide.rst
Korijn: I understand, I just worry that our readers may take it literally :)

docs/guide.rst
Korijn: I think users will want to use `look_at` in all kinds of scenarios to update the orientation of a camera, not just to center it on the scene, and will most commonly want to manage the extent themselves. Is this new behavior optional?

almarklein:
I think I might be able to factor `extent` out of the api, which might make it easier to digest.

docs/guide.rst
almarklein: Good point, I'll see if I can work this out.

almarklein:
> I think the guide should also briefly explain how the near and far plane settings work, would you agree?

Yes.

> [...] it may also give the impression that the controller will keep a referenc

Maybe something like:
```py
controller = gfx.OrbitController(camera, register_events=renderer_or_viewport)
```

docs/guide.rst
```diff
@@ -219,9 +219,16 @@ Multiple world objects may share a geometry.
...
-All visible world objects have a material. This material object defines the
-appearance of the object. Examples can be its color, how it behaves under lighting,
-what render-mode is applied, etc. Multiple world objects may share a material.
+All world objects that have an appearance, have a material that defines
+that apperance. (Objects that do not have an apperance are for example
+groups or cameras.) For each type of object there are typically a few
+different material classes, e.g. for meshes you have a
+``MeshBasicMaterial`` that is not affected by lights, a
+``MeshPhongMaterial`` that applies the Phong light model, and the
+``MeshStandardMaterial`` that implements the most realistic light model.
```
Korijn, line 228:
```suggestion
``MeshStandardMaterial`` that implements a physically-based light model.
```

almarklein:
Ok, so I managed to hide the extent stuff. Still works but is much simpler to explain.

> Can you give an example? How would I determine a proper value for extent, as a user?

So now you can just specify either the width or height in addition to the aspect. But the recommended approach is to use any of the `camera.show_xx` methods.

> I think users will want to use look_at in all kinds of scenarios to update the orientation of a camera, not just to center it on the scene, and will most commonly want to manage the extent themselves. Is this new behavior optional?

Good point. I left the `look_up` untouched, and introduced `show_pos` that's basically `look_at` but it also sets width and height. Now all methods that start with `show_` nicely prep the camera for use with a controller.

Remaining point is the registering of events. A proposal:
```
# Rename add_default_event_handlers -> register_events
controller = gfx.OrbitController(camera)
controller.register_events(renderer_or_viewport)

# And allow as a shortcut
controller = gfx.OrbitController(camera, register_events=renderer_or_viewport)
```

Korijn:
Proposal is :+1:

almarklein:
@Korijn I want to make sure that you approve the current state of the PR before I update the remaining examples (since that would make the PR harder to review, and I want to avoid duplicate work).

Korijn:
I think you did a great job here, after playing around with this and reviewing the code I have the following to say:

* OrbitController: traditionally mouse wheel zooming would be towards the center of orbit, not towards the mouse. Can you use a modifier for zooming to the mouse and make zooming towards the center of orbit the default? It's usually intentional the your camera is fixed on the center of orbit when you pick the orbit controller.
* I had to really look at the code to discover what features and modifiers are available; maybe add something to the guide/docs about what the controls are?

pygfx/cameras/_perspective.py
```diff
@@ -1,64 +1,467 @@
...
 
     def set_view_size(self, width, height):
         self._view_aspect = width / height
 
     def update_projection_matrix(self):
-        # Get the reference width / height
-        size = 2 * self.near * tan(pi / 180 * 0.5 * self.fov) / self.zoom
-        # Pre-apply the reference aspect ratio
-        width = size * self.aspect**0.5
-        height = size / self.aspect**0.5
```
almarklein, line 43: This logic was wrong. It's fixed below. This fix changes how the image is mapped to screen, depending on aspect and window size. This is why some validation examples need their screenshot updated.

almarklein:
All green 🚀

Korijn (approved):

PR #459, opened 2023-02-17
Make the root transform identity

almarklein:
See discussion in https://github.com/pygfx/pylinalg/pull/43

May need some discussion to align our ideas on what an object's transform really means :)  And the tests need updating.


almarklein:
Closing. This would not solve the actual problem.

It's an equally valid choice to include the root's transform. If we want to avoid potential confusion of this, it'd better to force the use of a `Scene` object that is not a `WorldObject`, or at least an object that cannot have a transform, but is merely a collection of objects. But this has downsides by itself.

PR #462, opened 2023-02-17
Tweak sponza lighting

### LINKED ISSUE #393: Shadow map bug
In the screenshot below you can see that shadow mapping works properly in the green circle, but incorrectly in the red circle. There should be a shadow on the portion of the pillar below the light source as it is occluded.

I think this is caused by the near plane for the shadow map cameras.

image

### LINKED ISSUE #453: Sponza Rendering needs a second look
xref: https://github.com/pygfx/pygfx/pull/442#issuecomment-1404951278

Even after adjusting the near plane when rendering the sponza scene, there still seems to be a glitch/bug in the lighting that needs solving:

image

I don't know what the cause is, but could it be that the model of the lamp holder is not closed at the bottom and as such light can slip through?

Korijn:
* Closes #393 
* Closes #453 

image

I couldn't really get rid of that little dot on the floor, I think it's actually a hole in the mesh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,5 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+glTF-Sample-Models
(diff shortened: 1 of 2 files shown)


almarklein (approved):
Nice!

PR #464, opened 2023-02-20
Pyinstaller hook

### LINKED ISSUE #454: PyInstaller hooks needed
When bundling pygfx with PyInstaller into an application, there are some data files and some dynamic libs missing.

We need to bundle:

* dynamic libs of `freetype` (collect_dynamic_libs)
* data from `pygfx.pkg_resources` (collect_data_files)

I noticed that when just using `collecting_data_files("pygfx.pkg_resources")` results in a name clash with `pkg_resources` module from setuptools. We might consider replacing the use of `pkg_resources` with `importlib.resources`?

Korijn:
Closes #454

Note: I renamed `pkg_resources` to `data_files` and not `resources` because that is already an existing subpackage


.github/workflows/ci.yml
```diff
@@ -87,6 +87,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
+    - name: Install package and dev dependencies
+      run: |
+        python -m pip install --upgrade pip
```
ivoflipse, line 101: Is the current pip outdated?

Korijn: I just copy pasted this, will remove it

Korijn: Actually, no, I'll keep it around. The other jobs use the command as well.

pygfx/__pyinstaller/test_pygfx.py
```diff
@@ -0,0 +1,6 @@
+def test_pyi_pygfx(pyi_builder):
+    pyi_builder.test_source(
+        """
+        import pygfx
```
almarklein, line 4:
```suggestion
        import pygfx as gfx
        ff = gfx.font_manager.select_font("foo", gfx.font_manager.default_font_props)[0][1]
        print(ff.codepoints)
        assert len(ff.codepoints) > 100
```

almarklein (approved):

Issue #460, opened 2023-02-17
Define the coordinate system


almarklein:
We should:
* Decide on a coordinate system
* Document it

### What do others use?

**pygfx currently** we seem to have a right-handed system with right +X, up +Y and forward -Z.

**glTF** uses a right-handed coordinate system. glTF defines +Y as up, +Z as forward, and -X as right; the front of a glTF asset faces +Z. The units for all linear distances are meters. All angles are in radians. Positive rotation is counterclockwise.

**OpenGL** by convention uses a right-handed system. What this basically says is that the positive x-axis is to your right, the positive y-axis is up and the positive z-axis is backwards.

**Blender** [uses](http://www.rab3d.com/tut_blen_started.php) the [right hand](http://en.wikipedia.org/wiki/Right-hand_rule) coordinate system with the Z axis pointing upwards. This is common with the coordinate systems used by most common 3D CAD packages.

From my understanding, CAD software like 3D studio max and blender historically have Z point upwards, whereas game and gfx related software has Y point upwards. Not sure why glTF swaps directions for X and Z ...

Korijn:
Can you clarify in each instance whether or not the coordinate system applies to model/view/clip space? Because the conventions are different between those spaces usually as well.

FirefoxMetzger:
> pygfx currently we seem to have a right-handed system with right +X, up +Y and forward -Z.

😆 sigh, so which is it now, does `forward` mean positive Z or negative Z? 😅

Korijn:
I found this nice picture: :)

![image](https://user-images.githubusercontent.com/1882046/219696037-28d75bd6-ffd8-4075-abd4-41fa69e07b36.png)

I would like to add that Three.js is Y-up, and right-handed.

Also, you can actually change the coordinate system in three like this:

`THREE.Object3D.DefaultUp = new THREE.Vector3(0,0,1);`

But that's all besides the point. We just need to pick something and document it. What do we want?

almarklein:
> 😆 sigh, so which is it now, does forward mean positive Z or negative Z? 😅

Sorry, I know! I was confused between NDC and model coords! 😆 

To quote Linkin Park: In the end, it doesn't even matter (what we choose). Users can always set the camera's up value, and/or flip the camera's scale to move from a left-handed to right-handed system. The main issue (as I understand it) is that when a user loads left-handed models into a right-handed system (or vise versa) the models will be inside-out. And that only is really a problem when people set the `material.side = 'front'`.

[...]

PR #466, opened 2023-02-22
Fix skybox

panxinmiao:
Our Skybox rendering has always had some problems in the past, details can be found in #353.

I noticed that we restated the coordinate system of pygfx in #460. This helps to make some of our logic clearer, 😄 

In this PR, I corrected the error based on the current Skybox rendering logic (different from the way that `Skybox` is regarded as a common world object in #353).

I left some instructions in the code. 

The examples provided in #353 ("[skybox_validate.py](https://github.com/pygfx/pygfx/pull/353/files#diff-f43428565bced2d45c5714e26f96ee161903967950dc4dfae632a424fe0ba384)" and "[validate_skybox_cubemap.py](https://github.com/pygfx/pygfx/pull/353/files#diff-35f79eff8d6d61c820555f9de7c9df25d54df1c36b52a8419a909df1f83f02c1)") can be used to verify the correctness of the results.


panxinmiao:
> Although the definitions state that the default up vector is +Y, we do want to make it possible to make e.g. up be +Z. This will likely be by changing the up vector of the camera. In a similar way, we should be able to rotate the skybox.

Er, I don't quite understand the specific meaning, 😅 

My understanding is that the up attribute is just a marker, which only specifies which dimension of the object is upward. When the up attribute of an object is changed, it will not produce actual rotation (changing up will not directly change the transform matrix of the object), nor will it change the attitude of the object directly. What really changes the attitude of the object is the "look_at" function, the up attribute works through the "look_at" function. 

If you want the Skybox can be affected by the rotation of the camera, this is already the case in our current implementation. The method of "ndc_to_world_pos" actually contains the rotation information of the camera.

If you want the sky box itself to have the rotation attribute (in theory, it should be so, the sky box should only be unable to translate freely relative to the camera, and other transformations should be free), it is not true now, because we do not regard skybox as a world object, nor use Skybox's transform matrix. The implementation in #353 is precisely because of this consideration. See: https://github.com/pygfx/pygfx/pull/353#discussion_r1063079728

almarklein:
> My understanding is that the up attribute is just a marker, which only specifies which dimension of the object is upward. 

Fair point, maybe it should be a different property, or indeed just use lookat. I think what's best depends to some extend on some moving parts in the cameras.

Korijn:
@almarklein I'll leave it up to you to approve this PR

panxinmiao:
I modified the code. Now Skybox supports arbitrary rotation instead of relying on the `up` property as the "direction" (now it consistent with the behavior of other world objects). 

In this case, I think maybe we don't need to deal with the `up` property of Skybox any more?

almarklein:
That's a nice solution!

Would be nice to have the examples from #353 as well. Could be in a separate PR though. Or we can merge this, and then you can refactor #353 to only add the examples. Whatever works best for you!

almarklein (approved):
Let's merge this! We can add the examples from https://github.com/pygfx/pygfx/pull/353 later.

examples/feature_demo/skybox_rotate.py
```diff
@@ -0,0 +1,57 @@
...
+camera = gfx.PerspectiveCamera(70)
+
+camera.position.set(0, 4, 20)
+
+controller = gfx.OrbitController(camera.position.clone())
+controller.add_default_event_handlers(renderer, camera)
+
+
+def animate():
+    controller.update_camera(camera)
```
almarklein, line 47:
```suggestion
```

PR #467, opened 2023-02-22
Fix alpha blending

### LINKED ISSUE #465: Blend modes are broken
Objects that are semi-transparent are (often?) not drawn.

Let's also add a validation example for the blend modes ...

almarklein:
Closes  #465

I introduced this regression during the update to the latest wgpu-py. I added a test to prevent that in the future.


almarklein:
Lol. Looks like our example testing has been disabled for a while. All tests (and generations) are skipped. Looks like it does not detect lavapipe correctly.

almarklein:
Lol, a few example screenshots are set back to what they were before #468.

PR #468, opened 2023-02-22
Add check to confirm lavapipe

almarklein:


almarklein:
Step 1 works:
```
=================================== FAILURES ===================================
_________________________ test_that_we_are_on_lavapipe _________________________

    def test_that_we_are_on_lavapipe():
        print(wgpu_backend)
        if os.getenv("PYGFX_EXPECT_LAVAPIPE"):
>           assert is_lavapipe
E           assert False
```

Now to fix the lavapipe detection.

almarklein:
One problem remains: certain text is not visible in the validation examples.

Korijn (approved):

Korijn:
You probably need to port this to wgpu-py repo as well

almarklein:
> One problem remains: certain text is not visible in the validation examples. It looks like with the Ubuntu-builtin LavaPipe, the text is thinner, to the extent that some text just vanishes.

It looks like these issues are related to alpha calculation/blending. There is now one test (the one for text anchor) that does not really test anything because of this. This should be temporary.

The example test build uses `ubuntu-latest`, so as soon as 22.04 becomes available, that one is used, which is likely to have a newer Lavapipe that hopefully has less such bugs.

almarklein:
> There is now one test (the one for text anchor) that does not really test anything because of this.

Try to fix it in #469

PR #469, opened 2023-02-23
Enable text anchor text again

almarklein:


almarklein:
Nevermind. The problem is fixed in #467 😅

PR #473, opened 2023-03-05
add Texture format for uint32

kushalkolar:
unsigned int32 uses big `"I"` whereas `int32` uses `"i"`

```python
import numpy as np
import pygfx

shape = 100, 100
size = shape[0] * shape[1]

a = np.random.randint(0, 255, size, dtype=np.int32).reshape(shape)
au = np.random.randint(0, 255, size, dtype=np.uint32).reshape(shape)

print(memoryview(a).format, memoryview(au).format)

# returns: i I
```

works for images

```python
tex = pygfx.Texture(au, dim=2).get_view()

geo = pygfx.Geometry(grid=tex)
mat = pygfx.ImageBasicMaterial(clim=(0, 255))

wo = pygfx.Image(geo, mat)

pygfx.show(wo)
```

image

### DIFF
--- a/pygfx/resources/_texture.py
+++ b/pygfx/resources/_texture.py
@@ -248,8 +248,7 @@ def format_from_memoryview(mem, size):
         "h": "i2",
         "H": "u2",
         "i": "i4",
-        "I": "i4",
-        "U": "u4",
+        "I": "u4",
         "e": "f2",
         "f": "f4",
     }
--- a/pygfx/resources/_texture.py
+++ b/pygfx/resources/_texture.py
@@ -248,6 +248,7 @@ def format_from_memoryview(mem, size):
         "h": "i2",
         "H": "u2",
         "i": "i4",
+        "I": "i4",
         "U": "u4",
         "e": "f2",
         "f": "f4",


pygfx/resources/_texture.py
```diff
@@ -248,6 +248,7 @@ def format_from_memoryview(mem, size):
         "h": "i2",
         "H": "u2",
         "i": "i4",
+        "I": "i4",
         "U": "u4",
```
almarklein, line 252: I think you're right. Memoryview uses the struct format characters, and there is not "U" there: https://docs.python.org/3/library/struct.html#format-characters

Probably a typo from my end.

But should it not be like this then?
```suggestion
        "I": "u4",
```

kushalkolar: ah, seems right, didn't see the u2 and u1. Is there a table of the format mappings specified by pygfx? 

related, what texture formats are actually supported, int8 and int16 seem to work but [`WGSL`](https://www.w3.org/TR/WGSL/#texture-builtin-functions) seems to only specify int32, uint32, and float32, but wgpu supports more? https://docs.rs/wgpu/latest/wgpu/enum.TextureFormat.html

Does uint8 for example get converted to uint32 before writing to the GPU?

kushalkolar:
related, what texture formats are actually supported, int8 and int16 seem to work but [WGSL](https://www.w3.org/TR/WGSL/#texture-builtin-functions) seems to only specify int32, uint32, and float32, but wgpu supports more? https://docs.rs/wgpu/latest/wgpu/enum.TextureFormat.html

Does uint8 for example get converted to uint32 before writing to the GPU?

almarklein:
> Is there a table of the format mappings specified by pygfx?

https://github.com/pygfx/pygfx/blob/0e12d34d384f5b18d30c2ff27638c4e259c07b51/pygfx/renderers/wgpu/_utils.py#L64-L81

> what texture formats are actually supported, [...] Does uint8 for example get converted to uint32 before writing to the GPU?

Indeed, in the shader, when you sample or load from a texture, you get either f32, i32 or u32.

almarklein (approved):
Sorry, this went off my radar, LGTM!

Korijn (approved):

PR #477, opened 2023-03-10
Fix shadow camera cull mode

### LINKED ISSUE #453: Sponza Rendering needs a second look

### LINKED ISSUE #476: Shadow glitch
Hello everyone, been playing with your examples. Remarkable code; god level commenting; thank you.
Looking at your 'Types of Cylinders' example [https://pygfx.readthedocs.io/en/latest/_gallery/feature_demo/cylinder.html#sphx-glr-gallery-feature-demo-cylinder-py](url), I'm convinced the shadows cast are incorrect. The full cylinder seems okay, but the other two objects appear to be missing shadows that ought to be cast by the faces closest the light source.
Don't know code well enough to contribute there I'm afraid; just trying to help :)

Korijn:
* Closes #476
* Closes #453 

Before:
image
image
image

After:
image
image
image

### DIFF
--- a/examples/other/sponza_scene.py
+++ b/examples/other/sponza_scene.py
@@ -58,8 +58,7 @@ sunlight = gfx.DirectionalLight()
 sunlight.position.set(-14.5, 31, 4.5)
 sunlight.target.position.set(5.3, -1.4, -2.5)
 sunlight.cast_shadow = True
-sunlight.shadow.camera.near = 0
-sunlight.shadow.camera.far = 50
+sunlight.shadow.camera.depth_range = (0, 250)
 sunlight.shadow.camera.update_projection_matrix()
 scene.add(sunlight)
 
@@ -73,8 +72,10 @@ for pos in [
     torch = gfx.PointLight("#ff7700", decay=2.5)
     torch.position.set(*pos)
     torch.cast_shadow = True
-    torch.shadow.camera.near = 0.01
+    torch.shadow.camera.depth_range = (0.01, 200)
     torch.shadow.camera.update_projection_matrix()
+    torch.shadow.cull_mode = "none"
+    torch.shadow.bias = 0.001
     # torch.add(gfx.PointLightHelper(size=0.01))
     scene.add(torch)
 


Korijn:
Converting to draft as @panxinmiao shared some new insights in #476

Korijn:
As per #467 my next move is to see if the cylinders example can be fixed purely through configuration.

Korijn:
@almarklein how would I go about making the cull mode a settable property that is properly hooked into the reactivity mechanics?

almarklein:
If you store the internal variable on `self._store` it is tracked. However, the tracking only happens for stuff drawn by the renderer's pipeline building logic. It looks like the shadow stuff builds its own pipeline as a sort of pre-processing step.

Korijn:
I've made `light.shadow.cull_mode` configurable and managed to resolve all the rendering artifacts just by setting `cull_mode` and `bias` to appropriate values for each example.

almarklein:
> `# TODO: move bias and cull_mode to Light so they can be reactive?`

That would not help. We could add the `Trackable` mechanics on this class, but in this particular case I think it would be easier to just set a flag in the setters and check that somewhere to see whether the shadow pipeline needs to be rebuild.

almarklein (approved):
Only comment is maybe include that link (https://learnopengl.com/Advanced-Lighting/Shadows/Shadow-Mapping) somewhere. Though we could also find it back on the issue or this PR.

Korijn:
> > `# TODO: move bias and cull_mode to Light so they can be reactive?`
> 
> That would not help. We could add the `Trackable` mechanics on this class, but in this particular case I think it would be easier to just set a flag in the setters and check that somewhere to see whether the shadow pipeline needs to be rebuild.

I've created #489 to track this.

> Only comment is maybe include that link (https://learnopengl.com/Advanced-Lighting/Shadows/Shadow-Mapping) somewhere. Though we could also find it back on the issue or this PR.

Well... I think we would do better to write a short guide that explains how to configure shadow mapping. But not for now.

PR #479, opened 2023-03-10
Generate class docs in normal Sphinx way

almarklein:
The current documentation for classes shows a list of methods and attributes, which is nice in the sense that they are separated, but it only shows the summary (first line) of their docstrings, and does not show method signatures. You can also not crossref to methods or properties.

This PR is a proposal to render documentation for classes in the common Sphinx way using:

```
.. autoclass:: lib.Foo
   :members:
   :show-inheritance:
   :member-order: bysource
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -2,6 +2,8 @@
 
 import os
 import sys
+import shutil
+
 from sphinx_gallery.sorting import ExplicitOrder
 import wgpu.gui.offscreen
 
@@ -46,6 +48,9 @@ extensions = [
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
+
+# Just let autosummary produce a new version each time
+shutil.rmtree(os.path.join(os.path.dirname(__file__), "_autosummary"), True)
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,6 @@ author = "Almar Klein, Korijn van Golen"
 # The full version, including alpha/beta/rc tags
 # release = '0.1.0'
 
-
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
@@ -104,6 +103,3 @@ sphinx_gallery_conf = {
 html_static_path = ["_static"]
 html_favicon = "_static/pygfx.ico"
 html_logo = "_static/pygfx.svg"
-
-if not (os.getenv("READTHEDOCS") or os.getenv("CI")):
-    html_theme = "sphinx_rtd_theme"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -99,3 +99,6 @@ sphinx_gallery_conf = {
 html_static_path = ["_static"]
 html_favicon = "_static/pygfx.ico"
 html_logo = "_static/pygfx.svg"
+
+if not (os.getenv("READTHEDOCS") or os.getenv("CI")):
+    html_theme = "sphinx_rtd_theme"
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -12,6 +12,11 @@ BUILDDIR      = _build
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+clean :
+	-rm -r _build
+	-rm -r _autosummary
+	-rm -r _gallery
+
 .PHONY: help Makefile
 
 # Catch-all target: route all unknown targets to Sphinx using the new
--- a/docs/_templates/custom_layout.rst
+++ b/docs/_templates/custom_layout.rst
@@ -8,52 +8,9 @@
 .. automodule:: {{ fullname }}
 {% else %}
 .. autoclass:: {{ objname }}
-   {% block methods %}
-   {% if methods %}
-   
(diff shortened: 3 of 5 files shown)


almarklein:
All green now.

----

For some reason, running the docs locally generate tons of warnings in two categories:
* `/Users/almar/dev/py/pygfx/docs/_gallery/introductory/plot_offscreen.rst:115: WARNING: download file not readable: /Users/almar/dev/py/pygfx/docs/_gallery/introductory/plot_offscreen.py`
* `/Users/almar/dev/py/pygfx/docs/_gallery/introductory/plot_cube.rst: WARNING: document isn't included in any toctree`

It's a mystery to me why this happens locally, but not on CI or RTD. It's also annoying if you have to search for that one warning that causes the CI doc build to fail.

Not a blocker for this PR, but if someone has an idea what's going on, that'd be great :)

FirefoxMetzger:
> The current documentation for classes shows a list of methods and attributes, which is nice in the sense that they are separated, but it only shows the summary (first line) of their docstrings, and does not show method signatures. You can also not crossref to methods or properties.

Yes, because most methods we have on classes don't have a docstring (beyond maybe a slug/first line), so the docs would become comparatively ugly without adding a lot of help so I chose to not build them in the documentation (beyond listing them in the method/attribute table).

We can easily add pages for each member function that can be crossref'ed or we can dump a list of methods/attributes at the end of a classes's doc page (what `autoclass` is doing when using `:members:`). I am a bit uncertain though as to what this addition buys us.

> This PR is a proposal to render documentation for classes in the common Sphinx way using:

At the time, we explicitly chose to exclude inherited members from the docs (hence the slightly lengthy template). It adds a lot of redundancy and occludes the "important" properties, i.e., the ones that are exclusive to an object. For example, it is much more interesting to know that a `SpotLight` has `SpotLight.decay` than to know that it has `SpotLight.position` (or `SpotLight.transform.position`) which is true for every WorldObject.

We can add docs for inherited properties if this is something that you value; however, I would probably do this in a separate section at the bottom, e.g., "Inherited Properties" and list all inherited methods and attributes there.

docs/conf.py
```diff
@@ -47,6 +49,9 @@ def _ignore_offscreen_run():
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
 
+# Just let autosummary produce a new version each time
+shutil.rmtree(os.path.join(os.path.dirname(__file__), "_autosummary"), True)
+
```
FirefoxMetzger, line 54: @almarklein Could you elaborate why we want to delete the cache each time we build?

Autosummary already has a mechanism of invalidating the generated files (it compares last modified timestamps of the source and the generated file), so from my perspective, it looks like this mostly adds build time.

FirefoxMetzger:
> /Users/almar/dev/py/pygfx/docs/_gallery/introductory/plot_cube.rst: WARNING: document isn't included in any toctree

This is likely an artifact of your environment @almarklein . Sphinx warns if any rst that is included in the generation isn't explicitly linked to, which can happen for autogenerated files if the source is renamed. You get a new name for the autogenerated file and the old autogenerated one lingers as an artifact.

Try deleting the `docs/_gallery` folder which will regenerate the entire gallery and delete any old/stale files. This should make the warning disappear.

> /Users/almar/dev/py/pygfx/docs/_gallery/introductory/plot_offscreen.rst:115: WARNING: download file not readable: /Users/almar/dev/py/pygfx/docs/_gallery/introductory/plot_offscreen.py

Same story here. IIRC `plot_XXX.py` examples were renamed in a PR that we merged a while ago so likely this warning will disappear if you clear/refresh the gallery cache (by deleting `docs/_gallery`).

docs/conf.py
almarklein: > Autosummary already has a mechanism of invalidating the generated files

If so, it does not work properly, because I was seeing all kinds of warnings from old modules that have changed at some point. Preferably I would delete that dir on `make clean`, not sure how hard it'd be to make that work.

FirefoxMetzger: Yes, if you modify the names of classes or modules you will generate new pages (since each class and each module gets its own doc page). autosummary doesn't delete files (not sure why), so the files matching the old names will linger and create warnings.

However, if you do normal coding and modify/update the docstrings or narrative docs then this caching saves you a couple of seconds each time you rebuild to inspect changes.

almarklein:
> Yes, because most methods we have on classes don't have a docstring (beyond maybe a slug/first line)

That is true, but there are also methods that do have more than one line, and there will be more of those as we improve our docs. These are all invisible to the user with the current setup 😄 

> We can easily add pages for each member function that can be crossref'ed or we can dump a list of methods/attributes at the end of a classes's doc page (what autoclass is doing when using :members:). I am a bit uncertain though as to what this addition buys us.

I find cross-referencing to methods pretty useful. Especially in the guide, where you can explain the use of a particular function or method, and the user can click on it to get the details. I'd very much prefer for the complete docs of a class (all methods) to be on a single page, to be honest.

 In short, I appreciate the idea for the docs to be concise, but I find it more important that they're complete. 

> At the time, we explicitly chose to exclude inherited members from the docs (hence the slightly lengthy template). 

The standard Sphinx approach (that this PR proposes) does not include inherited members either - you'd have to add `:inherited-members:` for that (which we don't).


> This is likely an artifact of your environment @almarklein . 

Ah, that seems to be it indeed, thanks! I just pushed another commit that makes `make clean` remove all directories that contain build caches.

Korijn (approved):

PR #482, opened 2023-03-13
Refactor pygfx using pylinalg

FirefoxMetzger:
Closes #40
Closes #54
Closes https://github.com/pygfx/pylinalg/issues/6
Closes https://github.com/pygfx/pylinalg/issues/21
Closes https://github.com/pygfx/pylinalg/issues/44

This PR updates pygfx to use pylinalg. It does so by introducing a `AffineTransform` class in `pygfx.utils.transform` which stores all the transform-related operations. 

It also does some quite substantial refactor of `pygfx.WorldObject`. It cleans up the caching of the world matrix (will be reintroduced later in this PR), and it also removes a lot of the getters and setters which (as far as I could tell) are just simple one-line wraps around calling the private version of the attribute. This adds a lot of boilerplate that I think we don't need.

Before merging, we will have to release a new version of pylinalg and then require it, because there are a lot of new features that this PR depends on.

Also, this is still a draft and a lot of work still has to be done. Putting it here to start tracking progress.

PR #484, opened 2023-03-14
Require pylinalg

almarklein:
In preparation for using pylinalg in #457 and #482, let's require it in a separate PR. I adopted the same strategy as we use for wgpu, so that people using dev installs are also notified at runtime.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: Test py37
-            pyversion: '3.7'
           - name: Test py38
             pyversion: '3.8'
           - name: Test py39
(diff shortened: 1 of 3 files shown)


almarklein:
Ah, the change that added `version_info` to pylinalg was not in `main` yet when I pushed the release tag. 

Another issue is that pylinal requires Python >= 3.8. Shall we drop 3.7 in pygfx too?

almarklein:
🚀

ivoflipse:
> Ah, the change that added `version_info` to pylinalg was not in `main` yet when I pushed the release tag. 
> 
> Another issue is that pylinal requires Python >= 3.8. Shall we drop 3.7 in pygfx too?

I presume because of the oldest supported Numpy?

Korijn (approved):

Issue #485, opened 2023-03-17
Stats helper


Korijn:
Introduce a Stats helper that can be added to the scene to display debugging information related to drawing performance.

As a starting point, it would be interesting to display a simple text box anchored to a corner or edge of the screen, with the following information:

* [X] Frames Per Second: Amount of frames rendered in the last second
* [X] Draw Time: Milliseconds it took to render the last frame
* [ ] RAM: Amount of RAM currently allocated in MB
* [ ] VRAM: Amount of VRAM currently allocated in MB
* [X] For each of these, it would also be cool to show the min and max value over the last X seconds (30 by default).

So, for example:
```
60 FPS (59-61)
8 MS (5-12)
386 MB RAM (80-400)
510 MB VRAM (10-510)
```

I think 3/4 metrics can be queried without new dependencies, but the amount of allocated RAM for the main python process may require psutil. If so, that would be a reason not to implement it.

* [ ] As a second step it would be useful to display a simple bar chart of how these numbers have changed over the last 30 seconds. But this is definitely not necessary, it would already be super useful without charts.

For inspiration:

* [stats.js](http://mrdoob.github.io/stats.js/)

These stats will also be useful in #482 since they allow you to identify if drawing is slow (high draw time numbers) or if work in between is slow (low draw time numbers, but also low fps).

In the first version merged in #486 I noticed that the first frame includes all the start-up overhead and therefore messed up the min/max values. So I decided to exclude it from the stats. However, it is actually a super interesting metric to see, what was the time to start? So:

* [ ] Display startup time in milliseconds.
* [ ] Additionally, it would be cool if users could add custom time metrics to the object.

Korijn:
I'm closing this for now, we can open new tickets in the future when we feel the need to.

PR #486, opened 2023-03-20
Stats helper

Korijn:
* Lays groundwork for #485 
* Adds two examples:
  * One showcasing manual integration into the render loop (useful, since it demonstrates how you can control the piece of code that is profiled):
    image
  * And another showing the new `stats=True` flag for `Display` which handles everything for you:
    image
* Note that it does not measure time spent rendering itself (intentionally).
* Obviously I hope this will help settle performance arguments more easily.


examples/feature_demo/helpers_stats_display.py
```diff
@@ -0,0 +1,34 @@
...
+
+def animate():
+    # Rotate the cube
+    rot = gfx.linalg.Quaternion().set_from_euler(gfx.linalg.Euler(0.005, 0.01))
+    box.rotation.multiply(rot)
+
+
+if __name__ == "__main__":
+    disp = gfx.Display()
+    disp.stats = True
```
almarklein, line 32: Maybe a bit much to create a new example for this? What about we just add `stats=True` to the cube example?

pygfx/helpers/_stats.py
```diff
@@ -0,0 +1,141 @@
...
+        self.ms = Text(
+            TextGeometry(
+                text="",
+                screen_space=True,
+                font_size=font_size,
+                anchor="topleft",
+            ),
+            text_material,
+        )
+        self.fps = Text(
```
almarklein, line 63: I see this would benefit from the TextGeometry being able to do multi-line ;)

pygfx/helpers/_stats.py
```diff
@@ -0,0 +1,141 @@
...
+            self._init = True
+            return
+
+        t = time.perf_counter_ns()
+        self._frames += 1
+
+        delta = round((t - self._tbegin) / 1_000_000)
+        self._tmin = min(self._tmin, delta)
+        self._tmax = max(self._tmax, delta)
+        self.ms.geometry.set_text(f"{delta} MS ({self._tmin}-{self._tmax})")
```
almarklein, line 123: This now reads as Mega-seconds 😉 , let's use "ms". Also, since the numbers would often be in the order of a few ms, would it it make sense to show one decimal?

pygfx/helpers/_stats.py
Korijn: In very big ways, yes

examples/feature_demo/helpers_stats_display.py
Korijn: Sure!

pygfx/helpers/_stats.py
almarklein: * Also being able to add more measurements without managing `n` text objects.

pygfx/helpers/_stats.py
almarklein: Ok 👍 let's try like this.

PR #488, opened 2023-03-21
Add look-at example

### LINKED ISSUE #487: Main is broken
The last pylinalg release broke pygfx main :)

```
❯ .venv/Scripts/python .\examples\feature_demo\look_at.py
QWindowsWindow::setDarkBorderToWindow: Unable to set dark window border.
  File "C:\Users\kvang\dev\pygfx\examples\feature_demo\look_at.py", line 55, in <module>
    disp.show(group)
  File "c:\users\kvang\dev\pygfx\pygfx\utils\show.py", line 192, in show
    self.camera.show_object(object, up=up, scale=1.4)
  File "c:\users\kvang\dev\pygfx\pygfx\cameras\_perspective.py", line 379, in show_object
    view_dir = la.quaternion_rotate((0, 0, -1), rotation)
AttributeError: module 'pylinalg' has no attribute 'quaternion_rotate'
```

See https://github.com/pygfx/pylinalg/pull/53

Replacing every call to `la.quaternion_rotate` with `la.vector_apply_quaternion` fixes the problem.

Korijn:
* Adds a look_at example
* Closes #487

This example really stretches the performance limits of pygfx a little bit more, so I hope it will prove useful in combination with the new `Stats` helper.

You can just increase the amount of cones in the for loop to push the CPU and GPU even more.

look-at

### DIFF
--- a/examples/feature_demo/look_at.py
+++ b/examples/feature_demo/look_at.py
@@ -1,0 +1,72 @@
+"""
+Look At
+=======
+
+In this example shows how to do the rendering world's hello world: Show a 3D
+Cube on screen.
+"""
+# sphinx_gallery_pygfx_animate = True
+# sphinx_gallery_pygfx_target_name = "disp"
+from time import perf_counter_ns
+import numpy as np
+import pygfx as gfx
+
+scene = gfx.Scene()
+
+sphere = gfx.Mesh(
+    gfx.sphere_geometry(100, 20, 20),
+    gfx.MeshPhongMaterial(color="#336699"),
+)
+sphere.receive_shadow = sphere.cast_shadow = True
+scene.add(sphere)
+
+plane = gfx.Mesh(
+    gfx.plane_geometry(4000, 4000),
+    gfx.MeshPhongMaterial(color="#336699"),
+)
+plane.position.set(0, -2500, 0)
+plane.rotation.set_from_euler(gfx.linalg.Euler(-np.pi / 2))
+plane.receive_shadow = True
+scene.add(plane)
+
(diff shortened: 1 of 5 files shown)


examples/feature_demo/look_at.py
```diff
@@ -0,0 +1,69 @@
...
+    cone.position.set(*(np.random.rand(3) * 4000 - 2000))
+    cone.scale.set_scalar(np.random.rand() * 4 + 2)
+    cone.receive_shadow = cone.cast_shadow = True
+    cones.add(cone)
+
+scene.add(cones)
+
+
+def animate():
+    t = perf_counter_ns() // 1_000_000 * 0.0005
```
almarklein, line 54: Let's ditch some zeros:
```suggestion
    t = perf_counter()  * 0.5
```

examples/feature_demo/look_at.py
```diff
@@ -0,0 +1,69 @@
+"""
+Look At
+=======
+
+This example shows how the look_at function can be used.
+"""
+# sphinx_gallery_pygfx_animate = True
+# sphinx_gallery_pygfx_target_name = "disp"
+from time import perf_counter_ns
```
almarklein, line 9:
```suggestion
from time import perf_counter
```

examples/feature_demo/look_at.py
almarklein: From what I understand `perf_counter_ns`'s use-case over `perf_counter` is that its an int.

almarklein (commented):
Minor suggestion. Otherwise 👍

almarklein (approved):

Issue #321, opened 2022-07-05
Manage updates to buffers and textures more efficiently


almarklein:
Right now, the renderer iterates over all resources of a world-object, and updates each resource if needed.

A better approach would be to have a global store (per pygfx render system, one for wgpu), that resources subscribe to when they have pending changes. The renderer can then iterate over all objects in that store to update them.

PR #490, opened 2023-03-23
Make controllers configurable and apply damping

almarklein:
New features:

* [x] Controller smoothing.
* [x] A system to configure the mouse and key bindings.
* [x] Key bindings can have 3 modes: apply a change (drag), apply change while it is pressed (peak), apply a change on each tick (push). The latter will be used by the fly camera. 
* [x] Can can control the controller programmatically.
* [x] Multiple control actions can occur at once.

It needs some work:
* [x] Clean up.
* [x] Make tick-update take the passed time into account. 
* [x] Docs.
* [x] Fix Orbit controller.
* [x] Example with custom controller config.
* [x] Example with controller but no external draw requests.
* [x] Fix `before_draw` (also in relation to viewports) in a seperate PR.

### How does the controller work?

When an interaction starts, an `action` object is created. In some cases (e.g. dragging) the action's target value is updated as new events come in. There can be multiple actions simultaneously, e.g. you can zoom while panning. Though only one drag action is allowed at one time, otherwise that just gets weird.

During each "tick" a number of things happen:
* The real action values are moved towards the target value. This is the smoothing part. 
* The action's corresponding update method is called. This updates a camera-state object using the action's value.
* Then all cameras that the controller controls are updated with that state.

A bit more on that state object: the camera's state is obtained using `camera.get_state()`, and is kept as a private var on the controller. It is updated at each tick (possibly by multiple actions). This controller's internal state  is used as long as there are active actions. This means that changes to the cameras (e.g. pushing a button to rotate it), have no effect while the controller is doing its thing. In practice the inertia from the smoothing is short, so this should not be a problem for changes invoked by user interaction.

So how is this kept up-to date? When an event occurs that changes the state, a new draw is requested (unless `auto_update` is false).

The tick mentioned above happens right before each draw, using a new event emitted by the renderer. The tick itself also requests a new draw when there are still active actions.

When you don't make use of `controller.register_events()`, this would work somewhat differently. We may need to change a few things around to make that work well.


pygfx/renderers/wgpu/_renderer.py
```diff
@@ -797,6 +797,7 @@ def convert_event(self, event: dict):
     "wheel": WheelEvent,
     "key_down": KeyboardEvent,
     "key_up": KeyboardEvent,
+    "before_draw": WindowEvent,
```
Korijn, line 800: I have a question about this; before_draw is a risky event to rely on, especially in applications without a steady frame rate. I assume you're aware of this and taking that into account?

Korijn:
@berendkleinhaneveld we need to test this branch with your project before merging. I'm a little concerned it won't be compatible with your push-based render loop.

pygfx/renderers/wgpu/_renderer.py
almarklein: >  I assume you're aware of this and taking that into account?

Yes. The controller requests new draws when it know that it needs them (also when the mouse has been released, but the "animation" has not finished yet due to the smoothing). I'll make sure we have at least one example without external draw requests.

I will also look into before_draw in relation to viewports. I might create a separate PR for that.

almarklein:
This is ready (an no longer requires a change in wgpu-py).

almarklein:
I added an example with a very wide texture, and a camera config that only allows horizontal panning. Funnily enough, the horizontal-only panning revealed that the panning also moved vertically, due to a bug where we calculate the panning vectors.

pygfx/renderers/wgpu/_renderer.py
```diff
@@ -465,6 +465,20 @@ def render(
...
+        if clear_color:
+            ev = WindowEvent(
+                "before_render",
+                target=None,
+                root=self,
+                width=logical_size[0],
+                height=logical_size[1],
+                pixel_ratio=self.pixel_ratio,
+            )
+            self.dispatch_event(ev)
```
Korijn, line 480: I'm still not convinced of the benefit of this event over just making users call `controller.update()` in the `animate` function, [as done here](https://github.com/mrdoob/three.js/blob/47b28bc564b438bf2b80d6e5baf90235292fcbd7/examples/misc_controls_orbit.html#L130) ([example](https://threejs.org/examples/#misc_controls_orbit)).

Korijn:
Another question: is it possible to disable damping?

almarklein:
> Another question: is it possible to disable damping?

Yes, just set it to 0.

pygfx/renderers/wgpu/_renderer.py
Korijn: Well, is that true though? This is the first "renderer event" that isn't an "input event".

Korijn: Fine, let's roll with this, and we'll see how well it works!

Korijn:
In that case, I'd like a test from berend to see how well this works on his project before we merge.

berendkleinhaneveld:
I'll try it out, but it will have to wait till next week. Unless we are in a hurry to get this in?

almarklein:
> I'll try it out, but it will have to wait till next week. Unless we are in a hurry to get this in?

No, not in a hurry. 

BTW, once this is merged I'll add a Trackball controller (which is just a few lines) and a FlyController (which also should be easy now).

Korijn:
So I discussed this with Berend and here's what would need to change in order for collagraph to retain full control over the update-render loop:

* don't use the before_render event
* don't run update_cameras
* controller.auto_update = False
* call handle_tick in independent update loop (16.7ms), independent from render loop
  * in that loop, copy controller._last_cam_state to camera state object (not the pygfx camera instance, but an observ camera state object)

There's two essential requirements here:

* Controllers do not mutate cameras
* Handle_tick is not triggered by before_render (this trigger prevents running state updates independently from draws)

almarklein:
> don't use the before_render event
> don't run update_cameras

Is the controller hooked up to the renderer events in the collagraph case? If not, a lot of the requirements are met, which means that we can keep the controller Just Work when using `register_events`, and with just a few tweaks provide the flexibility needed by e.g. collagraph. 

> call handle_tick on regular intervals

Would it work if we made `handle_tick` public, and leave it to the application? It could return the camera state, so the application can process it the way it wants.

Korijn:
> Is the controller hooked up to the renderer events in the collagraph case?

The controller is listening to mouse/keyboard events, but nothing else. I guess that's what you mean?

> If not, a lot of the requirements are met, which means that we can keep the controller Just Work when using `register_events`, and with just a few tweaks provide the flexibility needed by e.g. collagraph.

Yes, I agree.

> Would it work if we made `handle_tick` public, and leave it to the application? It could return the camera state, so the application can process it the way it wants.

Yes, that would work. We could also make the camera state a property, instead of returning it. I think that's a little more flexible.

almarklein:
> The controller is listening to mouse/keyboard events, but nothing else. I guess that's what you mean?

How are these events fed into the controller? By calling `handle_event()` explicitly?

> We could also make the camera state a property, instead of returning it. I think that's a little more flexible.

I'd rather not, because the internal camera state is only valid as long as any actions are active, and we'd have to figure out whether to e.g. set it to None when there are no actions, etc.

almarklein:
I just pushed a commit that I think should do the trick. The `tick()` method is not public. And when `auto_update` is False, the controller won't update the camera either.

Korijn:
Maybe #321?

pygfx/controllers/_base.py
```diff
@@ -101,10 +184,93 @@ def register_events(self, viewport_or_renderer: Union[Viewport, Renderer]):
...
+            action.tick(factor)
+            if action.is_at_target and action.done:
+                to_pop.append(key)
+            self._apply_action(action)
+
+        # Remove actions that are done
+        for key in to_pop:
+            self._actions.pop(key)
+
+        return self._update_cameras()
```
Korijn, line 226: I see how this returns camera state, but it seems like it still always calls `camera.set_state(...)` inside `self._update_cameras()` regardless of `auto_update`?

pygfx/controllers/_base.py
```diff
@@ -114,6 +280,308 @@ def add_default_event_handlers(self, *args):
...
+
+        rect = viewport.rect
+        need_update = False
+
+        type = event.type
+        if type.startswith(("pointer_", "key_", "wheel")):
+            modifiers = sorted([m.lower() for m in event.modifiers])
+            modifiers_prefix = "+".join(modifiers + [""])
+
+        if type == "before_render":
```
Korijn, line 299: should this also be conditional on `auto_update`?

almarklein: Yeah, that makes sense 👍

pygfx/controllers/_base.py
almarklein: Woops! I updated the docs, but forgot the code 😄

almarklein:
I added the reactive example and adjusted it for the new controller.

examples/other/reactive_rendering.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+frames = 0
+
+
+def render_frame():
+    global frames
+    frames += 1
+    print(f"frames: {frames}")
+    controller_tick()
```
Korijn, line 97: Would this call not be necessary if controller tick was executed on a timer?

examples/other/reactive_rendering.py
```diff
@@ -0,0 +1,160 @@
...
+    update_scene()
+    canvas.request_draw()
+
+
+if __name__ == "__main__":
+    # restore state from previous session
+    # NOTE: ideally we would persist visual_state
+    # and restore visual_state only,
+    # but the Controller class unfortunately makes this
+    # impossible without also redundantly tracking its state
```
Korijn, line 111: Well, you solved that :) so this note can be removed. Nice.

Korijn (approved):

almarklein:
> Would this call not be necessary if controller tick was executed on a timer?

Exactly. Because of the inertia, it needs to be called *somehow*  in addition to (mouse) events. A timer would work. At draw-time works too.

PR #491, opened 2023-03-28
Show depth buffer

almarklein:
This is a bit of an experiment to what we'd need to make the values of the depth buffer available (for users, via an API, and for post-processing). It also contains a few tweaks to the blender code to avoid resetting the depth buffer in transparent passes.

For most examples, this nicely shows the depth:
image

Some observations:
* Transparent fragments do not have a depth value, unless the blend_mode is "opaque"  or "ordered1". 
* For subplots, only the last drawn plot has a depth, because the full depth buffer is reset for each subplot.
* Similarly, for overlays, only the depth of the final pass is visible.


almarklein:
Closing to clean up. I left it open primarily to be able to find back the code to use when tackling #492.

PR #493, opened 2023-03-30
Remove TextureView from public api

### LINKED ISSUE #350: Allow only texture or texture-view
In some places we allow both a texture and a texture-view, but this can be confusion and is a source of bugs, e.g. #335. Let's be explicit where it matters.

Also see the discussion that lead to this issue here: https://github.com/pygfx/wgpu-py/issues/280

almarklein:
Closes #350 

We listed a few options in #350, but I take another approach here. The `TextureView` that we had, combined `GPUTextureView` and `GPUSampler`.  The former defines the view in the data, e.g. see it as it is, or view the 3D texture as 2D slices, or as a cube texture. The shader can usually make this decision. The latter defines filtering and address mode (what happens outside of texcoords 0..1), which is related to the appearance, so it should be either decided by the shader or defined on the material.

### Lib

* [x] The `gfx.TextureView` is no more.
* [x] The `gfx.Texture.get_view()`  raises a `DeprecationWarning`.
* [x] New `gfx.renderers._utils.wgpu.GfxTextureView` for internal use. 
* [x] Image and Volume materials have new `interpolation` prop.
* [x] Improve/fix the texture mimap utility.
* [x] Split the wgpu object creation from the buffer/texture update mechanic. 
* [x] All materials with a `map` also get a `map_interpolation` prop.
 
### Examples

* [x] Updated all examples to not use texture views.
* [x] Removed `volume_slice2.py`, which relied on a texture view. Renamed `volume_slice3.py` to `volume_slice2.py`. 
* [x] Tweaked the `dynamic_env_map.py` example a bit (it did not actually show the static env map). 
* [x] Push the example-changes. I'm leaving this for last, to make review easier.


almarklein:
We don't have an "improvements to Texture" issue, but if we had one, this'd close it 😄

Korijn:
> We don't have an "improvements to Texture" issue, but if we had one, this'd close it 😄

Maybe #321?

almarklein:
> Maybe https://github.com/pygfx/pygfx/issues/321?

No, that one's up next, I think.

almarklein:
All green. Two screenshots needed updating. I think this is because the edge-filter param is different now for these examples, which causes a slightly different result for some reason.

pygfx/materials/_background.py
```diff
@@ -136,6 +137,7 @@ def map(self):
 
     @map.setter
     def map(self, map):
+        assert map is None or isinstance(map, Texture)
```
Korijn, line 140: At some point, we should configure a linting rule that prevents assertions outside `tests` folders. :) Generic `AssertionError`s are not a great UX.

Korijn (approved):

pygfx/materials/_background.py
almarklein: Good point. I admit this is mostly out of lazyness, because they're so mich easier to write than `if ... raise ...` 😛

PR #500, opened 2023-04-07
Add reactive rendering example

Korijn:
* This example is intentionally based on the last commit (0c4f058) before #457 was merged
* It validates all the requirements for a reactive render loop
* @almarklein I would like to request that you refactor this example to make it work on your branch in #490 to validate that reactive rendering is still supported

reactive-rendering


Korijn:
See #490

PR #507, opened 2023-04-14
Skybox validation

almarklein:
* Renames the `examples/models` directory to `examples/data`. I don't feel much for having separate directories for models, data, etc. We don't plan on having that much data; let's just put it into one dir.
* Adds cubemap image.
* Use that to validate the skybox and env map. Example adopted from #353


examples/validation/validate_skybox.py
```diff
@@ -0,0 +1,63 @@
+"""
+Validate Skybox
```
almarklein, line 2: @panxinmiao I invite you to make a suggestion somewhere. That way your name gets attached to the commit for this PR, and you'd get the credits/stats for your work.

examples/data/cubemap.jpg
Korijn, line 1: If I look at #460 it says:

> glTF uses a right-handed coordinate system. glTF defines +Y as up, +Z as forward, and -X as right; the front of a glTF asset faces +Z.

So then the left/right faces are wrong in this cubemap, right?

examples/validation/validate_skybox.py
panxinmiao: Thank you, you are so considerate, 😄

almarklein: Only trying to be fair. I initially wanted to cherry-pick a commit from your PR, but everything has evolved too far to make that work :)

examples/data/cubemap.jpg
panxinmiao: By convention, cubemaps are specified in a left-handed coordinate system. We use a right-hand coordinate system (consistent with glTF regulations), so the left and right faces of this cubemap seem to be opposite. We have taken this into account in the code and made the conversion.

almarklein: There's also this:
https://github.com/pygfx/pygfx/blob/45f2c642336c47feae38960a1a2219fcea3482e7/pygfx/renderers/wgpu/backgroundshader.py#L105-L109

panxinmiao: See: https://github.com/pygfx/pygfx/blob/45f2c642336c47feae38960a1a2219fcea3482e7/pygfx/renderers/wgpu/backgroundshader.py#L105-L109
and
https://github.com/pygfx/pygfx/blob/45f2c642336c47feae38960a1a2219fcea3482e7/pygfx/utils/cube_camera.py#L49-L52

almarklein: Thanks for the clarification @panxinmiao! One thing that I'm not sure about is whether we should flip the x or z dimension. Said differently, do we want an unrotated camera to look at the +Z or -Z face?

panxinmiao: I have also considered this issue before. In fact, flipping the Z dimension would be equivalent to rotating the current cube map 180 degrees around the Y-axis. For users, it doesn't make much difference when used. Moreover, although cubemap specifies the use of a left-hand coordinate system, it does not specifically specify whether the Z-axis faces inside or outside the screen. (Even on Wiki, there is an example of the Z-axis facing left).

Quoted from Wiki:

>Note that the coordinate system for cubemaps is left-handed. If X is forward, and Y is up, then Z is to the left.

To this end, I specifically checked  the code of `three.js` and decided to keep it consistent with its logic (flipping x).

almarklein: In case of doubt, keeping things consistent with ThreeJS is usually a good idea 😄 . Good as it is from my end, then.

examples/validation/validate_skybox.py
```diff
@@ -0,0 +1,63 @@
...
+from wgpu.gui.auto import WgpuCanvas, run
+
+
+env_map_path = Path(__file__).parents[1] / "data" / "cubemap.jpg"
+data = imageio.imread(Path(env_map_path))
+
+h = data.shape[0] // 3
+w = data.shape[1] // 4
+
+posx = np.ascontiguousarray(data[1 * h : 2 * h, 2 * w : 3 * w])
```
panxinmiao, line 31:
```suggestion
"""
The layout of the example cubemap looks like this:
 ┌────┬────┬────┬────┐
 │    │ +Y │    │    │
 ├────┼────┼────┼────┤
 │ -X │ +Z │ +X │ -Z │
 ├────┼────┼────┼────┤
 │    │ -Y │    │    │
 └────┴────┴────┴────┘
 """
posx = np.ascontiguousarray(data[1 * h : 2 * h, 2 * w : 3 * w])
```

panxinmiao: I think maybe we can explain this code a bit, 😄

Korijn (approved):

Issue #272, opened 2022-03-28
Logic in materials is based on details about the wgpu renderer


almarklein:
While looking at #257 I ran into some abstraction leaks. We try to make the worldObject/geometry/material logic free from logic about the renderer. Even though we only have one real renderer (wgpu) this abstraction has helped us (I think) keep the code clean. But there are some things where it's off:

*Last updated 18-04-2023*

* Some material properties are stored just as a property, while others are stored in the uniform buffer. Ideally this is a detail of the renderer.
* Same for world object, though so far only the `render_mask` property.
* `Buffer` and `Texture` set a few wgpu-specific private attributes. Theses are only used by the renderer, though a user could technically use `_wgpu_usage` when needed.
* Actually, there are valid cases where a user *should* be able to set `wgpu_usage`, see #574.
* `Buffer` and `Texture` accept a format that is generic, but also accept a `wgpu.BufferFormat` or `wgpu.TextureFormat` respectively.
* The resource registry for buffers and textures uses wgpu-specific attributes of resources (writing this as I am working on #508, so maybe that has changed). 
* `Material._wgpu_get_pick_info()`
* `WorldObject._wgpu_get_pick_info()`
* utils/cube_camera.py -> uses/wraps a `WgpuRenderer`.
* The `show()` util uses the `WgpuRenderer`, but fair enough since that's the default.

I'm not sure what best to do about this. Just accept it, we use the abstraction as a "rule of thumb"  to keep on track? Or fix it, by moving the offending logic to the renderer?

Korijn:
You're right, it's all supposed to be managed and tracked as part of the renderer. But we don't have to fix this right away.

Korijn:
Decisions:

* wgpu calls will be confined to renderers/backends.
* render state may be managed on public API objects.
* wgpu enums/flags may be used in public API objects. (e.g. culling mode)
* We will embrace the fact that pygfx is based on the class of rendering backends that uses WGPU and similar implementations. As such "leaking abstractions" like properties that go into buffers or trigger shader compilations are OK to have in public API objects. It will be something that all foreseeable rendering backends require anyway.
* As an example, the SVG renderer can still be implemented even if wgpu flags are used in public API objects, since wgpu can be imported without importing the rust binaries and without creating GPU devices/objects. We will not directly further invest in the SVG but it is free for community members to further develop it.

PR #508, opened 2023-04-14
More efficient Buffer and Texture updates

### LINKED ISSUE #321: Manage updates to buffers and textures more efficiently

almarklein:
Closes #321

### Todo

* [x] Introduce a registry that is aware of all buffers and textures, and which ones need an update (uses weakrefs).
* [x] Make the renderer use that instead of checking *all* resources whether they need to be synced.
* [x] Removed a bit of wgpu-specific logic from lights.py
* [x] Updated the list in #272 along the way.
* [x] Removed a print-call from the fly controller. 

### How it now works

(also added this to the code in a comment).

* When a resource is first created, it's _wgpu_object attribute is None
  and its _wgpu_flags is unset.
* When the resource is actually being used somewhere, it will end up
  in the logic that creates a pipeline object (e.g. in _pipeline.py), which
  sets the appropriate usage flags (because that code knows how the resource
  is used) and then uses ensure_wgpu_object to create the object.
* Resources that need to be synced are tracked in the resource_registry,
  but only go into it when they have their _wgpu_object set (i.e. when
  the resource actually exists on the GPU).
* Right before the renderer performs a draw, it queries the registry
  and calls update_resource on each.

This PR basically makes the above work, and then removes the old system (which required updates more spread out over the code).

### DIFF
--- a/pygfx/resources/_base.py
+++ b/pygfx/resources/_base.py
@@ -1,3 +1,5 @@
+import weakref
+
 from ..utils.trackable import Trackable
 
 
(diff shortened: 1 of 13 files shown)


almarklein:
About done. Two validation examples fail. Will look into that.

pygfx/resources/_base.py
```diff
@@ -0,0 +1,77 @@
...
+        self._all = weakref.WeakSet()
+        self._syncable = weakref.WeakSet()
+
+    def _gfx_register(self, resource):
+        """Add a resource to the registry."""
+        if not isinstance(resource, Resource):
+            raise TypeError("Given object is not a Resource")
+        self._all.add(resource)
+
+    def _gfx_mark_for_sync(self, resource):
```
Korijn, line 44: I don't really see why these methods on registry need to be private?

almarklein: It's more that I don't see a reason to make them public, so a private method seems safer (can make it public later if needed, but the reverse is less true).

Korijn (approved):

PR #511, opened 2023-04-19
more informative exception message for `Buffer.update_range` valid types

kushalkolar:
I run into this often since `numpy.integer` types (such as `np.int64`) often end up being sent to `update_range()`. I thought about potentially just allowing `update_range()` to take `numpy.integer` types, however that might open up a can of worms where it is expected to support integer types from all array libraries?


almarklein (approved):
Nice! we indeed need to rewrite a few assertions to `if ... raise ...`.

Korijn (approved):

Issue #492, opened 2023-03-28
Proposal for renderer / viewport improvements


almarklein:
### Overview of issues

There are currently a few issues related to how we currently deal with viewports and overlays.

* It currently prevents using the values in the depth buffer.
  * Details and some experiments: #491
  * An earlier discussion: #320
  * Another related discussion: #318
  * This also affects post-processing that needs the depth buffer, like fog: #72
* When rendering to a viewport, code that uses events may need to offset mouse positions for viewport rect
  * E.g. in controllers #490 
  * Probably the Gizmo too?
* Subplots must share renderer props like blend_mode. Users may be forced to use multiple widgets/canvases.

### Proposal

This is a proposal that I think fixes these issues.

```

  ┌───────────────────────────────────────────┐
  │                         Canvas            │
  │ ┌───────────────┐                         │
  │ │  Renderer     │                         │
  │ │               │                         │
  │ │               │                         │
  │ │               │                         │
  │ │               │        ┌───────────┐    │
  │ └───────────────┘        │  Renderer │    │
  │                          │           │    │
  │                          │           │    │
  │                          │           │    │
  │                          │           │    │
  │                          └───────────┘    │
  │                                           │
  └───────────────────────────────────────────┘

```

* We drop the viewport class.
* A renderer can render to a specific region of the canvas.
* The renderer offsets event coordinates, so that everything is relative to the renderers rect. (We can still expose the canvas coords if that's useful.)
* The renderer keeps its own internal render textures, (which may be smaller than the canvas).
* We introduce `rendeer.render_overlay()` that allows rendering stuff like fps counters, gizmo's, axis ticks, labels, etc. This overlay pass does not clear the depth buffer. 
* The renderer gets an API to add post-processing steps. When the renderer flushes its color buffer to the canvas, these steps are applied in the same shader (the flush render pass). (This step is less important and can be implemented later.)
* We may also add the ability for an overlay pass to the total canvas. (May not be needed; let's look at this later.)

This should make that each renderer has a depth buffer that is "intact" and can be sampled by the user. In the flush step, the depth can also be made available. I guess you could still render with the same renderer multiple times for specific use-cases, but then the depth buffer contains the values of the last pass.

[...]

PR #512, opened 2023-04-19
add example of how to map screen to world

kushalkolar:
Simple example of how to map event locations from screen space to world space, includes viewports. As someone who isn't really a graphics person it took me a while to figure out :smile: 

https://user-images.githubusercontent.com/9403332/232991611-56cdcf8c-3b11-4298-990c-ac0f9d1aa157.mp4


almarklein:
Nice example to have! Note that this will change / get easier with #492.

kushalkolar:
I think this should be good to go, or I can wait until #492 and update it then.

almarklein:
Would you mind waiting at least for #482?

kushalkolar:
Yes of course (I haven't been following that PR, it is massive :smile: )

examples/feature_demo/map_click_events_to_world.py
```diff
@@ -0,0 +1,132 @@
...
+            pos[0] - viewport.rect[0],
+            pos[1] - viewport.rect[1],
+        )
+
+        vs = viewport.logical_size
+
+        # convert position to NDC
+        x = pos_rel[0] / vs[0] * 2 - 1
+        y = -(pos_rel[1] / vs[1] * 2 - 1)
+        pos_ndc = Vector3(x, y, 0)
```
FirefoxMetzger, line 102: Setting the third component to 0 means that you will always place points/objects onto the near plane of the camera. This works well for orthographic projections but may give odd results once you switch to perspective or parallel projections.

Unfortunately, picking a good number for the depth is always opinionated because you are literally pulling a number out of thin air (cursor gives you X,Y, but no Z). One way to do this that could be "natural" (depending on your application) is to do a ray-plane intersection with the XZ plane (going through the origin) and place objects there.

Korijn:
> Yes of course (I haven't been following that PR, it is massive 😄 )

It's been merged now, so if you can refactor your example to be compatible, we can have another look and merge it.

kushalkolar:
Digging through `pylinalg` and can't quite find the equivalent for this important bit:

```python
pos_world = (
    camera.position.clone().project(camera).add(pos_ndc).unproject(camera)
)
```

This is not quite right:

```python
# this is the click position in canvas space
pos = (ev.x, ev.y)
print(f"position click: {pos}")

for viewport, scene, camera in zip(viewports, scenes, cameras):
    if not viewport.is_inside(*pos):
        continue

    # get position relative to viewport
    pos_rel = (
        pos[0] - viewport.rect[0],
        pos[1] - viewport.rect[1],
    )

    vs = viewport.logical_size

    # convert position to NDC
    x = pos_rel[0] / vs[0] * 2 - 1
    y = -(pos_rel[1] / vs[1] * 2 - 1)
    pos_ndc = (x, y)

    vec = vec_unproject(
        pos_ndc, camera.camera_matrix
    )

    x = vec[0] + x
    y = vec[1] + y

    pos_world = vec_transform((x, y, 0), camera.view_matrix)
```

https://github.com/pygfx/pygfx/assets/9403332/7c40e2ee-ca2f-449d-803f-dba4e3f99798

almarklein:
You'd have to use the `camera.camera_matrix` in both cases. I think it'd be something like (untested):
```py
abs_pos_ndc = vec_transform(camera.world.position, camera.camera_matrix) + pos_ndc
pos_world = vec_unproject(abs_pos_ndc, camera.camera_matrix)
```

kushalkolar:
That almost does it, thanks! Ready for a look again. I think a click exampe for perspective projections can be separate since that's a bit more involved.

almarklein (approved):

Korijn (approved):

Korijn:
Thanks again!

PR #516, opened 2023-04-21
New cache class, apply to mipmaps

almarklein:
See #322 

This PR introduces a new GPUCache class, intended for use within the `renderers.wgpu` subpackage. The cache uses weakrefs, so the objects being cached should still be stored somehow, e.g. by attaching them to an world object or Buffer.

The stats of all caches can also be obtained. The `gfx.print_wgpu_report()` function uses it to now also print cache stats!

In this PR I only apply the new cache to the mipmap code. I plan to adopt it in other places in new PR's. Because of the new cache, the MipmapUtil class become absolete, falling apart into a few functions, which I also reordered for readability. Other than that, the mipmap util should not have any logical changes to the code.

While searching the code-base for "cache" I found same comments which I tweaked in a few cases.

### DIFF
--- a/pygfx/renderers/wgpu/_mipmapsutil.py
+++ b/pygfx/renderers/wgpu/_mipmapsutil.py
@@ -1,4 +1,4 @@
-import math
+import numpy as np
 import wgpu
 from ...resources._texture import Texture
 from ._utils import GfxTextureView, GpuCache
@@ -59,9 +59,10 @@ mipmap_source = """
 
 def get_mip_level_count(texture):
     """Get the number of mipmap levels from the texture size."""
-    assert isinstance(texture, Texture)
+    if not isinstance(texture, Texture):
+        raise TypeError("Expecting a Texture object.")
     width, height, _ = texture.size
-    return math.floor(math.log2(max(width, height))) + 1
+    return int(np.floor(np.log2(max(width, height))) + 1)
 
 
 def generate_texture_mipmaps(device, target):
--- a/pygfx/renderers/wgpu/_mipmapsutil.py
+++ b/pygfx/renderers/wgpu/_mipmapsutil.py
@@ -1,7 +1,10 @@
 import math
 import wgpu
 from ...resources._texture import Texture
-from ._utils import GfxTextureView
+from ._utils import GfxTextureView, GpuCache
+
+
+MIPMAP_CACHE = GpuCache("mipmap_pipelines")
 
 
 mipmap_source = """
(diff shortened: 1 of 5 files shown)


pygfx/renderers/wgpu/_mipmapsutil.py
```diff
@@ -54,155 +57,144 @@
...
-            pass_encoder.dispatch_workgroups(dst_size[0], dst_size[1])
-            pass_encoder.end()
+    device.queue.submit([command_encoder.finish()])
 
-            src_view = dst_view
 
-        self.device.queue.submit([command_encoder.finish()])
+def get_mipmap_pipeline(device, texture):
+    format = texture._wgpu_object.format
+    pipeline = MIPMAP_CACHE.get(format)
```
almarklein, line 137: This function shows the gist of the new cache.

pygfx/renderers/wgpu/_mipmapsutil.py
```diff
@@ -54,155 +57,144 @@
...
 
-class MipmapsUtil:
-    def __init__(self, device) -> None:
-        self.device = device
-
-        # Cache pipelines for every texture format used.
-        self.pipelines = {}
+def get_mip_level_count(texture):
+    """Get the number of mipmap levels from the texture size."""
+    assert isinstance(texture, Texture)
```
Korijn, line 62: I think you can guess what I have to say about `assert` :)

pygfx/renderers/wgpu/_mipmapsutil.py
```diff
@@ -54,155 +57,144 @@
...
+    width, height, _ = texture.size
+    return math.floor(math.log2(max(width, height))) + 1
```
Korijn, line 64: why not numpy?

pygfx/renderers/wgpu/_mipmapsutil.py
```diff
@@ -54,155 +57,144 @@
...
-    # If this looks like a cube or stack, generate mipmaps for each individual layer
-    if isinstance(target, Texture) and target.dim == 2 and target.size[2] > 1:
-        for i in range(target.size[2]):
-            generate_texture_mipmaps(
-                device, GfxTextureView(target, layer_range=(i, i + 1))
-            )
-        return
+    # Strore a ref of the pipeline, the cache uses weak refs.
+    # I.e. the pipeline for this format is in the cache for as long as
+    # any textures that use it are alive.
```
Korijn, line 166: I'm wondering, if the possible number of pipelines is low (e.g. there are only X possible pipelines to be generated, let's say X < 10), why would you ever evict them from cache?

pygfx/renderers/wgpu/_utils.py
```diff
@@ -193,6 +195,38 @@ def generate_uniform_struct(dtype_struct, structname):
...
+        GpuCache._caches[name] = self
+
+        self._objects = weakref.WeakValueDictionary()
+
+    def get_count(self):
+        """Get the number of (alive) objects in the cache."""
+        return len(list(self._objects.values()))
+
+    def get(self, hash):
+        """Get the cached object or None."""
```
Korijn, line 220: I would recommend to refer to `key` instead of `hash`, since python dicts actually compute the hash of the key internally. It's a little confusing to read `hash` here, it seems to imply that you're working with a handcrafted dictionary implementation.

pygfx/renderers/wgpu/_mipmapsutil.py
almarklein: Sorry 😄

pygfx/renderers/wgpu/_mipmapsutil.py
almarklein: This was already written like that. Converted to numpy.

pygfx/renderers/wgpu/_mipmapsutil.py
almarklein: That's a fair point in this case. But in other cases (that I plan to implement later) this will matter. I reckoned that its better to be consistent from the start.

pygfx/renderers/wgpu/_utils.py
almarklein: 👍

Korijn (approved):

PR #525, opened 2023-05-08
Caching in _pipeline.py

almarklein:
### Changes

* [x] Shader modules (was already cached, make it use new cache).
* [x] Avoid the wgsl generation (i.e. apply templating) on a shader-module cache hit.
* [x] pipeline objects.
* [x] bind groups and bind group layouts.

### Results 🐎 

Measuring the time to the first draw of the `collection_line.py` example:

* Caching disabled: ~8.7 s
* On `main`: ~ 8.4 s
* With new shader module caching: ~ 3.5 s
* Also caching pipelines, bind groups and bind group layouts: ~ 2.9

With `multi_select.py`:

* On `main`: ~1.4 s
* On this branch: ~0.7 s

Notes:
* Interesting, the caching that we did in `main` did not help much; the wgsl generation (templating) is much more expensive that turning wgsl into a wgpu module. Kudos to Naga!

### DIFF
--- a/examples/other/collection_line.py
+++ b/examples/other/collection_line.py
@@ -9,8 +9,6 @@ this is still performant.
 
 # sphinx_gallery_pygfx_render = True
 
-import time  # noqa
-
 import numpy as np
 from wgpu.gui.auto import WgpuCanvas, run
 import pygfx as gfx
@@ -20,7 +18,6 @@ canvas = WgpuCanvas(max_fps=999)
 renderer = gfx.WgpuRenderer(canvas, show_fps=True)
 scene = gfx.Scene()
 
-t0 = time.perf_counter()
 
 # Define number of vertices
 cols = 20
@@ -57,16 +54,10 @@ stats = gfx.Stats(viewport=renderer)
 
 
 def animate():
-    global t0
     with stats:
         renderer.render(scene, camera, flush=False)
     stats.render()
     canvas.request_draw()
-
-    if t0 is not None:
-        etime = time.perf_counter() - t0
-        print(f"Time to first draw: {etime:0.2f} s")
-        t0 = None
 
 
 if __name__ == "__main__":
--- a/examples/other/collection_line.py
+++ b/examples/other/collection_line.py
@@ -18,8 +18,9 @@ import pygfx as gfx
 
 canvas = WgpuCanvas(max_fps=999)
 renderer = gfx.WgpuRenderer(canvas, show_fps=True)
-
 scene = gfx.Scene()
+
+t0 = time.perf_counter()
 
 # Define number of vertices
 cols = 20
@@ -30,6 +31,7 @@ use_thin_lines = True
 print(nvertices * rows * cols, "vertices in total")
 
 x = np.linspace(0.05, 0.95, nvertices, dtype=np.float32)
+
 
 for row in range(rows):
     for col in range(cols):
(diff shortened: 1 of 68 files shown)


examples/other/collection_line.py
```diff
@@ -55,11 +57,17 @@
...
 def animate():
+    global t0
     with stats:
         renderer.render(scene, camera, flush=False)
     stats.render()
     canvas.request_draw()
 
+    if t0 is not None:
+        etime = time.perf_counter() - t0
+        print(f"Time to first draw: {etime:0.2f} s")
```
Korijn, line 68: we could add this as a feature to the Stats helper pretty easily, to output that number on the console

almarklein:
Some examples fail now. Can reproduce this locally - I am investigating. Maybe something is being cached too aggressively?

almarklein:
This is working and ready, except for that one point about mesauring time to first draw.

pygfx/renderers/wgpu/_utils.py
```diff
@@ -195,30 +197,99 @@ def generate_uniform_struct(dtype_struct, structname):
...
+
+    # If the strings is small enough, just use that. Otherwise use fast
+    # sha1 hashing to produce a 40-char hash. This costs a bit more
+    # time, but safes memory. Can also set this to True for debugging purposes.
+    if len(s) < 128:
+        return s
+    else:
+        h = hashlib.sha1()
+        h.update(s.encode())
+        return h.hexdigest()
```
Korijn, line 225: Would standard `hash()` be faster and simpler here?

pygfx/renderers/wgpu/_shaderbase.py
```diff
@@ -262,14 +262,25 @@ def __setitem__(self, key, value):
...
         """
-        h = hashlib.sha1()
-        h.update(repr(self.kwargs).encode())
-        h.update(self.code_definitions().encode())
-        return h.hexdigest()
+        # The full name of this shader class.
+        fullname = self.__class__.__module__ + "." + self.__class__.__name__
+
+        # Does it look like its builtin (into a lib)? This excludes e.g. __main__.CustomShader
+        is_builtin = fullname.count(".") >= 2
```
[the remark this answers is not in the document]
almarklein: In general it should be safe to use the name. The only case where it breaks down is when the class is re-defined, as can happen in an interactive session. This is easily checked by counting dots. It's not 100% bulletproof, since in theory a module can be reloaded, but that's rare and bad practice anyway.

I will change the comments here to clarify.

pygfx/renderers/wgpu/_utils.py
almarklein: `hash()` is faster, but only by a bit. Sha1 is more secure and maps to a larger space (40 chars representing 160 bits, rather than an int representing 64 bits). Also because we assume the hash is a string, we can just use the json string when its short, saving the hash step altogether.

pygfx/renderers/wgpu/_shaderbase.py
```diff
@@ -262,14 +262,33 @@ def __setitem__(self, key, value):
...
+        # optimization.
+
+        name_probably_defines_code = fullname.count(".") >= 2
+
+        if name_probably_defines_code:
+            # Faster, but assumes that the produced code only depends on kwargs.
+            return hash_from_value([fullname, self.code_definitions(), self.kwargs])
+        else:
+            # More reliable (e.g. in an interactove session).
+            return hash_from_value([self.get_code(), self.kwargs])
```
almarklein, line 291: Better?

examples/other/collection_line.py
almarklein: I added it to the renderer - when `show_fps` is set, it also prints the time to the first draw. Also ref #528.

pygfx/renderers/wgpu/_utils.py
Korijn: Ok. I don't think security is a concern here, and I also don't think we will have to worry about hash collisions as the number of objects is not going into the millions.

almarklein: That's fair. I'll update to use `hash()` instead.

Korijn (approved):

PR #527, opened 2023-05-09
Fix ups

almarklein:
Fix some leftover references to up. See #482.

* [x] Fixed some docstrings that still mentioned minus/reverse from when `reference_up`  was called `gravity`.
* [x] Fixed that `ob.world.reference_up`  was initially a tuple (its an array everywhere else).
* [x] Fixed that setting `ob.world.reference_up` was undone when the local transform changed (the change in `world.reference_up` was not propagated to the `local` transform.
* [x] Added a test to make sure the above stays working, based using `camera.show_pos`.
* [x] Prefixed two callback functions with an underscore.  
* [x] Update all (I think) cases that still used `ob.up` instead of `ob.world.reference_up`.

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -503,7 +503,7 @@ class RecursiveTransform(AffineBase):
 
     This transform behaves semantically identical to an ordinary
     ``AffineTransform`` (same properties), except that users may define a
-    ``parent`` transform which proceeds the ``matrix`` used by the ordinary
+    ``parent`` transform which preceeds the ``matrix`` used by the ordinary
     ``AffineTransform``. The resulting ``RecursiveTransform`` then controls the
     total transform that results from combinign the two transforms via::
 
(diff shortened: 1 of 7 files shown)


pygfx/utils/transform.py
```diff
@@ -590,30 +590,38 @@ def __init__(
...
-        self.own._reference_up = own_ref - parent_origin
-
+    def _parent_updated(self, parent: AffineBase):
+        self._propagate_reference_up()
         self.flag_update()
 
+    def _propagate_reference_up(self):
+        new_ref = la.vec_transform(self.reference_up, self.parent.inverse_matrix)
+        origin = la.vec_transform((0, 0, 0), self.parent.inverse_matrix)
+        self.own._reference_up = new_ref - origin
```
Korijn, line 612: does this resulting vector potentially need normalizing?

tests/cameras/test_cameras.py
```diff
@@ -79,6 +79,35 @@ def test_camera_show_methods():
...
+    camera.show_pos((0, 1, 0), up=(1, 0, 0))
+    assert tuple(camera.world.reference_up) == (1, 0, 0)
+
+    # Not specifying up, keeps the current up (no revert to default)
+    camera.show_pos((0, 1, 1))
+    assert tuple(camera.world.reference_up) == (1, 0, 0)
+
+    # Back to +Y
+    camera.show_pos((1, 0, 0), up=(0, 1, 0))
+    assert tuple(camera.world.reference_up) == (0, 1, 0)
```
Korijn, line 108: this test relies on the fact that show_pos sets up implicitly, I would recommend to also test by setting camera.world.reference_up directly.

pygfx/utils/transform.py
almarklein: Good point.

almarklein: Done!

tests/cameras/test_cameras.py
almarklein: Done!

Korijn (approved):

Issue #152, opened 2021-10-03
Mesh per-face coloring


almarklein:
Proposing to add something like `geometry.face_colors` that specifies a color for each triangle (as defined by the indices). Implementing per-face colors is normally not trivial, but because of how we render meshes, we have access to the face index :)  so this is easy to implement.

ivoflipse:
Here's an example of what I've done in the past
![2016-11-10 11_35_00-Figure 6~2.png](https://user-images.githubusercontent.com/485186/135750741-a69417ea-b745-44e3-b7b4-5ef7f6338ed2.png)

It's the distance between 2 meshes drawn on top of one of the two meshes

Korijn:
But that's vertex coloring, not face coloring

almarklein:
In #529 face colors are implemented, but we should take some time to think about the API.

cc @sharon92

### Current color modes

* uniform color (`material.color`).
* mapped via per-vertex texcoords (`geometry.texcoords` is sampled with vertex index, and result used to sample in `material.map`).
* vertex colors (`geometry.colors` is sampled with vertex index).

API-wise, vertex colors are only activated when `material.vertex_colors` is set to `True`. The colormap mode is automatically activated when `material.map` is not `None`. It is assumed that there are `texcoords` on the geometry (will error if not).

### New face color mode(s)

Ideally, I'd like to add:

* mapped via per-face texcoords (`geometry.texcoords` is sampled with face index, and result used to sample in `material.map`).
* face colors (`geometry.colors` is sampled with face index).

Some options for the API that I could think of:

#### Option 1
We could introduce a `material.face_colors` property, but then we can't distinguish between the different colormapped modes.

#### Option 2

We could introduce `material.color_mode` to explicitly set the mode ("uniform", "map", "vertex", "face-map", "face").

Perhaps including mode "auto" for backwards compatibility, that switches between "uniform" and "map" depending on whether `material.map` is present. 

#### Option 3

We could decide whether to use `geometry.colors` as vertex or face colors based on the size of the buffer: if it matches the number of vertices, they are vertex colors, if it matches the number of faces, they are face colors. Same for texcoords. We can then rename `vertex_colors` to `geometry_colors`.

I thought it might feel a bit too magical at first, but it is easy to document this behavior and JustWorks.

@Korijn thoughts?

Korijn:
[...]

PR #529, opened 2023-05-11
Quad drawing in Mesh Shader, when the input index has shape(nx4)

sharon92:


pygfx/renderers/wgpu/meshshader.py
```diff
@@ -68,9 +68,18 @@ def get_bindings(self, wobject, shared):
         if getattr(geometry, "normals", None) is not None:
             normal_buffer = geometry.normals
         else:
-            normal_data = normals_from_vertices(
-                geometry.positions.data, geometry.indices.data
-            )
+            if geometry.indices.data.shape[1] == 4:
```
Korijn, line 71:
```suggestion
            if geometry.indices.data.shape[-2] == 4:
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -68,9 +68,18 @@ def get_bindings(self, wobject, shared):
...
+                n1 = normals_from_vertices(
+                    geometry.positions.data, geometry.indices.data[:,:3]
```
Korijn, line 73:
```suggestion
                    geometry.positions.data, geometry.indices.data[..., :3]
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -68,9 +68,18 @@ def get_bindings(self, wobject, shared):
...
+                )
+                n2 = normals_from_vertices(
+                    geometry.positions.data, geometry.indices.data[:,[0,2,3]]
```
Korijn, line 76:
```suggestion
                    geometry.positions.data, geometry.indices.data[..., [0,2,3]]
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -68,9 +68,18 @@ def get_bindings(self, wobject, shared):
...
+                )
+                normal_data = (n1+n2)/2
```
Korijn, line 78: do you really need to compute the normals for both triangles of every quad though? aren't they always going to be equal for each triangle that makes up the quad?

I'd say you can just compute the normals for one triangle and work with that, right?

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -131,6 +140,11 @@ def get_render_info(self, wobject, shared):
         material = wobject.material
 
         n = geometry.indices.data.size
+        self['indexer'] = 3
+        if geometry.indices.data.shape[1]==4:
+            n =  geometry.indices.data.shape[0]*2*3
+            self['indexer'] = 6
```
Korijn, line 146:
```suggestion
        if geometry.indices.data.shape[-2] == 4:
            n =  geometry.indices.data.shape[0] * 2 * 3
            self['indexer'] = 6
```

pygfx/renderers/wgpu/meshshader.py
almarklein: Do you mean -1?

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -210,8 +246,18 @@ def code_vertex(self):
...
+            var sub_index = index % {{indexer}};
+            
+            // for quads assuming the vertices are oriented, the triangles are 0 1 2 and 0 2 3
+            if sub_index == 3 {
+                sub_index = 0;
+                } else if sub_index == 4 {
+                sub_index = 2;
+                } else if sub_index == 5 {
+                sub_index = 3;
+                }
```
almarklein, line 260: Its good practice to avoid if-statements in shaders. Maps are usually more performant. And in this case the code is actually simpler :)

Also, lets use templating to only use this code for quads.

```suggestion
            // for quads assuming the vertices are oriented, the triangles are 0 1 2 and 0 2 3
            $$ if indexer == 3
            var quad_map = array<f32,6>(0, 1, 2, 0, 2, 3);
            sub_index = quad_map[sub_index];
            $$ endif
```

almarklein, line 260: Its good practice to avoid if-statements in shaders. Maps are usually more performant. And in this case the code is actually simpler :)

Also, lets use templating to only use this code for quads.

```suggestion
            // for quads assuming the vertices are oriented, the triangles are 0 1 2 and 0 2 3
            $$ if indexer == 3
            var quad_map = array<f32,6>(0, 1, 2, 0, 2, 3);
            sub_index = quad_map[sub_index];
            $$ endif
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -221,7 +267,7 @@ def code_vertex(self):
             // set with the public API and we assume that it does not include a flip.
             let winding_world = get_sign_of_det_of_4x4(world_transform);
             let winding_cam = get_sign_of_det_of_4x4(u_stdinfo.cam_transform);
-            sub_index = select(sub_index, -1 * (sub_index - 1) + 1, winding_world * winding_cam < 0.0);
+            //sub_index = select(sub_index, -1 * (sub_index - 1) + 1, winding_world * winding_cam < 0.0);
```
almarklein, line 270: This was not meant to be removed, I think?
```suggestion
            sub_index = select(sub_index, -1 * (sub_index - 1) + 1, winding_world * winding_cam < 0.0);
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -234,11 +280,19 @@ def code_vertex(self):
...
+                        let world_pos{{ i }} = world_transform * vec4<f32>(raw_pos{{ i }}, 1.0);
+                        let ndc_pos{{ i }} = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos{{ i }};
+                    $$ endfor
+                $$ elif indexer == 6
+                    $$ for i in (1, 2, 3, 4)
+                        let raw_pos{{ i }} = load_s_positions(i32(ii[{{ i - 1 }}]));
+                        let world_pos{{ i }} = world_transform * vec4<f32>(raw_pos{{ i }}, 1.0);
+                        let ndc_pos{{ i }} = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos{{ i }};
+                    $$ endfor
+                $$ endif
```
almarklein, line 295: Would something like this work? Otherwise we can introduce a template var `verts_per_face`.
```suggestion
            $$ for i in ((1, 2, 3) if indexer == 3 else (1, 2, 3, 4))
                let raw_pos{{ i }} = load_s_positions(i32(ii[{{ i - 1 }}]));
                let world_pos{{ i }} = world_transform * vec4<f32>(raw_pos{{ i }}, 1.0);
                let ndc_pos{{ i }} = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos{{ i }};
            $$ endfor
```

almarklein, line 295: Would something like this work? Otherwise we can introduce a template var `verts_per_face`.
```suggestion
            $$ for i in ((1, 2, 3) if indexer == 3 else (1, 2, 3, 4))
                let raw_pos{{ i }} = load_s_positions(i32(ii[{{ i - 1 }}]));
                let world_pos{{ i }} = world_transform * vec4<f32>(raw_pos{{ i }}, 1.0);
                let ndc_pos{{ i }} = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos{{ i }};
            $$ endfor
```

almarklein, line 295: Would something like this work? Otherwise we can use a new template var `verts_per_face`.
```suggestion
                $$ for i in ((1, 2, 3) if indexer == 3 else (1, 2, 3, 4))
                    let raw_pos{{ i }} = load_s_positions(i32(ii[{{ i - 1 }}]));
                    let world_pos{{ i }} = world_transform * vec4<f32>(raw_pos{{ i }}, 1.0);
                    let ndc_pos{{ i }} = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos{{ i }};
                $$ endfor
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -308,9 +396,18 @@ def code_vertex(self):
 
             varyings.pick_id = u32(pick_id);
             varyings.pick_idx = u32(face_index);
-            var arr_pick_coords = array<vec3<f32>, 3>(vec3<f32>(1.0, 0.0, 0.0), vec3<f32>(0.0, 1.0, 0.0), vec3<f32>(0.0, 0.0, 1.0));
-            varyings.pick_coords = vec3<f32>(arr_pick_coords[sub_index]);
-
+            if {{indexer}} == 3 {
```
almarklein, line 399: Using templating (`$$ if indexer == 3`) would make the resulting shading code simpler, and likely more efficient.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -308,9 +396,18 @@ def code_vertex(self):
...
+                var arr_pick_coords = array<vec3<f32>, 3>(vec3<f32>(1.0, 0.0, 0.0), vec3<f32>(0.0, 1.0, 0.0), vec3<f32>(0.0, 0.0, 1.0));
+                varyings.pick_coords = vec3<f32>(arr_pick_coords[sub_index]);
+            } else {
+                var arr_pick_coords = array<vec3<f32>, 6>(vec3<f32>(1.0, 0.0, 0.0),
+                                                          vec3<f32>(0.0, 1.0, 0.0),
+                                                          vec3<f32>(0.0, 0.0, 1.0),
+                                                          vec3<f32>(1.0, 0.0, 0.0),
+                                                          vec3<f32>(0.0, 1.0, 0.0),
+                                                          vec3<f32>(0.0, 0.0, 1.0));
+                varyings.pick_coords = vec3<f32>(arr_pick_coords[sub_index_ori]);
```
almarklein, line 409: I think we'd need changes elsewhere to make picking work for quads.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -192,11 +206,33 @@ def code_vertex(self):
...
             let m3 = mat3x3<f32>(m[0].xyz, m[1].xyz, m[2].xyz);
             return sign(determinant(m3));
         }
 
+        fn dist_pt_line(x1: f32, y1: f32, x2: f32, y2: f32, x3: f32, y3: f32) -> f32 {
+         // Distance of pt (x3,y3) to line with coords(x1,y1) (x2,y2)
+         return abs((x2 - x1) * (y1 - y3) - (x1 - x3) * (y2 - y1)) / sqrt((x2-x1)*(x2-x1) + (y2-y1)*(y2-y1));
+        }
+
+        fn dist_pt_line_segment(x1: f32, y1: f32, x2: f32, y2: f32, x3: f32, y3: f32) -> f32 {
```
almarklein, line 219: Looks like this function is not used.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -285,16 +339,50 @@ def code_vertex(self):
...
-                    vec3<f32>(dist1, 0.0, 0.0), vec3<f32>(0.0, dist2, 0.0), vec3<f32>(0.0, 0.0, dist3)
-                );
-                varyings.wireframe_coords = vec3<f32>(arr_wireframe_coords[sub_index]);  // in logical pixels
+                $$ if indexer == 3
+                    $$ for i in (1, 2, 3)
+                        let p{{ i }} = (ndc_pos{{ i }}.xy / ndc_pos{{ i }}.w) * u_stdinfo.logical_size * 0.5;
+                    $$ endfor
+                    let dist1 = abs((p3.x - p2.x) * (p2.y - p1.y) - (p2.x - p1.x) * (p3.y - p2.y)) / distance(p2, p3);
+                    let dist2 = abs((p3.x - p1.x) * (p1.y - p2.y) - (p1.x - p2.x) * (p3.y - p1.y)) / distance(p1, p3);
+                    let dist3 = abs((p1.x - p2.x) * (p2.y - p3.y) - (p2.x - p3.x) * (p1.y - p2.y)) / distance(p2, p1);
```
almarklein, line 348: Let's also use `dist_pt_line` here 👍

almarklein (commented):
This is awesome!

Most of my comments are shader-specific details.

As @Korijn said, it woud be great to have an example that just used the `gui.auto`. Perhaps using some data generated by the example itself. Or you can put data in the `examples/data` folder. An example showing both normal shaded quads as well as the wireframe would be really nice 😄 ! Could be two separate examples if that makes more sense for you.

almarklein:
BTW: this repo is autoformatted with `black`, and we check linting errors with flake8. You can install both using `pip install -U black flake8`. Then you can use `black .`  and `flake8 .` to autoformat and lint, respectively.

pygfx/renderers/wgpu/meshshader.py
sharon92: hmm I think it wont work for ndim >2 or? but anyways yes it should be -1

Korijn: You're right! I updated the suggestions

pygfx/renderers/wgpu/meshshader.py
sharon92: I am averaging them because of this post:
https://stackoverflow.com/questions/5410143/calculating-face-normals-with-4-vertices

The quads in my mesh are not planar so I average them.

pygfx/renderers/wgpu/meshshader.py
sharon92: yep forgot to uncomment it sorry!

pygfx/renderers/wgpu/meshshader.py
sharon92: yes i left it there to see if it makes a difference but it doesnt I will remove it.

pygfx/renderers/wgpu/meshshader.py
Korijn: I never realized such quad geometry existed, I always assumed they would be planar! In that case, it's possible the two triangles have different surface area, correct? Should it be a weighted average to correct for that, or is the error assumed to be very small and thus can it be safely ignored?

Korijn: After looking at this again, it's already correct.

sharon92: Yes the area of both the triangles are different. These are irregular convex quads so the normals can be a bit different in both the triangles. I read that cross product of the diagonals would be the most apt but I haven't tested that.

Korijn: I think this should do the trick; first you triangulate all the quads, then you compute the vertex normals based on the triangles in a single step. Now the weighted sum will be correct.

```suggestion
                triangle_indices = np.concatenate([
                    geometry.indices.data[:, :3],
                    geometry.indices.data[:, [0,2,3]],
                ], axis=0)
                normal_data = normals_from_vertices(
                    geometry.positions.data,
                    triangle_indices,
                )
```

Korijn: I played around with this, there is an improvement to be made, I added a separate suggestion comment below.

sharon92:
> Thanks for contributing!
> 
> Some feedback in order to get this merged:
> 
> * You'll have to remove the `quad_test` folder and create a minimal example script in the `examples` folder instead.
> * Since you've changed the implementation of `MeshBasicMaterial` to support quads geometry, we need to check and evaluate what happens if you disable wireframe mode. Does that also work properly?

Yep, without the wireframe mode it paints the quads as expected.

pygfx/renderers/wgpu/meshshader.py
sharon92: corrected this to 
```
   $$ if indexer == 6
            var quad_map = array<i32,6>(0, 1, 2, 0, 2, 3);
```

pygfx/renderers/wgpu/meshshader.py
sharon92: I updated _wgpu_get_pick_info for quads

sharon92:
Hello @almarklein @Korijn with your latest commit, I can't sync my fork because of conflicts. And I am not sure if should delete this pull request or make a new one.

This pull request is now kind of stale.

Would you be interested if I make a new one? Although my fork has tweaks that you may not want:
- Quads Support for Mesh Objects i.e indices with shape nx4 instead of nx3 in Geometry object.
- Qt6 Example for Quads drawing
- Added "face_color" option for Mesh objects. This way it's possible to color each triangle or a quad from a unique ID in map. Texcoords are supplied with the map keys and are the same shape as number of triangles/quad in the mesh object.
- Tweaked "picking_info" so that if a quad is picked, additionally the corresponding "sub triangle" can be identified with the keyword "face_index"

While it would be comfortable for me to merge my fork to main, and just use the main branch, this is not an absolute necessity. 
Thank you either way for your help!

almarklein:
Hey @sharon92,

I resolved the conflict (new commit at the end).

I think it would be good to get these contributions merged. We're definitely interested in both the quads and the face colors. Ideally we'd have one PR for each, but I think it may be a bit tricky to untangle, and they're not huge changes, so its fine to do in this single PR 👍 

I had a quick look over the changes, and have two main points. Other than that I have a few more questions, but nothing we can't resolve. 

### The example

One problem is that we prefer our examples to be minimal and using the `wgpu.gui.auto` backend so that they can run for everyone. If the current example can be replaced with a simple example that demonstrates drawing a mesh with quads, that'd be great!

If your new example still uses the added data, we'd have to make sure that the license for the data allows us to redistribute it.

### Face colors

Your implementation now uses an indirection via the texcoords rather then using the colors directly. This is inconsistent with how colors currently work. Ideally it'd have both the options to use per-face texcoords (and lookup in the `material.map`) as well as per-face colors. But we'd need to discuss the API for that. Let's do that in #152.

sharon92:
@almarklein Thanks for resolving the conflict!

**Example**
I have made the example minimal, and now generate the quads inside the script.

wireframe:
image

color by vertex:
image

color by face:
image

also picking info will be printed in console.

`wgpu.gui.auto` somehow doesnt work for me, I removed the line for rust backend, and it still works, would that be enough?

**face colors**
I think i implemented look up colors in `geometry.colors` because implementing this in `materials.map` was complicated for me in the mesh shader code. But i definitely think that implementing it in `materials.map `will be a cleaner solution.

almarklein:
> `wgpu.gui.auto` somehow doesnt work for me, I removed the line for rust backend, and it still works, would that be enough?

When I meant "keep them simple" I also meant without using Qt. Sorry, I could've been more clear 😄. Should be relatively easy to do from the current state of the example. Basically remove the Qt stuff, and replace button click handlers with something like this:

```py
@renderer.add_event_handler("key_down")
def on_key(e):
    if e.key == "1":
        make_wireframe()
    if e.key == "2":
        make_vertex_color()
    ...
```

almarklein:
> I think i implemented look up colors in geometry.colors because implementing this in materials.map was complicated for me in the mesh shader code. But i definitely think that implementing it in materials.map will be a cleaner solution.

I posted a summary for the colors in #152, which may be a bit more than what you opted for. To be honest I'd rather focus on polishing the quads logic. If @Korijn agrees we keep the color logic as it is now, and refactor it to use `color_mode` in a later PR.

Korijn:
> > I think i implemented look up colors in geometry.colors because implementing this in materials.map was complicated for me in the mesh shader code. But i definitely think that implementing it in materials.map will be a cleaner solution.
> 
> I posted a summary for the colors in #152, which may be a bit more than what you opted for. To be honest I'd rather focus on polishing the quads logic. If @Korijn agrees we keep the color logic as it is now, and refactor it to use `color_mode` in a later PR.

As long as we finish that before the next release :)

sharon92:
@almarklein I have removed qt and updated the example.

almarklein (commented):
My only suggestions are related to the picking mechanics. In particular to how picking info is encoded. Going back to 20 bits instead of 26 to encode the face index, means that a mesh with over 1M faces cannot be reliably picked. 1M is a lot, but not unrealistic at all. I think we can make this work without sacrificing any bits though!

### Encoding pick info

Instead of introducing the `pick_face` element, I'd like the `pick_idx` to be:
```py
pick_idx = u32(face_index) * 2 + face;  // face is the triangle-index, maybe we can rename it for clarity?
```

And then the `pick_coords` can simply refer to the real picking coords for that triangle, so I think the triage around line 430 can be dropped. 


### Decoding pick info

I think that the most convenient way for the user to receive pick info (for a quad mesh) is that the pick coords is 4 elements instead of 3 (of which one is always zero). This is something we can discuss though.

In any case, I think we need to know in  `_wgpu_get_pick_info()` whether the mesh consists of quads or triangles. The material cannot know this. Therefore we should move `_wgpu_get_pick_info()` to the `Mesh` class in `objects/more.py`. (There are more objects that have their picking decoding logic on the WorldObject, I suspect we'll eventually move all of them there...)

The decoder would then become something like:
```py
    def _wgpu_get_pick_info(self, pick_value)
        values = unpack_bitfield(
            pick_value, wobject_id=20, index=26, coord1=6, coord2=6, coord3=6
        )
        face_index = values["index"]
        face_coord = [
            values["coord1"] / 64,
            values["coord2"] / 64,
            values["coord3"] / 64,
        ]
        if self.geometry.indices.data is not None and self.geometry.indices.data.shape[-1] == 4:
            triangle_index = face_index % 2
            face_index = face_index // 2
            if triangle_index == 0:
                face_coord.append(0.0)  # 0, 1, 2, << 3 >>
            else:
                face_coord.insert(1, 0.0)  # 0, << 1 >>, 2, 3

        return {"face_index": face_index, "face_coord": tuple(face_coord)}
```

sharon92:
> means that a mesh with over 1M faces cannot be reliably picked

This explains why picking for big meshes was so strange and didnt quite work! :') 

I have updated the fork with your suggestions and some corrections.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -333,9 +413,15 @@ def code_vertex(self):
...
-            varyings.pick_coords = vec3<f32>(arr_pick_coords[sub_index]);
+            varyings.pick_idx = u32(face_index*2 + face_sub_index);
+
+            var arr_pick_coords = array<vec3<f32>, 6>(vec3<f32>(1.0, 0.0, 0.0),
+                                                      vec3<f32>(0.0, 1.0, 0.0),
+                                                      vec3<f32>(0.0, 0.0, 1.0),
+                                                      vec3<f32>(1.0, 0.0, 0.0),
+                                                      vec3<f32>(0.0, 1.0, 0.0),
+                                                      vec3<f32>(0.0, 0.0, 1.0));
+            varyings.pick_coords = vec3<f32>(arr_pick_coords[sub_index_ori%3]);
```
almarklein, line 424: This should work now, I think. But there appears something else is broken, so I haven't been able to test it properly.
```suggestion
            var arr_pick_coords = array<vec3<f32>, 3>(vec3<f32>(1.0, 0.0, 0.0),
                                                      vec3<f32>(0.0, 1.0, 0.0),
                                                      vec3<f32>(0.0, 0.0, 1.0),
                                                      );
            varyings.pick_coords = vec3<f32>(arr_pick_coords[sub_index]);
```

almarklein (commented):
Two small comments. But more importantly, when I run the example, I get this, which does not seem quite right ... 🤔 

image

almarklein:
I was on holiday, so could not check earlier.

> Could the example issue originate from bleeding edge wgpu? I can't reproduce it, I have the wgpu version from pypi.

I tried with wgpu-py from pypi, but nope, same result :)


> Can you move around the scene? because there is a controller implemented. Can you switch between modes when pressing 1,2,3?

Everything works as expected, except that it's a saw-shape instead of a strip of quads.

I suspect there's a subtle detail in the shader that makes certain values zero. I'll have a more proper look early next week.

almarklein:
Is this correct for nr 3?

image

The issue was that the indices were int64, which the GPU cannot handle. Why it does work for you is beyond me. We should add a check to guard against that ... I created #579 for that. For this PR, you can simply `indices.astype(np.int32)` somewhere.

Korijn:
Just wanted to chime in and show some appreciation here; the persistence and collaboration is a joy to see! Awesome to have quad support in pygfx soon. We're almost there!

sharon92:
@almarklein that doesn't look right. 

The face colour should be a gradient stop from white to black (higher Z to lower Z values). 
There must be something else going on.

I did a fresh install on my home pc, and it's working as expected. 

https://github.com/pygfx/pygfx/assets/51754340/65d36b7a-3749-4b2d-9b45-42e97148597f

almarklein:
> @almarklein that doesn't look right.

I was afraid it didn't 😄. I looked into it, and it's the same problem: the texcoords are int64.

sharon92:
> > @almarklein that doesn't look right.
> 
> I was afraid it didn't 😄. I looked into it, and it's the same problem: the texcoords are int64.

It's weird that it works on both my work and home computer. 
Maybe my python + numpy version converts the ints to int32 and yours dont?
or my gpu supports int64?

I tried starting the example by defining the dtype as int64 and it doesnt load.

```
    gfx.Geometry(indices=indices.astype(np.int64),
  File "C:\Python311\Lib\site-packages\pygfx\geometries\_base.py", line 60, in __init__
    format = resource.format
             ^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\pygfx\resources\_buffer.py", line 140, in format
    self._store["format"] = format_from_memoryview(self.mem)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\pygfx\resources\_buffer.py", line 236, in format_from_memoryview
    raise TypeError(
TypeError: Cannot convert 'q' to vertex format. Maybe specify format?
```

I am using python 3.11.3, numpy > 1.24, gpu: nvidia 1030gt on my work computer
Home pc - python 3.11.3 numpy > 1.24, gpu amd radeon 570

almarklein:
Wow, there's two things at play here: I think indeed your numpy version defaults to int32 for the arrays created in the example.

Also, on my system, when the int64 array is converted to a memoryview, that memotyview reports a format of "l", which means 32bit. See https://docs.python.org/3/library/struct.html#format-characters. This is strange. Your machine reports "q" as expected.

sharon92:
@almarklein you must be using Linux right? I read that on windows the numpy dtype defaults to in32 unless it cant be represented as int32. On other platforms it defaults to int64.

But I guess thats a good bug caught on memory view :)

almarklein:
> I read that on windows the numpy dtype defaults to in32 unless it cant be represented as int32. On other platforms it defaults to int64.

How convenient 😕. So if a Windows user posts code online, any Linux/MacOS user that want to run that code, may get an error from pygfx ...

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -236,8 +269,18 @@ def code_vertex(self):
...
+            var face_sub_index = 0;
+
+            // for quads assuming the vertices are oriented, the triangles are 0 1 2 and 0 2 3
+            $$ if indexer == 6
+                var quad_map = array<i32,6>(0, 1, 2, 0, 2, 3);
+                //face_sub_index returns 0 or 1 in picking for quads. So the triangle of the quad can be identified.
+                var face_map = array<i32,6>(0, 0, 0, 1, 1, 1);
+                face_sub_index = face_map[sub_index];
+                sub_index = quad_map[sub_index];
+            $$ endif
```
almarklein, line 283: The sub_index tweaks to fix the winding just below also needs to be adjusted. If I flip the camera scale, I get nice art!

image


Looking into it.

pygfx/materials/_mesh.py
```diff
@@ -108,6 +96,15 @@ def vertex_colors(self):
     def vertex_colors(self, value):
         self._store.vertex_colors = bool(value)
 
+    @property
+    def face_colors(self):
+        """Whether to use the vertex colors provided in the geometry."""
```
sharon92, line 101: """Whether to use the face colors provided in the geometry."""

almarklein: Will include this fix.

almarklein:
Ok, I tested the example again:
* Added `dtype` where needed to make it work.
* Moved example to feature directory and updated the docstring so it appears nicely in the gallery docs.
* I found that the winding-fixing did not work for quads, so I fixed that.
* The picking was also not correct - because I have wrong/incomplete directions - they now also work.

almarklein (approved):
LGTM!

sharon92:
@almarklein Thanks a lot for the fixes! I am excited to see this in the gallery section! :D

Korijn (approved):

Korijn:
Many thanks for your contribution @sharon92 ! <3

almarklein:
Yeah, great to see these improvements landed!

PR #535, opened 2023-05-16
Final caching tweaks

### LINKED ISSUE #322: Caching of shader modules and pipeline objects

almarklein:
This closes #322.

* [x] In blender/flusher: No need to store bind group layouts - the pipeline remembers them. 
* [x] Faster mipmaps by storing bind group objects.
* [x] Did some ~research~ testing to check whether caching texture views and bind groups helps performance. -> it does, especially for the bind group.
* [x] Validated that no new objects are created at draw time (except for the first draw). Disclaimer: I put a print-statement in all the `create_` methods in wgpu-py and ran two examples (`lights_shadow.py`and `dynamic_env_map.py`, which cover light, shadows, mipmaps, and more).

### DIFF
--- a/examples/feature_demo/dynamic_env_map.py
+++ b/examples/feature_demo/dynamic_env_map.py
@@ -82,8 +82,6 @@ camera.show_object(scene, scale=2)
 
 controller = gfx.OrbitController(camera, register_events=renderer)
 
-etimes = []
-
 
 def animate():
     t = time.time()
@@ -105,12 +103,7 @@ def animate():
 
     ob2.local.rotation = la.quat_mul(rot, ob2.local.rotation)
 
-    t0 = time.perf_counter()
     cube_camera.render(scene)
-    etime = time.perf_counter() - t0
-    etimes.append(etime)
-    etimes[:-20] = []
-    print(sum(etimes))
     renderer.render(scene, camera)
     renderer.request_draw()
 
--- a/examples/feature_demo/dynamic_env_map.py
+++ b/examples/feature_demo/dynamic_env_map.py
@@ -82,6 +82,8 @@ camera.show_object(scene, scale=2)
 
 controller = gfx.OrbitController(camera, register_events=renderer)
 
+etimes = []
+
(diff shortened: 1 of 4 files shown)


kushalkolar:
Going to try fpl with this soon (once I catch up with pylinalg refactor). You guys are amazing! :D

Korijn (approved):

PR #536, opened 2023-05-17
Refactor usage of global device

### LINKED ISSUE #531: How to use the device object?
## Purpose
In pygfx we make use of a single global `device` object. One device per process. The motivation is that this allows sharing GPU data easily. This does mean that users won't be able to cleanly isolate GPU stuff, but I'm pretty sure that's ok.

This means that we can define one global, and use that throughout the code. Nice and simple. We do this in a few cases. In most cases, however, the code makes no assumptions about the origin of the device, and it's passed along from one function to the next. The motivation for this has been to allow tests to be performed on a separate device.

Let's do a bit of research to see whether there's actually a benefit for allowing a separate device. If so, we may want to adjust code to clearly mark points where a custom device can be passed. My expectation is that there is not that much of a benefit though, and we can simplify the code in quite a few places.

FWIW isolated tests are still possible by running then in a subprocess.

## Notes

There is `wgpu.utils.get_default_device()`, but we don't use that in pygfx. 

In Pygfx we have the `Shared` class that contains some global stuff for the renderer. An instance of this class is instantiated when the first WgpuRenderer is instantiated. That shared object creates the device:
https://github.com/pygfx/pygfx/blob/f3816b29915422fe18235c194473d9a5c3bfb29a/pygfx/renderers/wgpu/_shared.py#L51-L56

I think it makes sense to only create a device when a renderer is created, e.g. so pygfx can at least be imported (and have some of the units test run) on systems without a backend.

I think we could do something like: create a function that returns the global device object. It creates the device the first time it is called.

almarklein:
Closes #531 

* [x] Introduces a new function `get_shared()` that returns the globally shared object. This object contains the global adapter, device, std uniform buffer, access to the glyph atlas, and maybe more in the future. The first invocation creates the singleton.
* [x] Adds a test to make sure that `get_shared()` is not called during import.
* [x] Refactor code to make use of it. Some rules of thumbs that I used:
  * The device object is not passed as an argument to a function that is defined in another module.
  * Within a module avoid sprinkling `get_shared()` calls by passing the object as an argument (to the module-level private functions).
  * When the logic is structured in a class (e.g. blender, flusher, renderer), store the device on the object for easy use.

### DIFF
--- a/pygfx/renderers/wgpu/__init__.py
+++ b/pygfx/renderers/wgpu/__init__.py
@@ -3,7 +3,7 @@
 # The engine
 from ...objects._base import RenderMask
 from ._utils import registry, register_wgpu_render_function, gpu_caches
-from ._shared import get_shared, stdinfo_uniform_type, print_wgpu_report
+from ._shared import Shared, get_shared, stdinfo_uniform_type, print_wgpu_report
 from ._renderer import WgpuRenderer
 from ._pipeline import Binding
 from ._shader import BaseShader, WorldObjectShader
--- a/pygfx/renderers/wgpu/__init__.py
+++ b/pygfx/renderers/wgpu/__init__.py
@@ -3,7 +3,7 @@
 # The engine
 from ...objects._base import RenderMask
 from ._utils import registry, register_wgpu_render_function, gpu_caches
-from ._shared import Shared, stdinfo_uniform_type, print_wgpu_report
+from ._shared import get_shared, stdinfo_uniform_type, print_wgpu_report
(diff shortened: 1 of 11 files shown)


Korijn (approved):
I don't mind the change but the improvement is also not obvious to me

almarklein:
The main pain point (as I experienced it) was that the device is a global object, but was also passed around from one function to another a lot. My intention with this PR is to make its use more obvious, and tracking the device less of a problem, leading (I hope) to clearer code.

PR #541, opened 2023-05-25
Show line caps on line segments separated by nan

almarklein:
A convenient feature of lines is that one can introduce a nan-point to separate two line segments. This feature is also supported by e.g. matplotlib. This just happened to work for our current line shader 😄. This PR also adds caps to these loose line endings.

Implementation note: apparently wgsl has dropped `isNan` and co. The `isFinite` still exists, but it has been removed from Naga januari this year, which is why I rolled our own little `is_nan()`. See e.g. https://github.com/gpuweb/gpuweb/pull/2311.

### DIFF
--- a/pygfx/renderers/wgpu/lineshader.py
+++ b/pygfx/renderers/wgpu/lineshader.py
@@ -191,6 +191,10 @@ class LineShader(WorldObjectShader):
             let world_pos = u_wobject.world_transform * vec4<f32>(raw_pos.xyz, 1.0);
             return u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos;
         }
+
+        fn is_nan(v:f32) -> bool {
+            return !(0.0 < v) && !(0.0 > v);
+        }
         """
 
     def code_vertex(self):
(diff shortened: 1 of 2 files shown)


pygfx/renderers/wgpu/lineshader.py
```diff
@@ -191,6 +191,10 @@ def code_helpers(self):
             let world_pos = u_wobject.world_transform * vec4<f32>(raw_pos.xyz, 1.0);
             return u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos;
         }
+
+        fn is_nan(v:f32) -> bool {
+            return !(0.0 < v) && !(0.0 > v);
```
Korijn, line 196: Wouldn't this return true for 0.0?

Korijn (approved):

pygfx/renderers/wgpu/lineshader.py
almarklein: Sharp!

almarklein:
I think I need to do a bit more reading to make sure we can actually reliably check for nan.

Korijn:
> I think I need to do a bit more reading to make sure we can actually reliably check for nan.

You could glean the implementation from naga?

almarklein:
Let's go with the current solution, and look into the global if this turns out not to work on some devices. I don't like the idea to add buffers or increase buffer sizes just to implement this.

BTW:
image

Korijn (approved):

PR #542, opened 2023-05-26
Option to make a solid torus knot

almarklein:
* [x] Implementation works, our morph-code (wip) confirms it is fully closed and correctly winded.
* [ ] The terminology and docs need some work.
  * The term "solid" does not seem to be used for what I mean by it here.
  * The term "closed" is better.
  * The mathematical term seems to be that the mesh is *manifold*, although I'm not sure yet whether that implies that its closed?

### DIFF
--- a/pygfx/geometries/_toroidal.py
+++ b/pygfx/geometries/_toroidal.py
@@ -98,7 +98,7 @@ def klein_bottle_surface(u, v):
 
 
 def torus_knot_geometry(
-    scale=1.0, tube=0.4, tubular_segments=64, radial_segments=8, p=2, q=3
+    scale=1.0, tube=0.4, tubular_segments=64, radial_segments=8, p=2, q=3, solid=False
 ):
     """Generate a torus knot.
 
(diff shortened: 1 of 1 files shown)


Korijn:
Some definitions:

* A boundary edge is part of exactly one face.
* A manifold edge is part of exactly two faces.
* A mesh is edge-manifold if each edge is bounding either one or two faces.
* A mesh is connected if it is exactly one connected component.
* A mesh is manifold if it is connected and edge-manifold.
* A mesh is oriented if it is manifold and no two faces are connected to an edge in the same direction (with respect to winding order).
* A mesh is closed if it is manifold and it contains no boundary edges.
* A mesh is orientable if it is manifold, and the faces can be oriented such that all faces are connected to an edge in opposite directions (with respect to winding order). An example of an unorientable mesh is a Klein Bottle.

almarklein:
I had a quick stab at making the Klein bottle actually closed (and non-orientable), but this proofed tricky. It would be an interesting test-object though! Maybe introduce a (proper) mobius strip first, which is open and non-orientable, and therefore also interesting for both testing our viz and the morph logic.

Korijn (approved):

PR #553, opened 2023-06-22
fix image plus points example, flip camera call

kushalkolar:
`camera.local.scale_y = 1` has to be called after `camera.show_rect()`, I guess it changed during the #482

### DIFF
--- a/examples/feature_demo/image_plus_points.py
+++ b/examples/feature_demo/image_plus_points.py
@@ -37,8 +37,8 @@ points = gfx.Points(
 scene.add(points)
 
 camera = gfx.PerspectiveCamera(0)
-camera.local.scale_y = -1
 camera.show_rect(-10, 522, -10, 522)
+camera.local.scale_y = -1
 
 controller = gfx.PanZoomController(camera, register_events=renderer)
 


almarklein:
Nice find. I think that `show_rect` should take any axii being flipped into account though - it should not be needed to flip the axis afterwards.

almarklein:
Correction, I think the correct way is to set `camera.local.reference_up = 0, -1, 0`, and the `show_` method honor that setting. The only issue is that `show_rect()`  has a bug for flipped axii. The other show methods do work as expected.

almarklein:
No wait, then it only changes what's up without actually flipping the axis, so the flag would be on the wrong side.

almarklein:
I thought I traced this down to the lookat function, and then the `transform.forward` setter. But then I found this:
```py
>>> camera.local.scale
array([1., 1., 1.])

>>> camera.local.scale_y = -1

>>> camera.local.scale
array([-1.,  1.,  1.])
```

which is probably because of this code in pylinalg:

https://github.com/pygfx/pylinalg/blob/287cdba8af366489f64636dc1f661d13a6813845/pylinalg/matrix.py#L360-L362
```py
    scaling[:] = np.linalg.norm(matrix[:-1, :-1], axis=0)
    if np.linalg.det(matrix) < 0:
        scaling[0] *= -1
```

almarklein:
We can close this, since this has been fixed by #563. All examples that show images are upside-up again 😄

Issue #474, opened 2023-03-05
events not called if WorldObject material color has an alpha value less than 1.0


kushalkolar:
Simplest way to reproduce this is with the `multi_selection.py` example. Set the color for `default_material` on line 35 to something with an alpha < 1.0, and events are not called. It is true for lines as well, so I'm guessing this happens across many types of objects.

`wgpu-py` version is `0.9.4`, pygfx version is the current head of `main` on github.

```python
from functools import partial
from random import randint, random

from wgpu.gui.auto import WgpuCanvas, run
import pygfx as gfx


canvas = WgpuCanvas()
renderer = gfx.renderers.WgpuRenderer(canvas)

camera = gfx.PerspectiveCamera(70, 16 / 9)
camera.position.z = 400

controller = gfx.OrbitController(camera.position.clone())
controller.add_default_event_handlers(renderer, camera)

scene = gfx.Scene()
scene.add(gfx.AmbientLight(), camera.add(gfx.DirectionalLight()))

geometry = gfx.box_geometry(40, 40, 40)
default_material = gfx.MeshPhongMaterial(color=(1., 0., 1., 0.2))
selected_material = gfx.MeshPhongMaterial(color="#FF0000")
hover_material = gfx.MeshPhongMaterial(color="#FFAA00")

outline = gfx.BoxHelper(thickness=3, color="#fa0")
scene.add(outline)

selected_objects = []


def set_material(material, obj):
    if isinstance(obj, gfx.Mesh):
        obj.material = material


def select(event):
    # when this event handler is invoked on non-leaf nodes of the
    # scene graph, event.target will still point to the leaf node that
    # originally triggered the event, so we use event.current_target
    # to get a reference to the node that is currently handling
    # the event, which can be a Mesh, a Group or None (the event root)
    obj = event.current_target

    # prevent propagation so we handle this event only at one
    # level in the hierarchy
    event.stop_propagation()

    # clear selection
    if selected_objects and "Shift" not in event.modifiers:
        while selected_objects:
            ob = selected_objects.pop()
            ob.traverse(partial(set_material, default_material))

    # if the background was clicked, we're done
    if isinstance(obj, gfx.Renderer):
        return

    # set selection (group or mesh)
    selected_objects.append(obj)
    obj.traverse(partial(set_material, selected_material))


def hover(event):
    obj = event.target
    if event.type == "pointer_enter":
        obj.add(outline)
        outline.set_transform_by_object(obj, "local", scale=1.1)
[...]

PR #555, opened 2023-06-22
Enable picking of transparent pixels

almarklein:
Relates to discussions in #474. Also takes #492 and #491 into consideration.

* [x] The render pass in `blender.py` can decide whether to be pickable, independent on the render mask (previously it enabled picking only when the render mask indicated that it rendered opaque fragments).
* [x] Refactored the mechanism to clear buffers, so multiple passes using the same buffers behave better. Also for the depth buffer, which is a small step towards #474.
* [x] The front-most transparency pass allows picking.

This means that to enable picking of transparent fragments, one needs to:
* Use `renderer.blend_mode = "weighted_plus"`.
* Make sure the alpha value is nonzero, because such fragments are dropped. So you cannot make completely invisible fragments, but you can uses an alpha of 0.001 or so. (Note that to accommodate for roundoff errors, there is a small margin of 1e-6.)

The motivation for refactoring the clearing of the depth buffers is that it allows doing the front-most transparency pass last. If you'd do it first (as before this pr), if you click a transparent fragment, and there is an opaque fragment behind it, you'd click the opaque one.

### DIFF
--- a/examples/feature_demo/picking_mesh.py
+++ b/examples/feature_demo/picking_mesh.py
@@ -4,6 +4,10 @@ Mesh Picking
 
 Example showing picking a mesh. Showing two meshes that can be clicked
 on. Upon clicking, the vertex closest to the pick location is moved.
+
+One of the shown objects is semi-transparent. In order for picking to
+work on such objects, the ``renderer.blend_mode`` must be set to
+"weighted_plus".
 """
 # sphinx_gallery_pygfx_render = True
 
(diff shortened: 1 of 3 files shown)


Korijn (approved):

sharon92:
@almarklein  

> Use renderer.blend_mode = "weighted_plus"

This produces an error:
```
Shader error: label:  Some("")
Parsing error: invalid field accessor `color`

    ┌─ wgsl:164:16
    │
164 │             out.color = vec4<f32>(out_color);
    │                 ^^^^^ invalid accessor
    │
    = note:


Draw error
```

sharon92:
This happens in the Backgroundshader.py

if I don't paint a background, I can pick the transparent pixels fine.

almarklein:
@sharon92 I can indeed reproduce that behavior. Apparently backgrounds have never worked with the weighted blend modes. I prepared a fix in #559.

PR #560, opened 2023-07-05
Tweaks to better support dynamic meshes

almarklein:
Relies on (and currently includes) #561.

* [x] Support meshes with zero faces.
* [x] Bounding box resets when buffer is reset (by making the `rev` counter global). 
* [x] Introduce new `.view` prop so only a (contiguous) part of a mesh is shown.
* [x] Is `.view` a good name? I think ppl may confuse it with a numpy view or a texture view. Maybe `.range` or something?
* [x] Double-check that it also works for mesh slices.
* [x] Support the new prop for other objects, like line and points. (Could be a separate pr.) 
* [x] Replace `vertex_byte_range` with the new prop.


Korijn:
> * [ ]  Is `.view` a good name? I think ppl may confuse it with a numpy view or a texture view. Maybe `.range` or something?

It's `drawRange` in ThreeJS: https://threejs.org/docs/index.html?q=buff#api/en/core/BufferGeometry.drawRange

almarklein:
Also for lines and points!

image

pygfx/resources/_buffer.py
```diff
@@ -135,15 +144,32 @@ def format(self):
...
+        )
+
+    @property
+    def draw_range(self):
+        """The range to data (origin, size) expressed in items."""
+        return self._store.draw_range
+
+    @draw_range.setter
+    def draw_range(self, draw_range):
+        origin, size = draw_range
```
Korijn, line 164: * Why not make the signature `origin, size`?
* Why do values need to be multiplied by itemsize? Numbers should be expressed as the amount of indices to render for indexed geometry, or the amount of vertices for non-indexed geometry.

```suggestion
    def draw_range(self, origin, size):
```

Korijn (approved):

PR #562, opened 2023-07-07
Fix spotlight helper

panxinmiao:
As shown in the picture, the scaling of SpotLightHelper did not take effect correctly.

Fixed:

image

### DIFF
--- a/pygfx/helpers/_lights.py
+++ b/pygfx/helpers/_lights.py
@@ -212,7 +212,7 @@ class SpotLightHelper(Line):
             Geometry(positions=positions),
             LineSegmentMaterial(thickness=1.0),
         )
-        self.world.on_update(self._update)
+        self.update_id = self.world.on_update(self._update)
 
     @callback
     def _update(self, transform: AffineBase):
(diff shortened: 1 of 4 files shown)


pygfx/helpers/_lights.py
```diff
@@ -227,4 +227,7 @@ def _update(self, transform: AffineBase):
 
         cone_length = light.distance or 1000
         cone_width = cone_length * math.tan(light.angle)
-        self.scale = (cone_width, cone_width, cone_length)
+
+        self.world.remove_callback(self.update_id)
+        self.local.scale = (cone_width, cone_width, cone_length)
+        self.update_id = self.world.on_update(self._update)
```
almarklein, line 233: Why is the callback removed and added back here?

panxinmiao: Changing the value of `local.scale` triggers the `world.on_update` callback, which leads to infinite recursive calls.

pygfx/helpers/_lights.py
```diff
@@ -227,4 +227,7 @@ def _update(self, transform: AffineBase):
 
         cone_length = light.distance or 1000
         cone_width = cone_length * math.tan(light.angle)
-        self.scale = (cone_width, cone_width, cone_length)
+
+        self.world.remove_callback(self.update_id)
```
almarklein, line 231:
```suggestion
        # Temporarily remove callback to avoid infinite recursion when setting `self.local.scale`.
        self.world.remove_callback(self.update_id)
```

pygfx/helpers/_lights.py
almarklein: Ah I see. I added a comment for future reference.

almarklein (approved):

Korijn (approved):

PR #563, opened 2023-07-10
Preserve scaling factors on transforms

### LINKED ISSUE #556: Pass `scale` vector to `mat_decompose` when a `scale_XXX` setter is called on a transform
This will help preserve the `scale` vector in a roundtrip from components -> matrix -> components.

This should resolve the problem observed in #553

This requires a fix in pylinalg, see https://github.com/pygfx/pylinalg/issues/81

Korijn:
Closes #556

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -135,6 +135,7 @@ class AffineBase:
         self.update_callbacks = {}
         self.is_camera_space = int(is_camera_space)
         self._reference_up = np.asarray(reference_up, dtype=float)
+        self._scaling_signs = np.asarray([1, 1, 1], dtype=float)
 
     @property
     def matrix(self):
@@ -151,7 +152,7 @@ class AffineBase:
 
     @cached
     def _decomposed(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
-        return la.mat_decompose(self.matrix)
+        return la.mat_decompose(self.matrix, scaling_signs=self._scaling_signs)
 
     @cached
     def _directions(self):
@@ -297,7 +298,9 @@ class AffineBase:
 
     @scale.setter
     def scale(self, value):
-        self.matrix = la.mat_compose(self.position, self.rotation, value)
+        m = la.mat_compose(self.position, self.rotation, value)
+        self._scaling_signs = np.sign(value)
+        self.matrix = m
 
     @reference_up.setter
     def reference_up(self, value):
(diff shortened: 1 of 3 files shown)


pygfx/utils/transform.py
```diff
@@ -658,3 +661,7 @@ def __matmul__(self, other):
             return RecursiveTransform(other, parent=self)
         else:
             return np.asarray(self) @ other
+
+    @cached
+    def _decomposed(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+        return la.mat_decompose(self.matrix)
```
almarklein, line 667: Scale factors not not needed here?

tests/objects/test_world_object.py
```diff
@@ -390,3 +391,19 @@ def test_bounding_box():
...
+    ob.local.scale = s
+    # without scale preservation in matrix compose -> decompose roundtrip
+    # ob.local.scale becomes (-1, 2, 3)
+    npt.assert_array_equal(ob.local.scale, s)
+
+    child = gfx.WorldObject()
+    ob.add(child)
+    npt.assert_array_equal(child.local.scale, [1, 1, 1])
+    # we don't preserve the WORLD scale component
+    npt.assert_array_equal(child.world.scale, [-1, 2, 3])
```
almarklein, line 409: What does this mean in practice? That it's fine to flip the scale factors of e.g. the camera (which is the most common scenario, I think), but not of objects that (may) have children?

pygfx/utils/transform.py
Korijn: This is on Recursive transform, a subclass of Affine transform. We use this for world transforms.

tests/objects/test_world_object.py
Korijn: It means that we can guarantee scaling will be preserved if you set it on local.scale, but not if you set it on world.scale.

almarklein: But this test also seems to show that negative scale factors don't transfer (correctly) to their children, right?

Korijn: Well, they're equivalent, so I wouldn't say they're incorrect.

almarklein: Right, and that's ok, because in this case there is no roundtrip.

almarklein (approved):

tests/objects/test_world_object.py
Korijn: Exactly!

Korijn:
To be clear; I'm sure we _could_ propagate scaling signs down, but I don't think gives any benefits :)

PR #565, opened 2023-07-13
Propagate scaling signs and handle manually set transforms

Korijn:
I realized we didn't support manually set transform matrices properly, and I figured out how to propagate scaling signs. :)

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -150,9 +150,19 @@ class AffineBase:
     def _inverse_matrix(self) -> np.ndarray:
         return np.linalg.inv(self.matrix)
 
+    @property
+    def scaling_signs(self):
+        return self._scaling_signs
+
     @cached
     def _decomposed(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
-        return la.mat_decompose(self.matrix, scaling_signs=self._scaling_signs)
+        try:
+            return la.mat_decompose(self.matrix, scaling_signs=self.scaling_signs)
+        except ValueError:
+            # the matrix has been set manually
+            # and so there is no scaling component to preserve
+            # any decomposed scaling is acceptable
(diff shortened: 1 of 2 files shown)


pygfx/utils/transform.py
```diff
@@ -663,5 +673,5 @@ def __matmul__(self, other):
             return np.asarray(self) @ other
 
     @cached
-    def _decomposed(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
-        return la.mat_decompose(self.matrix)
+    def scaling_signs(self):
+        return self._parent.scaling_signs * self.own.scaling_signs
```
almarklein, line 677: 🤩

pygfx/utils/transform.py
```diff
@@ -150,9 +150,19 @@ def matrix(self):
...
     @cached
     def _decomposed(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
-        return la.mat_decompose(self.matrix, scaling_signs=self._scaling_signs)
+        try:
+            return la.mat_decompose(self.matrix, scaling_signs=self.scaling_signs)
+        except ValueError:
+            # the matrix has been set manually
+            # and so there is no scaling component to preserve
+            # any decomposed scaling is acceptable
+            return la.mat_decompose(self.matrix)
```
almarklein, line 165: How does the `ValueError`  come about? Can/should we be more precise in detecting when the matrix has been manually set?

Korijn: The ValueError is raised when the provided signs don't align with the determinant of the matrix (even/odd number of signs).

All codepaths use the matrix setter so that makes detecting manual matrix setting inconvenient.

I figured, if we're going to set an arbitrary sign anyway, it doesn't matter if we use self._scaling_signs or not! So we can just try it and fall back to not using it.

almarklein (approved):

PR #570, opened 2023-07-19
Fix normal interpolation

panxinmiao:
This issue has been bothering me for some time, as shown in the following image:

image

The lighting calculation is a bit strange, as specular lights are discontinuous at the junction of vertex and face edges on the spherical mesh. The "MeshPhongMaterial" have this issue too.

I have written some test cases using the most basic Phong lighting model for testing, and the issue still persists:
image

It is easy to assume that there is an error in the normal calculation. The normal of each fragment is interpolated from the vertex normal, which I have checked to be correct. However, the normal interpolation seems to be not smooth. I once thought it was a floating-point calculation precision issue. However, after a period of investigation, I finally discovered the problem, which was quite simple but covert: the normal vectors were not correctly normalized in the fragment shader. Although the vertex normals were already normalized in the vertex shader, the interpolated normals could not guarantee normalization and needed to be normalized again.

Now, we have a smooth specular, 😅:

image
image

Additionally, I have slightly adjusted the display effect of the “pbr2.py” example.

### DIFF
--- a/examples/feature_demo/pbr2.py
+++ b/examples/feature_demo/pbr2.py
@@ -92,7 +92,8 @@ while alpha <= 1.0:
 
 # Create camera and controls
 camera = gfx.PerspectiveCamera(45, 640 / 480)
-camera.show_object(scene)
+camera.show_object(scene, view_dir=(-2, -1.5, -3), scale=1.2)
+
 controller = gfx.OrbitController(camera, register_events=renderer)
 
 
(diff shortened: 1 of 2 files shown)


Korijn (approved):
@almarklein if you also think this looks good, go ahead and merge

almarklein (approved):
Thanks for this! I actually observed this in the pas days too. I kept it to a 🤔 and carried on. I'm very glad to see you got this fixed! 🙏

PR #580, opened 2023-08-07
Supports lightmap

panxinmiao:
Support for LightMap has been implemented separately for `MeshBasicMaterial`, `MeshPhongMaterial`, and `MeshStandardMaterial`. Now, these three materials can support pre-baked lighting for static scenes. 

**Please note that LightMap requires a second set of texture coordinates (`geometry.texcoords1`)**.

`Trimesh` is currently unable to read the second set of texture coordinates from GLTF files, so it needs to be saved separately and loaded manually. (In fact, `Trimesh` only implements a small portion of the GLTF model attributes loading, so it may be necessary for us to develop a more comprehensive GLTF model loader specifically designed for gfx in the future. )

The "lightmap.py" example shows the effects.

image


examples/feature_demo/lightmap.py
```diff
@@ -0,0 +1,108 @@
+"""
+PBR Rendering 1
+===============
+
+
+This example shows a complete PBR rendering effect.
+The cubemap of skybox is also the environment cubemap of the helmet.
```
almarklein, line 7: Title and description are copied from another example and need an update.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -117,6 +131,19 @@ def get_bindings(self, wobject, shared):
                 material, "env_mapping_mode", "CUBE-REFLECTION"
             )
 
+        # set lightmap configs
+        if getattr(material, "light_map", None):
+            if getattr(geometry, "texcoords1", None) is None:
```
almarklein, line 136: Would this work?
```suggestion
            if self.get("use_texcoords1", None) is None:
```

almarklein (commented):
Nice!

Why is a second set of texcoords needed? The name `texcoords1` seems somewhat arbitrary, would it make sense to call it `texcoords_lightmap` instead? Or are there other aplications for these auxillary texcoords?

panxinmiao:
> Why is a second set of texcoords needed? The name `texcoords1` seems somewhat arbitrary, would it make sense to call it `texcoords_lightmap` instead? Or are there other aplications for these auxillary texcoords?

Unlike manually assigned textures by a modeler, the texture coordinates used for lightmaps are typically automatically generated during the lightmap baking process in the rendering engine. In many cases, the same lightmap is shared across the entire scene or multiple meshes. 

It's also common for multiple textures to share the same set of texture coordinates, such as lightmaps and aomaps. Therefore, the name "texcoords_lightmap" may not be ideal.

In many rendering engines, different sets of texture coordinates are represented using indices, such as uv0, uv1, or texcoord0, texcoord1, and so on. This convention is also followed in the GLTF file specification.
For example, in "lightmap/scene.gltf":

```
      "primitives": [
        {
          "mode": 4,
          "attributes": {
            "POSITION": 0,
            "NORMAL": 1,
            "TEXCOORD_0": 2,
            "TEXCOORD_1": 3
          },
          "material": 0
        },
        {
          "mode": 4,
          "attributes": {
            "POSITION": 0,
            "NORMAL": 1,
            "TEXCOORD_0": 2,
            "TEXCOORD_1": 3
          },
          "material": 1
        },
        {
          "mode": 4,
          "attributes": {
            "POSITION": 0,
            "NORMAL": 1,
            "TEXCOORD_0": 2,
            "TEXCOORD_1": 3
          },
          "material": 2
        }
      ]
```

almarklein:
> In many cases, the same lightmap is shared across the entire scene or multiple meshes.

Right, and each mesh having differente texcoords to sample in the global map, right?

> It's also common for multiple textures to share the same set of texture coordinates, such as lightmaps and aomaps. 

Would it be reasonable to assume that they always apply to some lighting feature (aomaps also being related to lighting)?

panxinmiao:
> Right, and each mesh having differente texcoords to sample in the global map, right?

Yes, it is.

> Would it be reasonable to assume that they always apply to some lighting feature (aomaps also being related to lighting)?

I'm not entirely certain about this. Typically, aomap and lightmap are baked simultaneously, so they share the same set of texture coordinates.

In terms of the model itself, texture mapping should correspond to its respective texture coordinates. Different textures of the same model may not necessarily adhere to the same set of texture coordinates, and I'm unsure if there are specific conventions regarding this in the model file format. For some rendering engines that utilize a node-based material system for visual material editing, the textures of material nodes can be bound by users in the editor to a specific set of texture coordinates on the mesh.

almarklein:
>  Different textures of the same model may not necessarily adhere to the same set of texture coordinates, and I'm unsure if there are specific conventions regarding this in the model file format. 

But doesn't hard-coding to `texcoords1` give us the same problem then? In other words, when we add aomaps, will we require its texcoords to be `geometry.texcoords1`too, or something else?

We can also be specific and use `texcoords_lightmap`, and users can still set the geometry so that `geometry.texcoords_lightmap is geometry.texcoords` or `geometry.texcoords_lightmap is geometry.texcoords_aomap` ... 🤔

panxinmiao:
> In other words, when we add aomaps, will we require its texcoords to be `geometry.texcoords1`too, or something else?

Yes, aomap also needs `geometry.texcoords1`. I'm planning to open another PR to modify aomap to use the second set of texture coordinates. 

> We can also be specific and use `texcoords_lightmap`, and users can still set the geometry so that `geometry.texcoords_lightmap is geometry.texcoords` or `geometry.texcoords_lightmap is geometry.texcoords_aomap` ... 🤔

Yes, now we need to decide whether to use indices to differentiate between different texture coordinate sets or use a more descriptive attribute name like `texcoords_lightmap`.

I tried to search for some related information, It seems that there are no specific standards for this, but there are some common conventions.:

For a mesh, there are typically several scenarios:

1. Only one set of main texture coordinates (UV0). In this case, only a single color/material texture needs to be sampled.

2. Two sets of texture coordinates: main texture coordinates (UV0) and lightmap coordinates (UV1). In this case, UV0 is used for sampling color textures, while UV1 is used for sampling lightmap textures.

3. Multiple sets of texture coordinates: main texture coordinates (UV0), lightmap coordinates (UV1), and additional texture coordinates (UV2, UV3, etc.). In this case, besides UV0 and UV1, additional sets like UV2 and UV3 can be used to sample extra textures.

In some special cases, such as when sampling a large number of textures or when anisotropic filtering is required, a mesh can even have more than 8 sets of texture coordinates.

almarklein:
@korijn and I just discussed this. ThreeJS seems to do exactly as you proposed, and it seems the most pragmatic solution to have `texcoords` and then additional texture coordinates as `texcoords1`, `texcoords2`, etc.

Also see:
https://github.com/mrdoob/three.js/blob/0b169138a99655148ed9601f87c685da54997155/src/renderers/webgl/WebGLPrograms.js#L159-L161

The current state of the PR already does it this way. I'll have a quick look in a minute, but I think its about done!

examples/feature_demo/lightmap.py
```diff
@@ -0,0 +1,107 @@
...
+
+def animate():
+    vp1.render(scene1, camera)
+    vp2.render(scene2, camera)
+    vp3.render(scene3, camera)
+    vp4.render(text_scene, text_camera)
+    renderer.flush()
+
+
+# canvas.request_draw(animate)
```
Korijn, line 103: I think this line can be removed?

pygfx/materials/_mesh.py
```diff
@@ -327,6 +330,29 @@ def env_mapping_mode(self, value):
...
 
+    @property
+    def light_map(self):
+        """The light map to define pre-baked lighting (in srgb). Default is None.
+        It requires a second set of texture coordinates (geometry.texcoords1)."""
+        return self._store.light_map
+
+    @light_map.setter
+    def light_map(self, map):
+        assert map is None or isinstance(map, Texture)
```
Korijn, line 341:
```suggestion
        if map is None or isinstance(map, Texture):
            raise ValueError(f"light_map must be a Texture or None, received: {type(map)}")
```

pygfx/renderers/wgpu/_shaderlib.py
```diff
@@ -532,7 +532,14 @@ def lighting_phong(self):
                     }
                 }
             $$ endif
-            let irradiance = getAmbientLightIrradiance( ambient_color );
+            var irradiance = getAmbientLightIrradiance( ambient_color );
+
+            // Light map (pre-backed lighting)
```
Korijn, line 537:
```suggestion
            // Light map (pre-baked lighting)
```

pygfx/renderers/wgpu/_shaderlib.py
```diff
@@ -656,7 +663,7 @@ def lighting_pbr(self):
 
             // Light map (pre-backed lighting)
```
Korijn, line 664:
```suggestion
            // Light map (pre-baked lighting)
```

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -136,6 +150,19 @@ def get_bindings(self, wobject, shared):
                 material, "env_mapping_mode", "CUBE-REFLECTION"
             )
 
+        # set lightmap configs
+        if getattr(material, "light_map", None):
+            if "use_texcoords1" not in self.kwargs:
+                warnings.warn(
+                    "Light map requires a second set of texture coordinates (geometry.texcoords1), but none were found, so it will be ignored."
+                )
```
Korijn, line 158: I'm not sure where we stand currently with respect to warnings @almarklein ? Do we already use them in the codebase?

If not, I would suggest raising a `ValueError` instead.

Korijn:
Some minor comments from my end, looks great to me! Nice work! ✨

pygfx/renderers/wgpu/meshshader.py
almarklein: I agree this should be a `ValueError`.

examples/feature_demo/lightmap.py
```diff
@@ -0,0 +1,107 @@
+"""
+Lightmap
+===============
```
almarklein, line 3:
```suggestion
Lightmap
========
```

almarklein:
I have no other comments. Really nice-looking example!

pygfx/renderers/wgpu/meshshader.py
almarklein: re warnings:  #584

Korijn (approved):

PR #586, opened 2023-08-17
Supports aomap

panxinmiao:
Modified aomap to use the second set of texture coordinates and improved the support for aomap in MeshBasicMaterial, MeshPhongMaterial, and MeshStandardMaterial.

image

### DIFF
--- a/examples/feature_demo/aomap.py
+++ b/examples/feature_demo/aomap.py
@@ -1,6 +1,6 @@
 """
 Ambient occlusion
-========
+=================
 
 
 This example demonstrates the ambient occlusion map effects for MeshBasicMaterial, MeshPhongMaterial, and MeshStandardMaterial.
--- a/examples/feature_demo/aomap.py
+++ b/examples/feature_demo/aomap.py
@@ -1,0 +1,108 @@
+"""
+Ambient occlusion
+========
+
+
+This example demonstrates the ambient occlusion map effects for MeshBasicMaterial, MeshPhongMaterial, and MeshStandardMaterial.
+"""
+
+################################################################################
+# .. warning::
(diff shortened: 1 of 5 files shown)


examples/feature_demo/aomap.py
```diff
@@ -0,0 +1,108 @@
+"""
+Ambient occlusion
+========
```
almarklein, line 3: Title underline must be the same length as the title (this makes Sphinx complain and causes the doc build to fail 😄).
```suggestion
Ambient occlusion
=================
```

pygfx/materials/_mesh.py
```diff
@@ -356,6 +363,28 @@ def light_map_intensity(self, value):
...
+
+    @ao_map.setter
+    def ao_map(self, map):
+        if map is not None and not isinstance(map, Texture):
+            raise ValueError(f"ao_map must be a Texture or None, received: {type(map)}")
+        self._store.ao_map = map
+
+    @property
+    def ao_map_intensity(self):
+        """Intensity of the ambient occlusion effect. Default is 1.0 Zero is no occlusion effect."""
```
almarklein, line 380: Missing comma/dot.
```suggestion
        """Intensity of the ambient occlusion effect. Default is 1.0, zero is no occlusion effect."""
```

almarklein (commented):
Very nice work again! I have just two small textual suggestions.

examples/feature_demo/aomap.py
panxinmiao: Ah, I finally know the reason.

almarklein (approved):

Korijn (approved):
Amazing 🤩

PR #587, opened 2023-08-23
Enhance MeshPhongMaterial

panxinmiao:
`MeshPhongMaterial` has been improved to unify with `MeshStandardMaterial`. 

Support for `specular_map`, `emissive_map`, and `normal_map` has been added to `MeshPhongMaterial`.

Additionally, `MeshBasicMaterial` now also supports `specular_map` (which affects the environment mapping).

I have also added an example `earth.py`, which provides a more complete demonstration of the `MeshPhongMaterial`.

image


panxinmiao:
It seems that in the new version of the Sphinx configuration file, the `html_theme` field must be configured. 

I tried uncommenting the line `html_theme = "alabaster"` in the `docs/conf.py` file, but the generated documentation looks different from before in terms of style, 🤔

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -128,6 +131,35 @@ def get_bindings(self, wobject, shared):
         sampler = GfxSampler(material.map_interpolation, "repeat")
         texturing = "texture/auto"
 
+        # todo, check uv is present and has the right shape. (introduce uv channel for texture)
+
+        # specular map configs, for basic and phong materials
+        if (
+            getattr(material, "specular_map", None)
+            and type(material) is not MeshStandardMaterial
+        ):
```
almarklein, line 140: This shader is for `MeshBasicMaterial`, so I would expect that all properties  that are available on `MeshBasicMaterial` can be accessed via `material.xx`. There are a few cases where `getattr()` is still used, from before we refactored the material class hierarchy.


```suggestion
        if material.specular_map and not isinstance(material, MeshStandardMaterial):
```

examples/feature_demo/earth.py
```diff
@@ -0,0 +1,125 @@
...
+earth_material.normal_scale = 0.85, -0.85
+
+earth = gfx.Mesh(earth_geometry, earth_material)
+earth.local.rotation = la.quat_from_euler((0, -1.0, 0), order="XYZ")
+scene.add(earth)
+
+# Clouds
+clouds_material = gfx.MeshPhongMaterial()
+clouds_map_tex = load_texture(
+    model_dir / "planets" / "earth_clouds_2048.png", flip=True
```
almarklein, line 89: This example adds over 5 MB of data to the repo. Any idea why the cloud map is especially large?

almarklein (commented):
You now added `specular_map` to the `MeshBasicMaterial` and e.g. `emissive_map` to the `MeshPhongMaterial`. How did you make this distinction?

I see a lot of duplicate properties between the `MeshPhongMaterial` and the `MeshStandardMaterial`. Since both inherit from `MeshBasicMaterial`, and are the only classes inherited from it, I think it would make sense to change the purpose of the `MeshBasicMaterial` a bit. 

Right now it represents a mesh rendered as a 3D surface. The material itself is not affected by lights, but it does have properties related to lights, like `env_map` and `reflectivity`. I think it would make sense to move all properties that are common to both Phong and PBR to this class.

That'd mean that the `MeshStandardMaterial` represents a 3D model with everything needed to apply lights, but the lighting model is implemented in its subclasses, and the class itself is still unaffected by lights. @panxinmiao @Korijn what do you think?

panxinmiao:
> You now added `specular_map` to the `MeshBasicMaterial` and e.g. `emissive_map` to the `MeshPhongMaterial`. How did you make this distinction?

`MeshBasicMaterial` is used to apply the simplest way of shading to a mesh. It does not involve any real-time lighting calculations (hence it does not require normal maps or related properties) and instead relies solely on sampling from pre-existing texture maps. 
Because `MeshBasicMaterial` also supports environment mapping, and the `specular_map` property affects the appearance of the environment map, so it is included. 
On the other hand, the `emssive_map` property, even though it is also a texture sample, is used in conjunction with the `emssive` property to represent the inherent color emitted by the object itself, unaffected by lighting. Since the color of the `MeshBasicMaterial` itself is not affected by lighting, the emssive property has no significance for it.

> 
> I see a lot of duplicate properties between the `MeshPhongMaterial` and the `MeshStandardMaterial`. Since both inherit from `MeshBasicMaterial`, and are the only classes inherited from it, I think it would make sense to change the purpose of the `MeshBasicMaterial` a bit.
> 

I am also considering this issue, and the current inheritance relationship of Materials maybe not well-designed. 
There doesn't seem to be a clear attribute inheritance relationship between `MeshBasicMaterial`, `MeshPhongMaterial`, and `MeshStandardMaterial`. 
`MesPhongMaterial` and `MeshStandardMaterial` have many common attributes that do not exist in `MeshBasicMaterial`. `MeshBasicMaterial` also has certain properties that do not exist in others, such as `specular_map`.

pygfx/renderers/wgpu/meshshader.py
```diff
@@ -128,6 +131,35 @@ def get_bindings(self, wobject, shared):
         sampler = GfxSampler(material.map_interpolation, "repeat")
         texturing = "texture/auto"
 
+        # todo, check uv is present and has the right shape. (introduce uv channel for texture)
```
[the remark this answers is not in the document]
panxinmiao: I plan to complete it in another PR. I intend to introduce a "texture channel" property in the `Texture` object, which will determine which set of texture coordinates of the geometry the texture uses.

See: https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/figures/gltfOverview-2.0.0b.png
image

examples/feature_demo/earth.py
panxinmiao: I believe it's because the image format is PNG, and the cloud map has transparency with an alpha channel.

pygfx/renderers/wgpu/meshshader.py
panxinmiao: `MeshNormalShader`, `MeshSliceShader`, and other shaders inherit from this shader, but these materials do not have the "specular_map" property.

Korijn:
> That'd mean that the `MeshStandardMaterial` represents a 3D model with everything needed to apply lights, but the lighting model is implemented in its subclasses, and the class itself is still unaffected by lights. @panxinmiao @Korijn what do you think?

Personally, I was surprised to see there is any inheritance at all between the mesh materials. In my mind they are just API objects used to configure the rendering pipeline for meshes.

If there is redundancy between them, that is something we can potentially address with mixins. I don't think inheritance is the right solution here. A MeshStandardMaterial is not a special kind of MeshBasicMaterial.

pygfx/renderers/wgpu/meshshader.py
almarklein: Ah, right ... the materials don't but the *shader classes* do ...

almarklein:
> Personally, I was surprised to see there is any inheritance at all between the mesh materials. In my mind they are just API objects used to configure the rendering pipeline for meshes.

I agree it does not make a lot of sense from an API perspective to use inheritance. It was done for code-reuse. So if I understand you (Korijn) correctly, you'd prefer to use mixins instead, and then e.g. tweak the sphinx docs to show all props and methods of a material (including anything inherited), so that the mixins are hidden / an implementation detail.

Korijn:
Yes, that's right, although I'm not sure it's actually worthwhile to go through all the effort to implement mixins properly

panxinmiao:
The content of this PR has been incorporated into #850, so it can now be closed.

Maybe we can open a new issue to continue the discussion on the inheritance structure, code reuse, and API design for Mesh Materials.

almarklein:
Superseded by #850

PR #590, opened 2023-09-18
Let load_scene also work for single meshes

almarklein:
The `gfx.load_scene` expects a file containing an actual scene. This pr also allows using it to load a simple stl file. I later realized one can do `trimesh.load()` and then `gfx.geometry_from_trimesh()`, though this change does make that a bit easier, and ppl wont have to know about trimesh.

### DIFF
--- a/pygfx/resources/_buffer.py
+++ b/pygfx/resources/_buffer.py
@@ -176,6 +176,7 @@ class Buffer(Resource):
                 f"`offset` and `size` must be native `int` type, you have passed: "
                 f"offset: <{type(offset)}>, size: <{type(size)}>"
             )
+        offset, size = int(offset), int(size)  # convert np ints to real ints
 
         if size == 0:
             return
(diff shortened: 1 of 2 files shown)


Korijn:
I propose to rename the function to the following signature: `gfx.load(path, loader="trimesh")` since it's just going to try to load whatever it is in the path with trimesh

almarklein:
> I propose to rename the function to the following signature: `gfx.load(path, loader="trimesh")` since it's just going to try to load whatever it is in the path with trimesh

Perhaps `load` is a bit too generic? I think `load_meshes` would be the most correct.

Korijn:
> Perhaps `load` is a bit too generic? I think `load_meshes` would be the most correct.

Works for me!

almarklein:
Done. Also added `load_mesh()`, because there I found `load_meshes(...)[0]` particularly ugly.

Korijn (approved):

PR #591, opened 2023-09-19
Fix gaps in mesh slice

almarklein:
When the mesh slice intersects with an edge, the line for it is not drawn properly. This case is typically quite rare. Although if you set the plane based on picking a vertex, the plane already precisely intersects that vertex, so it's more likely to happen. If you then have artificial data (e.g. a generated sphere), you may actually see it a lot :)

Tasks:
* [x] Fix missing line pieces (by fixing the case where all three edges of a triangle pass the plane).
* [x] Fix spurious points (by discarding a missed case where only one edge passed the plane).
* [x] Fix that the pick coords sometimes jump to another location (by fixing pick coord encoding).  
* [x] Fix that pick coords were sometimes all zeros (fixed by the above).
* [x] Fix the somewhat wobbly movement of the face coords as you move over a mesh slice. (Fixed the jumps, not the wobbling.)
* [x] Add a visual test. 

Before:
image

After:
image

Note that we're dealing with weird shader arithmetic here. In this line:
```wgsl
let t1 = -(plane.x * p.x + plane.y * p.y + plane.z * p.z + plane.w) / dot(n, u);
```

The `dot(n, u)` can be zero, but the resulting `t1` wont be `inf`. We can detect the case and fix it relatively easily, but I agree that it looks a bit weird because it makes assumptions about the resulting `t1 t2 t3` that feel somewhat counterintuitive.


almarklein:
Mmm, seeing some occasional artifacts, so perhaps this now shows pieces of faces that it should not.

Korijn:
Wow, that looks like a very subtle bug.

almarklein:
> Wow, that looks like a very subtle bug. How'd you figure it out?

The one with the coor value encoding? In the mesh morph example, as you move the mouse over the mesh slice, the location being picked is shown in the 3D view (using the pick coords). I saw the location sometimes jumped when transitioning to a new face. After double-checking the logic that handles the pick coords (and not finding anything) I checked the encoding.

almarklein:
I still observed some wobbly behavior in how the position indicator moves over the 3D surface as you move the mouse along the mesh slice line:
* A minor "shaking" behavior seen everywhere.
* A jump when approaching/leaving a new face.

I found that the first is just because of the 6-bit precision. Because of this, the contribution of the three vertices, as defined by the pick coords is rounded, and as such there is a limited number of spots on the face that can be addressed.

The observed jump is because the joins (and ends) of the line pieces are round, extending beyond the actual edge. Inside the join, the pick coords are set to the last value on the edge, but since the join covers the next line piece, a jump is observed.

image

I looked into fixing this:
* Not drawing joins. This looks quite a bit worse, and when you want to click om the mesh slice, there is a chance that you click right in between two line pieces.
* Drawing the joins with alpha 0.99, so that they are visible, but not clickable. Although this helps solve the problem, there is now a new problem that it looks like you click the mesh slice, but no picking occurs.
* As I wrote this down I realized that we can also curve the joins away from the camera, so they won't overlap with the next line piece. This worked!

almarklein:
Rrrready!

Korijn (approved):

Korijn:
Very nice! 😊

PR #592, opened 2023-09-22
Implement color_mode for per-face coloring, also for lines, points and meshslice

### LINKED ISSUE #152: Mesh per-face coloring

almarklein:
Closes #152

### Tasks
* [x] Implement `material.color_mode` for mesh, meshslice, line, and point.
  * Options are 'uniform', 'vertex', 'face', 'vertex_map' and 'face_map'.
  * With 'vertex' and 'face`, the `geometry.colors` are used.
  * With 'vertex_map` and `face_map`, the `geometry.texcoords` are used, and then looked up into `material.map`.
  * Also has 'auto' which means use 'uniform' unless the material has a `.map`, in which case use `vertex_map`.
* [x] Meshslice now supports all color modes (previously only uniform color).
* [x] Lines now also support face colors, which means the line-pieces can be individually colored.
* [x] The `material.vertex_colors` flag raises a deprecationwarning for now.
* [x] The mesh `material.face_colors` (added in #529) is simply removed.
* [x] Fixed that the anti-aliased fragments were previously dropped because of an incorrect `render_mask`.
  * This makes the points look slightly larger (i.e. they were a bit too small earlier).

### Noes on aa
* The mesh shader does not use aa, so a mesh with an opaque color is fully opaque.
* Same for mesh slice at the moment.
* Lines uses aa, so they *always*  have transparent fragments.
* Same for points. However, the `render_mask` was previously set to ignore the transparent fragments, which can be considered a bug.
* Text also uses aa.


examples/screenshots/validate_image2.png
Korijn, line 1: Perhaps my eyes are fooling me but the points look slightly larger, is that right?

Korijn (approved):

examples/screenshots/validate_image2.png
almarklein: You're eyes work great! This pr also fixes a bug where the transparent fragments were not drawn. I updated the top post to give an overview of the changes.

Korijn (approved):

PR #597, opened 2023-10-10
Run CI tests with can_use_wgpu_lib

almarklein:
We have no unit-test-builds with lavapipe enabled, so some of the tests are never run. It seems that locally, I mostly run the examples tests, because I only found out today that a few unit tests are actually broken 😢 (and confirmed its not due to the recent changes in wgpu-py)

### DIFF
--- a/pygfx/renderers/wgpu/_pipeline.py
+++ b/pygfx/renderers/wgpu/_pipeline.py
@@ -497,8 +497,10 @@ class PipelineContainer:
             changed.update(("bindings", "pipeline_info", "render_info"))
 
         if "bindings" in changed:
+            self.shader.unlock_hash()
             with wobject.tracker.track_usage("!bindings"):
                 self.bindings_dicts = self.shader.get_bindings(wobject, self.shared)
+            self.shader.lock_hash()
             self._check_bindings()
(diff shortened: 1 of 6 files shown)


almarklein:
So there was actually a bug in the code related to shader cache invalidation. It's fixed and I also made some changes to prevent this kind of error from being introduced.

pygfx/renderers/wgpu/_pipeline.py
```diff
@@ -497,8 +497,10 @@ def update_shader_data(self, wobject, changed):
             changed.update(("bindings", "pipeline_info", "render_info"))
 
         if "bindings" in changed:
+            self.shader.unlock_hash()
             with wobject.tracker.track_usage("!bindings"):
                 self.bindings_dicts = self.shader.get_bindings(wobject, self.shared)
+            self.shader.lock_hash()
```
Korijn, line 503: This looks like something that needs a try/finally?

almarklein: No, if an exception is raised in `get_bindings()` then the code does not proceed anyway, so there is no need to apply the lock.

Korijn (approved):

PR #599, opened 2023-10-12
better error message for texture data array shapes that don't make sense

kushalkolar:
This just provides an error message for a previous `assert`, I've been wanting to catch this for months and saw that you put in the check recently :smile: 

Catching this is useful in interactive visualizations where the user ends up creating a slice that doesn't make sense. For us it's if you end up with an array of shape `(1, 1)`, i.e. just one pixel in one or all channels so that's why I put that as the "interpretive" error message. Not sure if that's the only case that's covered here but it's what I can think of. 

closes: https://github.com/fastplotlib/fastplotlib/issues/193 

trying to see if cross-repo closing is possible :smile:

### DIFF
--- a/pygfx/resources/_texture.py
+++ b/pygfx/resources/_texture.py
@@ -84,7 +84,10 @@ class Texture(Resource):
             if len(shape) == len(collapsed_size) + 1:
                 nchannels = shape[-1]
             else:
-                assert len(shape) == len(collapsed_size)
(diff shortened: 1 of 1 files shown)


almarklein (approved):
Thanks! I take the blame for using `assert` too freely sometimes :)

Korijn (approved):

PR #608, opened 2023-10-26
Unable to load model without UV.

sltruman:
When the added mesh object does not have UV coordinates, set a default to it.

### DIFF
--- a/pygfx/geometries/_compat.py
+++ b/pygfx/geometries/_compat.py
@@ -36,6 +36,9 @@ def geometry_from_trimesh(mesh):
         # the coordinate origin is in the upper left corner, while the opengl coordinate
         # origin is in the lower left corner.
         # trimesh loads textures according to the opengl coordinate system.
+        if mesh.visual.uv is None:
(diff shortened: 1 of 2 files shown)


pygfx/geometries/_compat.py
```diff
@@ -36,6 +36,9 @@ def geometry_from_trimesh(mesh):
         # the coordinate origin is in the upper left corner, while the opengl coordinate
         # origin is in the lower left corner.
         # trimesh loads textures according to the opengl coordinate system.
+        if mesh.visual.uv is None:
```
Korijn, line 39: I don't know enough about trimesh to understand why `mesh.visual.uv is None` when the kind is `"texture"`. Do you know?

sltruman: I made a model with textures and no UV mapping in Blender, which triggers this problem.

sltruman: You are right, this is problem of Trimesh.

PR #624, opened 2023-11-29
Improve overlay rendering

### LINKED ISSUE #620: Enabling overlay rendering with render_order and depth_read
We are able to render overlays in separate passes (see [example](https://pygfx.readthedocs.io/en/stable/_gallery/feature_demo/scene_overlay.html)), but we should also make it possible to do that within a single scene and a single render call.

If we add `depth_read` property on `Material`, you could render on top like this:

```python
import numpy as np
import pygfx as gfx
import pylinalg as la

# Create a canvas and renderer
canvas = WgpuCanvas(size=(500, 300))

# Compose a 3D scene
scene1 = gfx.Scene()

geometry1 = gfx.box_geometry(200, 200, 200)
material1 = gfx.MeshPhongMaterial(color=(1, 1, 0, 1.0))
cube1 = gfx.Mesh(geometry1, material1)
scene1.add(cube1)

camera1 = gfx.OrthographicCamera(300, 300)
scene1.add(camera1.add(gfx.DirectionalLight()))

# Render a line on top
positions = np.array(
    [
        [-1, -1, 0.5],
        [-1, +1, 0.5],
        [+1, +1, 0.5],
        [+1, -1, 0.5],
        [-1, -1, 0.5],
        [+1, +1, 0.5],
    ],
    np.float32,
)
geometry2 = gfx.Geometry(positions=positions * 0.9)
material2 = gfx.LineMaterial(thickness=5.0, color=(0.8, 0.0, 0.2, 1.0), depth_read=False) # <<< CHANGED LINE
line2 = gfx.Line(geometry2, material2)
line2.render_order = 1   # <<< CHANGED LINE
scene1.add(line2)

def animate():
    rot = la.quat_from_euler((0.005, 0.01), order="XY")
    cube1.local.rotation = la.quat_mul(rot, cube1.local.rotation)

    renderer.render(scene1, camera1)   # <<< CHANGED LINE

    canvas.request_draw()


if __name__ == "__main__":
    canvas.request_draw(animate)
    run()
```

### LINKED ISSUE #621: Overlay rendering broken
It seems like depth is not cleared in between render calls, even though it says this should be the case in the code:

https://github.com/pygfx/pygfx/blob/01b707f59adf2781346e910903b622296ea14850/pygfx/renderers/wgpu/_renderer.py#L405-L408

This is making it really difficult to put together visualizations that depend on layers.

image

I've adjusted the `scene_overlay.py` example to render the lines in the same space by sharing the camera and increasing the size of the lines a bit:

```python
"""
Scene Overlay
=============

Example showing a 3D scene with a 2D overlay.

The idea is to render both scenes, but clear the depth before rendering
the overlay, so that it's always on top.
"""
# sphinx_gallery_pygfx_render = True

import numpy as np
import pygfx as gfx
import pylinalg as la


# Create a canvas and renderer

canvas = WgpuCanvas(size=(500, 300))

# Compose a 3D scene

scene1 = gfx.Scene()

geometry1 = gfx.box_geometry(200, 200, 200)
material1 = gfx.MeshPhongMaterial(color=(1, 1, 0, 1.0))
cube1 = gfx.Mesh(geometry1, material1)
scene1.add(cube1)

camera1 = gfx.OrthographicCamera(300, 300)
scene1.add(camera1.add(gfx.DirectionalLight()))

# Compose another scene, a 2D overlay

scene2 = gfx.Scene()

positions = np.array(
    [
        [-1, -1, 0.5],
        [-1, +1, 0.5],
        [+1, +1, 0.5],
        [+1, -1, 0.5],
        [-1, -1, 0.5],
        [+1, +1, 0.5],
    ],
    np.float32,
)
geometry2 = gfx.Geometry(positions=positions * 100)   # <<<<< CHANGED LINE
material2 = gfx.LineMaterial(thickness=5.0, color=(0.8, 0.0, 0.2, 1.0))
line2 = gfx.Line(geometry2, material2)
scene2.add(line2)


def animate():
    rot = la.quat_from_euler((0.005, 0.01), order="XY")
    cube1.local.rotation = la.quat_mul(rot, cube1.local.rotation)

    renderer.render(scene1, camera1, flush=False)
    renderer.render(scene2, camera1)   # <<<<< CHANGED LINE

    canvas.request_draw()


if __name__ == "__main__":
    canvas.request_draw(animate)
    run()
```

almarklein:
* [x] Fixes that depth buffer was not clear before each render.
  * Closes #621
* [x] Implements `Material.depth_test`. 
  * Closes #620

image


almarklein:
🚀

Korijn:
I'd recommend to also use render_order in combination with depth_test to have a validation example that validates that multiple layers render on top of each other in the correct order.

almarklein:
Great to add that to a validation example, because it appears to be broken 😅

almarklein:
False alarm, had to turn on `renderer.sort_objects`.

Korijn:
> False alarm, had to turn on `renderer.sort_objects`.

Just wondering why that isn't enabled by default?

Korijn (approved):

almarklein:
>> False alarm, had to turn on renderer.sort_objects.
>
> Just wondering why that isn't enabled by default?

Performance

almarklein:
> Performance

To add a bit more info: when this property is set, it sorts by both the `object.render_order` as well as the z-position from the pov of the camera. Especially the last one can hurt performance. Perhaps we can make this property an enum, and sort by `render_order` by default. That should not suffer much from overhead, because it can be done as one flattens the scenegraph.

Korijn:
> To add a bit more info: when this property is set, it sorts by both the `object.render_order` as well as the z-position from the pov of the camera. Especially the last one can hurt performance. Perhaps we can make this property an enum, and sort by `render_order` by default. That should not suffer much from overhead, because it can be done as one flattens the scenegraph.

That makes sense to me. It seems like render_order not being used silently is a little misleading.

almarklein:
> It seems like render_order not being used silently is a little misleading.

I was confused by it not working, so I have to agree 😄  I'll make an issue

Issue #49, opened 2020-07-08
Improvements to line material


almarklein:
* [ ] depth attenuation
* [ ] Being able to set the width per-vertex
* [ ] Being able to set the color per-vertex

PR #628, opened 2023-12-05
Line dashing

almarklein:
Ref  #49

* [x] Enable shaders to do CPU-bound work on each draw. We need this to bake the cumulative distance along the line.
* [x] Implement dashing, both for screen-space and model-space.
* [x] Make continuous at shallow corners.
* [x] Make cumdist proceed nicely over the corners, without skewing at the segments. 
* [x] There seems to be a small glitch at the start/end caps. Let's put all "degenerate vertices" in the same (first/last) spot, instead of creating going back and forth. 
* [x] Support for dashes in world coords.
* [x] Proper dashing in the line-caps.
* [x] Enable antialiasing again.
* [x] Update other line shaders to use 6 vertices per position.
* [x] Support for line width in world coords.
* [x] Dashing in LineSegmentMaterial.
* [x] I put the shader into their own shader file.
* [x] Check todos.
* [x] API: dash_size, dash_ratio vs dash_cap.
* [x] per-vertex colors need similar treatment as cumdist.
* [x] Blog posts up: https://almarklein.org/line_rendering.html
* [x] Add validation examples: alpha, thickness, continuous vs broken joins, dashing, ...

Updated  #49 to check what we did, and add an item for a benchmark, and faster lines in special cases.


pygfx/renderers/wgpu/lineshader.py
```diff
@@ -543,12 +562,110 @@ def bake_function(self, wobject, camera):
         cum_distances = np.cumsum(distances)
```
Korijn, line 562:
```suggestion
        np.cumsum(distances, out=distance_array[1:])
```

pygfx/renderers/wgpu/lineshader.py
```diff
@@ -557,14 +557,13 @@ def bake_function(self, wobject, camera, logical_size):
...
+        # Calculate distances
         distances = np.linalg.norm(vertex_array[1:] - vertex_array[:-1], axis=1)
         distances[~np.isfinite(distances)] = 0.0
-        np.cumsum(distances, out=distance_array[1:])
 
-        # Apply
+        # Store cumulatives
         distance_array[0] = dash_offset
-        distance_array[1:] = cum_distances + dash_offset
+        np.cumsum(distances, out=distance_array[1:])
```
Korijn, line 566: don't forget to add dash_offset (inplace :)

pygfx/renderers/wgpu/lineshader.py
```diff
@@ -311,138 +333,453 @@ def code_vertex_core(self):
...
+        r_offset, r_size = positions_buffer.draw_range
+
+        # Prepare arrays
+        positions_array = positions_buffer.data[r_offset : r_offset + r_size]
+        distance_array = self.line_distance_buffer.data[r_offset : r_offset + r_size]
+
+        # Get vertices in the appropriate coordinate frame
+        if wobject.material.dash_is_screen_space:
+            xyz = la.vec_transform(positions_array, camera.camera_matrix)
+            vertex_array = xyz[:, :2] * 0.5 * np.array(logical_size)
```
Korijn, line 719: minor optimization thing:

```suggestion
            vertex_array = xyz[:, :2] * (0.5 * np.array(logical_size))
```

almarklein:
Progress: dashing over joins.

Screen Recording 2024-01-18 at 22 01 48

The dashes have caps that get deformed in the joins, as expected. When dashing is enabled, the join should probaly be broken into individual segments at smaller angles.

Getting here was a tough ride. Code should be converging from here on.

almarklein:
On dashing API:

* In html canvas:
  * One would do `ctx.setLineDash([30, 20])`, setting pairs of stroke-size and gap-size. The total size of one dash cycle is the sum of these.
  * Sizes are expressed in (screen?) coord, independent of line width. 
  * Caps reduce the (visual) size of the gaps.
* In ThreeJS ([docs](https://threejs.org/docs/index.html?q=dash#api/en/materials/LineDashedMaterial), [example1](https://threejs.org/examples/?q=lines#webgl_lines_fat), [example2](https://threejs.org/examples/?q=dash#webgl_lines_dashed)):
  * There is `dash_size`and `gap_size`, where `dash_size` is the *total* size, i.e. the strike is `dash_size - gap_size`. 
  * There also is `dash_scale` to scale the lot.
  * Dash size (and gap) are expressed in world coordinates (and then scaled with `dash_scale`), independent of line width.
* Matplotlib:
  *  has a `set_dashes(seq)` for providing on-off points, but also has presets via `set_linestyle()` such as "--", ":" and "-.".
  *  Dash sizes are relative to the line width, which has the advantage that when the line width is increased, the line looks similar. Also a dash stroke-size of zero (need to set it to 0.01 in mpl) with round caps results in a dot at each line width.

almarklein:
Have now implemented `material.dash_pattern = <sequence>`.  The sequence can be as long as you want - it's inserted as an array into the shader using templating. The values are expresses as line-thicknesses,  just as in MPL. I think this will also work out nicely for 3D scenes with thickness in world coords, but I have yet to implement and test that.

almarklein:
Thickness and dashing in world coords:
image

pygfx/materials/_line.py
```diff
@@ -188,6 +188,15 @@ def map_interpolation(self, value):
         self._store.map_interpolation = value
 
 
+class LineDebugMaterial(LineMaterial):
+    """Line debug material.
+
+    A material that renders the triangles that the line is made up off.
+    """
```
Korijn, line 195: Is this like wireframe mode?

pygfx/geometries/_base.py
```diff
@@ -71,8 +71,8 @@ def __init__(self, **kwargs):
                     if not ("x" not in format or format.startswith(("1x", "2x", "3x"))):
                         raise ValueError("Expected Nx1, Nx2 or Nx3 data for texcoords")
                 elif name == "colors":
-                    if not format.startswith("4x"):
-                        raise ValueError("Expected Nx4 data for colors")
+                    if not (format.startswith("4x") or format.startswith("3x")):
```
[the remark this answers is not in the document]
almarklein: Nice catch, will fix.

pygfx/renderers/wgpu/lineshader.py
```diff
@@ -113,6 +113,59 @@ def __init__(self, wobject):
...
 
+        # Handle dashing
+        if material.dash_pattern:
+            # Set dash props
+            self["dashing"] = True
+            self["dash_pattern"] = tuple(wobject.material.dash_pattern)
+            self["dash_count"] = len(wobject.material.dash_pattern) // 2
+            # For line segments we can calculate the distance between nodes in the shader.
+            # For normal lines, we need a cumulative distance.
+            if not isinstance(material, LineSegmentMaterial):
```
Korijn, line 124: and LineThinSegmentMaterial right?

almarklein: The thin lines don't do dashing at all. But we might indeed need a check for it somewhere around here. Should probably add a validation example for thin lines.

almarklein:
Artifacts when nodes are close together (here a sine wave consisting of 1000 positions):
image

pygfx/renderers/wgpu/lineshader.py
Korijn: That makes sense.

examples/feature_demo/line_basic.py
```diff
@@ -2,7 +2,6 @@
 Line Drawing
 ============
 
-
 Some basic line drawing.
```
Korijn, line 5: this docstring carries a different meaning now :)

pygfx/renderers/wgpu/_pipeline.py
```diff
@@ -403,6 +405,8 @@ def update(self, wobject, environment, changed):
                     self.render_containers.append(RenderPipelineContainer(shader))
                 else:
                     raise ValueError(f"Shader type {shader.type} is unknown.")
+                if shader.needs_bake_function:
+                    self.bake_functions.append(shader.bake_function)
```
Korijn, line 409: is it possible somehow for this function to be appended more than once, accidentally? just wondering if such a codepath is possible and if it would create issues

pygfx/renderers/wgpu/_shader.py
```diff
@@ -1,15 +1,24 @@
+from pkg_resources import resource_filename
+
 from ...resources import Buffer, Texture
 from ._utils import to_vertex_format, to_texture_format, GfxSampler, GfxTextureView
 from ._shaderbase import BaseShader
 
 
+def load_shader(shader_name):
+    filename = resource_filename("pygfx.data_files", f"shaders/{shader_name}")
```
Korijn, line 9: do you need to change anything about setup.py (and perhaps pyinstaller) to make sure the files in that folder are packaged?

pygfx/materials/_line.py
```diff
@@ -8,50 +8,59 @@ class LineMaterial(Material):
 
     Parameters
     ----------
+    thickness : float
+        The line thickness expressed in logical pixels. Default 2.0.
+    thickness_space : str
+        The coordinate space in which the thickness is expressed. Default "screen".
```
Korijn, line 14: what are the other options?

pygfx/materials/_line.py
```diff
@@ -8,50 +8,59 @@ class LineMaterial(Material):
...
-        The line thickness expressed in logical pixels.
     color_mode : enum or str
         The mode by which the line is coloured. Default 'auto'.
     map : Texture
         The texture map specifying the color for each texture coordinate. Optional.
     map_interpolation: str
-        The method to interpolate the color map. Either 'nearest' or 'linear'. Default 'linear'.
+        The method to interpolate the color map ('nearest' or 'linear'). Default 'linear'.
+    dash_pattern : tuple
+        The pattern of the dash. Defaults to an empty tuple, meaning no dashing.
```
Korijn, line 24: can you explain what valid patterns look like?

pygfx/materials/_line.py
```diff
@@ -130,20 +150,46 @@ def vertex_colors(self, value):
 
     @property
     def thickness(self):
-        """The line thickness expressed in logical pixels."""
+        """The line thickness.
+
+        The interpretation depends on `thickness_space`, which is in logical
+        pixels by default.
```
Korijn, line 156: and in what otherwise?

Korijn:
I'm [beyond excited](https://youtu.be/SbXmGgJePsk?si=TqH4-KGaB8qIs5bK) to see this come together! Absolutely amazing job @almarklein !

examples/feature_demo/line_basic.py
almarklein: Updated.

pygfx/renderers/wgpu/_pipeline.py
almarklein: Not really. The `self.bake_functions` is reset above. But I understand the concern, since this path is quite critical. I refactored this code a bit to store it as a tuple instead, which makes it less likely its appended to unintentionally.

pygfx/renderers/wgpu/_shader.py
almarklein: I will add a test for this to make sure.

pygfx/materials/_line.py
almarklein: done

pygfx/materials/_line.py
almarklein: done

pygfx/materials/_line.py
almarklein: Also  #661

examples/screenshots/validate_depth_overlay1.png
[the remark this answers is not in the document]
almarklein: mmm, no it does not 🤔

pygfx/renderers/wgpu/_shader.py
```diff
@@ -1,15 +1,25 @@
...
+
+from ...utils import get_resources_dir
 from ...resources import Buffer, Texture
 from ._utils import to_vertex_format, to_texture_format, GfxSampler, GfxTextureView
 from ._shaderbase import BaseShader
 
 
+def load_shader(shader_name):
+    filename = os.path.join(get_resources_dir(), "shaders", shader_name)
+    with open(filename, "rb") as f:
```
Korijn, line 11: Maybe add encoding utf-8?

almarklein: oh wait, this reads as binary and then calls `.decode()` on the bytes (which is utf-8 by default).

almarklein:
Looks like lavapipe does not handle caps very well. On closer inspection some other screenshots don't look good either.

almarklein:
This does not look good either:
image

Korijn (approved):

Korijn (dismissed):

almarklein:
Hold on. Seeing a glitch on Windows still, seems not to detect nans correctly.

almarklein:
New nan-detection confirmed to work on MacOS, Windows and Linux.

PR #638, opened 2024-02-01
Renderer.pixel_ratio always has a value

almarklein:
Needs #637.

* Makes `renderer.pixel_ratio` always have a value. Previously it can always be `None`. Now the default case is simply handled in the setter.
* Adds `renderer.pixel_filter` to control filtering in the flusher. 
  * Can e.g. be set to higher value for a more blurry image. 
  * Or to zero as requested in #640.
  * Or set to zero and use a `pixel_ratio` of 0.125 to get huge pixels to e.g. work on line aa :)

### DIFF
--- a/pygfx/renderers/wgpu/_renderer.py
+++ b/pygfx/renderers/wgpu/_renderer.py
@@ -129,6 +129,7 @@ class WgpuRenderer(RootEventHandler, Renderer):
             raise TypeError(
                 f"Render target must be a Canvas or Texture, not a {target.__class__.__name__}"
             )
+        self._flusher_filter = True  # Whether the flusher applies a filter
         self._target = target
         self.pixel_ratio = pixel_ratio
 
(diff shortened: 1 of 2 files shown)


pygfx/renderers/wgpu/_renderer.py
```diff
@@ -199,9 +200,9 @@ def pixel_ratio(self):
         """The ratio between the number of internal pixels versus the logical pixels on the canvas.
 
         This can be used to configure the size of the render texture
-        relative to the canvas' logical size. By default (value is None) the
-        used pixel ratio follows the screens pixel ratio on high-res
-        displays, and is 2 otherwise.
+        relative to the canvas' logical size. Can be set to None to set the default.
+        By default the pixel_ratio follows the screens pixel ratio on high-res
+        displays, and is 2.0 otherwise.
```
Korijn, line 205: I don't understand why you chose 2.0 as the default?

almarklein: It's best to have an integer number so the pixel grids align. With 1 there is no aa, with 3 the memory consumption might become a thing. Most high-res displays have a factor 2 as well.

almarklein:
I added  `renderer.pixel_filter` to make it public API. Updated top post accordingly. CI is green.

PR #643, opened 2024-02-13
Replace pkg_resources with importlib.resources

almarklein:
Like we did in https://github.com/pygfx/wgpu-py/pull/421. But in this case we require the data_dir to be on the file system, because we want to search it for font files.


almarklein:
Triggered by getting `ModuleNotFoundError: No module named 'pkg_resources'` on a fresh Python on Windows.

Issue #600, opened 2023-10-13
Improve buffer/texture update_range


almarklein:
During the work in (https://github.com/pygfx/gfxmorph) we realized two issues:

* It would be nice if the uploading of data could be more precise, rather than uploading the full range from minimal to maximal index. The most obvious solution is to define chunks to get a balance between number of upload calls and the size of the uploads.
* It may be convenient to specify indices to upload using a mask or list of indices, rather than a min-max-range.
* We observed that when using update_range to do precise uploads, some indices that changed were not updated, resulting in glitches. We've not yet found the cause, but it may be in pygfx.
* Also make a benchmark test, so we can compare speeds.
* Look into using the queue and e.g. buffer mapping. Note that `buffer.map()` is the busy-wait-sync version of `buffer.map_async()`, perhaps that makes it slow. On the other hand, using the async version might make things fast.
* See https://github.com/fastplotlib/fastplotlib/pull/473#issuecomment-2119817761 for some logic we can probably re-use from fpl.

kushalkolar:
This could be useful! Right now in fastplotlib, if precise indices are provided (such as with a binary mask with arbitrary indices), `update_range()` is just called for each one. I haven't benchmarked the performance since we haven't run into any yet. 
 https://github.com/fastplotlib/fastplotlib/blob/master/fastplotlib/graphics/_features/_base.py#L328

almarklein:
Thought that I had from #666: right now we check for updates right before a frame, which means that the drawing is delayed, and this delay may be variable. We could also update date right after a frame is rendered, resulting in a more stable fps. A downside might be that the display of new data is delayed. So should probably be optional? Also relates to https://github.com/pygfx/pygfx/issues/495.

hmaarrfk:
I think the reason that parts of the texture were not updated has to do with data alignment on the GPU

Issue #19, opened 2020-03-04
Improvements to Buffer


almarklein:
* [x] The `BufferWrapper` can probably just store the `usage` parameter, and let the renderer set a sensible default value when the gpu buffer is created.
* [x] Do we expose an API for mapping arrays? See #17 (No, we can't)
* [x] Property to use only a part of the geometry (mapping to `render_pass.draw()`args), see https://github.com/pygfx/pygfx/issues/414#issuecomment-1361006308
* [x] Use a separate buffer subclass for buffers that only live on the GPU?
  * No, but add a flag to buffers that allows us to semantically identify such buffers and make this explicit. This makes it easier to implement property validation, and to triage the kind of buffer in other places in the code.
* [x] Use a separate buffer subclass for uniform buffers, since they have no format?
  * No, but revise the `update_range` method. Let's rename it to `update`. Without arguments it should update the entire buffer. Then the arguments should make it possible to also mark ranges and bitmasks for an update.
  * Optimize the internals to support more fine-grained updates. See #600
* [x] ~Store less attributes on `self._store` since the size won't change (we initially were not sure whether buffer sizes could change).~ -> the reason for using `_store` for constants is that they still "change" from the pipeline's perspective, when a buffer is swapped with another.
* [x] Explain buffers and textures in the Guide in the docs (transferred from #114).
* [x] Support updating the buffer/texture data without copying data (to the `buffer.data` array).
  * Add a `data` property setter to update the data without needing to copy via `buffer.data[:] = other_data` as it is today.
* [x] Can we support other endianness? See #714

Postponed indefinitely until we identify valid use cases:

* [ ] Leverage numpy to define structured arrays and buffer vs view?
* [ ] Support for interleaved buffers (should follow from the above).
* [ ] Explicit properties for `stride` and `element_size`, or use the numpy array for that?

almarklein:
From Slack, how it works in threeJS:
```js
var arrayBuffer = new ArrayBuffer( particles * 16 );

var interleavedFloat32Buffer = new Float32Array( arrayBuffer );
var interleavedUint8Buffer = new Uint8Array( arrayBuffer );

// (this is not needed for non-interleaved buffers)
// InterleavedBuffer( array : TypedArray, stride : Integer )
var interleavedBuffer32 = new THREE.InterleavedBuffer( interleavedFloat32Buffer, 4 );
var interleavedBuffer8 = new THREE.InterleavedBuffer( interleavedUint8Buffer, 16 );

[...]

PR #647, opened 2024-02-15
Add texture's data to be set

hmaarrfk:
This helps avoid a copy when users overwrite the data.

I'm not sure if this is allowed, but it seems to "work".

### DIFF
--- a/examples/feature_demo/synced_video.py
+++ b/examples/feature_demo/synced_video.py
@@ -47,7 +47,7 @@ for i in range(3):
 def animate():
     # update with new random image
     for img in images:
-        img.geometry.grid.data[:] = np.random.rand(512, 512).astype(np.float32) * 255
+        img.geometry.grid.data = np.random.rand(512, 512).astype(np.float32) * 255
         img.geometry.grid.update_range((0, 0, 0), img.geometry.grid.size)
 
     renderer.render(scene, camera)
--- a/pygfx/resources/_texture.py
+++ b/pygfx/resources/_texture.py
@@ -126,6 +126,16 @@ class Texture(Resource):
         """
         return self._data
 
+    @data.setter
+    def data(self, data):
+        self._data = data
+        self._mem = memoryview(data)
+        self._store.nbytes = self._mem.nbytes
+        self._store.size = self._size_from_data(self._mem, self.dim, self.size)
+        Resource._rev += 1
+        self._rev = Resource._rev
+        self._gfx_mark_for_sync()
+
     @property
     def mem(self):
         """The data for this buffer as a memoryview. Can be None if


hmaarrfk:
i can add checks for continuity and same shape if you want.

almarklein:
I'm a bit hesitant, because currently the array that's passed to the texture stays the data placeholder. We even mention that in the docs "Note: the data is the same reference that was given to instantiate this object". It also says "but this may change", but I don't know what code in pygfx and downstream libs assume that the array does not change.

This relates to #19, which lists some possibilities for buffers, but some also apply to textures. Also relates to #600. If we are to change it, it'd probably be best to do it together with other (possibly breaking) changes to buffer/texture.

I assume the main point is speed by avoiding data copies?

hmaarrfk:
> I assume the main point is speed by avoiding data copies?
Yes the main point is to avoid data copies.

To be precise, when working with "4k" sized images this has a very noticeable effect on performance for me.

Thank you for pointing me to both those issues. I've subscribed and will try to contribute to those discussions however I can.

From my leanings this past week:

1. Changing the size of the data should not be allowed, at least not without other logic. 
    * NB. This was possible with vispy / opengl (though it is unclear if this was a bug we took advantage of).
2. I haven't really tested with anything other than just Intel integrated graphics + linux. So I too fear that there may some things that aren't OK with other processing pipelines.

Memory copies are really important to consider. So hopefully I can learn more about GSL and WGPU to be able to better contribute.

Thank you for the discussion. I'll close this and discuss on the aforementioned issues.

almarklein:
AFAIK the size of a texture in OpenGL is fixed, but I think we handled size changes in vispy at a relatively low level. In pygfx we opted (for now) to more "closely" wrap the GPU texture.

Do you think there are use-cases where being able to change the size has advantages? Under the hood a new texture object will have to be created anyway, so in terms of performance it won't differ from creating a new `gfx.Texture`.

hmaarrfk:
> Do you think there are use-cases where being able to change the size has advantages?

I don't really think so.

Personally, my biggest challenge right now is dealing with the "-0.5" offset in the images that seems to be occuring from somewhere in the pipeline ;) it will take me some time, but i'll figure it out. Your examples have been tremendously helpful.

I think that overtime, we got better at managing things between vispy and now pygfx so the advantages that seemed good 5 years ago when we first built our display pipeline are not actually useful.

For "changing the texture size" (that is the number of bytes) isn't so useful since removing and adding textures seems very very fast in pygfx.

almarklein:
> Personally, my biggest challenge right now is dealing with the "-0.5" offset in the images

The `gfx.Image` indeed has an offset so that the pixel center's are aligned with the grid, i.e. the bottomleft pixel is on top of the origin. This is hard-coded in the shader:
https://github.com/pygfx/pygfx/blob/329586c59e1dd98519794b0ef6159552ccd19ef6/pygfx/renderers/wgpu/imageshader.py#L63-L76

But you should be able to undo it by doing `image.local.position = 0.5, 0.5, 0`.

> [...] removing and adding textures seems very very fast in pygfx.

Yes it should be pretty fast. The bindings will be updated, and therefore the pipeline object recomposed, but there is no shader recompilation, which would be much more expensive.

hmaarrfk:
I should say that 
```
             let pos1 = vec2<f32>(-0.5); 
```

is probably the right way to interpret images. it is just a little annoying to do this change at the visualization level in our codebase.

hmaarrfk:
I've been listening and reading alot about WGPU and why it was created over vulkan.

Their primary desire was to ensure safety for the web, and rust really prides itself on  being memory safe.

I would be surprised if, given the little code I wrote that the proposed operation was not safe. It mostly avoids a python side copy in the data transfer chain.

almarklein:
I agree. My main reason from being hesitant is related to expectations about how the pygfx Texture object relates to the data it holds. I also agree we should probably support a path to avoid data copies when updating all data. I added a note to the top post of #19.

hmaarrfk:
A small question. Is the whole `_rev` stuff supposed to be a unique global counter? or just "each object's rev should be different when updating a particular object"

almarklein:
The id of a worldObject is globally unique. The `_rev`'s are per-object and should be monotonously increasing at each update.

Issue #651, opened 2024-02-19
Considering making properties like position return read only arrays


hmaarrfk:
https://github.com/pygfx/pygfx/blob/329586c59e1dd98519794b0ef6159552ccd19ef6/pygfx/utils/transform.py#L247

I learned my lesson trying to assign to the returned values of these arrays.

But maybe future users don't have to learn like I did?

Making read only views  is possible

We should be able to do this with:

```python
.flags['WRITEABLE'] = False
```

But honestly, i'm pretty sure it will break somebody else's code.

Anyway, thanks for the nice API.

Korijn:
It might be a good idea. Why don't you try it in a draft PR? I doubt it would "break" any more than making users aware of a mistake.

hmaarrfk:
Hmm I think it may already have caught some quirks in some python/numpy interactions:


```
__________________________________________________________________________________________ test_examples_run[multi_slice1] ___________________________________________________________________________________________

module = PosixPath('/home/mark/git/pygfx/examples/feature_demo/multi_slice1.py'), force_offscreen = None

    @pytest.mark.parametrize("module", examples_to_run, ids=lambda x: x.stem)
    def test_examples_run(module, force_offscreen):
        """Run every example marked to see if they can run without error."""
        # use runpy so the module is not actually imported (and can be gc'd)
        # but also to be able to run the code in the __main__ block
    
        # (relative) module name from project root
        module_name = module.relative_to(ROOT).with_suffix("").as_posix().replace("/", ".")
    
        # Reset logged warnings/errors
        log_handler.records = []
    
>       runpy.run_module(module_name, run_name="__main__")

examples/tests/test_examples.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../miniforge3/envs/mcam_dev/lib/python3.10/runpy.py:227: in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
[...]

PR #652, opened 2024-02-19
Make returned arrays of gometric transforms read only

hmaarrfk:
This should help users accidentally avoid writing to temporary arrays

Closes https://github.com/pygfx/pygfx/issues/651

### DIFF
--- a/pygfx/helpers/_axes.py
+++ b/pygfx/helpers/_axes.py
@@ -79,7 +79,12 @@ class AxesHelper(Line):
             arrow_head.local.position = pos
             # offset by half of height since the cones
             # are centered around the origin
-            arrow_head.local.position += arrow_size / 2 * la.vec_normalize(pos)
+            # Do not use +=
+            # see: https://github.com/pygfx/pygfx/issues/651
+            arrow_head.local.position = (
+                arrow_head.local.position +
+                arrow_size / 2 * la.vec_normalize(pos)
+            )
             arrow_head.local.rotation = la.quat_from_vecs(
                 (0, 0, 1), la.vec_normalize(pos)
             )
--- a/pygfx/helpers/_axes.py
+++ b/pygfx/helpers/_axes.py
@@ -82,8 +82,7 @@ class AxesHelper(Line):
             # Do not use +=
             # see: https://github.com/pygfx/pygfx/issues/651
             arrow_head.local.position = (
-                arrow_head.local.position +
-                arrow_size / 2 * la.vec_normalize(pos)
+                arrow_head.local.position + arrow_size / 2 * la.vec_normalize(pos)
             )
             arrow_head.local.rotation = la.quat_from_vecs(
                 (0, 0, 1), la.vec_normalize(pos)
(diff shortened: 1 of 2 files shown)


pygfx/helpers/_axes.py
```diff
@@ -79,7 +79,12 @@ def __init__(self, size=1.0, thickness=2):
             arrow_head.local.position = pos
             # offset by half of height since the cones
             # are centered around the origin
-            arrow_head.local.position += arrow_size / 2 * la.vec_normalize(pos)
+            # Do not use +=
+            # see: https://github.com/pygfx/pygfx/issues/651
+            arrow_head.local.position = (
+                arrow_head.local.position +
+                arrow_size / 2 * la.vec_normalize(pos)
```
Korijn, line 86: Well, that's an annoying and surprising side effect of this change! :/ I guess I knew that ndarrays overload this to be in-place...

pygfx/utils/transform.py
```diff
@@ -157,12 +157,20 @@ def scaling_signs(self):
...
             # and so there is no scaling component to preserve
             # any decomposed scaling is acceptable
-            return la.mat_decompose(self.matrix)
+            decomposed = la.mat_decompose(self.matrix)
+
+        # Stop the user from accidentally writing the temporary arrays
+        # that we return from these operations
+        # https://github.com/pygfx/pygfx/issues/651
+        for m in decomposed:
+            m.flags['WRITEABLE'] = False
```
Korijn, line 171: Should this flag also be set for other ndarrays in this class?

hmaarrfk: I think so. I just wanted to get feedback on the potential challenges before going too far in this.

It seems that searching for usages of `la.` will find most of them.

The 730 lines of code shouldn't be too hard to go through one by one either.

hmaarrfk:
Note that one of the arrays was already setting this flag with `flags.writeable` so I followed suit.

pygfx/utils/transform.py
```diff
@@ -544,6 +569,9 @@ def flag_update(self):
     @property
     def matrix(self):
         view_array = self.untracked_matrix.view()
+        # Stop the user from accidentally writing the temporary arrays
+        # that we return from these operations
+        # https://github.com/pygfx/pygfx/issues/651
         view_array.flags.writeable = False
```
Korijn, line 575: Does this flag only the view or does it also flag the original array?

hmaarrfk: only the view:

```python
In [1]: import numpy as np

In [2]: a = np.zeros((2, 3))

In [3]: b = a.view()

In [4]: b.flags.writeable = False

In [5]: assert a.flags.writeable

In [6]: a.flags.writeable
True

In [7]: assert not b.flags.writeable

In [8]: b.flags.writeable
False
```

Korijn:
I think this looks good to merge! @almarklein any thoughts?

Korijn (approved):

almarklein (approved):

PR #653, opened 2024-02-21
Use wgsl saturate instrinsic when we can.

hmaarrfk:
I searched for

```
rg "clamp\(.*0\.0, 1\.1 "
```

and it found a few occurances.

Honestly, this might be a micro optimization since the WGSL documentation states that "saturate expands to clamp(x, 0.0, 1.0)" but again, some hardware documentation lists that that the call to "saturate" is free.

I know this is a style thing, so feel free to close.

I did this cleanup in my own codebase so I figured I would oblige.

### DIFF
--- a/pygfx/renderers/wgpu/_shaderlib.py
+++ b/pygfx/renderers/wgpu/_shaderlib.py
@@ -24,16 +24,13 @@ class Shaderlib:
         return """
         // TODO smoothstep and saturate probably exists when Naga gets updated
         fn _smoothstep( low : f32, high : f32, x : f32 ) -> f32 {
-            let t = clamp( ( x - low ) / ( high - low ), 0.0, 1.0 );
+            let t = saturate( ( x - low ) / ( high - low ));
             return t * t * ( 3.0 - 2.0 * t );
-        }
-        fn _saturate( x: f32) -> f32 {
-            return clamp(x, 0.0, 1.0);
         }
         fn getDistanceAttenuation(light_distance: f32, cutoff_distance: f32, decay_exponent: f32) -> f32 {
             var distance_falloff: f32 = 1.0 / max( pow( light_distance, decay_exponent ), 0.01 );
             if ( cutoff_distance > 0.0 ) {
-                distance_falloff *= pow2( _saturate( 1.0 - pow4( light_distance / cutoff_distance ) ) );
+                distance_falloff *= pow2( saturate( 1.0 - pow4( light_distance / cutoff_distance ) ) );
             }
             return distance_falloff;
         }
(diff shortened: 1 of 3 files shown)


pygfx/renderers/wgpu/_shaderlib.py
```diff
@@ -24,16 +24,13 @@ def light_deps_basic(self):
...
             return t * t * ( 3.0 - 2.0 * t );
         }
-        fn _saturate( x: f32) -> f32 {
-            return clamp(x, 0.0, 1.0);
-        }
         fn getDistanceAttenuation(light_distance: f32, cutoff_distance: f32, decay_exponent: f32) -> f32 {
             var distance_falloff: f32 = 1.0 / max( pow( light_distance, decay_exponent ), 0.01 );
             if ( cutoff_distance > 0.0 ) {
-                distance_falloff *= pow2( _saturate( 1.0 - pow4( light_distance / cutoff_distance ) ) );
+                distance_falloff *= pow2( saturate( 1.0 - pow4( light_distance / cutoff_distance ) ) );
```
almarklein, line 33: Then we can remove the `_saturate()` function defined above as well.

almarklein: @panxinmiao I was wondering whether you deliberately avoided the wgsl builtin `saturate` here.

almarklein:
Ref https://www.w3.org/TR/WGSL/#saturate-float-builtin

I'd expect/hope that Naga would translate `clamp(x, 0.0, 1.0)`  to a `saturate(x)` call, if that indeed matters for performance. But just in case, and also because this makes the code easier to read, I agree with these changes.

pygfx/renderers/wgpu/_shaderlib.py
hmaarrfk: > Then we can remove the `_saturate()` function defined above as well.

I did that correct?

almarklein (approved):

Korijn (approved):

panxinmiao:
> @panxinmiao I was wondering whether you deliberately avoided the wgsl builtin `saturate` here. And if so, what the reason would have been :)

When writing these shader codes, `Naga` has not yet implemented the built-in `saturate` function (nor `smoothstep`) :)

pygfx/renderers/wgpu/_shaderlib.py
```diff
@@ -24,16 +24,13 @@ def light_deps_basic(self):
         return """
         // TODO smoothstep and saturate probably exists when Naga gets updated
```
hmaarrfk, line 25: i should have probably read english in addition to code...

PR #659, opened 2024-02-26
Using the built-in function "smoothstep"

panxinmiao:
See: https://github.com/pygfx/pygfx/pull/653#issuecomment-1963179267

When writing these shader codes, there are certain built-in functions in Naga that have not been implemented yet, including some that were later included in the WGPU specification.


Korijn (approved):

almarklein:
Thanks for the cleanup!

PR #660, opened 2024-02-26
Fix that colormaps were not sampled correctly

### LINKED ISSUE #655: Interpretation of colormaps
I was personally surprised with the results of using a very simple colormap `[(0, 0, 0), (1, 1, 1)]` with the image visuals.

image

What I found was that the colormap was sampled as if there were only "two" pixels placed at the position `0.25`, and `0.75`.

This sampling makes sense maybe for images, but it didn't jive with my expectation of colormaps.

In my own shader, I am shifting things and rescaling them so as to sample the full "range" of the colormap. 

1D example
```
           let sizef = f32(textureDimensions(t_colormap));
           let sample_coord = texcoord * (sizef - 1) / sizef + 0.5 / sizef;
           let value = textureSample(t_colormap, s_colormap, sample_coord);
```

For example, sampling a matplotlib colormap at a value of 0.1 will give you something much closer to 0.1 than the equivalent 2 value colormap in pygfx (pygfx will clamp it to 0).

<details><smmary>Matplotlib code</summary>

```python
import numpy as np
from matplotlib.colors import LinearSegmentedColormap

colors = [(0, 0, 0), (1, 1, 1)]

custom_cmap = LinearSegmentedColormap.from_list('custom_cmap', colors)

print(custom_cmap(0.1))
```

</details>

<details> <summary>PyGFX code</summary>

```python
import numpy as np
import pygfx as gfx

canvas = WgpuCanvas(size=(512 * 2 + 100, 400), max_fps=999)
renderer = gfx.renderers.WgpuRenderer(canvas, show_fps=True)
scene = gfx.Scene()
camera = gfx.OrthographicCamera(512 * 2 + 50 + 100, 550)
camera.local.y = 256
camera.local.scale_y = -1
camera.local.x = (512 * 2 + 50)/ 2

cmap = gfx.Texture(
    np.asarray([(0, 0, 0), (1, 1, 1)], dtype='float32'),
    dim=1
)

img = np.arange(256, dtype='float32').reshape(16, 16)
geometry = gfx.Geometry(grid=gfx.Texture(img, dim=2))

image0 = gfx.Image(
    geometry,
    gfx.ImageBasicMaterial(clim=(0, 255))
)
image0.local.scale = 512 / 16, 512/16, 1

scene.add(image0)

image1 = gfx.Image(
    geometry,
    gfx.ImageBasicMaterial(clim=(0, 255), map=cmap)
)
image1.local.scale = 512 / 16, 512/16, 1
image1.local.x = 50 + 512

scene.add(image1)

def animate():

if __name__ == "__main__":
    canvas.request_draw(animate)
    run()
```

</details>

Maybe there is an easier way to do this by changing how the GPU interpolates a texture?

Thank you for the discussion.

almarklein:
Closes #655

image

### DIFF
--- a/examples/validation/validate_image_colormap.py
+++ b/examples/validation/validate_image_colormap.py
@@ -4,8 +4,9 @@ Simple Colormap
 
 Show an image with a simple colormap.
 
-* You should see a square image with 3 equally sized vertical bands.
-* The bands should be red, green, and blue.
+* You should see two rectangles with a colormap.
+* The top goes smoothly from red to green to blue.
+* The bottom shows three equally thick bands (red, green blue).
 """
 
 # test_example = true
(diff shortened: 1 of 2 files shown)


Korijn (approved):

hmaarrfk:
Thanks for the quick fix

Issue #495, opened 2023-04-04
Update modes


Korijn:
Somewhat related to discussion in #490

I was able to pool the collective memory of HN on this topic: https://news.ycombinator.com/item?id=35435669

What really struck me as an interesting option is the way Bevy handles it: https://docs.rs/bevy/0.10.1/bevy/winit/enum.UpdateMode.html

But we should also remain attentive to the "reactive" methodology where a central state object is used to trigger redraws (as @berendkleinhaneveld has implemented in his project), and this method is not supported by Bevy either.

almarklein:
We are already close to the reactive mode (I think we're actually closer to `ReactiveLowPower`). The `request_draw` mechanism that we have is pretty neat in that it is more precise than e.g. drawing whenever an event happens (most button-less mouse move events don't require a redraw). It also works well for reactive rendering.

I do like the idea to queue events, and process them all right before a draw. We already rate-limit events, but bringing it under the update step together with drawing is even faster, and I suspect also leads to simpler code.

How this could work:

* Schedule updates at a fixed interval using the GUI's scheduler (e.g. at 30fps, 60 fps, or as fast as possible).
* On each update:
  * Process all events in the queue.
  * Emit a custom "update event" that animations can use. (we currently call this before_draw).
  * Both of these can request new draws.
  * Only when a new draw has indeed been requested, we do the actual draw.

This should also work nice with reactive rendering:
- Disable pygfx from calling `request_draw` by itself.
	- Setting `auto_update=False` on the controller.
	- Could even subclass the renderer to make `request_draw()` do nothing.
- Listen to events, and  use them to update the app's state object instead of changing the state of pygfx world objects.
- React to changes in the app state, two options:
	- 1: Watch state changes as they occur, and when the state changes, request a new draw.
	- 2: Check for state changes in the 'update' event, and request a new draw when state has changed.

In order to implement this, we 'd need the following changes:
* in wgpu:
  * implement an update-loop that runs at a steady interval, as outlined above.
* in pygfx:
  * replace the before_draw for the update event.

Korijn:
>   * Emit a custom "update event" that animations can use. (we currently call this "before_render").

I assume this would be emitted by the GUI layer, right?

almarklein:
Yes, I think the canvas would emit it, and the renderer would re-emit it just like it does for the other events.

[...]

PR #665, opened 2024-03-01
Basic input class

### LINKED ISSUE #664: lack of "key_hold" events
Is it possible to have a `"key_hold"` event that is triggered when a key is held in the "down" position? It seems like the `"key_down"` event is only triggered once when a key is pressed, [there doesn't seem to be an event](https://github.com/pygfx/pygfx/blob/main/pygfx/objects/_events.py#L15) that corresponds to a key being held down. 

One possible way is to add a animation function that is called before rendering each frame to check if a key is being held down, but this could be inefficient?

Korijn:
Closes #664

Adds a basic Input class with an example demonstrating how to implement simple FPS controls with it

### DIFF
--- a/pygfx/controllers/_input.py
+++ b/pygfx/controllers/_input.py
@@ -1,0 +1,35 @@
+from ..utils.viewport import Viewport
+
+
+class Input:
+    def __init__(self, register_events=None):
+        self.key_down = {}
+        self.key_down = {}
+        if register_events is not None:
+            self.register_events(register_events)
+
+    def register_events(self, viewport_or_renderer):
+        viewport = Viewport.from_viewport_or_renderer(viewport_or_renderer)
+        viewport.renderer.add_event_handler(
+            lambda event: self.handle_event(event, viewport),
+            # "pointer_down",
+            # "pointer_move",
+            # "pointer_up",
+            "key_down",
+            "key_up",
+        )
+
+    def handle_event(self, event, viewport):
+        type = event.type
+
+        if type.startswith("key_"):
+            # modifiers = {m.lower() for m in event.modifiers}
+            # modifiers.discard(event.key.lower())
+
+            if event.type == "key_down":
+                self.key_down[event.key.lower()] = True
+            elif event.type == "key_up":
+                self.key_down[event.key.lower()] = False
+
+    def is_key_down(self, key):
+        return self.key_down.get(key.lower(), False)
--- a/examples/feature_demo/input_state.py
+++ b/examples/feature_demo/input_state.py
@@ -1,0 +1,84 @@
+"""
+Input state
+===========
+
+Example showing how to use the Input class.
+"""
+
+# sphinx_gallery_pygfx_render = True
+# sphinx_gallery_pygfx_target_name = "disp"
+
+import numpy as np
+import pygfx as gfx
+from wgpu.gui.auto import WgpuCanvas, run
+
+
+renderer = gfx.renderers.WgpuRenderer(WgpuCanvas())
+scene = gfx.Group()
+
+# add background
+background = gfx.Background(None, gfx.BackgroundMaterial("#ccc"))
+scene.add(background)
+
+# add floor
+floor = gfx.Mesh(
+    gfx.plane_geometry(20, 20),
+    gfx.MeshPhongMaterial(color="#808080", side="Front"),
+)
+floor.local.euler_x = np.pi * -0.5
+floor.receive_shadow = True
+scene.add(floor)
+
+# add torus
+torus = gfx.Mesh(
+    gfx.torus_knot_geometry(1, 0.3, 128, 16),
+    gfx.MeshPhongMaterial(color="#fff", side="Front"),
+)
+torus.local.y = 2
+torus.cast_shadow = True
+scene.add(torus)
+
+# add lights
+ambient = gfx.AmbientLight("#fff", 0.1)
+scene.add(ambient)
+
+light = gfx.DirectionalLight("#aaa")
+light.local.x = -50
+light.local.y = 50
+light.cast_shadow = True
+light.shadow.camera.width = 100
+light.shadow.camera.height = 100
+light.shadow.camera.update_projection_matrix()
+scene.add(light)
+
+# add camera
+camera = gfx.PerspectiveCamera(70, 16 / 9)
+camera.local.position = 6, 2, 6
+camera.look_at((0, 2, 0))
+
+# add input
+input = gfx.Input(register_events=renderer)
+velocity = 0.1
+
+
+# setup animation loop
+def animate():
+    # implement basic FPS controls
+    movement = np.array([0.0, 0.0, 0.0])
+    if input.is_key_down("w"):
+        movement += camera.local.forward * velocity
+    if input.is_key_down("a"):
+        movement += camera.local.right * -velocity
+    if input.is_key_down("d"):
+        movement += camera.local.right * velocity
+    if input.is_key_down("s"):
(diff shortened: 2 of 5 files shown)


kushalkolar:
You guys are epic and so fast! With pointer events I can already see how it can simplify our [drag PR](https://github.com/fastplotlib/fastplotlib/pull/413) ! I didn't realize we can just subclass from the existing Controller :D

Korijn:
I'm working to at least include all the basic functionality in this PR with pointer state as well.

pygfx/controllers/_input.py
```diff
@@ -0,0 +1,35 @@
+from ..utils.viewport import Viewport
+
+
+class Input:
+    def __init__(self, register_events=None):
+        self.key_down = {}
+        self.key_down = {}
```
almarklein, line 7: Duplicate line.

Korijn:
This is now working. Could one of you give it a try locally and review?

@almarklein I was thinking to move the FPS controls into a controller class, but it will have a different architecture and design than the other controllers. I also don't want to refactor all of them. Do you have any ideas on how to deal with that? I was thinking about creating another subpackage `pygfx.input`...

examples/feature_demo/input_fps.py
```diff
@@ -0,0 +1,129 @@
...
+    if input.key("a"):
+        movement += camera.local.right * -velocity
+    if input.key("d"):
+        movement += camera.local.right * velocity
+    if not fly:
+        # project movement vector onto the floor
+        movement -= (
+            np.dot(movement, camera.local.reference_up) * camera.local.reference_up
+        )
+    camera.local.position = camera.local.position + movement
```
Korijn, line 117: @FirefoxMetzger I just wanted you to see how elegant this FPS controls implementation is thanks to your work <3

pygfx/controllers/_time.py
```diff
@@ -0,0 +1,40 @@
...
+        return perf_counter()
+
+    @property
+    def t0(self):
+        """Returns the time in seconds at which the previous frame was rendered."""
+        return self._t0
+
+    @property
+    def delta(self):
+        """Returns the time in seconds since the last frame."""
```
Korijn, line 39: This is actually the time the previous frame took to render 🤦‍♂️

almarklein: It's likely even longer than the time it took to render the frame (fps limiting).

Would we want to perhaps smooth this value, to avoid variations in it from affecting the way the interaction feels?

pygfx/controllers/_input.py
```diff
@@ -0,0 +1,121 @@
...
+            elif type == "key_up":
+                self._key_state[key] = False
+                self._key_up[key] = True
+
+        elif type == "wheel":
+            self._wheel_delta[event.pointer_id] = event.dy
+
+    def key_down(self, key):
+        """Returns True if the key_down event was received in the current frame"""
+        return self._key_down.get(key.lower(), False)
```
almarklein, line 89: Is there a benefit to this, compared to listening for key_down events directly? Also in theory a user can press a key twice, and using this method only see one. Or is that intentional?

examples/feature_demo/input_fps.py
```diff
@@ -0,0 +1,129 @@
...
+import pygfx as gfx
+from wgpu.gui.glfw import WgpuCanvas, run
+
+
+canvas = WgpuCanvas()
+renderer = gfx.renderers.WgpuRenderer(canvas)
+scene = gfx.Group()
+
+# hide the cursor, lock it to the window and keep it centered
+glfw.set_input_mode(canvas._window, glfw.CURSOR, glfw.CURSOR_DISABLED)
```
almarklein, line 23: Nice! Can we also listen to Esc to free the mouse again?

kushalkolar: Just played with it and I agree :laughing:

almarklein:
Tried it and it works really nice!  Also feels immersive because the mouse is grabbed.

>  I was thinking about creating another subpackage pygfx.input...

Yes, I think that makes the most sense. I think controllers are quite different because they closely relate to cameras. Perhaps we can (later) refactor the controllers to make use of the input class(es) to make them simpler. I think wrapping makes more sense than subclassing here, what do you think?

examples/feature_demo/input_fps.py
almarklein: Impressive with how few lines of code this takes! ❤️

kushalkolar:
The example works well for me along with small modifications to try out various things (moving the torus and the view, sensitivities, etc.) :smile:

pygfx/controllers/_input.py
Korijn: It's an abstraction that makes it possible for many once-per-frame update functions to avoid the boilerplate of receiving events and tracking state of inputs :)

examples/feature_demo/input_fps.py
Korijn: Haha, how often do you play shooter video games? Usually you alt tab out to get the mouse back. But yes, I can try to implement that.

pygfx/controllers/_input.py
almarklein: I see the convenience, but it also add the constraint that any code that uses this must run "in sync", i.e. exactly once per draw.

examples/feature_demo/input_fps.py
kushalkolar: Just tried and alt+tab works! my gaming muscle memory is usually not present when working and I just panic instead :P

pygfx/controllers/_input.py
Korijn: Yes, although that is intentional

pygfx/controllers/_time.py
Korijn: After thinking about this some more, I think it's correctly implemented:

* For every `after_flush` event, there is only 1 call to `perf_counter()`, so we don't "lose time"
* The `delta` should have the same value for all update handlers running in a given frame
* The `delta` should be equal to the time since the previous point in time where update handlers were executed, which is usually right after `after_flush`

examples/feature_demo/input_fps.py
Korijn: Added some code to release the mouse with escape (and ignore mouse events while it is released). Clicking on the window captures it again.

pygfx/controllers/_input.py
Korijn: I also wanted to mention, in an earlier state of this code I had used `input.key(...)` instead of `input.key_down(...)`. It was very tricky to toggle fly mode because pressing the key would make `input.key(...)` evaluate to `True` for multiple frames. In other words, it would rapidly toggle where I just wanted to toggle it once. That's the use case for `input.key_down(...)`.

Korijn:
* Added escape/mouse click handlers to release/capture the mouse
* Added acceleration to make movement smoother

The example is "done" I think, next step is reorganizing the code into a `pygfx.inputs` folder.

Korijn:
The more I think about this the more I feel like it should be a part of wgpu-py's GUI layer... I will see what that feels like by trying it out on a branch.

kushalkolar:
I'll give this a shot in wgpu if you can hint me at where it would go!

Korijn:
> I'll give this a shot in wgpu if you can hint me at where it would go!

`wgpu/gui/input.py` :)

almarklein:
Closing. We will get back to this when we work on #495. We see two approaches, but we'll have to see at the time:
* The animation event has info on what buttons are down.
* The canvas has API to check what buttons are down.

PR #669, opened 2024-03-05
add zmq example of how to compute data in one process and view it in another process

kushalkolar:
A simple example that uses `zmq` to send data from a compute process to a separate render process that pygfx uses to render the data. One could do this with python builtin multiprocessing too. ZMQ has the advantage that it works across languages.

related: #572, #667

https://github.com/pygfx/pygfx/assets/9403332/7a5b5cca-b4dc-467e-b42e-9fc39276b0d9


Korijn:
Awesome! Very nice solution. This is great.

Korijn (approved):

PR #672, opened 2024-03-07
Buffer and Texture test for c_contiguous

### LINKED ISSUE #307: dealing with non-contiguous arrays

almarklein:
Closes #307

Somewhat relates to #19


almarklein:
Interestingly, we use buffers with non-contiguous data in some of our examples.

tests/resources/test_texture.py
```diff
@@ -109,3 +109,14 @@ def test_unsupported_shapes():
         a = np.zeros((10, 10, 10, i), np.float32)
         with pytest.raises(ValueError):
             gfx.Texture(a, dim=3)
+
+
+def test_contiguous():
+    im0 = np.zeros((100, 100), np.float32)
+    im1 = im0[10:-10, 10:-10]
```
kushalkolar, line 116: I think that non-contiguous textures work because I do this all the time, unless we're doing an additional step in fastplotlib which makes it contiguous. I can check later.

kushalkolar:
> Interestingly, we use buffers with non-contiguous data in some of our examples.

Interesting, which examples?

almarklein:
My mistake. We do actually support non-contiguous buffer/texture data. The data is made contiguous when it's sampled for upload (in `._get_subdata()`). The problem was that there was a check for `.contiguous`, which should have been `.c_contiguous`.

Additionally, the corresponding check for textures was just copy-pasted from the buffer, and always failed. This means a copy was made in all cases. I fixed that too.

@hmaarrfk I think I removed a data-copy when uploading texture data to the GPU here. Curious if you can feel the difference 😉

Korijn (dismissed):

almarklein:
Sorry @Korijn, I spotted a few small improvements. I should've waited a sec with marking for review.

Korijn (approved):

hmaarrfk:
Sorrry I can’t spot obviously where a copy was removed here

Korijn:
> Sorrry I can’t spot obviously where a copy was removed here
> 
> to me it looks like checks for men.contiguous were changed to mem.c_contiguous

See `test_contiguous`

hmaarrfk:
maybe i misunderstood, you are avoiding data copies when data is transfered from the gpu, to the cpu is that correct?

I unfortunately do not really have any way to test this ;)

almarklein:
@hmaarrfk to clarify: in the `get_subdata()` func, there was a check whether the memoryview could be passed to the upload machinery as-is. But this check was completely broken for the Texture, causing it to always copy, because it thought the data was not contiguous 🤦

almarklein:
I think the copy we removed in this PR was an extra copy *on top* of the other one. The one you work around with #647 is when the user sets the data on the buffer. The one here is when the internals of pygfx get data to upload to the GPU.

hmaarrfk:
I think in the well behaved case:

```python
        if isinstance(self.data, np.ndarray):
            arr = self.data
        elif not self.mem.c_contiguous:
            raise ValueError(
                "Non-contiguous texture data is only supported for numpy array."
            )
        else:
            arr = np.frombuffer(self.mem, self.mem.format)
        arr = arr.reshape(self.size[2], self.size[1], self.size[0], -1)
        # Slice it
        slices = []
        for d in reversed(range(3)):
            slices.append(slice(offset[d], offset[d] + size[d]))
        sub_arr = arr[tuple(slices)]
        if pixel_padding is not None:
            padding = np.ones(sub_arr.shape[:3] + (1,), dtype=sub_arr.dtype)
            sub_arr = np.concatenate([sub_arr, pixel_padding * padding], -1)
        return memoryview(np.ascontiguousarray(sub_arr))
```

this code amounted to a no-op in my case, and just
```python
memorview(self.data)
```

from ipdb:

```python
> /home/mark/git/pygfx/pygfx/resources/_texture.py(255)_get_subdata()
    254             sub_arr = np.concatenate([sub_arr, pixel_padding * padding], -1)
--> 255         return memoryview(np.ascontiguousarray(sub_arr))
    256 

ipdb> sub_arr.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
```

almarklein:
Ah, yes, I think you're right. I figured `sub_arr = arr[tuple(slices)]` would mean a copy is made, but that'd be a view, of course. Sorry for the noise 😄

PR #675, opened 2024-03-14
Refactor wgpu renderer

almarklein:
* [x] Move `line.wgsl` into `renderers.wgpu.wgsl` package. This is what I started out with, but then one thing led to another.
* [x] Refactored the wgu renderer into a few subpackages, making the code easier to follow.
   * `engine`: the stuff that make pygfx do its magic.
   * `shader`: the base shader class, templating, common code, generating wgsl from bindings, for varyings, and outputs, etc.
   * `wgsl`: `xx.wgsl` files.
   * `shaders`: the builtin shaders. These now import all their needs from `gfx.renderers.wgpu` (i.e. from the public API, kinda).
* [x] Update setup.py
* [x] Fix some examples, test custom shader examples.
* [x] Probably some tests need updates too.
* [x] Fix, or just remove (?), the qt text overlay example.
* [x] Update docs.

I want to refactor the base shader class too, because it's a bit messy, and it would be nice to put some of the wgsl into actual `.wgsl`  files. But I'll do that in another PR, so that this one is (nearly) only concerned with moving modules to new places.


Korijn:
> Fix, or just remove (?), the qt text overlay example

Go ahead and remove it.

pygfx/renderers/wgpu/wgsl/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+def load_wgsl(shader_name):
+    """Load wgsl code from the file system."""
+    package_name = "pygfx.renderers.wgpu.wgsl"
+    if sys.version_info < (3, 9):
+        context = importlib.resources.path(package_name, shader_name)
+    else:
+        ref = importlib.resources.files(package_name) / shader_name
+        context = importlib.resources.as_file(ref)
+    with context as path:
+        with open(path, "rb") as f:
```
Korijn, line 17: Should this have `encoding="utf-8"`?

almarklein: No it opens as binary ("rb") and then `.decode()`  on the line below (which defaults to utf-8).

pygfx/renderers/wgpu/shaders/backgroundshader.py
```diff
@@ -1,10 +1,18 @@
 import wgpu  # only for flags/enums
 
-from . import register_wgpu_render_function, WorldObjectShader, Binding, RenderMask
-from ._utils import to_texture_format, GfxSampler, GfxTextureView
-from ...objects import Background
-from ...materials import BackgroundMaterial, BackgroundImageMaterial
-from ...resources import Texture
+from ....objects import Background
+from ....materials import BackgroundMaterial, BackgroundImageMaterial
+from ....resources import Texture
```
almarklein, line 5: These four dots hurt my eyes a bit. I thought about collapsing `renderers/wgpu/`  into `wgpurenderer/` to reduce one level of nesting. I half-like that idea, but when I think about possibly having more renderers in the future, I very much like the current structure ...

Korijn (approved):

PR #679, opened 2024-03-18
Documentation improvements

almarklein:
### Done

* [x] Fix that "Examples" heading was shown also when there was no mini-gallery to show.
* [x] Show mini-gallery at the bottom of the page (instead of between class docstring and its methods/props).
* [x] Fix that the mini-gallery showed no examples for certain groups of classes. 
* [x] Remove the need for setting `target_name` by making the canvas-selection a wee bit smarter. 
* [x] Animations behave much better w.r.t. fps.
* [x] Use webp for screenshots too, because it's much smaller than png, even in lossless mode.
* [x] Use `sphinx_gallery_pygfx_docs` comment to control per-file gallery config.
* [x] Use `sphinx_gallery_pygfx_test` comment to control per-file test config.
* [x] Both of these comments must be present in each example. This is tested for.
* [x] Add advanced docs for how to use our gallery scraper.
* [x] Updated all examples accordingly (this is mostly what makes this pr so big).


examples/HOWTO.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+## Testing
+
+Examples have three possible modes with regards to testing. For this purpose the
+`# example_testing:` comment must be present in each file.
+
+* `# example_testing: no` - not tested.
+* `# example_testing: run` - tested by running the code (including rendering a frame).
+* `# example_testing: compare` - tested by running the code, rendering a frame, and comparing it to the reference (i.e. image comparisons).
```
Korijn, line 28: can you update the text in the README to refer to this explanation?

examples/HOWTO.md
```diff
@@ -0,0 +1,45 @@
...
+example.
+
+* `example_gallery: hidden` - not present in the gallery.
+* `example_gallery: code` - present only as code (don't run the code when building docs).
+* `example_gallery: screenshot` - present in the gallery with a screenshot.
+* `example_gallery: animate 3s` - present in the gallery with an animation (length specified in seconds).
+
+Examples that are animated should ``from time import perf_counter``, and use it
+to determine the speed of the animation. The ``perf_counter`` variable is
+replaced during rendering to simulate time. This way the animation is also
```
Korijn, line 44: you mean "patched" right?

pygfx/utils/gallery_scraper.py
```diff
@@ -1,3 +1,5 @@
+import re
```
Korijn, line 1: why is the gallery_scraper.py file not in the `docs` folder actually? are we shipping it so dependents of pygfx can use it to build their own galleries?

almarklein: I first thought it was because Sphinx required it for some reason, but I think your explanation is correct. That way e.g. fastplotlib and others can create a gallery without replicating the scraping mechanics.

examples/HOWTO.md
almarklein: yeah patched is a better word here, though technically we replace the global variable in the example's scope. That way we don't have to patch the `time` module itself.

examples/HOWTO.md
almarklein: It looks like we don't refer to this anywhere. (In wgpu we do refer to it in the readme.)

almarklein: I can add a note in the contributing.md

almarklein:
I changed the comment to `# sphinx_gallery_pygfx = 'xxx'`, because that's the standard that sphinx-gallery. Sphinx also "sees" the comment and tries to `eval()` the value ... but it does not seem to pass that value to the scraper? 

I am a bit frustrated because my general feeling during this whole PR is that on the one hand sphinx-gallery is rather opinionated about how it should be used, but on the other hand that API feels incomplete / half-baked. 😕  I just wanted to simplify the way that examples are configured for the gallery, but it got me jumping hoops.

---

Aaaanyway ... I'm now on the fence about two options:
```py
# Kinda what we have now. The "problem" is that the second comment is shown in the code in the gallery.
# sphinx_gallery_pygfx = 'screenshot'
# example_testing = compare'

import pygfx as gfx
...
```
vs
```py
# If we re-use/mis-use the gallery option mechanism, sphinx will remove both comments from the gallery code.
# sphinx_gallery_pygfx_test = compare'

import pygfx as gfx
...
```

Korijn:
I honestly don't care. Either option is fine if we document it. People will just copy paste whatever we have anyway.

kushalkolar:
> People will just copy paste whatever we have anyway.

yup! does anyone actually know how sphinx works fully :laughing:

almarklein:
We don't have a gallery there, so I guess not? Perhaps if/when wgpu-py gets some more examples ...

Korijn:
Well, the changes you made to the example testing stuff for example.

almarklein:
Yeah, I can apply that to wgpu too, for consistency.

PR #680, opened 2024-03-18
Better line-render behavior for nonfintes

almarklein:
Matlab ignores inf just like it doe nans:
```py
import matplotlib.pyplot as plt
plt.plot([1, 2, 3, np.nan, 3, 2, 1, -np.inf, 1, 2, 3, np.inf, 3, 2, 1])
plt.show()
```

This PR makes our line renderer do the same.

### DIFF
--- a/pygfx/renderers/wgpu/shaders/lineshader.py
+++ b/pygfx/renderers/wgpu/shaders/lineshader.py
@@ -145,8 +145,7 @@ class LineShader(WorldObjectShader):
                 vertex_array_sub = xyz[:, :2] * (0.5 * np.array(logical_size))
             # Fix up
             if has_non_finites:
-                vertex_array = np.empty((positions_array.shape[0], 2), np.float32)
-                vertex_array.fill(np.nan)
+                vertex_array = np.full((len(positions_array), 2), np.nan, np.float32)
                 vertex_array[finites] = vertex_array_sub
             else:
                 vertex_array = vertex_array_sub


pygfx/renderers/wgpu/shaders/lineshader.py
```diff
@@ -125,13 +125,31 @@ def bake_function(self, wobject, camera, logical_size):
...
-            vertex_array = la.vec_transform(positions_array, wobject.world.matrix)
-        else:  # wobject.material.thickness_space == "screen":
-            xyz = la.vec_transform(
-                positions_array, camera.camera_matrix @ wobject.world.matrix
-            )
-            vertex_array = xyz[:, :2] * (0.5 * np.array(logical_size))
+        else:
+            # Prep
+            finites = np.isfinite(positions_array).all(axis=1)
+            has_non_finites = finites.sum() != finites.size
```
Korijn, line 131: I think `finites.any()` is sufficient and much faster since it can exit early

almarklein: Good one.

almarklein: Oh, actually, it needs to know if there are any non-finites. So it should be `not finites.all()` 😄

almarklein: lol, apparently was looking at the comment before it was edited.

Korijn (approved):

PR #685, opened 2024-03-21
Make pixel_ratio follow canvas if it changes

almarklein:
Changes the pixel ratio logic again (see e.g. #638).

* [x] If the `renderer.pixel_ratio` is set to None, this will now follow the canvas' ratio, e.g. as it moves to a different screen, instead of staying fixed to whatever it was at the start.
* [x] Fixes the raising of a ZeroDivisionException in the notebook, because the initial size of the canvas is `(0, 0)`.

@kushalkolar have you seen that bug before? Perhaps you just set an explicit pixel_ratio to prevent it?

### Implementation notes

* The `self._pixel_ratio` is not used outside of the property anymore.
* It is set to `None` to mean "auto mode".
* So `renderer.pixel_ratio` may change in the background, but it still always produces a value (so this pr does not undo #638).

### DIFF
--- a/pygfx/renderers/wgpu/engine/renderer.py
+++ b/pygfx/renderers/wgpu/engine/renderer.py
@@ -214,26 +214,21 @@ class WgpuRenderer(RootEventHandler, Renderer):
         with less jagged edges. Alternatively, this value can be set
         to e.g. 0.5 to *lower* the resolution.
         """
-        return self._pixel_ratio
+        if self._pixel_ratio is not None:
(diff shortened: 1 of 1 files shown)


Korijn (approved):

kushalkolar:
> @kushalkolar have you seen that bug before? Perhaps you just set an explicit pixel_ratio to prevent it?

Yup, we set it to 2 by default few weeks ago: https://github.com/fastplotlib/fastplotlib/pull/412

PR #687, opened 2024-03-25
Fix update issues related to render_mask

### LINKED ISSUE #682: Error changing material color when adding or removing alpha channel
Hi! I ran into an error trying to change mesh colors on-the-fly, i.e. on a running ,non-blocking window.

```python
>>> # Generate mesh
>>> cube = gfx.Mesh(gfx.box_geometry(200, 200, 200), gfx.MeshPhongMaterial(color="#336699"))
>>> # Add to already being rendered scene
>>> scene.add(cube)
>>> # Changing color like this works like a charm 
>>> cube.material.color = gfx.Color("r")

>>> # Now the same thing but we use a (r, g, b, a) color
>>> cube2 = gfx.Mesh(gfx.box_geometry(200, 200, 200), gfx.MeshPhongMaterial(color=(0, 0, 1, .5)))
>>> scene.add(cube2)
>>> cube2.material.color = gfx.Color("r")

Draw error: 0 (57)
No texture to present, missing call to get_current_texture()?
```

After a bit of experimentation it seems that if the material is initialized with an alpha channel, we break things by trying to assign a color without alpha channel. The other way around causes the same issue. I don't know whether that's expected and/or fixable but perhaps in the first instance the `.color` setter method could prohibit adding/removing the alpha channel?

almarklein:
Closes #682

* [x] Store value of `buffer.format` at `buffer._store.format`, so its value can be tracked.
* [x] In the shaders, avoid using `buffer.data.shape`, but use (in this case) `buffer.format` instead, so it is tracked.
* [x] Add a `nchannels_from_format` utility function for this purpose. 
* [x] In `pipelines.py`, the update path ends sooner by tracking whether the objects are up-to-date for a specific environment much earlier.
* [x] Also call `environment.register_pipeline_container()` earlier. This puts the environment-handling together, and out of the rest of the logic.
* [x] When an update *does* proceed, it iterates just a bit deeper, checking for each pass whether a shader must be compiled and/or pipeline must be composed. *This is the heart of the fix.*

Consider the scenario where an object goes from fully opaque to fully transparent, i.e. needing only a pipeline for the opaque pass vs only needing a pipeline for the transparent pass. There are two ways to make this work:
* Invalidate the whole set of pipelines (all passes).
* Fill in missing pipelines.

I went with the second option. It may cause having pipeline objects that are not actually used, but it prevents repeated shader compilation when i.e. the transparency of an object is changed to highlight it.

### DIFF
--- a/pygfx/renderers/wgpu/__init__.py
+++ b/pygfx/renderers/wgpu/__init__.py
@@ -37,6 +37,7 @@ Lower level functions that may or may not be needed in custom shaders:
     :toctree: _autosummary/renderers/wgpu
     :template: ../_templates/custom_layout.rst
 
+    nchannels_from_format
     to_vertex_format
     to_texture_format
 
@@ -50,6 +51,7 @@ Lower level functions that may or may not be needed in custom shaders:
 from ...objects._base import RenderMask
 from .engine.utils import (
     register_wgpu_render_function,
+    nchannels_from_format,
     to_vertex_format,
     to_texture_format,
     GfxSampler,
--- a/pygfx/renderers/wgpu/engine/pipeline.py
+++ b/pygfx/renderers/wgpu/engine/pipeline.py
@@ -163,8 +163,7 @@ def get_pipeline_container_group(wobject, environment):
         changed_labels = wobject.tracker.pop_changed()
 
     # Update if necessary - this part is defined to be fast if there are no changes
-    if changed_labels:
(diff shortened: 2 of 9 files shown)


Korijn:
Would it be worthwhile to add a regression unit test? Or not?

almarklein:
Yes, I"ll have a look.

Korijn (approved):

PR #691, opened 2024-03-26
Fix `look_at`

### LINKED ISSUE #678: Weird shadow flicker
Saw it in look_at.py example. Not sure if other examples are affected too.

panxinmiao:
Close #678 

Fixed a bug in the `look_at` method when the new `forward` vector is parallel to the current `forward` vector, taking into account precision errors when determining parallelism.

See: https://github.com/pygfx/pygfx/issues/678#issuecomment-2019600974

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -6,6 +6,8 @@ import functools
 
 from typing import Tuple, Union
 
+PRECISION_EPSILON = 1e-7
+
 
 class cached:  # noqa: N801
     """Cache for computed properties.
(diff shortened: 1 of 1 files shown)


panxinmiao:
Additionally, it may be worth considering where to define PRECISION_EPSILON more appropriately (perhaps as a constant in `pylinalg`?).

It may also be necessary to check if there are other parts of the code that need to account for precision errors.

@almarklein @Korijn

almarklein (approved):
Having the prevision constant here for now is fine by me. At a later time we can check if we use such epsilons in different places. But that's not urgent, I think.

Korijn (approved):

PR #693, opened 2024-03-28
Add point sprites

almarklein:
Not necessarily the ones useful for creating GUI elements. But more so for particle effects etc.

image

### DIFF
--- a/pygfx/materials/_points.py
+++ b/pygfx/materials/_points.py
@@ -242,7 +242,7 @@ class PointsSpriteMaterial(PointsMaterial):
     """A material to render points as sprite images.
 
     Renders the provided texture at each point position. The images are square
-    and sized just like with a PointMaterial. The texture color is multiplierd
+    and sized just like with a PointMaterial. The texture color is multiplied
     with the point's "normal" color (as calculated depending on ``color_mode``).
 
     The sprite texture is provided via ``.sprite``.
--- a/pygfx/materials/_points.py
+++ b/pygfx/materials/_points.py
@@ -259,12 +259,12 @@ class PointsSpriteMaterial(PointsMaterial):
         The dimensionality of the map must be 2D. If None, it just shows a
         uniform color.
         """
-        return self._sprite
+        return self._store.sprite
 
     @sprite.setter
     def sprite(self, sprite):
         assert sprite is None or isinstance(sprite, Texture)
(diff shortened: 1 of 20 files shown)


pygfx/materials/_points.py
```diff
@@ -238,5 +238,34 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+        self.sprite = sprite
+
+    @property
+    def sprite(self):
+        """The texture map specifying the sprite image.
+
+        The dimensionality of the map must be 2D. If None, it just shows a
+        uniform color.
+        """
+        return self._sprite
```
Korijn, line 262: should this be stored on store?

pygfx/materials/_points.py
```diff
@@ -238,5 +238,34 @@ class PointsGaussianBlobMaterial(PointsMaterial):
     """
 
 
+class PointsSpriteMaterial(PointsMaterial):
+    """A material to render points as sprite images.
+
+    Renders the provided texture at each point position. The images are square
+    and sized just like with a PointMaterial. The texture color is multiplierd
```
Korijn, line 245:
```suggestion
    and sized just like with a PointMaterial. The texture color is multiplied
```

pygfx/materials/_points.py
```diff
@@ -238,5 +238,34 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+    with the point's "normal" color (as calculated depending on ``color_mode``).
+
+    The sprite texture is provided via ``.map``.
```
Korijn, line 248: Isn't it `.sprite`?

pygfx/materials/_points.py
almarklein: Good point! I copied from `PointsMaterial.map`, but that one was faulty too :)

pygfx/materials/_points.py
almarklein: woops, yets  it is!

pygfx/materials/_points.py
Korijn: I think you missed `self._sprite` though?

almarklein: VSCode refused to save the file because of conflicts with the typo-change :P

Korijn (approved):

PR #697, opened 2024-03-28
Updated wx rotating cube example

cansik:
This PR updates the [cube_wx.py](https://github.com/pygfx/pygfx/blob/main/examples/other/cube_wx.py) example to be compatible with the most recent version of pygfx (world / local position, readonly vectors and so on). To be able to test it, I had to include the wgpu package, where the following PR has been applied: https://github.com/pygfx/wgpu-py/pull/486

Now everything works as expected (currently tested on MacOS):

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,6 @@ venv.bak/
 
 glTF-Sample-Models
 reactive_rendering_state.pkl
+
+# jetbrains idea
+.idea/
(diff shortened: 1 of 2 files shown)


almarklein:
Thanks! I think we need to hold this until your PR @ wgpu lands in a release. But this looks good!

almarklein:
This is ready to go!

Korijn (approved):

PR #700, opened 2024-04-01
Allow picking of background objects

hmaarrfk:
So maybe I'm hacking too much and feel free to reject this patch, but I'm overlaying one scene onto the main scene to create some kind of picture in picture preview.

It would be really useful to interact with the background of the scene since it indicates that the user has clicked on the picture in picture (but maybe not on an object).

I haven't really gone through the codepath to make picking background textures meaningful but this is already more than enough for me.

Otherwise, the pick just goes to the "object behind my picture in picture".

Thanks for considering


pygfx/renderers/wgpu/shaders/backgroundshader.py
```diff
@@ -170,6 +170,14 @@ def code_fragment(self):
             // written in the opaque pass in order for it to really be background.
             // So we fool the blender into thinking this fragment is opaque, even if its not.
             var out = get_fragment_output(varyings.position.z, vec4<f32>(out_color.rgb, 1.0));
+            $$ if write_pick
+            out.pick = (
+                // For a background, it doesn't make sense to report back
+                // anything other than the background id.
+                pick_pack(u32(u_wobject.id), 20)
+                // Maybe this can be different for for textured backgrounds?
```
Korijn, line 178: How about including the texcoords like in image shader?

https://github.com/pygfx/pygfx/blob/f163bc82a2d2b244d5aac4500143b7a0f76b34d3/pygfx/renderers/wgpu/shaders/imageshader.py#L258

hmaarrfk: I am happy to look into if you are OK with the general concept. 

The one feature I have the least experience with is the skybox background.

almarklein: +1 on the generic change to make backgrounds pickable.

The background can be a gradient, an image, or a cube/skybox.  Let's at least produce picking info that's consistent. Some options that come to mind:
* Only wobject id. Fine by me. We'll see if we need more later.
* Return 2D screen location, either in NDC or logical pixels.
* Return "texture coordinates": x y, z.

I think I like the latter most. Because a user can easily map the xy texture coordinates to NDC or logical pixels. And this way one can e.g. have a skybox and know what face the user clicked.

pygfx/renderers/wgpu/shaders/backgroundshader.py
```diff
@@ -171,12 +171,24 @@ def code_fragment(self):
...
-                // Maybe this can be different for for textured backgrounds?
-            );
+                $$ if texture_dim == 'cube'
+                    out.pick = (
+                        // For a background, it doesn't make sense to report back
+                        // anything other than the background id.
+                        pick_pack(u32(u_wobject.id), 20) +
+                        pick_pack(u32(varyings.texcoord.x * 16383.0), 14) +
+                        pick_pack(u32(varyings.texcoord.y * 16383.0), 14) +
+                        pick_pack(u32(varyings.texcoord.z * 16383.0), 14)
```
hmaarrfk, line 181: Hmmm... this doesn't seem to be working..

almarklein: The `z` is (semantically) an integer, I think.

pygfx/materials/_background.py
```diff
@@ -140,13 +140,37 @@ def map(self, map):
...
+            # Cube / Skybox
+            values = unpack_bitfield(pick_value, wobject_id=20, x=14, y=14, z=14)
+            texcoords_encoded = values["x"], values["y"], values["z"]
+            size = tex.size
+            x, y, z = [(v / 16383) * s - 0.5 for v, s in zip(texcoords_encoded, size)]
+            print(x, y, z)
+            ix, iy, iz = int(x + 0.5), int(y + 0.5), int(z + 0.5)
+            return {
+                "index": (ix, iy, iz),
+                "voxel_coord": (x - ix, y - iy, z - iz),
```
almarklein, line 167: Maybe just `coord`? If someone is only interested in the 2D location, write `info['coord'][:2]` and it works on all backgrounds.

hmaarrfk: ok all have been called coord now.

For the solid color backgrounds, they range from 0 to 1.

pygfx/renderers/wgpu/shaders/backgroundshader.py
hmaarrfk: hmmm, yeah that didn't help, with my debugging output  (enabled on this branch)

```
python examples/feature_demo/skybox.py
```

just prints clamped values most of the time. Sometimes..... it prints non-clamped values.

hmaarrfk:
ok so for the skybox, you basically sample with a direction vector, which isn't super intuitive and I'm not sure how you perfer to return this. The basics are in here. I need to review with a clear head to get something "right".

Some questions I still need to answer for myself:

- [ ] Always return a tuple of length 3?
- [ ] Is it ok to return a "direction vector" for the skybox? or do we want "u, v, face_index" or something of that sort.

almarklein:
Oh, I now realize I assumed that the `varyings.texcoord` always represented the full quad coords, and the `.z` field indicating the face of the cube, but this is not the case; it's the direction vector. Sorry, my earlier comments may have been confusing because of this.
 
I propose to keep it simple:
* In the vertex shader always do `varyings.texcoord = vec3<f32>(pos * 0.5 + 0.5, 0.0);`
* For cube texture use new `varyings.cubecoord`  (instead of using `textcoord` for it).
* Send this `texcoord` as picking info and expose it as a `coord` field in the pick info dict.
* Forget about cube-specific pick info for now.

hmaarrfk:
> I propose to keep it simple:
> * In the vertex shader always do varyings.texcoord = vec3<f32>(pos * 0.5 + 0.5, 0.0);
> * For cube texture use new varyings.cubecoord (instead of using textcoord for it).
> * Send this texcoord as picking info and expose it as a coord field in the pick info dict.
> * Forget about cube-specific pick info for now.

Are these mutually exclusive options? Sorry the choice of words makes it seem like they could be sequential.

As it stand "right now" (i might force push to cleanup the commit history), the normalized direction vector is being sent. After sleeping on it, it makes sense to me for a direction vector to get sent because in a "skybox" the most important thing might be "the direction where the user clicked".

The "6 sided cube trick" to me feels like a neat graphics trick, but I wonder if it helps or hinders the downstream application from being able to use the information effectively.

For example, I could see an application placing markers:
* Current position + skybox vector * 10 

away as the user clicks on the skybox.

It seems that the picking info is super material specific, so I'm not too worried about that.
I might suggest that the `BackgroundImage` material loose its ability to take in Cube matrial. When "inputs" change return values, to me that is where the API gets messy.

If:

* `BackgroundImageMaterial` throws an error on skybox textures, then it becomes a clear rule that the returned pick info is a 2D coordinate.
* `BackgroundSkyboxMaterial` always returns a 3D unit vector.

almarklein:
> Are these mutually exclusive options? Sorry the choice of words makes it seem like they could be sequential.

No I meant these as a package of choices to keep things simple. In other words, I said: let's forget about the skybox and always return a vector that can be used to see where on the screen the user clicked.

Now that I think about it, the position where the user clicked should (eventually) be part of the `info` always, the renderer can do that (once we figure out what to do with viewports). So maybe ... forget about *any* pick info for now! 🤷

hmaarrfk:
> So maybe ... forget about any pick info for now! 🤷


I'm down, i just want it to "respond" to "mouse move". Right now, moving the mouse over the Background just "ignores it"

hmaarrfk:
Keeping it so simple it seems my top level comment is still appropriate.

I added a comment pointing to this discussion. Not sure if you find that appropriate.

Thanks for your time again!

Korijn (approved):

Issue #702, opened 2024-04-02
Making objects "unpickable" when render mode is `ordered1`


hmaarrfk:
In our context, we might need to make certain "unpickable" so they don't "get in the way".

I noticed that this was maybe a concept:
https://github.com/pygfx/pygfx/blob/f163bc82a2d2b244d5aac4500143b7a0f76b34d3/pygfx/helpers/_gizmo.py#L262

However, when you set the render mode to `ordered1` it seems that this concept isn't respected.

To exaggerate this effect I am running the `image_click_event.py` demo with the following patch:

```patch
diff --git a/examples/feature_demo/image_click_events.py b/examples/feature_demo/image_click_events.py
index b8d691a..031edaa 100644
--- a/examples/feature_demo/image_click_events.py
+++ b/examples/feature_demo/image_click_events.py
@@ -14,7 +14,8 @@ import pygfx as gfx
 import numpy as np
 
 canvas = WgpuCanvas()
-renderer = gfx.renderers.WgpuRenderer(canvas)
+renderer = gfx.renderers.WgpuRenderer(canvas, blend_mode="ordered1")
+# renderer = gfx.renderers.WgpuRenderer(canvas)
 scene = gfx.Scene()
 
 # add image
@@ -44,6 +45,18 @@ points = gfx.Points(
 points.local.z = 1  # move points in front of the image
 scene.add(points)
 
+lines = gfx.Line(
+    gfx.Geometry(
+        positions=[(x, y, 0) for x, y in zip(xx, yy)],
+    ),
+    gfx.LineMaterial(
+        color=[0.3, 0.3, 0.3, 0.9],
+        thickness=30
+    ),
+)
+lines.local.z = 0.5
+scene.add(lines)
+
 
 previous_selection = None
 

```

When the mode is `ordered1` I can't click through the image. Any tips?

Korijn:
I think we should add a `pick_write` boolean flag on `Material`, just like we already have `depth_test` and are planning to add `depth_write` and `render_order` flags. Also see #492

hmaarrfk:
> I think we should add a pick_write boolean flag on Material

Understood.

> just like we already have depth_test and are planning to add depth_write and render_order flags.

When you do these kinds of changes, and if I were to make a PR to suggest the `pick_write` change, do you try to ensure you cover as much of your standard materials as possible? or is it ok for me to suggest the change and only implement it for Points and Lines renderers?

Korijn:
@almarklein what do you think?

[...]

PR #704, opened 2024-04-03
Making material pickability controllable

hmaarrfk:
My first observation are:

- My `pointer_move` event is no longer getting triggered for the one I attached to my scene.
- It still doesn't make the lines on the example code not block the pip. Maybe i missed something?

xref: https://github.com/pygfx/pygfx/issues/702

General todo list

- [x] Get pointer picking to work as expected on `examples/feature_demo/image_click_events.py`
- [x] Go through each of the demos to assess if they are broken or not.
- [x] Fix the broken demo files
- [x] Decide on appropriate default values for the pickability of different material.

### DIFF
--- a/docs/advanced_shaders.rst
+++ b/docs/advanced_shaders.rst
@@ -220,7 +220,18 @@ Picking
 -------
 
 The `output` struct of the fragment shader also has a ``pick`` field that can
-be set with picking info. It is an ``u64`` into which we can pack as many fields
+be set with pointer picking info. To enable picking for a material, use the
+``write_pick`` parameter at creation time.
+
+.. code-block:: python
+
+    cube = gfx.Mesh(
+        gfx.box_geometry(200, 200, 200),
+        gfx.MeshBasicMaterial(map=tex, opacity=0.8, pick_write=True),
+    )
+
+The picking info returnd can vary based on the shader. For all shaders,
+it is an ``u64`` into which we can pack as many fields
 as needed, using the ``pick_pack()`` function. The material needs to
 implement a corresponding ``_wgpu_get_pick_info()`` method
 to unpack the picking info. See e.g. the picking of a mesh:
--- a/docs/advanced_shaders.rst
+++ b/docs/advanced_shaders.rst
@@ -221,7 +221,7 @@ Picking
 
 The `output` struct of the fragment shader also has a ``pick`` field that can
 be set with pointer picking info. To enable picking for a material, use the
-``write_pick`` parameter at creation time.
+``pick_write`` parameter.
 
 .. code-block:: python
 
@@ -230,8 +230,8 @@ be set with pointer picking info. To enable picking for a material, use the
         gfx.MeshBasicMaterial(map=tex, opacity=0.8, pick_write=True),
     )
 
-The picking info returnd can vary based on the shader. For all shaders,
-it is an ``u64`` into which we can pack as many fields
+The picking info returned can vary based on the shader. For all shaders,
+it is a ``u64`` into which we can pack as many fields
 as needed, using the ``pick_pack()`` function. The material needs to
 implement a corresponding ``_wgpu_get_pick_info()`` method
 to unpack the picking info. See e.g. the picking of a mesh:
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -35,7 +35,13 @@ class Material(Trackable):
     )
 
     def __init__(
-        self, *, opacity=1, clipping_planes=None, clipping_mode="any", depth_test=True
+        self,
+        *,
+        opacity=1,
+        clipping_planes=None,
+        clipping_mode="any",
+        depth_test=True,
+        pick_write=False,
     ):
         super().__init__()
 
@@ -45,6 +51,7 @@ class Material(Trackable):
         self.clipping_planes = clipping_planes or []
         self.clipping_mode = clipping_mode
         self.depth_test = depth_test
+        self.pick_write = pick_write
 
     def _set_size_of_uniform_array(self, key, new_length):
         """Resize the given array field in the uniform struct if the
@@ -183,6 +190,17 @@ class Material(Trackable):
         if not isinstance(value, (bool, int)):
             raise TypeError("Material.depth_test must be bool.")
         self._store.depth_test = bool(value)
+
+    @property
+    def pick_write(self):
+        """Whether this material is picked by the pointer."""
+        return self._store.pick_write
+
+    @pick_write.setter
+    def pick_write(self, value):
+        if not isinstance(value, (bool, int)):
+            raise TypeError("Material.pick_write must be bool.")
+        self._store.pick_write = bool(value)
 
 
 class ColorMode(enum.Enum):
--- a/pygfx/renderers/wgpu/engine/blender.py
+++ b/pygfx/renderers/wgpu/engine/blender.py
@@ -32,7 +32,7 @@ class BasePass:
     render_mask = 3  # opaque end transparent
     write_pick = True
 
-    def get_color_descriptors(self, blender):
+    def get_color_descriptors(self, blender, material_write_pick):
         """Get the list of fragment targets for device.create_render_pipeline()."""
         # The result affects the wobject's pipeline.
         # Returning [] prevents this pipeline from being created.
@@ -77,7 +77,7 @@ class OpaquePass(BasePass):
     render_mask = 1  # opaque only
     write_pick = True
 
-    def get_color_descriptors(self, blender):
+    def get_color_descriptors(self, blender, material_write_pick):
         bf, bo = wgpu.BlendFactor, wgpu.BlendOperation
         return [
             {
@@ -91,7 +91,7 @@ class OpaquePass(BasePass):
             {
                 "format": blender.pick_format,
                 "blend": None,
-                "write_mask": wgpu.ColorWrite.ALL,
+                "write_mask": wgpu.ColorWrite.ALL if material_write_pick else 0,
             },
         ]
 
@@ -181,7 +181,7 @@ class SimpleSinglePass(OpaquePass):
     render_mask = 3  # opaque end transparent
     write_pick = True
 
-    def get_color_descriptors(self, blender):
+    def get_color_descriptors(self, blender, material_write_pick):
         bf, bo = wgpu.BlendFactor, wgpu.BlendOperation
         return [
             {
@@ -195,7 +195,7 @@ class SimpleSinglePass(OpaquePass):
             {
                 "format": blender.pick_format,
                 "blend": None,
-                "write_mask": wgpu.ColorWrite.ALL,
+                "write_mask": wgpu.ColorWrite.ALL if material_write_pick else 0,
             },
         ]
 
@@ -223,7 +223,7 @@ class SimpleTransparencyPass(BasePass):
     render_mask = 2  # transparent only
     write_pick = False
 
-    def get_color_descriptors(self, blender):
+    def get_color_descriptors(self, blender, material_write_pick):
         bf, bo = wgpu.BlendFactor, wgpu.BlendOperation
         return [
             {
@@ -312,7 +312,7 @@ class WeightedTransparencyPass(BasePass):
 
         self._weight_code = weight_code.strip()
 
-    def get_color_descriptors(self, blender):
+    def get_color_descriptors(self, blender, material_write_pick):
         bf, bo = wgpu.BlendFactor, wgpu.BlendOperation
         return [
             {
@@ -405,7 +405,7 @@ class FrontmostTransparencyPass(BasePass):
 
     write_pick = True
 
-    def get_color_descriptors(self, blender):
+    def get_color_descriptors(self, blender, material_write_pick):
         bf, bo = wgpu.BlendFactor, wgpu.BlendOperation
         return [
             {
@@ -419,7 +419,7 @@ class FrontmostTransparencyPass(BasePass):
             {
                 "format": blender.pick_format,
                 "blend": None,
-                "write_mask": wgpu.ColorWrite.ALL,
+                "write_mask": wgpu.ColorWrite.ALL if material_write_pick else 0,
             },
         ]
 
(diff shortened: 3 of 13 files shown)


hmaarrfk:
In building https://github.com/pygfx/pygfx/pull/700 i learned that a "pick" is generated everytime you move your mouse, so for the very least, the background of the scene should be made pickable to get the mouse movement events.

pygfx/materials/_base.py
```diff
@@ -184,6 +191,18 @@ def depth_test(self, value):
             raise TypeError("Material.depth_test must be bool.")
         self._store.depth_test = bool(value)
 
+    @property
+    def write_pick(self):
+        """Whether this material is picked by the pointer."""
+        return self._store.write_pick
```
almarklein, line 197: Reading it like this I think perhaps this prop should be named `pickable` or something similar. The fact that *something* has to be written to make it so, is a fact that we don't need to concern the user with. @Korijn what do you think?

almarklein:
I think I know what's missing. In `blender.py`, the `get_color_descriptors()` will produce a descriptor for the picking, but this should happen conditionally now. I think we can add a `material_write_pick` arg to  `get_color_descriptors()`.

pygfx/materials/_base.py
Korijn: As I proposed earlier I prefer the property to be named `pick_write` so that it will be consistent with other similar planned properties `depth_write` and `depth_test`. It's also more discoverable (with things like search and autocomplete)

hmaarrfk: Ok. I noticed that difference between your two comments. I meant to bring it up earlier. Sorry for not flagging.
I’ll make the change to the property name.

hmaarrfk:
I think I tried something like your suggestion in https://github.com/pygfx/pygfx/pull/704/commits/65cb368feeb1da08474bc299a5442214168813ce

but it seems that have crashed quite badly on me. 

I'll have to sleep on this. maybe read more about GPUs

almarklein:
> I think I tried something like your suggestion in https://github.com/pygfx/pygfx/commit/65cb368feeb1da08474bc299a5442214168813ce

Not exactly what I meant. Sorry for being not sufficiently clear. I meant that in the implementation of the `get_color_descriptors()` methods that produce a color descriptor for picking, this should be optional. In particular, these cases:

https://github.com/pygfx/pygfx/blob/0189f812673ea4aea044500681eb6263e1b7acce/pygfx/renderers/wgpu/engine/blender.py#L80-L96

https://github.com/pygfx/pygfx/blob/0189f812673ea4aea044500681eb6263e1b7acce/pygfx/renderers/wgpu/engine/blender.py#L184-L200

https://github.com/pygfx/pygfx/blob/0189f812673ea4aea044500681eb6263e1b7acce/pygfx/renderers/wgpu/engine/blender.py#L408-L424

And these methods of the passes are all called via this method of the blender object:

https://github.com/pygfx/pygfx/blob/0189f812673ea4aea044500681eb6263e1b7acce/pygfx/renderers/wgpu/engine/blender.py#L572-L573

almarklein:
> I'll have to sleep on this. maybe read more about GPUs

Not so much related to GPU's. The `pipeline.py` is probably the trickiest code in pygfx to follow, because it's where a lot of things come together.

hmaarrfk:
<details><summary>my notes</summary>

- validate_helpers2.py
  - PASS: Used the mouse to move the axes.
  - PASS: This continue to work even after write_pick is false
- validate_line_2d.py
  - PASS: can still control rotating the camera
- validate_line_dash.py
  - PASS: can still control camera
- validate_line_thickness_space.py
  - PASS: can still control camera
- validate_meshslice.py
  - FAIL: hover over mesh no longer produced text on the screen.
  - FIX: Needed to make the mesh pickable
- validate_skybox.py
  - FAIL: Skybox disappeared????? But the reflection in the ball is still good???
  - FIX: Neded to make the skybox pickable???? what??? why....
  - REASON: It seems that some feature of the skybox shader is using write_pick as a detection of the opaque pass.

- collection_line.py
  - PASS: can zoom in with mouse

- cube_wx.py
  - UNTESTED

- light_view.py
  - UNRELATED FAIL: AttributeError: module 'pylinalg' has no attribute 'Quaternion'
  - PASS: Camera control

- reactive_rendering.py:
  - UNREALTED: ModuleNotFoundError: No module named 'observ'
  - MUST CHECK: event_handlers added

- examples/other/sponza_scene.py
  - UNRELATED FAIL: RuntimeError: Could not find 'glTF-Sample-Models' directory.

- clipping_planes.py
  - PASS: can rotate things

- cube.py
  - FAIL: background disappeared
  - FIX: Neded to make the background pickable???? what??? why....
  - REASON: It seems that some feature of the background shader is using write_pick as a detection of the opaque pass.

- light_basic.py
  - seems like i can move around. seems like the light is working

- object_bounding_box.py
  - seems like the bounding box shows up

- orbit_camera.py
  - FAIL: background doesn't show up
  - FIX: Same as other background stuffs

- points_basic.py
  - FAIL: background not rendered
  - FIX: Background(pick_write=True)


## Feature

- aomap.py
- colormap_channels.py
- colormap_image.py
- colormap_mesh.py
- custom_object1.py
- custom_object2.py
- custom_object3.py
- cylinder.py
- dynamic_env_map.py -- Background pick_write bug
- env_maps.py -- Background pick_write bug
- flat_shaded_torus.py
- fly_controller.py
- geometry_cubes.py
- geometry_image.py
- geometry_klein_bottle.py
- geometry_mobius.py
- geometry_plane.py
- geometry_polyhedron.py
- geometry_polyhedron_subdivisions.py
- geometry_torus_knot.py
- helpers_gizmo.py -- fixed
- helpers_stats.py
- image_click_events.py -- You can now click the points!!! and it clicks through them!!!
- image_overlay.py
- image_plus_points.py
- instancing_mesh.py
- light_directional_shadow.py
- lightmap.py
- light_shadow.py
- light_spotlight_shadows.py
- light_spotlights.py
- line_basic.py
- line_performance.py
- line_segments.py
- line_thick.py
- look_at.py
- manual_matrix_update.py
- map_click_events_to_world.py -- looks black... add pick_write=True... -- background hack
- mesh_depth_material.py
- mesh_dynamic.py
- mesh_slice.py
- multiprocessing_zmq
- multi_select.py -- Had to add picking to cubes
- multi_slice1.py -- background hack
- multi_slice2.py -- background hack
- paint_to_texture.py -- background hack, image must be write_pick=True
- panzoom_camera.py -- background hack
- pbr2.py -- background skybox hack
- pbr.py -- background skybox hack
- physical_color.py
- picking_color.py -- background hack -- cube write pick
- picking_mesh.py -- background hack -- write pick on objects
- picking_points.py -- write_pick enabled
- quad_mesh.py -- background hack -- write pick on objects
- README.rst
- scene_in_a_scene.py -- background hack
- scene_overlay.py
- scene_subplots1.py
- scene_subplots2.py -- background hack

- scene_subplots_video.py
- show_image.py
- show_scene.py
- show_stl.py
- skybox.py -- background hack
- skybox_rotate.py -- background hack
- spheres.py
- synced_video.py

- text_contrast.py -- background hack
- text_snake.py -- and object pick write

- text_waterfall.py -- background hack
- transparency1.py
- transparency2.py
- volume_render1.py -- object pick write
- volume_render2.py
- volume_slice1.py
- volume_slice2.py
- volume_slice3.py -- pick_write object
- wireframe1.py
- wireframe2.py
- wireframe_material.py
- world_bounding_box.py

</details>

pygfx/renderers/wgpu/engine/pipeline.py
```diff
@@ -679,7 +683,7 @@ def _check_render_info(self):
         assert all(isinstance(i, int) for i in indices)
         assert len(indices) == 3
 
-    def _compile_shader(self, pass_index, env):
+    def _compile_shader(self, pass_index, env, material=None):
         """Compile the templateds wgsl shader to a wgpu shader module."""
         return get_cached_shader_module(self.device, self.shader, {})
```
hmaarrfk, line 688: now that the material can control the pickability, shouldn't this affect the result of the cache too?

hmaarrfk: hmm this is the compute shader, so i guess it isn't so important.

pygfx/renderers/wgpu/engine/blender.py
```diff
@@ -91,7 +91,7 @@ def get_color_descriptors(self, blender):
             {
                 "format": blender.pick_format,
                 "blend": None,
-                "write_mask": wgpu.ColorWrite.ALL,
+                "write_mask": wgpu.ColorWrite.ALL if material_write_pick else 0,
             },
```
almarklein, line 95: Another option is to drop this whole dict when `material_write_pick` is False. I'm not sure to what extend this is different, but it would probably require applying the same trick to `get_color_attachments()`.

So maybe first see if it works like this (with just setting `write_mask`).

hmaarrfk: I believe I tried this. I got some really strange errors. Looking at it again, I think you are correct that the `get_color_attachments` change is likely also necessary.

<details>

```
$ python examples/feature_demo/image_click_events.py
Detected skylake derivative running on mesa i915. Clears to srgb textures will use manual shader clears.
The fragment stage "fs_main" output @location(1) values are ignored
Draw error
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/wgpu/gui/_gui_utils.py", line 77, in log_exception
    yield
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/wgpu/gui/base.py", line 135, in _draw_frame_and_present
    self.draw_frame()
  File "/home/mark/git/pygfx/examples/feature_demo/image_click_events.py", line 86, in <lambda>
    canvas.request_draw(lambda: renderer.render(scene, camera))
  File "/home/mark/git/pygfx/pygfx/renderers/wgpu/engine/renderer.py", line 527, in render
    command_buffers += self._render_recording(
  File "/home/mark/git/pygfx/pygfx/renderers/wgpu/engine/renderer.py", line 657, in _render_recording
    render_pass.end()
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_api.py", line 2641, in end
    libf.wgpuRenderPassEncoderEnd(self._internal)
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_helpers.py", line 322, in proxy_func
    raise wgpu_error  # the frame above is more interesting ↑↑

Caused by:
    In wgpuRenderPassEncoderEnd
      note: encoder = `<CommandBuffer-(0, 1, Vulkan)>`
    In a set_pipeline command
      note: render pipeline = `<RenderPipeline-(1, 1, Vulkan)>`
    Render pipeline targets are incompatible with render pass
    Incompatible color attachments at indices []: the RenderPass uses textures with formats [Some(Rgba8UnormSrgb), Some(Rgba16Uint)] but the RenderPipeline uses attachments with formats [Some(Rgba8UnormSrgb)]
Draw error: Validation Error (2)
Draw error: Validation Error (3)
Draw error: Validation Error (4)
```
</details>

`get_color_attachments` is only used in 1 place where a whole `wobject_list` is passed and not a single object, so I'm worried that changing the attachments isn't really straightforward.
https://github.com/pygfx/pygfx/blob/main/pygfx/renderers/wgpu/engine/renderer.py#L639

almarklein: Oh right ... the color attachment is per-pipeline, the descriptor per object. Then the current solution should be the correct one 👍

pygfx/renderers/wgpu/engine/pipeline.py
```diff
@@ -771,6 +776,10 @@ def _compile_shader(self, pass_index, env):
         env_kwargs = env.get_shader_kwargs(env_bind_group_index)
         shader_kwargs = blender_kwargs.copy()
         shader_kwargs.update(env_kwargs)
+        if material is not None:
+            shader_kwargs["write_pick"] = (
+                shader_kwargs["write_pick"] and material.pick_write
+            )
```
Korijn, line 782: I think this is a rare opportunity to use the `&=` operator:

```py
>>> write_pick = True
>>> write_pick &= True
>>> write_pick
True
>>> write_pick &= False
>>> write_pick
False
```

hmaarrfk: > rare opportunity

so rare i dont even think about it ^_^

hmaarrfk:
Ok i added a note in the docs about the change in the default value. 

Hopefully this is good to go.

docs/advanced_shaders.rst
```diff
@@ -220,7 +220,18 @@ Picking
 -------
 
 The `output` struct of the fragment shader also has a ``pick`` field that can
-be set with picking info. It is an ``u64`` into which we can pack as many fields
+be set with pointer picking info. To enable picking for a material, use the
+``write_pick`` parameter at creation time.
```
Korijn, line 224:
```suggestion
``pick_write`` parameter at creation time.
```

docs/advanced_shaders.rst
```diff
@@ -220,7 +220,18 @@ Picking
...
+
+.. code-block:: python
+
+    cube = gfx.Mesh(
+        gfx.box_geometry(200, 200, 200),
+        gfx.MeshBasicMaterial(map=tex, opacity=0.8, pick_write=True),
+    )
+
+The picking info returnd can vary based on the shader. For all shaders,
```
Korijn, line 233:
```suggestion
The picking info returned can vary based on the shader. For all shaders,
```

docs/advanced_shaders.rst
```diff
@@ -220,7 +220,18 @@ Picking
...
+it is an ``u64`` into which we can pack as many fields
```
Korijn, line 234:
```suggestion
it is a ``u64`` into which we can pack as many fields
```

Korijn:
Have you confirmed toggling `pick_write` at runtime works as expected?

hmaarrfk:
> Have you confirmed toggling pick_write at runtime works as expected?

No.

almarklein (approved):

Korijn (approved):

Korijn:
> > Have you confirmed toggling pick_write at runtime works as expected?
> 
> No.
> 
> I need to think of a "demo" test for this.

We'll merge, would be great if you could still follow up on that check.

Thanks a ton!

almarklein:
> Have you confirmed toggling pick_write at runtime works as expected?

I just checked, and it does not 😛 Created #720 to track. I can have a look.

PR #705, opened 2024-04-03
Supports InstancedMesh Shadows

panxinmiao:
Resolved a legacy issue. Added shadow support for InstancedMesh.


examples/feature_demo/look_at_instanced.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    for i in range(100):
+        m = mesh.get_matrix_at(i)
+        dummy.world.matrix = m
+        dummy.look_at(sphere.world.position)
+        m = dummy.world.matrix
+        mesh.set_matrix_at(i, m)
+
+    mesh.instance_buffer.update_range(0, 100)  # update instances matrix buffer manually
```
panxinmiao, line 78: The 'set_matrix_at' method seems to be unable to automatically update the "GPU buffer". It needs to be manually updated here. @almarklein ,Is there a better approach?

panxinmiao: > I think `set_matrix_at` just needs an additional call at the end:

The original `pygfx` seems to have done it this way before Almar refactored the architecture and utilized the `trackable` mechanism.

almarklein: Probably not, since it's only used internally.

Korijn (approved):
Awesome work as always. 💝

PR #707, opened 2024-04-03
Improve trimesh conversion

schlegelp:
The purpose of this PR is to make the conversion from trimesh objects into pygfx Meshes more robust:
1. Trimesh scenes where geometries are used by multiple nodes are now correctly converted 
2. `material_from_trimesh` now also converts trimesh's `SimpleMaterials` 
3. `pygfx.load_meshes` now also works for remote sources (simply wrapping `trimesh.load_remote`)
4. Fixed a couple corner cases where the conversion would break before 

With these changes you can now do this:

```python
>>> import pygfx as gfx
>>> meshes = gfx.load_meshes("https://github.com/mikedh/trimesh/raw/main/models/CesiumMilkTruck.glb")
>>> scene = gfx.Scene()
>>> scene.add(*meshes)
>>> scene.add(*[gfx.AmbientLight(), gfx.PointLight()])
>>> gfx.show(scene)
```
Screenshot 2024-04-03 at 16 45 03

I have tried this with a couple of the [trimesh models](https://github.com/mikedh/trimesh/tree/main/models) and it seems to work across the board.

Two questions from my side:
1. Would you like to add tests for this?
2. Is there appetite for functionality to load the entire trimesh scene, i.e. including lights, camera, background, etc?

### DIFF
--- a/examples/feature_demo/lightmap.py
+++ b/examples/feature_demo/lightmap.py
@@ -40,7 +40,7 @@ import pygfx as gfx
 canvas = WgpuCanvas(size=(1200, 400), title="lightmap")
 renderer = gfx.renderers.WgpuRenderer(canvas)
 
-meshes = gfx.load_meshes(model_dir / "lightmap" / "scene.gltf")
+meshes = gfx.load_mesh(model_dir / "lightmap" / "scene.gltf")
 
 light_map = iio.imread(model_dir / "lightmap" / "lightmap-ao-shadow.png")
 light_map_tex = gfx.Texture(light_map, dim=2)
--- a/pygfx/utils/load.py
+++ b/pygfx/utils/load.py
@@ -11,12 +11,14 @@ from importlib.util import find_spec
 def load_meshes(path, remote_ok=False):
     """Load meshes from a file.
 
+    Deprecated: use load_mesh() instead.
+
     Parameters
     ----------
     path : str
         The location where the mesh is stored.
     remote_ok : bool
-        Whether to allow loading meshes from URLs, by default False.
+        Whether to allow loading meshes from URLs. Default is False.
 
     Returns
     -------
@@ -38,7 +40,7 @@ def load_mesh(path, remote_ok=False):
         The location where the mesh or scene is stored.
         Can be a local file path or a URL.
     remote_ok : bool
-        Whether to allow loading meshes from URLs, by default False.
+        Whether to allow loading meshes from URLs. Default is False.
 
     Returns
     -------
@@ -102,13 +104,13 @@ def load_scene(
         The filepath. Can be a local file or a URL.
     flatten : bool
         If True, will ignore any hierarchical structure in the scene graph and
-        instead import as a flat list of objects.
+        instead import as a flat list of objects. Default is False.
     meshes : bool
-        Whether to load meshes.
+        Whether to load meshes. Default is True.
     materials : bool
-        Whether to load materials.
+        Whether to load materials. Default is True.
     volumes : bool
-        Whether to load 3D image volumes.
+        Whether to load 3D image volumes. Default is True.
     lights :  "auto" | "file" | "none"
         Whether to import lights. "auto" (default) will always add lights to the scene
         even if the file does not contain any; "file" will import only lights defined in
@@ -118,7 +120,7 @@ def load_scene(
         even if the file does not contain any; "file" will import only cameras defined in
         the file; "none" will not add any cameras to the scene.
     remote_ok : bool
-        Whether to allow loading files from URLs, by default False.
+        Whether to allow loading files from URLs. Default is False.
 
     Returns
     -------
@@ -180,9 +182,9 @@ def meshes_from_trimesh(scene, materials=True, apply_transforms=True):
     scene : trimesh.Scene
         The scene to convert.
     materials : bool
-        Whether to import materials. If False, a default material will be created.
+        Whether to import materials. If False, a standard material will be created. Default is True.
     apply_transforms : bool
-        Whether to apply the scene graph transforms directly to the meshes, by default True.
+        Whether to apply the scene graph transforms directly to the meshes. Default is True.
 
     Returns
     -------
@@ -294,13 +296,13 @@ def scene_from_trimesh(
         A trimesh Scene.
     flatten : bool
         If True, will ignore any hierarchical structure in the scene graph and
-        instead import as a flat list of objects.
+        instead import as a flat list of objects. Default is False.
     meshes : bool
-        Whether to import meshes.
+        Whether to import meshes. Default is True.
     materials : bool
-        Whether to import materials.
+        Whether to import materials. Default is True.
     volumes : bool
-        Whether to load 3D image volumes.
+        Whether to load 3D image volumes. Default is True.
     lights :  "auto" | "file" | "none"
         Whether to import lights. "auto" (default) will always add lights to the scene
         even if the file does not contain any; "file" will import only lights defined in
@@ -311,7 +313,7 @@ def scene_from_trimesh(
         the file; "none" will not add any cameras to the scene.
     background : bool
         Trimesh scenes typically have a white background. If True, we will add a white background
-        to the pygfx scene. If False, no background will be added.
+        to the pygfx scene. If False, no background will be added. Default is True.
 
     Returns
     -------
@@ -507,10 +509,6 @@ def _volume_from_voxelgrid(vxl, cmap=None, clim="data"):
     ----------
     vxl : trimesh.voxels.VoxelGrid
         The voxel grid to convert.
-    camera : bool
-        Whether to add camera to the scene.
-    lights : "bool
-        Whether to add lights to the scene.
     cmap : pygfx.Texture, optional
         The colormap to use. If None, this will default to `pygfx.cm.viridis`.
     clim : "data" | "dtype" | tuple | None
--- a/pygfx/utils/load.py
+++ b/pygfx/utils/load.py
@@ -3,6 +3,7 @@ Utilities to load scenes from files, using trimesh.
 """
 
 import pygfx as gfx
+import numpy as np
 
 
 def load_meshes(path, remote_ok=False):
@@ -70,6 +71,7 @@ def load_scene(
     flatten=False,
     meshes=True,
     materials=True,
+    volumes=True,
     lights="auto",
     camera="auto",
     remote_ok=False,
@@ -77,8 +79,9 @@ def load_scene(
     """Load file into a scene.
 
     When reading file formats that can contain more than just meshes (e.g. `.glb`) this function will
-    attempt to import the entire scene, including lights, cameras, and materials. If the file format
-    is mesh-only (e.g. `.stl`) we will instead construct a minimal scene.
+    attempt to import the entire scene, including lights, cameras, volumes and materials. If the file
+    format is mesh-only (e.g. `.stl`) or volume-only (e.g. `.binvox`) we will instead construct a
+    minimal scene.
 
     This function requires the trimesh library.
 
@@ -93,6 +96,8 @@ def load_scene(
         Whether to load meshes.
     materials : bool
         Whether to load materials.
+    volumes : bool
+        Whether to load 3D image volumes.
     lights :  "auto" | "file" | "none"
         Whether to import lights. "auto" (default) will always add lights to the scene
         even if the file does not contain any; "file" will import only lights defined in
@@ -139,6 +144,7 @@ def load_scene(
         tm_scene,
         flatten=flatten,
         meshes=meshes,
+        volumes=volumes,
         materials=materials,
         lights=lights,
         camera=camera,
@@ -254,6 +260,7 @@ def scene_from_trimesh(
     flatten=False,
     meshes=True,
     materials=True,
+    volumes=True,
     lights="auto",
     camera="auto",
     background=True,
@@ -271,6 +278,8 @@ def scene_from_trimesh(
         Whether to import meshes.
     materials : bool
         Whether to import materials.
+    volumes : bool
+        Whether to load 3D image volumes.
     lights :  "auto" | "file" | "none"
         Whether to import lights. "auto" (default) will always add lights to the scene
         even if the file does not contain any; "file" will import only lights defined in
@@ -297,24 +306,34 @@ def scene_from_trimesh(
 
     import trimesh  # noqa
 
+    # Basic scene setup
+    gfx_scene = gfx.Scene()
+
     # Convet single meshes into a scene (this makes the code below much simpler)
     if isinstance(tm_scene, trimesh.Trimesh):
         tm_scene = trimesh.Scene(geometry=tm_scene)
+    elif isinstance(tm_scene, trimesh.voxel.VoxelGrid):
+        # VoxelGrids can not be in a trimesh.Scene directly (trimesh renders them
+        # as collection of cube meshes), so we need to handle them separately
+        if volumes:
+            vol = _volume_from_voxelgrid(tm_scene)
+            gfx_scene.add(vol)
+
+        # Generate an empty scene so we can continue with the rest of the function
+        # (i.e. potentially add lights and camera)
+        tm_scene = trimesh.Scene()
     elif not isinstance(tm_scene, trimesh.Scene):
         raise ValueError(f"Unexpected trimesh data: {type(tm_scene)}")
 
-    # Basic scene setup
-    gfx_scene = gfx.Scene()
+    # Use the graph representation of the scene - easier to handle
+    G = tm_scene.graph.to_networkx()
 
     # Take care of meshes
     # Note: we're traversing the scene graph only for meshes (i.e. not lights, cameras, etc.)
     # That's because as far as I can tell, trimesh's scene graph only applies to geometries
     # while lights and cameras are stored separate from 'world'.
-    if meshes:
+    if meshes and len(G):
         if not flatten:
-            # Use the graph representation of the scene - easier to handle
-            G = tm_scene.graph.to_networkx()
-
             # Load the geometries and materials
             gfx_geometries, gfx_materials = objects_from_trimesh(
                 tm_scene, materials=materials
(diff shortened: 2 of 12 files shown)


pygfx/utils/load.py
```diff
@@ -41,36 +41,60 @@ def load_meshes(path):
...
     meshes : list
         A list of loaded meshes.
 
     """
-
     import trimesh  # noqa
 
-    scene = trimesh.load(path)
+    # Trimesh's load() performs a similar check and refers
+    # loading from URLs to load_remote()
```
Korijn, line 55: If that's the case why can't we just use `trimesh.load`?

schlegelp: When `trimesh.load` detects a URL it raises a `ValueError` telling us to use `trimesh.load_remote`:

```python
>>> scene = tm.load("https://github.com/mikedh/trimesh/raw/main/models/cube_blender_uv.ply")
...
ValueError: use load_remote to load URL: https://github.com/mikedh/trimesh/raw/main/models/cube_blender_uv.ply
```

According to the docstring of `tm.load_remote` the logic is to avoid accidental network requests:

```
Signature: tm.load_remote(url, **kwargs)
Docstring:
Load a mesh at a remote URL into a local trimesh object.

This must be called explicitly rather than automatically
from trimesh.load to ensure users don't accidentally make
network requests.
```

I thought the convenience of having one function do both was preferable but I see their point. Any preference?

Korijn: How about you add a keyword argument `network_ok=False` to force users to opt-in to the magic behaviour :)

pygfx/utils/load.py
```diff
@@ -41,36 +41,60 @@ def load_meshes(path):
...
-            )
-            for m in scene.geometry.values()
-        ]
+
+            # Convert each geometry only once
+            if geometry_name not in gfx_geometries:
+                m = scene.geometry[geometry_name]
+                gfx_geometries[geometry_name] = gfx.geometry_from_trimesh(m)
+                if hasattr(m.visual, "material"):
+                    gfx_materials[geometry_name] = gfx.material_from_trimesh(
```
Korijn, line 87: should materials also be deduplicated? they can be shared in pygfx, at least

pygfx/utils/load.py
```diff
@@ -41,36 +41,60 @@ def load_meshes(path):
...
+                    )
+                else:
+                    gfx_materials[geometry_name] = gfx.MeshStandardMaterial()
+
+            # Generate a visual for each node
+            mesh = gfx.Mesh(gfx_geometries[geometry_name], gfx_materials[geometry_name])
+            mesh.local.matrix = transform  # set the node's transform
+
+            meshes.append(mesh)
+        return meshes
```
Korijn, line 98: I see we are generating a flat list of meshes, but you mention a scene graph concept exists in trimesh, just like it does in pygfx. You _can_ replicate the actual tree in trimesh if you want to.

Korijn:
First of all: awesome!! 🎉 Thank you so much for this wonderful contribution.

> Two questions from my side:
> 
> 1. Would you like to add tests for this?

If you feel like it that would be wonderful!

> 2. Is there appetite for functionality to load the entire trimesh scene, i.e. including lights, camera, background, etc?

Well, the more trimesh functionality we can leverage, the better! It would be super useful for all our users.

pygfx/utils/load.py
schlegelp: True! I suppose that's what I was going for with this question:

> Is there appetite for functionality to load the entire trimesh scene, i.e. including lights, camera, background, etc?

I'll have a look into it but my thinking was that a `load_meshes` is expected to return meshes and nothing else. 

I could implement this in a separate function or via an additional argument (perhaps something like `return_scene=True`). Any preference?

pygfx/utils/load.py
schlegelp: Added via ed99a85

pygfx/utils/load.py
Korijn: well. load_mesh was originally intended to load a single mesh. we already deviated from that quite a bit. personally I would expect a load function per filetype, but I guess in this case filetypes don't necessarily map to object types 1:1?

schlegelp:
No problem. I'll have a crack at that. Somewhat awkwardly there already is a `load_scene()` that has been deprecated:

```python
def load_scene(path):
    raise DeprecationWarning(
        "The load_scene() function is replaced with load_meshes()."
    )
```

I guess I'll be reviving it?

Korijn:
> No problem. I'll have a crack at that. Somewhat awkwardly there already is a `load_scene()` that has been deprecated:
> 
> ```python
> def load_scene(path):
>     raise DeprecationWarning(
>         "The load_scene() function is replaced with load_meshes()."
>     )
> ```
> 
> I guess I'll be reviving it?

That's fine! Here's my two cents on the API design:

Personally, I would expect a gfx library to offer functions for three core features:

* Load a texture/material (e.g. png or jpg files), `load_texture`
* Load a mesh (e.g. obj, ply stl files), `load_mesh`
* Load a scene (e.g. gltf files), `load_scene`

To me `load_meshes` is a little odd/unnecessary. Our users could assume its a special case of `load_mesh` which you could run in a loop (wrong, that's not what our implementation does -> surprising and confusing). In actuality it's a special case of `load_scene` where you are only getting a (flattened) portion of its output.

I would expect `load_scene` by default to return a tree structure scene graph with all of the elements it can load from the file.

If anything, we can add keyword arguments to control its behaviour. For example:

* `flatten=False` - return a flattened list instead of a scene graph
* `meshes=True` - load meshes
* `materials=True` - load materials
* `lights=True` - load lights

...and so on for other types of objects

schlegelp:
I had a crack at the proposed API design. Let me know what you think.

pygfx/utils/load.py
```diff
@@ -5,72 +5,432 @@
...
+        instead import as a flat list of objects.
+    meshes : bool
+        Whether to load meshes.
+    materials : bool
+        Whether to load materials.
+    lights : bool, optional
+        Whether to load lights:
+          - True (default): lights are imported if present but if not, minimal lights are added
+          - False: ignore any lights in the file but add minimal lights to scene
+          - None: no lights are loaded
```
almarklein, line 100: If find the "enum"  a bit awkward. Users that do not load the lights, can easily add a minimal light with a on-liner, I think?

examples/feature_demo/lightmap.py
```diff
@@ -40,7 +40,7 @@
 canvas = WgpuCanvas(size=(1200, 400), title="lightmap")
 renderer = gfx.renderers.WgpuRenderer(canvas)
 
-meshes = gfx.load_meshes(model_dir / "lightmap" / "scene.gltf")
+meshes = gfx.load_mesh(model_dir / "lightmap" / "scene.gltf")
```
almarklein, line 43: 🤔 I'm a bit confused. The method is singular but it returns a list. Would it not be better to deprecate `load_mesh()` instead?

@Korijn you made a comment related to this earlier. Do you mean we should deprecate/remove one of these, or specifically `load_meshes()`?

Also I found out (after writing the above) that there's `trimesh.load_mesh()`. If we want to align our API a bit with trimesh here, I think that's a valid argument.

pygfx/utils/load.py
```diff
@@ -5,72 +5,432 @@
...
+
+    """
+    import trimesh  # noqa
+
+    # If this is a single object we can just convert and return it
     if isinstance(scene, trimesh.Trimesh):
         m = gfx.Mesh(
             gfx.geometry_from_trimesh(scene),
-            gfx.MeshPhongMaterial(),
+            gfx.material_from_trimesh(scene.visual.material),
```
almarklein, line 168: This triggers an exception for the docs CI (which use trimesh 4.2.4):
```
../examples/feature_demo/show_stl.py unexpectedly failed to execute correctly:

      File "/home/docs/checkouts/readthedocs.org/user_builds/pygfx/checkouts/707/examples/feature_demo/show_stl.py", line 36, in <module>
        meshes = gfx.load_mesh(model_dir / "teapot.stl")
      File "/home/docs/checkouts/readthedocs.org/user_builds/pygfx/envs/707/lib/python3.9/site-packages/pygfx/utils/load.py", line 65, in load_mesh
        return meshes_from_trimesh(scene, apply_transforms=True)
      File "/home/docs/checkouts/readthedocs.org/user_builds/pygfx/envs/707/lib/python3.9/site-packages/pygfx/utils/load.py", line 168, in meshes_from_trimesh
        gfx.material_from_trimesh(scene.visual.material),
    AttributeError: 'ColorVisuals' object has no attribute 'material'
```

almarklein (commented):
Nice work!

pygfx/utils/load.py
schlegelp: Yeah, maybe it's a bit over-engineered. These are the use cases I was thinking of:

1. Users that just want a scene that works and looks good, no matter whether it all came from the file itself (`True`).
2. Users that want _exactly_ what's in the file - nothing more, nothing less (currently not covered).
3. Users that do not want to load any lights - e.g. perhaps because the lights in the file are broken or because they want to use their own lights (`None`).
4. Users that want none of the lights in the file but would like _some_ lights so the scene is ready to go (`False`).

I suspect the order of frequency would be 1 >> 2 > 3 > 4. The current behaviour is not actually capturing that. So perhaps the arguments should change to:
- `True` = 1
- `None` = 2
- `False` = 3

That said: I'm also perfectly happy to keep it simple with a just `True` and `False` for `1` and `3`, respectively.

examples/feature_demo/lightmap.py
Korijn: I would prefer `load_mesh` and I would prefer it to return a single object. However it seems there are various file formats that return multiple objects. I think it's preferred that return types don't vary based on what's in the file, so always returning a list is an okay compromise to me. So:

`list_of_meshes = load_mesh(filename)`

pygfx/utils/load.py
schlegelp: Whoops. It turns out `ColorVisuals` are used for vertex and face colors but they can also be empty. These visuals do not have a `.material` but we still need to translate them into a `pygfx` material. I fixed this with 
be5be00 and 7fe732 and that example now completes. 

Since `ColorVisual` is now explicitly converted, the aesthetics also changed slightly because trimesh uses fairly high default values for shininess:

Screenshot 2024-04-05 at 10 30 11

Not sure if you care about that?

pygfx/utils/load.py
almarklein: Then perhaps this argument is actually an enum, and we should make it an actual enum (allowing the user to select the option using a string).

examples/feature_demo/lightmap.py
almarklein: My argument is mostly, if `load_mesh()` returns a list of meshes (which may contain just one mesh), is not not better to call it `load_meshes()`? 

> We could add a keyword argument to control the return type:

It's actually shorter for a user to add `[0]`  after the call :)

pygfx/utils/load.py
schlegelp: Perhaps something like this then?
```
lights :  "auto" | "file" | "none"
    Whether to import lights. "auto" (default) will always add lights to the scene
    even if the file does not contain any; "file" will import only lights defined in
    the file; "none" will not add any lights to the scene.
```

examples/feature_demo/lightmap.py
Korijn: > My argument is mostly, if `load_mesh()` returns a list of meshes (which may contain just one mesh), is not not better to call it `load_meshes()`?
> 
> > We could add a keyword argument to control the return type:
> 
> It's actually shorter for a user to add `[0]` after the call :)

To be honest, I'm okay with either at this point, it feels like bikeshedding. What matters to me is that we offer the functionality in an unsurprising way, and that the API is predictable. Those things are at odds here.

load_mesh is more consistent (predictable) since other load functions are also singular, but load_meshes is less surprising because of the return type being a list.

So both are suboptimal and I don't really mind which one we pick. We'll have to document them well and then that will be that. :)

pygfx/utils/load.py
```diff
@@ -5,72 +5,432 @@
...
+                            gfx_geometries[geometry_name],
+                            gfx_materials[geometry_name],
+                        )
+                    else:
+                        mesh = None
+
+                    # If this is not a leaf node, we need to use a Group
+                    if not is_leaf:
+                        # Generate a new graph for this child
+                        child_group = gfx.Group()
```
Korijn, line 327: Just to be clear, you're not technically required to use a Group, all WorldObjects can have children. So if the current node has geometry, you could just create a Mesh and add children to it.

Not sure if that would complicate the code unnecessarily, but it would help to keep the node count low, which is beneficial for performance.

pygfx/utils/load.py
```diff
@@ -5,72 +5,432 @@
...
+        A dictionary of material objects. Keys are the names of the geometries.
+
+    """
+    gfx_geometries = {}
+    gfx_materials = {}
+    for name, mesh in scene.geometry.items():
+        gfx_geometries[name] = gfx.geometry_from_trimesh(mesh)
+        # If mesh has a material, convert it
+        if hasattr(mesh.visual, "material") and materials:
+            gfx_materials[name] = gfx.material_from_trimesh(mesh.visual.material)
```
Korijn, line 223: Is it possible to tell if two trimesh materials are identical/equivalent? Same question for geometry.

If possible, you can share the geometry and material objects between WorldObjects and reduce VRAM usage.

Korijn:
This is amazing code! Very neatly done.

pygfx/utils/load.py
schlegelp: Ah, good point! Originally I thought, I'd stick as close as possible to the trimesh scene graph in case the use of groups is intentional. However, it seems that in the trimesh scene graph WorldObjects can't have children directly, i.e. they _have_ to use groups.  

With f56bcf7 this is now only using `gfx.Groups` when absolutely necessary.

pygfx/utils/load.py
schlegelp: Yes, you are right: at least materials can be identical. I assume that can also be the case for geometries but I haven't seen that in the wild.

I'm not sure about about equivalent though - it doesn't look like trimesh materials/objects implement any special `__eq__` methods.

With 2e4b649 this is now de-duplicating materials and geometries based on identity but not equivalence.

pygfx/utils/load.py
schlegelp: @almarklein please have a look if you like the design introduced with 8a6ec2f better?

schlegelp:
Thanks @Korijn! A lot of my work has greatly benefitted from what you guys are doing here and e.g. over at `vispy`, and I'm glad I can give something back in return.

Also FYI: with dccdb39 `load_scene` now also works if trimesh loads a `VoxelGrid`:

```python
>>> scene = gfx.load_scene('https://github.com/mikedh/trimesh/raw/main/models/chair_model.binvox', remote_ok=True)
```

Screenshot 2024-04-08 at 12 27 38

I think I addressed all questions/issues raised so far?

Korijn:
I have no further feedback! Awesome work. Just gotta wait for @almarklein to resolve the remaining conversations.

schlegelp:
Looks like some tests are failing because `trimesh` has a soft-dependency on `httpx` for its `load_remote()` function. I must have gotten this from somewhere else in my local environment because the tests worked. 

I have added `httpx` to the `dev` requirements with 503fbfd.

schlegelp:
*sigh* OK, `networkx` was also missing but that should really be it. Now there are technically three soft dependencies:
- `trimesh` (for both `load_mesh` and `load_scene`)
- `httpx` (for both `load_mesh` and `load_scene` if source is a URL)
- `networkx` (for `load_scene`) 

Given that, would you prefer if I added verbose `try/except` import statements? Something like:

```python
if "https://" in str(path) or "http://" in str(path):
   try:
      import httpx
   except ImportError:
      raise ImportError("Loading scenes from remote sources requires the `httpx` library: `pip install httpx`")
```

I could probably re-write the implementation to not require networkx.

schlegelp:
No, no big deal at all:
- 8d9292d removed the necessity for having `networkx` installed 
- with 3070c16 we now raise informative errors if `trimesh` or `httpx` are missing

examples/feature_demo/lightmap.py
almarklein: 🤷 I agree. Let's move on :)

pygfx/utils/load.py
```diff
@@ -3,74 +3,564 @@
...
+
+    This function requires the trimesh library.
+
+    Parameters
+    ----------
+    path : str
+        The filepath. Can be a local file or a URL.
+    flatten : bool
+        If True, will ignore any hierarchical structure in the scene graph and
+        instead import as a flat list of objects.
```
almarklein, line 105: Would be nice to specify the default value for this and a few other args.

almarklein (changes requested):
Really nice work! I have just two comments/requests:

* The functions like `load_scene` have become really powerful with several args to control import behavior. Would be even nicer if their docstrings mentioned the argument's default values.
* Need to run `black .`:)

schlegelp:
> * The functions like `load_scene` have become really powerful with several args to control import behavior. Would be even nicer if their docstrings mentioned the argument's default values.
> * Need to run `black .`:)

Should both be fixed now!

pygfx/utils/load.py
almarklein: The new enums look great!

almarklein (approved):

Korijn (approved):
Awesome job! Thank you!

pygfx/utils/load.py
```diff
@@ -3,74 +3,562 @@
...
-            gfx.Mesh(
-                gfx.geometry_from_trimesh(m),
-                gfx.material_from_trimesh(m.visual.material),
-            )
-            for m in scene.geometry.values()
-        ]
+            mesh = gfx.Mesh(gfx_geometries[geometry_name], gfx_materials[geometry_name])
+
+            if apply_transforms:
+                mesh.local.matrix = transform  # set the node's transform
```
panxinmiao, line 224: Setting the `local.matrix` of an object will not have any effect, unless you load the entire scene as a whole.

If you want to load a Mesh independently, its `local.matrix` property is meaningless, because the `local.matrix` property is only useful when the object is part of a scene or in a coordinate system. In other words, the `local.matrix` only becomes meaningful when the Mesh is added to a parent object, but at that point, the matrix will be recalculated.

Korijn: Are you sure? Only `world.matrix` depends on the parent's transform.

panxinmiao: > Are you sure? Only `world.matrix` depends on the parent's transform.

Hmm, I'm not entirely sure, and it's possible I may have misunderstood.  but now the transform matrix of the Mesh is not functioning as expected when load from gltf file. If that's the case, the issue might be occurring elsewhere.

PR #708, opened 2024-04-04
Add PointsMarkerMaterial

almarklein:
## Intro

This adds a new points material to draw markers with different shapes, that optionally have an edge. I followed what MPL and Plotly do: the line that represents the edge is centered on what is normally the marker edge. I.e. not on the inside of the marker, or on outside of it, but halfway. For blending, if `edge_color` is semi-transparent, it will show the marker beneath. In other words, it's as if drawing circles over the markers.

## Refs

* Plotly marker types: https://plotly.com/python/marker-style/#custom-marker-symbols
* MPL marker types: https://matplotlib.org/stable/api/markers_api.html
* The marker paper by Nicolas: https://jcgt.org/published/0003/04/01/
* Codepen to play with markers in plotly: https://codepen.io/plotly/pen/LVMRMO

In de codepen, replace the `trace1` with the below, and hack around:
```
var trace1 = {
  x: [1, 2, 3, 4],
  y: [10, 15, 13, 17],
  mode: 'markers',
  marker: { color: 'red', size: 40, symbol: 'square', line: {width: 2, color: 'black'}}
};
```

## Tasks

* [x] Implement `material.edge_width` and `material.edge_color`.
* [x] Implement more marker shapes.
* [x] Check blending (in flycontroller example, the black edges become colored). -> track in #724
* [x] Add (validation) examples.
* [x] Make a proper enum for the available shapes?

----

Screen Recording 2024-04-04 at 13 30 56

image


pygfx/materials/_points.py
```diff
@@ -238,6 +238,118 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+            "o": "circle",
+            "s": "square",
+            "D": "diamond",
+            "+": "plus",
+            "x": "cross",
+            "^": "triangle_up",
+            "<": "triangle_left",
+            ">": "triangle_right",
+            "v": "triangle_down",
+            "♥": "heart",
```
kushalkolar, line 339: emojis in the source code, very modern 😂

almarklein: Just unicode :)

pygfx/materials/_points.py
```diff
@@ -238,6 +238,118 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+
+        * 'circle' or 'o"
+        * 'ring'
+        * 'square' or 's'
+        * 'diamond' or 'D'
+        * 'plus' or '+'
+        * 'cross' or 'x'
+        * 'asterix'
+        * 'triangle_up' or '^'
+        * 'triangle_down' or ''
```
Korijn, line 294: Missed one

pygfx/materials/_points.py
```diff
@@ -238,6 +238,118 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+        * 'triangle_left' or '<'
+        * 'triangle_right' or '>'
+        * 'heart' or '♥'
+        * 'spade'
```
Korijn, line 298: Come on! ♠️

pygfx/materials/_points.py
```diff
@@ -238,6 +238,118 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+        self.uniform_buffer.update_range(0, 1)
+
+    @property
+    def marker(self):
+        """The type/shape of the markers.
+
+        Possible values are:
+
+        * 'circle' or 'o"
+        * 'ring'
```
Korijn, line 287: 💍

pygfx/materials/_points.py
```diff
@@ -238,6 +238,118 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+        * 'square' or 's'
+        * 'diamond' or 'D'
+        * 'plus' or '+'
+        * 'cross' or 'x'
+        * 'asterix'
```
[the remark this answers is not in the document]
almarklein: The `*`  has 6 legs, not eight.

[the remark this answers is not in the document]
Korijn: ✳️

pygfx/materials/_points.py
Korijn: They're quite useful actually, as an example you can signal warnings with ⚠️ instead of using (complicated) shell text color output

kushalkolar: I had never thought of that!

pygfx/renderers/wgpu/wgsl/points.wgsl
```diff
@@ -259,3 +306,160 @@ fn fs_main(varyings: Varyings) -> FragmentOutput {
...
     return out;
 }
+
+
+fn get_signed_distance_to_shape_edge(coord_in: vec2<f32>, size: f32) -> f32 {
+    // Thank you Nicolas!
+    //
+    // The paper "Antialiased 2D Grid, Marker, and Arrow Shaders", by Nicolas
+    // Rougier (https://jcgt.org/published/0003/04/01/) has been instrumental to
+    // implement these marker shapes.
```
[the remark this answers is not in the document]
rougier: You're welcome :)

hmaarrfk:
If you think it is a appropriate: I would like to add a request:

* A transparent fill with a a non transparent outline.
* Fills that set alpha != to `0.0` or `1.0`
* Edge colors with alpha != `0.0` or `1.0`.

almarklein:
@hmaarrfk the fill and edge behave as they do in MPL and plotly and their alphas can be changed individually. So all of that should be possible :)

pygfx/materials/_points.py
almarklein: Done!

pygfx/materials/_points.py
```diff
@@ -216,6 +216,123 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+    @property
+    def marker(self):
+        """The type/shape of the markers.
+
+        Supported values:
+
+        * A string from :obj:`pygfx.utils.enums.MarkerShape`.
+        * Matplotlib compatible characters: "osD+x^v<>".
+        * Unicode symbols:
+        * Emoticons:
```
Korijn, line 279: Emojis is the modern day term I believe :)

pygfx/materials/_points.py
```diff
@@ -216,6 +216,123 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+            "◀": "triangle_left",
+            "▶": "triangle_right",
+            # Emoticons (these may look like their plaintext variants in your editor)
+            "❤️": "heart",
+            "♠️": "spade",
+            "♣️": "club",
+            "♦️": "diamond",
+            "💎": "diamond",
+            "💍": "ring",
+            "✳️": "asterix",
```
Korijn, line 324: 📍

pygfx/materials/_points.py
almarklein: I'm from the emoticon generation ;)

pygfx/materials/_points.py
```diff
@@ -216,6 +216,124 @@ class PointsGaussianBlobMaterial(PointsMaterial):
...
+    @property
+    def marker(self):
+        """The type/shape of the markers.
+
+        Supported values:
+
+        * A string from :obj:`pygfx.utils.enums.MarkerShape`.
+        * Matplotlib compatible characters: "osD+x^v<>".
+        * Unicode symbols:
+        * Emoticons:
```
Korijn, line 279:
```suggestion
* Emoticons:
```

pygfx/utils/enums.py
```diff
@@ -120,4 +121,24 @@ class CoordSpace(Enum):
...
+    cross = None  #: x A rotated plus symbol.
+    asterix = None  #: ✳️ A plus and a cross combined.
+    triangle_up = None  #: ▲
+    triangle_down = None  #: ▼
+    triangle_left = None  #: ◀
+    triangle_right = None  #: ▶
+    heart = None  #: ♥
+    spade = None  #: ♠
+    club = None  #: ♣
+    pin = None  #: A 'location' symbol.
```
Korijn, line 141:
```suggestion
    pin = None #: 📍
```

hmaarrfk:
Is the center of the triangle markers meant to be on the center of the diamond? Or on the tip of triangle where both sides of equal length mean? Or at the intere'sections of the medians?

Mostly asking if the location of the triangle marker origins is intentional

almarklein:
> Is the center of the triangle markers meant to be on the center of the diamond?

Good question. These triangles are created by the intersection of planes. It's easy to put them at the center with an offset though!

hmaarrfk:
So for what it's worth. It looks like matpkotlib and plotly have different markers that seem to cover variations of the options I listed out. 

On my phone on a flight now (which is why I have time to browse) so it's hard to really load pages.  But it seems like a good idea to just make sure that things "align" with plotly or matplotlib. 

For example. They had a markers to do the original behavior 

image

So this is somewhat to say that even the original markers were useful to some in that they made it to matplotlib!

almarklein:
I tried to align with MPL and Plotly where I could. But the way these markers are drawn (i.e. calculating an SDF from the point coordinates) has its limitations on what shapes we can support. I added all shapes from Nicolas' paper except chevron, tag, infinity, block arrow, ellipse.

So the list can be extended in the future. For shapes that we cannot draw with this technique, we can use a future `PointsSdfMaterial`.

almarklein:
Finally, all green 🚀

Korijn (approved):

examples/screenshots/validate_points_markers.png
Korijn, line 1: the pin looks cut off at the top, is that right?

almarklein: mmm, yeah, it looks like its just a bit too close to the edge.

pygfx/materials/_points.py
```diff
@@ -216,6 +216,123 @@ class PointsGaussianBlobMaterial(PointsMaterial):
     """
 
 
+class PointsMarkerMaterial(PointsMaterial):
```
hmaarrfk, line 219: I presume that one of the challenges with transparent colors and edges might be that opacity and is_transparent are still independent of the colors and the antialiasing. 

I believe that with anti-aliasing enabled, is_transparent should always be set true and similarly for transparent colors. 

This is to ensure that it is rendered in the correct pass (opaque vs transparent).

almarklein: Oh good point! This crossed my mind a few times, but I forgot to update the render_mask logic!

hmaarrfk: I think it might also be import for AA text but I haven't really looked into it

almarklein: I just checked. Could indeed take more into account:
https://github.com/pygfx/pygfx/blob/637bef6460dd47d7782c8e728ace69173c50e864/pygfx/renderers/wgpu/shaders/textshader.py#L73-L78

almarklein: pr on the way

hmaarrfk: I see the `shaders/*.py` can add other properties that they look at when deciding what the `render_mask` is.

PR #709, opened 2024-04-05
Add functions to select a specific gpu

almarklein:
* [x] Add `gfx.renderers.wgpu.select_power_preference()`
* [x] Add `gfx.renderers.wgpu.select_adapter()`
* [x] Add example, and test that these actually work.
* [x] Check whether the docs look good.

### DIFF
--- a/pygfx/renderers/wgpu/engine/shared.py
+++ b/pygfx/renderers/wgpu/engine/shared.py
@@ -49,19 +49,31 @@ class Shared(Trackable):
     # target platforms.
 
     _features = set(["float32-filterable"])
-
+    _selected_adapter = None
+    _power_preference = None
     _instance = None
 
     def __init__(self, *, canvas=None):
         super().__init__()
 
+        # Set this instance as the global one
         assert Shared._instance is None
         Shared._instance = self
 
-        # Create adapter and device objects - there should be just one per
+        # Select adapter to use.
+        if Shared._selected_adapter:
+            self._adapter = Shared._selected_adapter
+        else:
+            self._adapter = wgpu.gpu.request_adapter(
+                power_preference=Shared._power_preference or "high-performance"
+            )
+
+        # Create logical device from adapter. There should be just one per
         # process. Having a global device provides the benefit that we can draw
-        # any object anywhere.
-        self._adapter = wgpu.gpu.request_adapter(power_preference="high-performance")
+        # any object anywhere. Supporting different devices per renderer/canvas
+        # is technically possible, but would require an extra layer of
+        # indirection in the pipeline objects (device -> environment -> passes).
+        # So out of scope for the time being.
         self._device = self.adapter.request_device(
             required_features=list(Shared._features), required_limits={}
         )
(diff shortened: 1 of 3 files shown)


pygfx/renderers/wgpu/engine/shared.py
```diff
@@ -130,6 +142,66 @@ def _create_diagnostics(self):
...
+    The order of the adapters returned by ``wgpu.gpu.enumerate_adapters()`` is
+    such that Vulkan adapters go first, then Metal, then D3D12, then OpenGL.
+    Within each category, the order as provided by the particular backend is
+    maintained. Note that the same device may be present via multiple backends
+    (e.g. vulkan/opengl).
+
+    We cannot make guarantees about whether the order of the adapters matches
+    the order as reported by e.g. ``nvidia-smi``. We have found that on a Linux
+    multi-gpu cluster, the order does match, but we cannot promise that this is
+    always the case. If you want to make sure, do some testing by allocating big
```
kushalkolar, line 188: Maybe add the few lines to make a buffer to the docstring here?

almarklein: done

pygfx/renderers/wgpu/engine/shared.py
```diff
@@ -130,6 +142,86 @@ def _create_diagnostics(self):
...
+
+    Select an adapter as obtained via ``wgpu.gpu.enumerate_adapters()``, which
+    can be useful in multi-gpu environments.
+
+    For example::
+
+        adapters = wgpu.gpu.enumerate_adapters()
+        adapters_tesla = [a for a in adapters if "Tesla" in a.summary]
+        adapters_discrete = [a for a in adapters if "DiscreteGPU" in a.summary]
+        pygfx.renderers.wgpu.select_adapter(adapters_discrete[0])
```
kushalkolar, line 174: I guess this isn't exposed at top level in case other renderers are added in the future?

almarklein: Yes, the `pygfx.renderers.wgpu` is a public namespace for power-users.  It has a special place in the docs: https://pygfx.readthedocs.io/en/latest/advanced_api.html The `print_wgpu_report()` function is one exception that is promoted to the root namespace.

almarklein: We could promote other functions to the root namespace as well if there is a demand for that. But I'm a bit careful doing that since its much easier to add api then to remove it (and pissing off users).

kushalkolar:
LGTM! Thanks this will be very useful :smile:

pygfx/renderers/wgpu/engine/shared.py
kushalkolar: makes sense, people who want to select a GPU will probably dig and find it easy enough

almarklein: And the example will help 😄

Korijn (approved):

PR #711, opened 2024-04-06
Use the render_mask to control when the background shader is displayed

hmaarrfk:
This was found in https://github.com/pygfx/pygfx/pull/704

It seems that write_pick was previously used a sentinel to detect the type of pass that was getting used.

It is currently defined as true when:

* BasePass
* OpaquePass
* FullOpaquePass
* SimpleSinglePass
* FrontmostTransparencyPass

### DIFF
--- a/pygfx/objects/_more.py
+++ b/pygfx/objects/_more.py
@@ -55,10 +55,10 @@ class Background(WorldObject):
 
     """
 
-    def __init__(self, geometry=None, material=None, **kwargs):
+    def __init__(self, geometry=None, material=None, render_mask="opaque", **kwargs):
         if geometry is not None and material is None:
             raise TypeError("You need to instantiate using Background(None, material)")
-        super().__init__(None, material, **kwargs)
+        super().__init__(None, material, render_mask=render_mask, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


pygfx/objects/_more.py
```diff
@@ -55,10 +55,10 @@ class Background(WorldObject):
 
     """
 
-    def __init__(self, geometry=None, material=None, **kwargs):
+    def __init__(self, geometry=None, material=None, render_mask="opaque", **kwargs):
```
hmaarrfk, line 58: I understand that this Background WorldObject was thought to be a little special.

we could if you want 

* Allow the user to mess things up (as they can in the currently proposed implementation)
* redefine the property `render_mask` to simply be a constant instead of "settale" by the user.

almarklein (approved):

pygfx/objects/_more.py
hmaarrfk: I meant specifically redefining the property (attribute) of the `Background` material. 

however, I agree that flexibility is great, so I would suggest keeping it as proposed in this PR.

Korijn (approved):

PR #713, opened 2024-04-07
Recall the last used font properties with set_text

hmaarrfk:
I believe a common usecase is to generate the text geometry "once", then keep updating it with a call to `set_text` without always thinking about the font and other properties.

It would be good to start to "recall" the last used font properties, style and weight. We currently drag along the font style and weight every time we try to set the text in our application. We have code that changes the displayed text often, similar to the stats counter, but we would like it to maintain the same "font"

https://github.com/pygfx/pygfx/blob/2927712faeb867412a233b3efc8743ba1cc4cb11/pygfx/helpers/_stats.py#L123

Can become more commonplace.

Thanks for considering.

### DIFF
--- a/pygfx/geometries/_text.py
+++ b/pygfx/geometries/_text.py
@@ -167,11 +167,6 @@ class TextGeometry(Geometry):
         preferred in the given order. Characters that are not supported by any
         of the given fonts are rendered with the default font (from the Noto
         Sans collection).
-    style: str
-        The style of the font (normal, italic, oblique). Default "normal".
-    weight: str, int
-        The weight of the font. E.g. "normal" or "bold" or a number between
-        100 and 900. Default "normal".
     direction : str
         The text direction. By default the text direction is determined
         automatically, but is always horizontal. Can be set to 'lrt', 'rtl',
@@ -191,8 +186,6 @@ class TextGeometry(Geometry):
         line_height=1.2,
         text_align="left",
         family=None,
-        style=None,
-        weight=None,
         direction=None,
     ):
         super().__init__()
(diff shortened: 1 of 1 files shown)


almarklein:
I'm hesitant, especially for `style` and `weight, and this has mostly to do with `set_martkdown` (and possible other `set_xx` methods we may add in the future, like `set_html`). Should the previous/default style and weight be ignored, or used as the baseline?

This is a rather low-level object, and this change would add a kind of overridable state to it. As it is now, it's at least clear what it means when `set_text()` is used, what text attributes are bound to the geometry object (like size) and what are bound to the text-items.

So my feeling is that a convenience like this should be implemented higher up.

hmaarrfk:
> This is a rather low-level object

I totally get it.

> Should the previous/default style and weight be ignored, or used as the baseline?

Ultimately, I think requesting `FontProp(None, None, None)` is likely not what the user wants unless we provide a way to change the `default_font` (though I really dislike APIs that encourage changing some global state). 

> set_martkdown

`set_martkdown` uses the even lower level function `set_text_items`.

I mean, i'm happy to drag our font around, but I've taught a few people already:

> Always remember to specify the font family you want to use anytime you want to change the text of your label.


I've pushed a reduced set of changes that will only remember the font family.

However, I'm also going to close this PR and let you re-open it if you feel like it would be hurting the long term direction you have in mind.

almarklein:
> but I've taught a few people already:
> 
> >    Always remember to specify the font family you want to use anytime you want to change the text of your label.

I see how this is not ideal 🤔

PR #715, opened 2024-04-10
Add SkinnedMesh

panxinmiao:
I am introducing `SkinnedMesh` into `pygfx`.

Example:
Video_2024-04-10_102245

I have opened this draft PR in advance as there are some issues that need to be discussed in this work.


panxinmiao:
Some `Wobject` (such as `Skeleton`) need to update their own state before each frame is rendered, but we do not seem to have provided this method. 

_I remember that in the early days of pygfx, the Wobject had a method like `update_world_matrix()`, which was called before each frame was rendered to update its own transform matrix (based on the update flag to determine if the actual update was necessary). When an object needs to perform some additional state updates before each rendering, we could override this method and add some extra logic. But now the update of the model matrix is managed by the `RecursiveTransform` object automatically, no longer called before each frame rendering. The `wobject.world.on_update()` callback is only triggered when the model matrix changes, and cannot guarantee that it will be called before each frame rendering._

Could we add a hook method like `before_render()` in the `Wobject` class, which would be automatically called by the renderer before each rendering? Or any other good suggestions?

examples/feature_demo/skinned_mesh.py
```diff
@@ -0,0 +1,134 @@
...
+    gfx.OrbitController(camera, register_events=renderer)
+
+    def animate():
+        t = time.time()
+        for bone in mesh.skeleton.bones:
+            rotation_y = math.sin(t) * 2 / len(mesh.skeleton.bones)
+            bone.local.rotation = la.quat_from_euler((0, rotation_y, 0))
+
+        mesh.skeleton.update()
+        skeleton_helper.update()
```
panxinmiao, line 124: See: https://github.com/pygfx/pygfx/pull/715#issuecomment-2046493145
Now we need to manually update in the user program.

almarklein:
> Some `Wobject` (such as `Skeleton`) need to update their own state before each frame is rendered, but we do not seem to have provided this method.

Since recently, the shader classes have a hook like this. They can set `self.needs_bake_function` to `True` and that will make their `bake_function()` to be called. See the [line shader](https://github.com/pygfx/pygfx/blob/be623d2b4d77a8ca76d02ccebc0e5e2cd5e7d742/pygfx/renderers/wgpu/shaders/lineshader.py#L113) for an example.

almarklein:
To provide some context, we want to make such a hook opt-in, so that if you have 10000 objects in the scene, we can render them more efficiently if they do not change (using renderbuffers, eventually).

panxinmiao:
> Since recently, the shader classes have a hook like this. They can set `self.needs_bake_function` to `True` and that will make their `bake_function()` to be called. See the [line shader](https://github.com/pygfx/pygfx/blob/be623d2b4d77a8ca76d02ccebc0e5e2cd5e7d742/pygfx/renderers/wgpu/shaders/lineshader.py#L113) for an example.
> 
> Would that do the trick?

It can achieve the desired functionality, but it's not so convenient. The `bake_function()` is a shader-related method, whereas what we're looking for is a Wobject-specific method.

When different wobjects use the same shader, and they all require their own `bake_function()` method, we can foresee that the `bake_function()` method in the shader will have many similar conditional branches:

```
if isinstance(wobject, XXXX):
    # some updates
elif isinstance(wobject, YYYY):
    # some updates
...
```

Furthermore, when users want to implement their own specific logic for certain Wobjects, they would need to modify the underlying shader code, which is not an ideal solution.

For example, with `SkeletonHelper`, even though we may provide a built-in `SkeletonHelper` in the pygfx library, you can imagine how inconvenient it would be if users wanted to implement a similar custom object themselves. I believe allowing users to inherit from `Wobject` and implement their own specific logic is a very common requirement.

Korijn:
As an aside, we are thinking about the overarching issue of render loops, events and such in #495 We intend to address this around Q2 of this year, so I recommend to keep an eye on that.

The vision is that the renderer should be just that: a renderer. Events and wobject lifecycles should be implemented outside of the renderer. It's a complicated subject 😅 I apologize for the current state where it is just not really considered well

panxinmiao:
After searching for a solution for some time, I've decided to implement my own GLTF file loader for pygfx, completely based on the [GLTF 2.0 specification](https://raw.githubusercontent.com/KhronosGroup/glTF/main/specification/2.0/figures/gltfOverview-2.0.0d.png).

I've found a `gltflib` library that can handle basic GLTF file reading and writing, which should help simplify a lot of the work. The new loader I'm currently developing can already correctly load Meshes and Skeletons, and I hope it will eventually support the full GLTF 2.0 specification, including GLTF Extensions. However, this may not be completed within this PR. I will continue to improve it incrementally based on the requirements.

image

panxinmiao:
I've implemented the loading of GLTF animation data and updated the example.

We can use this example as a starting point to design the future pygfx animation system.




Additionally, I've had an ongoing concern. Our current RecursiveTransform system seems to generate a significant amount of unnecessary matrix calculations.

While the current design is very convenient, where modifying an object's transform properties automatically updates the matrix of all related objects, in many cases, we need to update multiple properties of multiple targets within a single frame. However, we only need to calculate the final transform matrix once before rendering, without the need to recursively calculate everything upon each property change, as that would incur a significant performance overhead.

Taking the code snippet in the skinning_animation.py as an example:

```
    # if property == "scale":
    #     target.local.scale = value
    # elif property == "translation":
    #     target.local.position = value
    # elif property == "rotation":
    #     # TODO: should use spherical linear interpolation instead
    #     target.local.rotation = value
    # elif property == "quaternion":
    #     target.local.quaternion = value / np.linalg.norm(value)  # normalize quaternion
    # else:
    #     print("unknown property", property)

    target_temp = getattr(target, "_temp", None)

    if target_temp is None:
        target._temp = {}
        target_temp = target._temp

    if property == "rotation":
        value = value / np.linalg.norm(value)  # normalize quaternion

    target_temp[property] = value
```
The commented-out section above is my initial code, where I set the position, rotation, and scale of the bones individually. This implementation only ran at under 10 FPS on my computer. I had to change it to the implementation below, which can achieve 20 FPS. However, this still involves a significant amount of unnecessary calculations.

Let's say we have a Skeleton with 10 nodes, for example, `bone1 --> bone2 --> ... --> bone10`. If we need to update the state of these 10 bones in a single frame, and we start by setting the transform property of bone1, the recursive calculation of all the child nodes is completely unnecessary, as the child nodes' states have not yet been updated. To update all these bones, we only need 10 matrix multiplications, but in the current design, in the worst case, we might end up calculating 10 + 9 + 8 + ... + 1 = 45 times. This becomes very evident when there are a large number of objects in the scene, especially with complex parent-child relationships.

panxinmiao:
In the early implementation of pygfx (and the current implementation of Three.js), the Wobject matrix updates were passive. When modifying the transform property of an object, it would not immediately recursively calculate all the matrices. Instead, it would only set a `matrix_need_update=True` flag on itself and all of its child nodes. The world matrix of all the changed objects in the scene would then be updated in a single pass, just before the final rendering (ensuring that the properties of all the nodes in the scene had been updated).

panxinmiao:
I conducted a test experiment where I used some tricks to block the automatic recursive update of the Bone object's world matrix, and instead manually updated it once before rendering. As a result, I was able to achieve a frame rate of over 80 FPS on my computer. 



This confirms the concerns I raised in my previous comments.

(By the way, The current interpolation lookup algorithm for animation keyframes has not yet been thoroughly optimized. After this optimization, I estimate that the frame rate for this use case can exceed 120 FPS on my computer. 😄 )

panxinmiao:
Please let me know your thoughts on the issues mentioned above. @almarklein @Korijn .

 As the discussion has evolved, this pull request seems to have expanded in scope. Do you think we should try to keep it more focused within defined boundaries?

almarklein:
Ok, it looks like our transformation system needs some work then. Great to hear that there is quite a lot to gain though 😄 

In terms of scope, you're already touching 3 topics here: skinning, glfw loading, transformation optimizations. Ideally this would be three prs. I think it should be relatively easy to move the glfw loader into a new pr that we merge first, would that be ok for you?

almarklein:
> you're already touching 3 topics here: skinning, glfw loading, transformation optimizations.

Not the least topics btw 🚀 ❤️ !

panxinmiao:
> I think it should be relatively easy to move the gltf loader into a new pr that we merge first, would that be ok for you?

Okay, I will do this.

panxinmiao:
**TBR:**

Completed tasks:

- `SkinnedMesh`
- `Skeleton` & `Bone`
- `SkeletonHelper`
- Added skins and animations loader support to the glTF loader.
- Examples: `skinned_mesh.py` and `skinning_animation.py`

To improve performance, I used some tricks to block the automatic recursive update of the `Bone` object's matrix. This is a temporary solution. see: https://github.com/pygfx/pygfx/pull/715#issuecomment-2053385803

**TBD:** Explore a mechanism to automatically update the objects (`Skeleton`, `SkeletonHelper`, etc.) dynamically. see: https://github.com/pygfx/pygfx/pull/715#issuecomment-2046493145

Korijn:
> without the need to recursively calculate everything upon each property change, as that would incur a significant performance overhead

While implementing the new transform system, I have been very adamant in my reviews about ensuring that updates propagate lazily through the system, only incurring the overhead of computation when some derived information is actually read/consumed.

The obvious downside of delaying all updates until just before rendering, is that there are actually valid use cases to have the updated matrices earlier than that, for example when picking with the mouse or as part of collision detection.

With the new transform system we tried to find a system that is **does not eagerly propagate** modified transformation parameters such as position, but only does so lazily on demand. I'm a little surprised that it still doesn't seem to work that way. :/

panxinmiao:
> The obvious downside of delaying all updates until just before rendering, is that there are actually valid use cases to have the updated matrices earlier than that, for example when picking with the mouse or as part of collision detection.

Perhaps we can delay the automatic update of matrices until before rendering, but at the same time provide a method for users to update matrices immediately. If users need to access these matrices before rendering, they need to ensure that they have been updated themselves.

panxinmiao:
> With the new transform system we tried to find a system that is **does not eagerly propagate** modified transformation parameters such as position, but only does so lazily on demand. I'm a little surprised that it still doesn't seem to work that way. :/

I'm pretty sure it's not working now...

> I'm 100% sure there is a system in there to operate like this, but perhaps its effect is only local, and does not propagate through recursive transform hierarchies correctly.

I noticed in the code that the computation of transforms is cached, which can avoid redundant calculations. However, it does not prevent the recursive propagation of matrix updates when properties are changed.

almarklein:
> I'm pretty sure it's not working now...

A good first step would be to extract a few cases that demonstrate that it does not work, and write them as unit tests.

panxinmiao:
> A good first step would be to extract a few cases that demonstrate that it does not work, and write them as unit tests.

Matrix multiplication happens here:
https://github.com/pygfx/pygfx/blob/eb0b0898bd3f5f07031cf1dedc4d8acb3c530648/pygfx/utils/transform.py#L743C1-L745C53

Add logging here:
```
    @cached
    def _matrix(self):
        print("matrix multiplied")
        return self._parent.matrix @ self.own.matrix
```
Run this test code:
```
import pygfx as gfx

a = gfx.WorldObject()
b = gfx.WorldObject()

a.add(b)
print('==============')
a.local.position = (1, 2, 3)

# output:

# matrix multiplied
# matrix multiplied
# ==============
# matrix multiplied
# matrix multiplied
```
You can see that matrix multiplication is called twice when executing a.add(b), and it is called twice again immediately when setting `a.local.position`.

Korijn:
To be honest, it is quite hard for me to review this properly. I'm a little concerned about the experimental nature/workarounds for the transform problems. I see two options:

# Option 1
1. Remove the workarounds for transform performance problems and accept the poor performance. Then we can review & merge this.
2. Create a separate PR to address the transform performance problems.

# Option 2
1. Create a separate PR to address the transform performance problems.
2. Update this PR afterwards and then we can review & merge this.

panxinmiao:
> Remove the workarounds for transform performance problems and accept the poor performance. Then we can review & merge this with a little smaller scope.

I think this approach is feasible. Let this PR maintain the minimal feature, which mainly completes the features related to `SkinnedMesh`.

Since Bone objects tend to have complex parent-child relationship graphs in most scenarios, the current automatic recursive calculation of transforms has a significant performance impact on Bones and Skeletons. I have adopted this temporary solution just to indicate the existing problem.

We definitely need to address the performance issues of the current Transform system eventually.

BTW, If you accept the solution of lazy updating (on before rendering) the `world matrix`, I can provide a PR.

Korijn:
> BTW, If you accept the solution of lazy updating (on before rendering) the `world matrix`, I can provide a PR.

Well, I do want to have lazy updates, but not through `update_world_matrix` or `before_render` :) I believe it can be solved internally in the transform module without any relationship to other modules.

panxinmiao:
> Well, I do want to have lazy updates, but not through `update_world_matrix` or `before_render` :) I believe it can be solved internally in the transform module without any relationship to other modules.

I don't think this is contradictory.

After checking the current code, it's clear that the current code is not designed for lazy updates, but rather is immediate and reactive, where any property changes immediately update all relevant matrices.

https://github.com/pygfx/pygfx/blob/41aff5b5f7ba51762e564906593c7eb22fe60c74/pygfx/utils/transform.py#L361C5-L367C71

https://github.com/pygfx/pygfx/blob/41aff5b5f7ba51762e564906593c7eb22fe60c74/pygfx/utils/transform.py#L583C1-L586C27

I think lazy updates mean that when a property is updated, instead of immediately updating the matrices, it only sets a flag marking the relevant matrices as dirty data. When accessing these dirty data (with the flag set), it triggers the actual update and clears the dirty flag.

A typical rendering engine's render loop includes these typical stages:

**1. User program:** 
Update the state and position information of sceen objects, which will make some data (that needs to be submitted to the GPU) become dirty data.

**2. Render preparation (CPU):** 
Prepare for the final rendering, perform various calculations and data preparation, including view transformation, scene culling, render queues, etc.

**3. GPU stage:** 
Geometry processing, lighting, rasterization, etc.

I thick the final calculation and synchronization of the transform matrices should happen in the **render preparation stage**, not in the **user program stage**.

panxinmiao:
> I don't think this is contradictory.
> ......
> I thick the final calculation and synchronization of the transform matrices should happen in the render preparation stage, not in the user program stage.

Anyway, the current pull request is unrelated to the topic of lazy updates :)

If you have any review comments on the current pull request, please let me know.

pygfx/objects/_skins.py
```diff
@@ -0,0 +1,206 @@
...
+    A skeleton is a collection of bones that are used to animate a SkinnedMesh.
+
+    Parameters
+    ----------
+    bones : List[Bone]
+        The array of bones.
+        Note this is a copy of the original array, not a reference, so you can modify the original array without effecting this one
+    bone_inverses : List[np.ndarray] | None
+        An array of matrix4x4 that represent the inverse of the world matrix of the individual bones.
+        If not provided, they will be auto calculated.
```
almarklein, line 37: Avoid confusion with ndarray:
```suggestion
    bones : List[Bone]
        The list of bones.
        Note that a copy will be made, so you can modify the original list without effecting this one.
    bone_inverses : List[np.ndarray] | None
        A list of matrix4x4 that represent the inverse of the world matrix of the individual bones.
        If not provided, they will be auto calculated.
```

pygfx/objects/_skins.py
```diff
@@ -0,0 +1,206 @@
...
+from ..utils.enums import BindMode
+from ..resources import Buffer
+from ._more import Mesh
+
+
+class Bone(WorldObject):
+    """A bone.
+
+    A bone is a part of a skeleton.The skeleton in turn is used by the SkinnedMesh.
+    Bones are almost identical to a blank WorldObject.
```
almarklein, line 16: Do these Bone objects *contain* `Mesh` objects? And if so, can they contain more than one mesh?

pygfx/objects/_skins.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @property
+    def bind_mode(self):
+        """
+        How a skinned mesh is bound to its skeleton. Either "attached" or "detached".
+
+        "attached" means the skinned mesh shares the same world space as the skeleton.
+
+        In contrast, "detached" is useful when sharing a skeleton across multiple skinned meshes.
+        Default is "attached".
```
almarklein, line 169: We can just refer to the docs of the enum.
```suggestion
        See :obj:`pygfx.utils.enums.BindMode`:. Default "attached".
```

almarklein (commented):
Just a few minor comments.

pygfx/objects/_skins.py
panxinmiao: No.
Bone objects are almost just blank 3D objects, invisible and not actually rendered. They only exist to manage and maintain world transforms

almarklein: Sorry, I meant are the actual meshes children of the `Bone` objects?

panxinmiao: No.
The bone objects and the actual Mesh are independent. 
Through the `bind` method of `SkinedMesh`, Skeleton can be bound to a specific Mesh, thereby driving the animation of the Mesh. 
The same skeleton can be bound to different Meshes, and theoretically, a Mesh can also be bound to different skeletons.

almarklein: 🤯 thanks for explaining!

almarklein (dismissed):

panxinmiao:
image

Added an example demonstrating the usage of the detached binding mode for SkinnedMesh. This mode allows multiple SkinnedMesh instances to share the same skeleton.

_Note: To achieve better performance and smooth animation, #746  and #756  are required_

almarklein (approved):

Korijn (approved):
I apologize for the lack of activity on my part. Some unexpected events occurred at work that cause me to be significantly distracted. I am relying on your own reviews and approvals as a proxy for the quality of the work for now. For the time being, I will not pay very detailed attention myself.

panxinmiao:
It's OK，Thanks for your approval and trust. : )

PR #718, opened 2024-04-10
Refactor enums

almarklein:
We kindof used enums in a few places, but with the large number of marker types in #708 it was time to have a good look at enums. This also touches on #272 a bit.

* [x] Implemented an enum class. 
   * I played a bit with Python's `enum.Enum` but I hate that `SomeEnum.field` provides ... an enum object.
   * In our custom enum it simply produces a string or an int (depening on enum / flag-like).
   * This also makes things simpler when these values reach the templated shader code.
   * I started by using a similar implementation [as in wgpu](https://github.com/pygfx/wgpu-py/blob/e8bef7db15462772c048ff2f325f618046eec2f8/wgpu/enums.py#L12-L27), which works perfectly, but it takes a lot of effort to make Sphynx document them properly.
   * So I used classes instead. As a result I needed a metaclass (yuk), but it's still very little code.
* [x] Moved (I think) all  current enums to `enums.py`.
* [x] I tweaked the docs template a bit. No effect in the end, but a bit clearer, I think. 
* [ ] ~Add some logic to nicely support aliases (e.g. "+" for "plus" in markers).~ Solve in #708

It makes the docs really nice, the circles below are links:
image

Follow-ups:
* Maybe use the same enum class in wgpu-py. 
* New enum for SizeMode (mode/world/screen).
* Proceed with #708.
* Probably more enums we can add ... also with #272.

### DIFF
--- a/docs/_templates/custom_layout.rst
+++ b/docs/_templates/custom_layout.rst
@@ -3,13 +3,8 @@
 .. currentmodule:: {{ module }}
 
 
-{% if objtype == "module" and name in [] %}
-{# Some modules look nicer as a single-page doc. #}
-
-.. automodule:: {{ fullname }}
-  :members:
-
-{% elif objtype == "module" %}
+{% if objtype == "module" %}
+{# Assume that the module docstring has an autosummary section to produce a list of members. #}
 
 .. automodule:: {{ fullname }}
 
--- a/docs/_templates/custom_layout.rst
+++ b/docs/_templates/custom_layout.rst
@@ -2,7 +2,13 @@
 
 .. currentmodule:: {{ module }}
 
-{% if objtype == "module" %}
+
+{% if objtype == "module" and name == 'enums' %}
+
+.. automodule:: {{ fullname }}
+  :members:
+
+{% elif objtype == "module" %}
 
 .. automodule:: {{ fullname }}
 
@@ -14,7 +20,7 @@
     :add-heading: Examples
     :heading-level: ^
 
-{% else %}
+{% elif objtype == "class" %}
 
 .. autoclass:: {{ objname }}
     :members:
@@ -23,5 +29,9 @@
 
 .. minigallery:: pygfx.{{ objname }}
     :add-heading: Examples
+
+{% else %}
+
+.. autodata:: {{ objname }}
 
 {% endif %}
--- a/docs/_templates/custom_layout.rst
+++ b/docs/_templates/custom_layout.rst
@@ -3,7 +3,8 @@
 .. currentmodule:: {{ module }}
 
 
-{% if objtype == "module" and name == 'enums' %}
+{% if objtype == "module" and name in [] %}
+{# Some modules look nicer as a single-page doc. #}
 
 .. automodule:: {{ fullname }}
   :members:
--- a/pygfx/materials/_line.py
+++ b/pygfx/materials/_line.py
@@ -124,7 +124,7 @@ class LineMaterial(Material):
     def color_mode(self, value):
         if value not in ColorMode:
             raise ValueError(
-                f"LineMaterial.Colormode must be a string in {ColorMode}, not {repr(value)}"
+                f"LineMaterial.color_mode must be a string in {ColorMode}, not {repr(value)}"
             )
         self._store.color_mode = value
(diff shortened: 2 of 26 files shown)


Korijn:
In principle I think a custom enum implementation is a bad idea, not worth all the lines of code, just to have its members be a string...

Also, py 3.11 has [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum), which I guess is what you're looking for?

docs/_templates/custom_layout.rst
```diff
@@ -2,7 +2,14 @@
 
 .. currentmodule:: {{ module }}
 
-{% if objtype == "module" %}
+
+{% if objtype == "module" and name in [] %}
```
Korijn, line 6: Isn't this condition always False?

pygfx/materials/_points.py
```diff
@@ -114,28 +115,16 @@ def aa(self, aa):
...
-                value = getattr(ColorMode, value.lower())
-            except AttributeError:
-                raise ValueError(f"Invalid color_mode: '{value}'")
-        else:
-            raise TypeError(f"Invalid color_mode class: {value.__class__.__name__}")
-        if value == ColorMode.face or value == ColorMode.face_map:
-            raise ValueError(f"Points cannot have color_mode {value}")
+        if value not in ColorMode:
+            raise ValueError(
+                f"PointsMaterial.Colormode must be a string in {ColorMode}, not {repr(value)}"
```
Korijn, line 126: color_mode?

pygfx/materials/_points.py
```diff
@@ -184,24 +174,16 @@ def size_space(self, value):
...
-                value = value.split(".")[-1]
-            try:
-                value = getattr(SizeMode, value.lower())
-            except AttributeError:
-                raise ValueError(f"Invalid size_mode: '{value}'")
-        else:
-            raise TypeError(f"Invalid size_mode class: {value.__class__.__name__}")
+        if value not in SizeMode:
+            raise ValueError(
+                f"PointsMaterial.SizeMode must be a string in {SizeMode}, not {repr(value)}"
```
Korijn, line 185: size_mode?

pygfx/objects/_base.py
```diff
@@ -283,20 +266,21 @@ def render_mask(self):
...
+            if value not in dir(RenderMask):
+                msg = f"WorldObject.render_mask must be one of {dir(RenderMask)} not {value!r}"
                 raise ValueError(msg) from None
+            value = RenderMask[value]
         else:
             raise TypeError(
                 f"WorldObject.render_mask must be int or str, not {type(value)}"
             )
+        # Store the value as an int, because this is a flag, but also for backwards compat.
+        assert isinstance(value, int)
```
Korijn, line 282: No assert in runtime code :)

pygfx/utils/enums.py
```diff
@@ -0,0 +1,129 @@
...
+
+    RenderMask
+    ColorMode
+    SizeMode
+
+"""
+
+__all__ = ["RenderMask", "ColorMode", "SizeMode"]
+
+# We implement a custom enum class that's much simpler than Python's enum.Enum,
```
Korijn, line 18: I don't think you can call anything that uses a metaclass simple :)

pygfx/materials/_line.py
```diff
@@ -115,28 +116,16 @@ def aa(self, aa):
...
-                value = value.split(".")[-1]
-            try:
-                value = getattr(ColorMode, value.lower())
-            except AttributeError:
-                raise ValueError(f"Invalid color_mode: '{value}'")
-        else:
-            raise TypeError(f"Invalid color_mode class: {value.__class__.__name__}")
+        if value not in ColorMode:
+            raise ValueError(
+                f"LineMaterial.Colormode must be a string in {ColorMode}, not {repr(value)}"
```
Korijn, line 127: color_mode

docs/_templates/custom_layout.rst
almarklein: Removed.

almarklein:
I addressed your points.

Korijn (approved):

PR #722, opened 2024-04-11
Fix that pick_write is not dynamic

### LINKED ISSUE #720: Make it possible to dynamically change material.pick_write
Follow up on #704

Problem: changing `pick_write` once the object is rendered, does not invoke a shader recompile.

almarklein:
Closes #720

The problem was (obviously) that the `material.pick_write` was not used under a tracking context. The fact that we had methods passing around a wobject (or its material) was a flag that information was used that was not tracked. In this case this also includes `depth_test`.

So I removed the passing around of the world object. Instead any info that is used from the world object must be put into a little dict (`self.wobject_info`) in a way that this is tracked

So this *also* fixes the same problem for `depth_test`, and avoids similar bugs in the future.

### DIFF
--- a/pygfx/renderers/wgpu/engine/pipeline.py
+++ b/pygfx/renderers/wgpu/engine/pipeline.py
@@ -462,11 +462,15 @@ class PipelineContainer:
         self.shared = get_shared()  # the globally Shared object
         self.device = self.shared.device
 
+        # Dict to store info on the wobject that affects shaders or pipeline.
+        # Fields are set in a tracking-context to make sure things update accordingly.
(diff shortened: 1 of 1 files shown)


almarklein:
I tested that it works in an interactive session.

Korijn (approved):

hmaarrfk:
Thank you for addressing this. I know this is the kind of stuff where "contributions" can explode the scope.

I really appreciate your help.

almarklein:
No problem. I think of this particular case more as that it made an existing problem in `pipelines.py` more explicit 😄

Issue #696, opened 2024-03-28
Would you consider miniver for version management?


hmaarrfk:
I know the versioneer transition was somewhat painful for many that relied on it, but it would help me work with your development version and report bugs more precisely.

https://github.com/jbweston/miniver

I can make the PR to enable it.

For those not familiar with versioneer: 

The advantage is that given that you use tags for your releases, and a linear development stype (i.e. no support branches) you could have the current main's version be

```
$ git describe --tags
v0.1.18-10-g34b26ac
```

would report as 

```
pygfx.__version__
'0.1.18.post10+g34b26ac'
```

or something like that.

let me know.

For me, i could start to add compatibility shims to my code that uses your repo without try except statements!

almarklein:
I'm open to the idea. @Korijn ?

almarklein:
Can you explain more about the problem that this would solve?

One advantage is being able to just push a tag without needing commits like #725.

Seeing #726, I also wonder if adding a `get_full_version()` function (or something like it) would also solve the problem, but in a simpler way. It could simply take the release version number (what we have now) and then adds a suffix calculated from git.

hmaarrfk:
> It could simply take the release version number (what we have now) and then adds a suffix calculated from git.

This hybrid approach is interesting. I'll try to give it a go when I have a second again. One constraint is to try to avoid the use of subprocess during a released version to ensure good import speed.

> I also wonder if adding a get_full_version() function (or something like it) 

I feel like `package_name.__version__` is the amount of cognitive load that I can bear when trying to identify the version of the code I'm running. But I'm happy to get to a middle ground.

Ideally, this is something you build once, and lasts 3-4 years so I'm not willing to give up on a good solution just yet.

@Korijn brought up complexity, which is a real concern, when `versioneer` broke a few years back (the code was copied into many repos), it caused alot of churn for packing efforts.

> Can you explain more about the problem that this would solve?

3 Aspects:

1. Downstream developers.
2. Automation of small tasks that add up cognitive load.
3. Faster resolution time for bugs reported by end users 

### Downstream code

The usecase is that in my downstream code I could for example start to be "compatible" with version 0.1.19 even before it comes out. 

Contributing users (this can be you in your own downstream project too) can start to have the conditions:
```python
if Version(pygfx.__version) > Version('0.1.18');
    print("do something special")
```

### Automation of routine tasks

Improved release process cannot be understated. For me, I release on a daily basis so the delay in the process:

1. agreeing to do a release.
[...]

PR #726, opened 2024-04-13
Add miniver to simplify development branch descriptions

hmaarrfk:
Closes: https://github.com/pygfx/pygfx/issues/696

```bash
pip install -e . -vv
# You might only need to run this once I believe. we don't run it daily on our dev machines
```

```python
In [1]: import pygfx

In [2]: pygfx.__version__
'0.1.18.post29+g9342016'

In [3]: from packaging.version import Version

In [4]: Version('0.1.18.post29+g9342016') > Version('0.1.18')
True

In [5]: assert Version('0.1.18.post29+g9342016') > Version('0.1.18')

In [6]: assert Version('0.1.18.post29+g9342016') < Version('0.1.19')
```

So I use this in my project and the github action that I use to upload the package to pypi will make the static version correctly.

You don't need to change the source when you create a new release. just tag then your CI will pickup the new version in the upload to pypi.

see: https://github.com/ramonaoptics/python-teensytoany
Xref: https://github.com/jbweston/miniver


pygfx/__init__.py
```diff
@@ -25,8 +25,8 @@
 
 del pylinalg
 
-__version__ = "0.1.18"
-version_info = tuple(map(int, __version__.split(".")))
+from ._version import __version__
+version_info = tuple(map(int, __version__.split(".")[:3]))
```
hmaarrfk, line 29: I'm not sure what behavior you want from this `version_info` tuple. I don't have such constructs in my code.

I typically use

```
from packaging import Version
```
for all my comparisons.

hmaarrfk:
oh man, the CIs made this more difficult than I wanted to.

.github/workflows/ci.yml
```diff
@@ -41,7 +44,10 @@ jobs:
     strategy:
       fail-fast: false
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
```
hmaarrfk, line 47:
```suggestion
    - uses: actions/checkout@v3
```

hmaarrfk:
hmm, i've never used pyinstaller ...

hmaarrfk:
This is actually the first time i read through the miniver code in earnest.

I understand how it is designed now, and I somewhat understand their design decisions of having 2 files.

But it is somewhat overly complicated.

It may be interesting to use https://github.com/proudust/gh-describe though this seems like getting more untrusted code in the repo.

pygfx/__init__.py
almarklein: Let's do this:
```suggestion
version_info = tuple(int(v) if v.isnumeric() else v for v in __version__.split("."))
```

pygfx/_version.py
```diff
@@ -0,0 +1,130 @@
...
+def get_version_from_git():
+    import subprocess
+
+    # git describe --first-parent does not take into account tags from branches
+    # that were merged-in. The '--long' flag gets us the 'post' version and
+    # git hash, '--always' returns the git hash even if there are no tags.
+    for opts in [["--first-parent"], []]:
+        try:
+            p = subprocess.Popen(
+                ["git", "describe", "--long", "--always", "--tags", "--dirty"] + opts,
```
almarklein, line 50: IIUC this means that if someone obtains pygfx in another way than via a package manager or git repo, the version cannot be established.

almarklein (commented):
I'm not overly enthusiastic 😅 but this may also be because I don't understand the problem that it solves. Let's discuss in #696

Korijn:
My general feeling about this is that it's adding a ton of code and complexity to solve a very small simple manual step... not too enthusiastic.

pygfx/_version.py
hmaarrfk: Correct. There become some "edge cases" that you have to start taking into account when people "receive" the source code. 

I just checked my project for example, the version downloaded from pypi burns things in due to the modifications to the setup.py but the archive from github would make it pretty difficult for people to use.

image

hmaarrfk:
> complexity and dependencies [...]  not too enthusiastic.

Understood. This is a fair criticism.

PR #730, opened 2024-04-15
Align and fix rendermask usage

almarklein:
Make all render magic calculations follow this pattern:
```py
        render_mask = 0
        if wobject.render_mask:
            render_mask = wobject.render_mask
        elif material.is_transparent:
            render_mask = RenderMask.transparent
        else:
            # Determine what passes are needed
            if material.something:
                render_mask |= xx
            else:
                render_mask |= yy
            if ... etc  
```

This makes these triages easier to follow, in part because we OR what passes are needed, and also because each shader follows the same pattern.

Fixes some cases for text and volume.

### DIFF
--- a/pygfx/materials/_text.py
+++ b/pygfx/materials/_text.py
@@ -92,7 +92,7 @@ class TextMaterial(Material):
         color = Color(color)
         self.uniform_buffer.data["color"] = color
         self.uniform_buffer.update_range(0, 1)
-        self._store.color_is_transparent = color < 1
+        self._store.color_is_transparent = color.a < 1
 
     @property
     def color_is_transparent(self):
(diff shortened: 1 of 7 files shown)


pygfx/materials/_text.py
```diff
@@ -92,11 +92,7 @@ def color(self, color):
         color = Color(color)
         self.uniform_buffer.data["color"] = color
         self.uniform_buffer.update_range(0, 1)
-        self._check_color_is_transparent()
-
-    def _check_color_is_transparent(self):
-        max_a = max(self.color.a, self.outline_color.a)
-        self._store.color_is_transparent = max_a < 0
```
almarklein, line 99: This `< 0` should have been `< 1` the whole time 😆

Korijn (approved):

PR #731, opened 2024-04-15
Make buttons and modifiers lists not tuples

hmaarrfk:
it seems that when these are set, they are lists not tuples.

I would prefer if they were tuples (immutable by the receiver of the event) but it seems that by the time it is received by convert_event, it is already a list:

https://github.com/pygfx/pygfx/blob/main/pygfx/renderers/wgpu/engine/renderer.py#L802

so the change to tuples would have to come from within wgpu.


Korijn:
So then why not open a PR on wgpu-py then? :)

@almarklein what do you think of the change?

almarklein:
I'd be in favor of them being tuples too. Would indeed best be solved by changes in `wgpu-py` for `glfw.py` and `qt.py`. Maybe also in `jupyter_rfb` for good measure.

almarklein:
Xref https://github.com/vispy/jupyter_rfb/pull/88

hmaarrfk:
> So then why not open a PR on wgpu-py then? :)

Not sure how big your user base is!

Well I tested the types in my application, seems like it "works".

almarklein (approved):

hmaarrfk:
> Not sure how big your user base is!

I guess if you are concerned about users "concatenating" to these the syntax to concatenate is different for both.

```python
event.modifiers + ["Shift",]
# vs 
event.modifiers + ("Shift",)
```

Thanks for the fast reviews!

Korijn:
> > Not sure how big your user base is!
> 
> I guess if you are concerned about users 

I don't really know how big the user base is, and I also don't think they are modifying events. Theoretically it's possible but theoretically they could also depend on internal APIs. :)

Korijn (approved):
Obviously this will depend on the new version of wgpu-py to land, but since it just a doc change we can merge

Issue #724, opened 2024-04-12
Transparency blending issue


almarklein:
I discovered this in #708, because markers with a transparent face and opaque edges produce the issue very well. But it can also be reproduced with normal points, of which some are made opaque black. The cyan lines and text in the below image are drawn on top as annotations. You'd expect the pink on the black circle to be equally dark as the pink on the background. But it's much brighter instead.

<img width="839" alt="image" src="https://github.com/pygfx/pygfx/assets/3015475/db5351c0-8235-4bd0-8db0-583efe7136f6">

hmaarrfk:
Haha. We struggle with this alot. Are these points all from the same material and geometry? Or are they distinct objects?

I ask because for distinct objects we've been able to get (more) reliable results by using z and the render order. 

Without both being set correctly, we would often be blending things with the background instead of the object that was placed behind it.

Korijn:
I wonder if this is related to #622

almarklein:
This is one points objects.

hmaarrfk:
Do you think it is also related to the AA artifacts 
I zoomed into the screenshot you took. It seems that some circles are blending with the background. 


![IMG_9301](https://github.com/pygfx/pygfx/assets/90008/24b1634f-0b89-43cc-b33c-b126526849c6)

almarklein:
I see three things going on:

* Colors blend differently with the bg than with black opaque objects. This is because we consider the framebuffer to be transparant, allowing to render rgba images. I think this is best solved by using a `gfx.Background` in examples that use transparency.
* My initial surprise was: why do black edges, when shown through a semi-transparent face become the color of the face, rather than staying black? This is actually the expected behavior; it perfectly follows the blend equation. The effect I was expecting falls in the category of advanced physical light simulation (is it called translucency?).
* The black edges due to aa. Not sure yet why they appear.

almarklein:
I think understand the dark edges too now: the faces of these dark disks/edges are opaque, so all transparent fragments that are between the face and that marker get blended. On the edges, however, aa makes the fragments semi-transparent, and therefore they get blended with the aforementioned transparent fragments, as well as the fragment that are *behind* the marker (up until an opaque fragment is encountered).

For `ordered2` this means that if the dark marker happened to be blended first, you'd get a dark edge. If its blended later, all is good. With `weighted` the dark semi-transparent fragment from the aa edge contribute to the transparency's combined color, making it darker. So it order-independently fails ;)

Issue #582, opened 2023-08-08
Responsibility of the Scene object


almarklein:
Right now, the `Scene` object is really just a `Group`. Although the `gfx.show()` function does behave differently depending on whether a `Scene` is given or a `Group`.

We could add more data to the scene that can then be shared by all objects in it. This possibility has been suggested/discussed in a fee places, both in issues and in meetings, so I figured we collect what we know here. Also, if we decide not to give the scene more responsibility, I think we should remove it and use `Group` instead.

Things that could be attached to a `Scene`:
* Background
  * As opposed to a background object that lives in the scene.
  * Counter argument: by making them objects, they can be customized, e.g. shaders that generate a horizon. 
* Clipping planes to apply to all object
  *  As opposed to making them objects that live in the scene.
  * Counter-argument: clipping planes very much have a position and orientation, so it kinda makes sense to have them in the scene?
* Ambient lights, perhaps
  * As opposed to a an `AmbientLight` object that lives in the scene.
  * Counter argument: all lights are world objects, so it'd be weird if the ambient light is not.
* Environment maps 
  * As opposed to a `MeshMaterial.env_map`.
  * Makes sense.
* Texture coords for light maps
  * As opposed to `mesh.geometry.texcoords1`, see #580.
  * Makes sense.

As can be seen, there is a lot that *can* be added to the `Scene`, but for many of them it's questionable whether its a good idea. For the things that remain, we have to decide whether it's worth it to make the `Scene` class heavier to support these cases. What is the cost for doing this?

panxinmiao:
I suggest keeping the Scene object, even if it exists as a special type of Group object. It should be distinguished from regular Group objects.

> ### Things that could be attached as a property to the `Scene`
> * Background

I agree with having it as an attribute of the Scene. Additionally, I think we could also consider keeping the `Background` object. The `Scene.background` attribute can be set as a `Background` object (customized), or it can be set as a regular color, image, or texture. When set as a color, image, or texture, we can encapsulate it as a `Background` object within the built-in rendering program.

> * Ambient lights, perhaps

I think it would be better to treat it as a standalone object that can be added to the scene, just like other lights.

> * Environment maps

I strongly agree because in many cases, the environment map can affect every object (or most objects) in the scene. However, MeshMaterial should also retain the `env_map property` and have a higher priority than `Scene.env_map`. Only when `MeshMaterial.env_map` is absent should `Scene.env_map` be used.

> * Light maps

[...]

PR #733, opened 2024-04-16
Add Background.from_color

almarklein:
## Intro

To avoid blending issues (as in #724) it should be recommended to use a background in most use-cases. But creating a simple uniform background is currently a bit verbose:
```py
scene.add(gfx.Background(None, gfx.BackgroundMaterial("#000")))
```
This is because the `Background` has the same geometry+material args as all world objects (except `Scene` and `Group`). I value that consistency, but it's also annoying to write out.

## Alternative options
One option could be to drop the `Geomety` arg. But that's still pretty long. And it creates a problem when at some point we get a new kind of background that *does* need geometry.
```py
scene.add(gfx.Background(gfx.BackgroundMaterial("#000")))
```

Another option is to allow something like this. I like the brevity, but documenting this is a bit awkward, I don't think this level of dynamic arguments works well here. Also what about a gradient?
```py
scene.add(gfx.Background("#000"))
```

So what I came up with is this. A bit longer than the previous, but in return its explicit and unambiguous.
```py
scene.add(gfx.Background.from_color("#000"))
```

## Tasks

* [x] Agreement on this proposal.
* [x] Update examples that currenly use a color background.

### DIFF
--- a/examples/feature_demo/fly_controller.py
+++ b/examples/feature_demo/fly_controller.py
@@ -18,6 +18,7 @@ import numpy as np
 canvas = WgpuCanvas()
 renderer = gfx.renderers.WgpuRenderer(canvas)
 scene = gfx.Scene()
+scene.add(gfx.Background.from_color("#000"))
 
 # Create a bunch of points
 n = 1000
--- a/examples/feature_demo/map_click_events_to_world.py
+++ b/examples/feature_demo/map_click_events_to_world.py
@@ -38,7 +38,7 @@ for i in range(4):
     scene = gfx.Scene()
     scenes.append(scene)
 
-    bg = gfx.Background(None, gfx.BackgroundMaterial(bg_colors[i]))
+    bg = gfx.Background.from_color(bg_colors[i])
     scene.add(bg)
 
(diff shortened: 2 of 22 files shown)


Korijn (dismissed):

panxinmiao:
How about adding a `background` property to the `scene`? 

Its value can accept `Color`, `Image`, `Texture(CubeTexture)` and `Background`, and pygfx will automatically handle the conversion.

With this, we can use the syntax `scene.background = (0, 0, 0)`

almarklein:
@panxinmiao That is an interesting idea, which is also discussed in #582 (and other issues linked therein). A decision about that has not been made yet. We want to iron out other parts of the API first. But for the time being we prefer to keep the `Scene` no more than a group object.

The current PR also does not exclude that path, if we want to go that direction later.

almarklein:
To add to that, I *really* like `scene.background = 'red'`, but if we go that path, we should consider other properties (lights, clipping planes, etc.) too, and that's a pretty profound API change.

Korijn (approved):
I'm not really a fan of scene.background... I prefer API consistency

PR #734, opened 2024-04-16
Add validation examples for blending

### LINKED ISSUE #724: Transparency blending issue

almarklein:
Closes #724

This adds to (validation) examples to demonstrate/document two cases from #724:
* The case that if you don't use a background, semi-transparent objects will look darker.
* The artifacts caused by aa edges under semi-transparent objects.
  * Occurs with `ordered2` when blend order is not in your favour.
  * Occurs with `weighted` always, and also with `weighted_plus` although slightly less obvious.

Requires #733


almarklein:
Docs issue fixed.

PR #736, opened 2024-04-19
Refactor baseshader

almarklein:
### Tasks

* [x] Refactor logic out of `base1.py`
* [x] Move wgsl in `base2.py` to a `.wgsl`  file.
* [x] Combine `base1.py` and `base2.py`.
* [x] Use `.wgsl` for shaderlib.
* [x] Implement a mechanism to include wgsl from other wgsl files. 
* [x] Check docstrings.
* [x] Check gallery and docs.  

### Notes

Most of our wgsl is now in `.wgsl` files. Some contains snippets/functions that can be imported into others using Jinja2's include mechanis (hooray jinja! 🚀 ), e.g. `{$ include 'pygfx.light_common.wgsl'$}`. This means that custom shaders, or shaders in downstream packages can *also* include our snippets. Even better, downstream, code can register a loader for a prefix so it can do e.g. `{$ include 'fastplotlib.special_stuff.wgsl'$}`.

### API changes

(for the release notes)

* Renamed `WorldObjectShader` to `BaseShader`.
* The `BaseShader` class no longer has `.code_definitions()` and `.code_common()`. Use `{$ include 'pygfx.std.wgsl' $}` instead.


pygfx/renderers/wgpu/shader/base.py
```diff
@@ -0,0 +1,285 @@
...
+          * "render_mask"
+          * "indices" (list of 2 or 4 ints).
+        """
+        return {
+            "indices": (1, 1),
+            "render_mask": 0,
+        }
+
+
+class BaseShader(ShaderInterface):
```
almarklein, line 88: This class is the combination of what was previously `BaseShader` and `WorldObjectShader`. Other changes to the code are mostly confined to the `generate_wgsl()` method.

pygfx/renderers/wgpu/engine/binding.py
```diff
@@ -0,0 +1,205 @@
...
+from .update import ensure_wgpu_object, ALTTEXFORMAT
+
+
+visibility_render = wgpu.ShaderStage.VERTEX | wgpu.ShaderStage.FRAGMENT
+visibility_all = (
+    wgpu.ShaderStage.VERTEX | wgpu.ShaderStage.FRAGMENT | wgpu.ShaderStage.COMPUTE
+)
+
+
+class Binding:
```
almarklein, line 17: `binding.py` was simply moved out of `pipeline.py`.

pygfx/renderers/wgpu/shader/bindings.py
```diff
@@ -0,0 +1,177 @@
...
+from ....resources import Buffer
+
+from ..engine.utils import (
+    to_vertex_format,
+    to_texture_format,
+    generate_uniform_struct,
+)
+
+
+class BindingDefinitions:
```
almarklein, line 13: This class was extracted out of the `BaseShader` class, for clarity.

pygfx/renderers/wgpu/shader/templating.py
```diff
@@ -0,0 +1,49 @@
...
+
+def register_wgsl_loader(context, func):
+    """Register a source for shader snippets.
+
+    When code is encountered that looks like::
+
+       {$ include 'some_context.name.wgsl' $}
+
+    The loader for "some_context" is looked up and used to load the wgsl to include.
+    This function allows registering a loader for your downstream package.
```
almarklein, line 25: cc @kushalkolar in case you guys will create your own shaders in the future.

almarklein (commented):
This PR does not do make a lot of changes in the strict sense, but it does move *a lot*  of code around. No wgsl code was changed. I added some comments to hopefully help review.

almarklein:
CI is having a bad day: fails at `sudo apt-get update -y -qq`.

almarklein:
When looking into jinja2's import mechanism, I also learned about its `import` keyword. It's a bit more advanced, but might be convenient for us: https://jinja.palletsprojects.com/en/3.1.x/templates/#import

panxinmiao:
> When looking into jinja2's import mechanism, I also learned about its `import` keyword. It's a bit more advanced, but might be convenient for us in some cases: https://jinja.palletsprojects.com/en/3.1.x/templates/#import

Recently, I also went through the Jinja documentation and found that Jinja supports macro definitions.
While I tried to extend texture support for pygfx (such as multiple UV channels, different texture mapping modes, etc.), I have been looking for an elegant way to implement dynamic shader code generation. The macro definition feature in Jinja seems can  provide great assistance for this. 
However, I am not very familiar with Jinja and am still learning and researching it.

almarklein:
The CI hiccup resolved itself, fortunately.

Korijn (approved):

PR #738, opened 2024-04-24
Color Conversion in PBR Example

panxinmiao:
I noticed that currently in the pygfx library, the values of Color are treated as being in the srgb color-space. When converting to the HSL or HSV color-spaces, the values need to be first converted to the linear-srgb space.


panxinmiao:
Uh, I made a mistake. All I need to do is modify the color space in the example.

Korijn (approved):

Korijn:
> Uh, I made a mistake. All I need to do is modify the color space in the example.

I thought so too but you and Almar are more knowledgeable on this topic so I was hesitant to immediately approve 😅

PR #742, opened 2024-04-25
Change event time_stamp unit to seconds

### LINKED ISSUE #741: Double click event triggers only once
Hi. I noticed that the double click event seems to only ever be triggered once.

Here's a minimal example to reproduce: you should see `"double_click"` printed to the console the first time you double click but not for any subsequent double clicks. 

```python
import pygfx as gfx

canvas = WgpuCanvas()
scene = gfx.Scene()

def animate():
    canvas.request_draw()

def trigger(event):
    print(event.type)

if __name__ == "__main__":
    renderer.add_event_handler(trigger, "double_click")
    canvas.request_draw(animate)
    run()
```

Tried with both `pygfx` `0.2.0` and current `main`.

panxinmiao:
Closes #741.

### DIFF
--- a/pygfx/objects/_events.py
+++ b/pygfx/objects/_events.py
@@ -1,13 +1,13 @@
 from collections import defaultdict
 from enum import Enum
-from time import perf_counter_ns
+from time import perf_counter
 from typing import Union
 from weakref import ref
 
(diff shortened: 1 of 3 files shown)


Korijn (approved):
Nice one!

Issue #739, opened 2024-04-24
Axes


almarklein:
## Intro

This is a meta-issue to collect thoughts about implementing axes.

Axes are important for downstream code like Fastplotlib. I think it makes sense to support the building blocks in pygfx itself, so that axes can be composed and configured with relative ease.

## Things that make up axes

Strict:

* Viewport (in 2D the axes is drawn over/around the scene.
* Lines for the axes.
* Ticks.
* Tick labels.
* Grid.

Somewhat broader:

* Title, x-label, y-label.
* Legend.
* Colorbar.
* UI elements.

## Things to solve or account for

* Tick spacing.
* Log scale.
* Polar scale.
* Map projection.
* Layout (i.e. positioning title and labels)
* Twin axes (Left and right side with diff y-scales)

## Links

Related issues and prs:

* #492 
* #683
* #582
* #118 

Techniques:

* Nicolas' shaders for infinite grids: https://jcgt.org/published/0003/04/01/

Implementations:

* [Axis in visvis](https://github.com/almarklein/visvis/blob/master/core/axises.py)
* [Axis in vispy](https://github.com/vispy/vispy/blob/main/vispy/visuals/axis.py)

----

Example from visvis (we can do better 😉):

<img width="539" alt="image" src="https://github.com/pygfx/pygfx/assets/3015475/10063cb1-da00-4fea-80a0-70e8bbb38557">

Example from vispy:

<img width="822" alt="image" src="https://github.com/pygfx/pygfx/assets/3015475/62d3376d-2493-4e9e-92cc-71dd2da2b9cd">

----

*I will update this text as I collect more info and converge on a plan.*

Korijn:
Do you also intend to support 3D rendering of grids (in world space with a perspective camera)?

And rendering grids and axes on deformed surfaces?

kushalkolar:
We've done some brainstorming on legends and a very primitive legend currently exists in fpl (most experimenting with things, we're far from happy with it): https://github.com/fastplotlib/fastplotlib/issues/403 

I can help out with legends later in the summer and contribute that to `pygfx` instead if that makes sense. Anyways I'm guessing the "Strict" things you pointed out would come first.

almarklein:
> Do you also intend to support 3D rendering of grids (in world space with a perspective camera)?

Yes, definitely that. And axii too (works in visvis).

> And rendering grids and axes on deformed surfaces?

[...]

PR #743, opened 2024-04-25
Grid object

almarklein:
*Relies on #736, will cherrypick/rebase once that is merged.*

Addresses one point in #739.

## Techniques

I'm looking (until now) at two approaches:
* The "Nico-grid" from [Nicolas Rougier's paper](https://jcgt.org/published/0003/04/01/).
* The Prestine-grid from [this post by Ben Golus](https://bgolus.medium.com/the-best-darn-grid-shader-yet-727f9278b9d8#5ef5) and [this shadertoy](https://www.shadertoy.com/view/mdVfWw).

I'm comparing these and will likely combine the best of both ... wip.

## Notes

* Grid range:
  * Sometimes you want a grid with a specific range. Sometimes infinite. 
  * This is more a matter of doing the right tricks in the vertex shader. Does not (really) matter for the grid-drawing.
* Line width:
  * Can be in screen coords, world coords. 
  * The Prestine grid mimics a texture grid, and is in world coords, but flat in the plane (not bill-boarded).
  * Can probably take code from line rendering to make it do whatever we want.
* Major and minor ticks:
  * Must be done in one shader; drawing two grids does not blend well.
  * The Nico-grid does this. Can probably make this work for Prestine grid.
* Performance:
  * Do a small benchmark to see if one is significantly slower. 
* AA:
  * super in both.
* Distance fading and suppression of Moiré artifacts:
  * Very good in the Prestine-grid. Not sure yet for the Nico-grid.

## Use-cases / scope

Focusing on grids for axes in this PR. Will also consider log, polar and map projections, to at least know how it will fit in.

Other use-cases (for later):

* Draw a surface (height map) and have planar grid lines or isosurface lines drawn on top.
* Draw gridlines over a mesh, the texcoords specify the grid space. Maybe a `MeshGridMaterial`.
* Render checkerboard instead of a grid.

----

Early result of the prestine grid:

image


Korijn:
> * Major and minor ticks:
>   
>   * Must be done in one shader; drawing two grids does not blend well.
>   * The Nico-grid does this. Can probably make this work for Prestine grid.

The prestine grid has this feature all the way at the bottom of the article, including a link to shader source:

image

almarklein:
Also easy to make prestine grid use screen-space line width.

kushalkolar:
That is indeed a very nice grid with minimal moire :open_mouth:

almarklein:
Almost done! Only Cartesian grids for now. Want to look at log/polar/maps later.

image

One problem is that the inf grid looks rather odd on LavaPipe ... 🤔 

image

Korijn:
Is the default orientation the same as a plane geometry? :)

almarklein:
> Is the default orientation the same as a plane geometry? :)

No, it's not. The default plane orientation is `xy`. I opted for a default `xz` because that way it can function as a floor in most our examples. I can change the default to match `xy`. We can force the user to specify the `orientation`, leaving less room for surprises.

Korijn:
> image

Something that stands out to me is that the infinite grid does not actually reaches out to the horizon :)

examples/validation/validate_grid1.py
```diff
@@ -0,0 +1,119 @@
...
+
+# An inf grid at the floor
+grid0 = gfx.Grid(
+    None,
+    gfx.GridMaterial(
+        thickness_space="world",
+        axis_thickness=0.1,
+        major_thickness=0.02,
+        minor_thickness=0.01,
+        inf_grid=True,
```
Korijn, line 30: How about just `infinite` as the argument name?

examples/validation/validate_grid1.py
```diff
@@ -0,0 +1,119 @@
...
+    ),
+    orientation="xz",
+)
+
+# A grid oriented in yz, showing major and minor ticks
+grid1 = gfx.Grid(
+    None,
+    gfx.GridMaterial(
+        major_step=1,
```
Korijn, line 39: Are the steps in thickness_space too?

almarklein:
> the infinite grid does not actually reach out to the horizon :) 

That's because it's not actually infinite, but scaled and positioned to appear infinite. If I scale the grid too much, the grid coords become unstable. But I'll have a look to see if I can crack it up a bit.

examples/validation/validate_grid1.py
almarklein: If the grid is aligned with a world axis, the grid' lines are always exactly on "whole steps" in world-space, regardless of scale or offset. See how the lines on the green and blue grid touch.

kushalkolar: I guess this seems to make sense right? For example, the grid lines would correspond to data in the positions buffer of a line?

Korijn:
So how does the prestine grid shader article achieve this effect then?

examples/validation/validate_grid1.py
Korijn: OK, that makes sense!

almarklein:
Updates:
* The horizon is much better now. The vertex shader now creates a square consisting of a small inner quad and a (much) larger outer region. That way we have accuracy nearby, and massive range for a proper horizon.
* Changed the grid's orientation to be in xy-plane if its not rotated. The `orientation` argument is mandatory.
* The glitched screenshot I posted earlier applied to Vulkan (not just Lavapipe). It was due to how the uv for axis-lines was clamped, Fixed.
* Also saw how the grid looked bad for large step sizes. Turned out the math was not correct. Fixed that too.

almarklein:
Ready from my end.

image

Korijn (approved):

kushalkolar:
Just played with it, it works really well! :partying_face:

almarklein:
Yeah, I'm really happy with the end result. The prestine grid is amazing! And mixing it with a few ideas from Nicolas (most notable handling edges of non-finite grids) makes it more generally useful.

kushalkolar:
10 million points and zooming along the grid :laughing: 

https://www.youtube.com/watch?v=7pv0zurv7x0

PR #746, opened 2024-04-28
Improve Skeleton's transform performance

panxinmiao:
This is a workaround to improve skeletal animation performance, serving as a temporary solution before we resolve the current performance issues with the Transform system. 

See: https://github.com/pygfx/pygfx/pull/715#issuecomment-2053385803
Since Bone objects tend to have complex parent-child relationship graphs in most scenarios, the current automatic recursive calculation of transforms has a significant performance impact on Bones and Skeletons.

This PR employs certain tricks to prevent the automatic recursive calculation of the matrix for Bone objects. 

We definitely need to address the performance issues of the current Transform system eventually. Once resolved, the contents of this PR can be removed, and the Bone objects will simply be regular blank WorldObjects.


panxinmiao:
Rebased now.
This is separated from #715.
The content of this PR can be safely removed once the performance issues of the Transform system are resolved.

Korijn (approved):

PR #751, opened 2024-05-07
Add morph targets support.

panxinmiao:
### Added Support for Morph Targets Animation

- **Geometry**:
  - `morph_attributes (dict)`: A dictionary of attributes containing details of the geometry's morph targets.
  - `morph_targets_relative (bool)`: Controls the behavior of morph targets. When set to True, the morph target data is treated as relative offsets rather than absolute positions/normals.

- **Mesh**:
  - `morph_target_influences (array-like)`: The weights, typically ranging from 0 to 1, that determine the extent to which each morph target is applied.

- **Additional**:
  - Morph target data is encoded into textures before rendering for use by shaders, improving performance.
  - The glTF loader now supports loading morph target data.

- **Examples**:
  - `morph_targets.py`
  - `morph_facecap.py`

For more details, please refer to the examples.

Example:
Video_2024-05-07_231033


panxinmiao:
Add a exampe to demonstrates how to animate a model with morph targets.

The model is a facecap model with morph targets for facial expressions (52 blend shapes, Apple ARkit).

Video_2024-05-08_173852

panxinmiao:
Ready from my end :)

panxinmiao:
Updated this PR, and merged the changes to `texture` in #795

pygfx/geometries/_base.py
```diff
@@ -82,6 +85,36 @@ def __init__(self, **kwargs):
             # Store
             setattr(self, name, resource)
 
+    @property
+    def morph_attributes(self):
```
almarklein, line 89: I am opposed to adding this property to the generic `Geometry` object, because it's rather specific, even for meshes, and it means nothing for e.g. a `Line` object.

I think there's plenty of options though:
1. A custom `MeshGeometry` or `MeshMorphGeometry`.
2. Just put the `morph_attributes` in the geometry in `load_gltf`. We do this for other data too, so why not for this.
3. Put `morph_positions`, `morph_normals`, and `morph_colors` separately on the geometry.

I think my preference would be 3, since it's transparent and flat.

almarklein (commented):
Very cool feature! I think it needs a bit of work to fit it in nicely, but we can make it work!

pygfx/geometries/_base.py
panxinmiao: > and it means nothing for e.g. a `Line` object

Morph targets are meaningful for `Line` and `Point` objects as well. Essentially, they record the 'vertex attribute transformation methods,' enabling vertex attribute deformation animations even for `Line` and `Point` objects.

I realize that perhaps I should put the shader logic related to morph targets into `BaseShader` rather than `MeshShader`. 😅 

Moreover, theoretically, any vertex attribute can change, not just position, color, and normal; for instance, texcoord can also vary. 
However, currently, we only support morph of position, color, and normal attributes, as these are the most common vertex deformations. so I think 3 is OK.

almarklein:
> Morph targets are meaningful for Line and Point objects as well. Essentially, they record the 'vertex attribute transformation methods,' enabling vertex attribute deformation animations even for Line and Point objects.

Fair point. Though I think we can treat this morph data similar to the other geometry attributes. Especially if we want to make this more generally available.

If I understand correctly, each morph attribute is a list of arrays with (either relative or absolute) values for the corresponding attribute. E.g. the positions morph data contains say 10 arrays that are each of equal size as the `geometry.positions.data`. They represent the keyframes that the shader interpolates to create motion. Is this correct?

I wonder whether we can simply put that data on the geometry in a buffer/texture, just like `geometry.positions` itself.

BTW: I worked with moving volumetric data during my PhD, I can see how such a thing can be supported via the same path 🚀

panxinmiao:
> If I understand correctly, each morph attribute is a list of arrays with (either relative or absolute) values for the corresponding attribute. E.g. the positions morph data contains say 10 arrays that are each of equal size as the `geometry.positions.data`. They represent the keyframes that the shader interpolates to create motion. Is this correct?

Yes, it‘s right.

> I wonder whether we can simply put that data on the geometry in a buffer/texture, just like geometry.positions itself.

We need to organize this data in some form on the CPU side. Typically, this data doesn't need to be accessed frequently once loaded.

A model can include many morph attributes; for example, the ARKit standard on iOS includes 52 blend shapes, while high-detail 3D models may have thousands. Storing each morph target attribute separately in a buffer isn't wise. Additionally, shader access and computation speed for these morph attributes' data should be considered.

Encoding morph targets data into morph textures (using vertex indices as texture coordinates) maximizes GPU parallel computing capabilities, speeding up vertex deformation and animation calculations. This is a widely used optimization technique.

examples/feature_demo/morph_facecap.py
```diff
@@ -0,0 +1,136 @@
...
+# sphinx_gallery_pygfx_docs = 'animate 4s'
+# sphinx_gallery_pygfx_test = 'run'
+
+import time
+import numpy as np
+import pygfx as gfx
+from scipy import interpolate
+from wgpu.gui.auto import WgpuCanvas, run
+
+gltf_path = model_dir / "facecap.glb"
```
almarklein, line 40: Do we need a license / attribution of the source?

almarklein:
I had another look (sorry it took so long).

The only thing standing in the way of a merge (IMO) is the geometry API.

The API of the geometry object is kinda weird in the first place, since we just set attributes and it's the responsibility of the user to use the correct names 😄  and we don't document them very well yet. Something to look into (perhaps documenting possible fields is sufficient). But for another pr.

Anyway, it feels a bit odd that for all the "normal" stuff we don't have properties, but we have for the morph stuff. So let's remove the properties, just set `geometry.morph_positions = ..`. And in the shader use `hasattr()` like we do with e.g. normals.

Probably needs a small change in `Geometry.__setattr__` to also exclude fields starting with "morph_".

Does this work for you?

panxinmiao:
>So let's remove the properties, just set `geometry.morph_positions = ..`. And in the shader use `hasattr()` like we do with e.g. normals.

Got it. I still have some local code changes that haven't been committed yet. I'll push them later :)

examples/feature_demo/morph_facecap.py
panxinmiao: The model is from [Face Cap](https://www.bannaflak.com/face-cap/documentation.html#1.5), and I've made some conversions and modifications. I'll add the attribution later.

pygfx/geometries/_base.py
```diff
@@ -83,13 +83,13 @@ def __init__(self, **kwargs):
             setattr(self, name, resource)
 
     def __setattr__(self, key, new_value):
-        if not key.startswith("_"):
+        if not key.startswith("_") and not key.startswith("morph_"):
```
almarklein, line 86:
```suggestion
        if not key.startswith(("_", "morph_")):
```

pygfx/geometries/_base.py
```diff
@@ -83,13 +83,13 @@ def __init__(self, **kwargs):
...
             if isinstance(new_value, Trackable) or key in self._store:
                 return setattr(self._store, key, new_value)
         object.__setattr__(self, key, new_value)
 
     def __getattribute__(self, key):
-        if not key.startswith("_"):
+        if not key.startswith("_") and not key.startswith("morph_"):
```
almarklein, line 92:
```suggestion
        if not key.startswith(("_", "morph_")):
```

panxinmiao:
Updated.

Also adjusted the rendering effects in the example and introduced ImGui to visually display the real-time changes in morph_target_influences during morph animations.

https://github.com/user-attachments/assets/ffd8704c-e331-4418-a28f-28a727886f50

The CI failed due to the introduction of the `imgui_bundle` library. Should we add `imgui_bundle` to the dependencies, or simply set `sphinx_gallery_pygfx_test = 'off'` to disable testing for this example?

almarklein (commented):
I made two suggestions for minor optimizations. Other than that, this looks good to go from my end!

almarklein:
> The CI failed due to the introduction of the imgui_bundle library. Should we add imgui_bundle to the dependencies, or simply set sphinx_gallery_pygfx_test = 'off' to disable testing for this example?

Let's add it to the `extras_require["examples"]` dependencies.

examples/feature_demo/morph_facecap.py
panxinmiao: Updated the document and added model attribution.

almarklein (approved):

Korijn (approved):

PR #752, opened 2024-05-07
feat: additive blending mode

### LINKED ISSUE #750: additive blending?
Apologies for what feels like a very basic question.  I've looked through the docs and source code and I'm struggling to find something like an additive blending mode (might just be semantics and I'm missing it?)

I'm using a `QWgpuCanvas` with a `WgpuRenderer`, and have added a few `pygfx.Image` to the scene, each with a `ImageBasicMaterial` (using `depth_test=False` for what it's worth) and each with a different colormap (e.g. Red, Green, and Blue).  What I'd like is for the colors in these three images to render to something similar to or the same as what I would get with a single RGB image (but for my application I want to retain individual addressable images with independent colormaps).  I've tried `renderer.blend_mode`: "opaque", "ordered1", "ordered2", "weighted", "weighted_depth", "weighted_plus"... and I do get something that appears sort of additive with the `weighted` variants.  but it's "muted" compared to what I'd expect (i.e. R,G,B don't seem to add up to white; they saturate at something around #9C9C9C)

thanks!

tlambert03:
closes #750 

@almarklein, i can confirm that this behaves as expected for my application... but I'm not that confident in the details, so please have a look and let me know if I should change anything.  Also, let me know how/where to add a test if you'd like me to add one


almarklein:
Very nice!

A validation example (see examples/validation/) is probably the easiest way to make sure it keeps working, and also provides an example for interested users.

pygfx/renderers/wgpu/engine/blender.py
```diff
@@ -911,3 +911,74 @@ def _create_combination_bind_group(self, bind_group_layout):
...
+        }
+
+    def get_shader_code(self, blender):
+        return """
+        struct FragmentOutput {
+            @location(0) color: vec4<f32>,
+        };
+        fn get_fragment_output(depth: f32, color: vec4<f32>) -> FragmentOutput {
+            var out : FragmentOutput;
+            out.color = color;
```
tlambert03, line 975: i did have one question about alpha, here and in the line `"alpha": (bf.one, bf.one, bo.add),`.  I'm aware that there are varying strategies for adding alpha (pre-multiplied, etc).  Is the current implementation what you would expect for a "standard" additive blending mode?  should i handle alpha differently, and/or be more specific in the naming of the blending mode?

tlambert03:
ok, added an example:

Screenshot 2024-05-08 at 10 02 47 AM

hmaarrfk:
A question. Is the idea here that your whole scene is blended "additively"? Is it possible to in general mix and match? maybe have negative alpha?

tlambert03:
Yep, exactly. The final fragment color is simply the linear addition of all other fragments.  

> Is it possible to in general mix and match? maybe have negative alpha?

I don't believe negative alphas are supported in general.  They'll be coerced to 0-1 [here](https://github.com/pygfx/pygfx/blob/41aff5b5f7ba51762e564906593c7eb22fe60c74/pygfx/utils/color.py#L147)

examples/validation/validate_blend_additive.py
```diff
@@ -0,0 +1,41 @@
+"""
+Additive Blending
+===============
+
+This example draws overlapping circles of reference colors with additive blending.
+"""
+
```
almarklein, line 7: Let's also add it to the docs, and compare the screenshot on CI 😄 
```suggestion

# sphinx_gallery_pygfx_docs = 'screenshot'
# sphinx_gallery_pygfx_test = 'compare'

```

pygfx/renderers/wgpu/engine/blender.py
almarklein: ### Context

In e.g. the transparent pass, the color is premultiplied with the alpha:
```py
out.color = vec4<f32>(color.rgb * color.a, color.a);
```
With the blend function ...
```py
"blend": {
                    "alpha": (bf.one, bf.one_minus_src_alpha, bo.add),
                    "color": (bf.one, bf.one_minus_src_alpha, bo.add),
                },
```
... this results in a color that is `src_a * src_rgb + (1-src_a) * target_rgb`. I can't remember exactly why I did it this way, ad not use `(bf.src_alpha, bf.one_minus_src_alpha, bo.add)`, but one reason could be thatit results in the blend-functions for alpha and color to be equal.

### Back to your question

I haven't done anything with additive blending so I'm not sure what the expected behavior is. But I *think* the idea is that it all just adds up, so the current approach is correct.

examples/validation/validate_blend_additive.py
tlambert03: thanks1

examples/validation/validate_blend_additive.py
```diff
@@ -0,0 +1,44 @@
+"""
+Additive Blending
+===============
```
almarklein, line 3:
```suggestion
Additive Blending
=================
```

almarklein: fixes docs build

almarklein (approved):
I added the screenshot generated by the screenshots build.

Korijn (approved):

tlambert03:
what is your typical release cycle?  could this be released in the near future?   thanks :)

almarklein:
Godo question! I'll try to do a release this week.

almarklein:
https://github.com/pygfx/pygfx/releases/tag/v0.3.0

Issue #745, opened 2024-04-26
Selection helper


schlegelp:
I originally had added this as a comment to #739 but it felt a bit too off topic. Hence now in a separate issue.

I've been working on a simple selection helper (mimicking the implementation of the `TransformGizmo`) for another project. It needs a bit of cleaning-up but I thought something like this might be useful to have in `pygfx` itself - even if just as demo for how to do it. Would you consider a PR?

https://github.com/pygfx/pygfx/assets/7161148/5021cedd-9bb8-40bd-b058-02caa8ead356

kushalkolar:
This looks very nice! Curious to see how you implemented it, in fastplotlib we'll also be working on more selection tools soon.

almarklein:
Very nice! I can definitely see this as an example. And we can also consider moving (some of) it as a helper class in `pygfx`.

PR #753, opened 2024-05-08
Add a Selection Gizmo

schlegelp:
This is a draft PR for adding a selection gizmo (see #745). 

```bash
python3 examples/feature_demo/selection_helper.py
```

https://github.com/pygfx/pygfx/assets/7161148/989ca670-7eff-4f5f-b9d2-a7692bf52ce1

Potential TODOs:
- [ ] check if this works in 3D 
- [ ] explore the possibility of a 3D selection gizmo (i.e. a box where we need to click twice)
- [ ] lasso selection gizmo (this would be fairly easy)
- [ ] allow multiple modifiers for starting the drag (e.g. shift-control click to start dragging)
- [ ] handle WorldObjects in `SelectionGizmo.is_inside` 

I'm aware that this might be out of the scope for `pygfx` itself as you may not want to weigh yourself down with having to maintain too many non-essential bits and pieces. If that's the case, no hard feeling! Also no worries if this implementation is going in the wrong direction.

### DIFF
--- a/pygfx/helpers/_selection.py
+++ b/pygfx/helpers/_selection.py
@@ -1,0 +1,457 @@
+"""
+A selection gizmo to select world objects.
+"""
+
+import numpy as np
+import pylinalg as la
+import pygfx as gfx
+
+from pygfx.objects import WorldObject
+from pygfx.utils.viewport import Viewport
+from pygfx.utils.transform import AffineTransform
+
+
+class SelectionGizmo(WorldObject):
+    """Gizmo to draw a Selection Box.
+
+    To invoke the Gizmo:
+
+    * Shift-click on the canvas and start dragging to draw a selection rectangle
+
+    Parameters
+    ----------
+    renderer : Renderer | Viewport
+        The renderer or viewport to use for the gizmo.
+    camera : Camera
+        The camera to use for the gizmo.
+    scene : Scene, optional
+        The scene to render the gizmo in. This can also be a Group.
+    callback_after_drag : callable, optional
+        A callback function to call after the selection is completed.
+    callback_during_drag : callable, optional
+        A callback function to call during the selection, i.e. as we
+        drag the selection box.
+    modifier : str, optional
+        The modifier key to use to activate the gizmo. Default "Shift".
+    edge_color : str | tuple, optional
+        The color to use for the edge of the selection box.
+    fill_color : str | tuple, optional
+        The color to use for the fill of the selection box. Set to
+        None (default) to disable the fill.
+    line_width : float, optional
+        The width of the selection box outlines.
+    line_style : "solid" | "dashed" | "dotted", optional
+        The line style to use for the selection box outlines.
+    force_square : bool, optional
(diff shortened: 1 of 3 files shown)


pygfx/helpers/_selection.py
```diff
@@ -0,0 +1,457 @@
...
+        if event.type != "before_render":
+            return
+
+        if self._viewport and self._camera and self._active:
+            self._update_ndc_screen_transform()
+
+    def _update_ndc_screen_transform(self):
+        # Note: screen origin is at top left corner of NDC with Y-axis pointing down
+        x_dim, y_dim = self._viewport.logical_size
+        screen_space = AffineTransform()
```
kushalkolar, line 221: I did not realize this exists, I have been doing it manually with `pylinalg` all this while :joy:

kushalkolar: nevermind I thought this does the vector transform

pygfx/helpers/_selection.py
```diff
@@ -0,0 +1,457 @@
...
+        screen_space.position = (-1, 1, 0)
+        screen_space.scale = (2 / x_dim, -2 / y_dim, 1)
+        self._ndc_to_screen = screen_space.inverse_matrix
+        self._screen_to_ndc = screen_space.matrix
+
+    def add_default_event_handlers(self):
```
kushalkolar, line 227: I think a `clear_handlers` method would help with garbage collection

pygfx/helpers/_selection.py
```diff
@@ -0,0 +1,457 @@
...
+        """Register Gizmo callbacks."""
+        self._viewport.renderer.add_event_handler(
+            self.process_event,
+            "pointer_down",
+            "pointer_move",
+            "pointer_up",
+        )
+        # Not sure we actually need to update the gizmo during rendering
```
kushalkolar, line 235: true I'm a bit confused by this as well

kushalkolar:
What do you think about drawing a rectangle (or arbitrary shape with lasso) and allowing it to stay and then be moved around?

schlegelp:
> What do you think about drawing a rectangle (or arbitrary shape with lasso) and allowing it to stay in the scene so it can be moved around? (That could be another selection tool, I'm just to think if it'd be easy to do here 🤔 )

Certainly possible. I've added it to the potential TODOs.

kushalkolar:
A few other things I can think of are resize handlers and a circular/elliptical selection tool. Anyways we can chat more on Friday. This actually helps us a lot with the fastplotlib roadmap!

kushalkolar:
Hi, I don't want to be blocking this PR if you guys want to merge it!

sharon92:
this is so cool!

My implementation in 3D view looks like this:

https://github.com/user-attachments/assets/7623dfa0-cf24-40d2-8600-427894aa066f

I am drawing the selection area by using the nearest object in my scene.

It would be great to have your implementation worked out, so I could get rid of mine :D

PR #756, opened 2024-05-15
KeyframeTrack and Interpolants

panxinmiao:
This Pull Request introduces keyframe animation functionality and implements a fast interpolation algorithm.

This "Interpolant" based on binary search and caching mechanism can effectively improve the efficiency of interpolation lookups.

- For linear access crossing at most 2 intervals, the time complexity is O(1);
- For random access cases, the time complexity is O(log2(n)), where n is the number of positions.

Meanwhile, following the glTF standard, four specific interpolation methods are implemented:
See: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#appendix-c-interpolation

- Linear Interpolation
- Spherical Linear Interpolant
- Step Interpolation
- Cubic Spline Interpolation

The "skinning_animation.py" example and "morph_facecap.py“example have been updated to use the new keyframe interpolation algorithm, which can provide very smooth animation effects.

Additionally, with this Pull Request, the frame rate of the "skinning_animation.py" example on my computer has improved from 80fps to 150+fps; while the frame rate of the "morph_facecap.py“ example can easily reach 500+fps.

Video_2024-05-15_141511
image


panxinmiao:
The example runs perfectly fine, but the CI tests fail, indicating a FloatinPointError. I can also reproduce this issue locally using "pytest -v examples". 
image

Does anyone know what might be causing this? @almarklein @Korijn

panxinmiao:
Using `np.seterr(under="ignore")` temporarily ignores this exception to pass the CI tests (not sure if it's appropriate, 😅 ).

This PR aims to implement an interpolation algorithm for keyframe animation to fully and accurately present the original effects of GLTF model animation data. However, to facilitate demonstration, I've included a very simple animation system (and objects).
﻿
I also have some additional code related to the animation system (like the animation mixer) that hasn’t been submitted yet, which still needs more organization and consideration. To make code review easier, I prefer not to expand the scope of this PR, I plan to continue developing the animation system in separate PRs.

almarklein (dismissed):
Awesome!

almarklein:
I clicked the "Update branch" button, because I wondered whether that would work, given the many formatting changes applied recently. But it worked!

Korijn (dismissed):

examples/feature_demo/gltf_animations.py
```diff
@@ -0,0 +1,153 @@
...
+import pygfx as gfx
+from wgpu.gui.auto import WgpuCanvas, run
+
+from wgpu.utils.imgui import ImguiRenderer
+from imgui_bundle import imgui
+
+import numpy as np
+
+# Ignore underflow runtime warning for pytest, See: https://github.com/pygfx/pygfx/pull/756#issuecomment-2370715504
+np.seterr(under="ignore")
```
hmaarrfk, line 45: i think the error is due to numpy's new strictness in numpy2. 

can we avoid this kind of hackery. i'm happy to give it a look this weekend to see how we can get around the error.

hmaarrfk: We are dealing with this kind of new error in our codebase due to overflows with `uint8`. numpy used to let you just "
```
a = np.zeros(3, dtype='uint8')
a[:] = 0, 128, 256
```

happily, but now will raise warnings and or error.

hmaarrfk: Ok it seems that `np.seterr` is enabled by the test suite:

https://github.com/pygfx/pygfx/blob/main/conftest.py#L37

almarklein: My proposal is to change the line in conftest because there simply are cases where we need to cast float64 to a float32 for the GPU:
```py
np.seterr(all="raise", under="ignore")
```
And then we can remove the `np.seterr()` line in this example.

Makes sense @hmaarrfk?

hmaarrfk: I don’t really follow. But I feel like you understand so that’s good enough for me

almarklein: From what I understand, the code in this PR triggers a floatingpoint underflow in Numpy, because we need to cast to float32 in a few places. I think you're right that this is a general problem that this pr just happens to trigger. Except its not really a problem, but just a fact (GPU's want float32) we need to live with.

So my suggestion is to change the `np.seterr()` call to not trip over these underflows.

hmaarrfk: Yes. That makes sense

Korijn: If there are only a few specific spots where underflows happen, I would prefer to mark those spots with `with np.seterr(...)` locally so that we explicitly allow underflow in those parts of the codebase.

It's very useful to have this check in the test suite because _silently happening_ numerical errors can make it very difficult to trace a problem to its origin.

hmaarrfk: The problem is every transform poses the risk of underflow.

almarklein: We could mark mark the spots where a transform is being stored into a uniform buffer (this is where he cast to float32 happens):
```py
 @callback
    def _update_uniform_buffers(self, transform: AffineBase):
        orig_err_setting = np.seterr(under="ignore")
        self.uniform_buffer.data["world_transform"] = transform.matrix.T
        self.uniform_buffer.data["world_transform_inv"] = transform.inverse_matrix.T
        self.uniform_buffer.update_full()
        np.seterr(**orig_err_setting)
```

There's (at least) one other spot; where the camera transform is set in the std uniform, but for now, the problem is with the transform of the world objects.

panxinmiao:
I just returned from my vacation and took some time to investigate this issue in depth.

There are two points I’ve found a bit puzzling.

The first is that the underflow occurs in the `AffineTransform` callback function, which should happen when setting the Transform (position, rotation, scale) of a `WorldObject`. However, in the skeletal animation demonstrated in the example, there is no transformation of WorldObjects; only the Transform updates of Bone objects are involved. Due to PR #746 , Bone objects have their own special logic and do not use AffineTransform, so this code should not be called, which makes it quite strange. 

I traced the code and found that there are indeed some non-bone WorldObject nodes in the animation track. They exist in the glTF model file and the animation data, but not in the scene. I suspect these nodes might be rigid body objects used for the physics engine or control endpoints for IK. The issue is that these objects have not actually been added to the scene, so their matrix calculations are completely unnecessary. Currently, our matrix updates occur immediately (when setting position, rotation, or scale) rather than being based on the scene graph before rendering, which leads to this problem. Typically, in most engines, these rigid bodies are not added to the scene graph, so their matrix updates are generally managed by the physics engine instead of the renderer.

The second point is that I am confident the glTF loader and keyframe interpolator use float32 data types, and the outputs are also float32. I had traced the code again to confirm this. So, how does the conversion from float64 to float32 occur?
Upon further investigation, I found instances in our code where ndarray is initialized with `dtype=float` ,

For example:
https://github.com/pygfx/pygfx/blob/94e59391c521dc1212e881b04cb547df45e67bc3/pygfx/utils/transform.py#L564-L565

I believe that 64-bit precision is unnecessary, and we should perhaps standardize on using `dtype=np.float32` to initialize them.

almarklein:
> I believe that 64-bit precision is unnecessary, and we should perhaps standardize on using dtype=np.float32 to initialize them.

I don't expect it will matter much in performance, but using float64 will likely avoid a range of issues concerning float32 precision.


> The issue is that these objects are not actually added to the scene, making any matrix calculations involving them unnecessary. However, our matrix updates occur immediately (when setting position, rotation, or scale) rather than being based on the scene graph before rendering, which leads to this problem.

This is an interesting problem. Something to consider when we attack the problem of the transforms being slow. 

For now I hope we can find a quick solution that is good enough to wrap up this PR :)

panxinmiao:
> I don't expect it will matter much in performance, but using float64 will likely avoid a range of issues concerning float32 precision.

Yes, that's what I meant. And I think 32-bit precision is sufficient because data is also computed with 32-bit precision in GPUs. Unifying internal data types in Pygfx should also have some benefits.

> For now I hope we can find a quick solution that is good enough to wrap up this PR :)

I have updated the code as per your comments [here](https://github.com/pygfx/pygfx/pull/756#discussion_r1785741410).

The investigation above was only meant to understand the root cause of the issue and is not related to the main content of this PR, 😅

almarklein (approved):

Korijn (approved):

PR #762, opened 2024-05-21
Tweak wgpu abstractions

### LINKED ISSUE #272: Logic in materials is based on details about the wgpu renderer

### LINKED ISSUE #574: API for specifying buffer and texture wgpu usage
Specific use-cases, perhaps using custom shaders, may require buffers or textures to have additional usage flags, see e.g. #567. There is  `resource._wgpu_usage`, which is a private flag intended for internal usage. We *could* expose it, but that would add wgpu- stuff into our main API, which we're trying to avoid ... 🤔 (there's a few more places where this happens)

xref #272

almarklein:
Closes #272, closes #574.

### DIFF
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -98,6 +98,11 @@ class Material(Trackable):
 
     @property
     def uniform_buffer(self):
+        """The uniform buffer object for this material.
+
+        Properties that map are represented in the buffer can be updated cheaply
+        (i.e. without requiring shader compilation).
+        """
         return self._store.uniform_buffer
 
     @property
--- a/pygfx/materials/_base.py
+++ b/pygfx/materials/_base.py
@@ -27,6 +27,13 @@ class Material(Trackable):
         against the depth buffer and also not writing to it.
     """
 
+    # Note that in the material classes we define what properties are stored as
+    # a uniform, and which are stored on `._store` (and will likely be used as
+    # shader templating variables). This can be seen as an abstraction leak, but
+    # abstracing it away is inpractical and complex. So we embrace wgpu as the
+    # primary rendering backend, and other backends should deal with it.
+    # See https://github.com/pygfx/pygfx/issues/272
+
     uniform_type = dict(
         opacity="f4",
         clipping_planes="0*4xf4",  # array<vec4<f32>,3>
--- a/pygfx/materials/_mesh.py
+++ b/pygfx/materials/_mesh.py
@@ -3,7 +3,7 @@ from ._base import Material
 from ..resources import Texture
 from ..utils import logger
 from ..utils.color import Color
-from ..utils.enums import ColorMode
+from ..utils.enums import ColorMode, VisibleSide
 
 
 class MeshAbstractMaterial(Material):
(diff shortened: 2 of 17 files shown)


pygfx/materials/_base.py
```diff
@@ -91,6 +98,11 @@ def _wgpu_get_pick_info(self, pick_value):
 
     @property
     def uniform_buffer(self):
+        """The uniform buffer object for this material.
+
+        Properties that map are represented in the buffer can be updated cheaply
```
Korijn, line 103: "Map" is a typo here I think?

pygfx/materials/_mesh.py
```diff
@@ -155,11 +152,12 @@ def side(self):
...
-        if side in ("FRONT", "BACK", "BOTH"):
-            self._store.side = side
-        else:
-            raise ValueError(f"Unexpected side: '{value}'")
+        value = (value or "both").lower()
+        if value not in VisibleSide:
+            raise ValueError(
+                f"MeshMaterial.side must be a string in {VisibleSide}, not {repr(value)}"
+            )
+        self._store.color_mode = value
```
Korijn, line 160: color_mode?

pygfx/resources/_texture.py
```diff
@@ -147,12 +149,18 @@ def size(self):
...
+        """The texture format.
+
+        Usually a pygfx format specifier (e.g. 'u2' for scalar uint16, or '3xf4'
+        for RGB float32), but can also be a value from ``wgpu.TextureFormat``.
         """
         return self._format
 
+    @property
+    def usage(self):
+        """Bitmask indicating how the texture            can be used in a wgpu pipeline."""
```
Korijn, line 161: Missing a newline?

pygfx/utils/enums.py
```diff
@@ -141,4 +151,35 @@ class MarkerShape(Enum):
...
+class VisibleSide(Enum):
+    """The VisibleSide enum specifies what side of a mesh is visible.
+
+    Note that this is the inverse of the "CullMode", as it specifies what
+    side is visible rather than what side is culled.
+    """
+
+    front = None  #: The front is visible.
+    back = None  #: The back is visible.
+    Both = None  #: Both the front and back are visible.
```
Korijn, line 182: Why is both capitalized?

pygfx/materials/_base.py
almarklein: yes. Fixed.

pygfx/materials/_mesh.py
almarklein: Nice catch!

pygfx/utils/enums.py
almarklein: typo

Korijn (approved):

PR #765, opened 2024-05-24
Audio visualization

panxinmiao:
An Example of Audio Visualization.

This PR implements a simple audio analyzer based on W3C standards and demonstrates how to create an audio visualization application.
See: https://www.w3.org/TR/webaudio/#AnalyserNode


almarklein:
Nice! A smaller sound file might be nice though 😅

panxinmiao:
> Nice! A smaller sound file might be nice though 😅

To prevent large binary files from being included in the PR, this example has been modified to use a network address for music playback. To avoid the long startup delay caused by downloading the entire file before playing, I have implemented a simple streaming solution that allows downloading and playing simultaneously. This functionality works well on my computer, but I'm unsure about the experience under poor network conditions. Achieving a better experience in such scenarios might require estimating the buffer size based on network speed and remaining playback time, though I don't think this is crucial for this example.

Additionally, this example also supports playing local files and loading network addresses in a non-streaming manner ( downloading the entire file before playing). You can achieve this by simply modifying the parameters in the example.

almarklein (approved):

Korijn (approved):

PR #769, opened 2024-05-30
Measure gpu time for all render passes

almarklein:
This adds support to measure the GPU time for the different render passes, which can be useful in benchmarks to see how much strain each pass in the rendering process puts on the GPU, and how certain choices affect the GPU usage. This feature is opt-in, one must set `renderer.measure_gpu_times = True`. This flag is undocumented/experimental for now.

Note1: This relies on https://github.com/pygfx/wgpu-py/pull/505, which is not yet released. But IMO that's ok, since for now this feature is undocumented and only used by ourselves (meaning devs who checkout the `main`of wgpu-py).

Note2: Using this feature requires `gfx.renderers.wgpu.enable_wgpu_features("timestamp-query")`.

This PR:
* [x] Adds a `GpuTimeMeasurer` to abstract away the details, so that the rendering code has minimal extra stuff to do the measurements.
* [x] The renderer creates an instance for each `render()` call, and uses it to measure the render passes, blender combine pass (if applicable), flush pass, and shadow passes.
* [x] The renderer then publishes the result in a new `stats` prop at `renderer.stats["gpu_times"]`.  

An example output from the benchmark code I'm working on. You can see how the number of measurements changes depending on the presence of shadows and used blend mode.
```      
       weighthed_plus blending - cpu:  6.0  bcombine:  0.5  flush:  1.6  pass1:  0.2  pass2:  0.3  pass3:  0.3
                        shadow - cpu:  7.6  flush:  1.4  pass1:  0.2  pass2:  0.2  shadow:  0.4
```

### DIFF
--- a/pygfx/renderers/wgpu/engine/blender.py
+++ b/pygfx/renderers/wgpu/engine/blender.py
@@ -606,7 +606,7 @@ class BaseFragmentBlender:
         """Get the number of passes for this blender."""
         return len(self.passes)
 
-    def perform_combine_pass(self):
+    def perform_combine_pass(self, time_measurer):
         """Perform a render-pass to combine any multi-pass results, if needed."""
 
         # Get bindgroup and pipeline. The creation should only happens once per blender lifetime.
@@ -623,6 +623,11 @@ class BaseFragmentBlender:
 
         command_encoder = self.device.create_command_encoder()
 
+        timestamp_writes = None
+        if time_measurer:
+            time_group = time_measurer.create_group(self.device, "bcombine", 1)
+            timestamp_writes = time_group.get_timestamp_writes(0)
+
         # Render
         render_pass = command_encoder.begin_render_pass(
             color_attachments=[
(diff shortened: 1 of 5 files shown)


almarklein:
Interesting ... in a benchmark that renders 1M points, the passes (obviously) take more gpu time, but the flush-pass also reports a much larger time. In other words, the measured gpu-time is affected by previous render passes.

Nice reallity-check that the GPU and CPU operate asynchronously: the CPU submits work, and the GPU then does it. The longer time for the flush-pass means just that all passes where submitted in quick succession, but the flush pass finishes later, because the render-pass took longer ...

In practice, we sync the CPU with the GPU (i.e. the CPU waits for the GPU to finish) when we read from the buffer containing the timestamps. 

There's a lot to gain by using async code instead ... Also, using start+end times (instead of their just difference) can help us create graphs that show where CPU and GPU are doing work!! Handy when we work on #495.

almarklein:
Moving back to draft. Don't need this right now for benchmarks. But these timestamps are very useful when we look at scheduling cpu vs gpu work, see #495

PR #773, opened 2024-06-01
Add isosurface volume material

kevinyamauchi:
As discussed in #86, this PR adds a material for volume iso surfaces. In this PR, I aim to add the iso surface shader. I propose to add lighting that is integrated with the scene lighting system in a follow-up PR. 

https://github.com/pygfx/pygfx/assets/1120672/24e12f1e-8d5c-45ef-8716-e83663e9fdf3

To do:

- [x] expose surface threshold as a uniform
- [ ] add some simple lighting


kevinyamauchi:
https://github.com/pygfx/pygfx/assets/1120672/99ddc36d-9919-4070-a7b2-c71b19116fc9

<details>
  <summary>Example script</summary>
  
```python
import imageio.v3 as iio
import numpy as np
from PySide6 import QtWidgets
from wgpu.gui.qt import WgpuCanvas

import pygfx as gfx


class Main(QtWidgets.QWidget):
    def __init__(self):
        super().__init__(None)
        self.resize(640, 480)

        # Creat button and hook it up
        initial_threshold = 1000
        self._slider = QtWidgets.QSlider(parent=self)
        self._slider.setMinimum(0)
        self._slider.setMaximum(2000)
        self._slider.setValue(initial_threshold)
        self._slider.valueChanged.connect(self._on_slider_changed)

        # Create canvas, renderer and a scene object
        self._canvas = WgpuCanvas(parent=self)
        self._renderer = gfx.WgpuRenderer(self._canvas)
        self._scene = gfx.Scene()

        voldata = iio.imread("imageio:stent.npz").astype(np.float32)

        geometry = gfx.Geometry(grid=voldata)
        material = gfx.VolumeIsoMaterial(clim=(0, 2000), isosurface_threshold=initial_threshold)

        self.volume =  gfx.Volume(geometry, material)
        self._scene.add(self.volume)

        self._camera = gfx.PerspectiveCamera(70, 16 / 9)
        self._camera.show_object(self._scene, view_dir=(-1, -1, -1), up=(0, 0, 1))

        self._controller = gfx.OrbitController(
            self._camera, register_events=self._renderer
        )

        # Hook up the animate callback
        self._canvas.request_draw(self.animate)

        layout = QtWidgets.QHBoxLayout()
        self.setLayout(layout)
        layout.addWidget(self._slider)
        layout.addWidget(self._canvas)

    def _on_slider_changed(self):
        self.volume.material.isosurface_threshold = self._slider.value()

    def animate(self):
        self._renderer.render(self._scene, self._camera)
        self._canvas.request_draw()


app = QtWidgets.QApplication([])
m = Main()
m.show()


if __name__ == "__main__":
    app.exec()

```
</details>

^ demo of changing the isosurface threshold with a slider

### isosurface threshold uniform
I have attempted to make the isosurface threshold a uniform and I'm not sure how to do so with the current setup where the different volume render snippets (e.g., MIP, isosurface) are in the same file. When I do so, this causes issues for the other volume ray modes when building the shader because they don't have the `isosurface_threshold` field. See the error message below caused by running the `volume_render1.py` example.

```
Device::create_shader_module error: 
Shader '' parsing error: invalid field accessor `isosurface_threshold`
    ┌─ wgsl:451:36
    │
451 │     let iso_threshold = u_material.isosurface_threshold;
    │                                    ^^^^^^^^^^^^^^^^^^^^ invalid accessor


Draw error
  File "/Users/kyamauch/miniconda3/envs/pygfx/lib/python3.10/site-packages/wgpu/gui/_gui_utils.py", line 77, in log_exception
    yield
  File "/Users/kyamauch/miniconda3/envs/pygfx/lib/python3.10/site-packages/wgpu/gui/base.py", line 135, in _draw_frame_and_present
    self.draw_frame()
  File "/Users/kyamauch/Documents/pygfx/examples/feature_demo/volume_render1.py", line 56, in animate
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/renderer.py", line 511, in render
    container_group = get_pipeline_container_group(wobject, environment)
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 159, in get_pipeline_container_group
    pipeline_container_group.update(wobject, environment, changed_labels)
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 239, in update
    container.update(wobject, environment, env_hash, changed)
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 324, in update
    raise err
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 321, in update
    self.update_wgpu_data(environment, env_hash, changed)
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 392, in update_wgpu_data
    env_shaders[pass_index] = self._compile_shader(pass_index, environment)
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 591, in _compile_shader
    return get_cached_shader_module(self.device, self.shader, shader_kwargs)
  File "/Users/kyamauch/Documents/pygfx/pygfx/renderers/wgpu/engine/pipeline.py", line 72, in get_cached_shader_module
    result = device.create_shader_module(code=wgsl)
  File "/Users/kyamauch/miniconda3/envs/pygfx/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_api.py", line 1349, in create_shader_module
    id = libf.wgpuDeviceCreateShaderModule(self._internal, struct)
  File "/Users/kyamauch/miniconda3/envs/pygfx/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_helpers.py", line 322, in proxy_func
    raise wgpu_error  # the frame above is more interesting ↑↑

Caused by:
    In wgpuDeviceCreateShaderModule

Shader '' parsing error: invalid field accessor `isosurface_threshold`
    ┌─ wgsl:451:36
    │
451 │     let iso_threshold = u_material.isosurface_threshold;
    │                                    ^^^^^^^^^^^^^^^^^^^^ invalid accessor


    invalid field accessor `isosurface_threshold`
```


What do you think is the best path forward? Move the isosurface rendering shader from the `volume_ray.wgsl` to its own file? I suppose we could also add the `isosurface_threshold` uniform to the `VolumeBasicMaterial`, but adding a field that won't be used by other modes doesn't seem like the best approach.

kevinyamauchi:
> What do you think is the best path forward? Move the isosurface rendering shader from the volume_ray.wgsl to its own file? I suppose we could also add the isosurface_threshold uniform to the VolumeBasicMaterial, but adding a field that won't be used by other modes doesn't seem like the best approach.

I moved the iso function to a separate file and have a conditional include (`mode == 'iso'`). This seems to fix the problem. Does this seem reasonable? https://github.com/pygfx/pygfx/pull/773/commits/d4275513e2eaf9100e9bfebaec78b487dd641095

kevinyamauchi:
I had a go at adding some simple lighting (https://github.com/pygfx/pygfx/pull/773/commits/9bc3da761b0fce027a2c88f934adbf00d244bc15). This is basically a copy of the `lighting_phong` function in `light_phong_simple.wgsl` - I just omit the emissive term. I think it would be nice to unify these functions. I did a quick search and it didn't seem like any existing shaders still depend on `light_phong_simple.wgsl`, so it seems safe to make breaking changes to it, but I may have missed something so I wanted to check in first.

https://github.com/pygfx/pygfx/assets/1120672/64c90c4b-a7cc-4873-884f-f208177604a9

almarklein (commented):
❤️  Thanks for working on this! Awesome results already 😃 

> What do you think is the best path forward? Move the isosurface rendering shader from the `volume_ray.wgsl` to its own file? 

I prefer to have the render functions together in the same file. But we have to fix that error regarding the uniform. So maybe we can rename `render_mode_{{ mode }}()`  to simply `raycast()`  or something. And then use this trick:
```wgsl

$$ if mode == 'mip'

    fn raycast(...) {
     ..
    }

$$ elif mode == 'iso'

     fn raycast(...) {
     ..
    }

$$ endif
```

> Do we want to include lighting like this? 

We can start with the simple lighting. I had a quick look at the `lighting_phong()`  from `light_phong.wgsl`, but that must be refactored to have args for e.g. `world_pos` instead of taking the `varyings`, so that we can pass the correct value. Let's do that later.

So let's go with the simple-phong in this PR. You can `{$ include 'pygfx.light_phong_simple.wgsl' $}` and use the function define in it. Fine to modify it if needed, because its indeed not used anywhere.

> Should we expose all of the constants as uniforms on the material?

What constants do you mean?

pygfx/renderers/wgpu/wgsl/volume_ray.wgsl
```diff
@@ -144,12 +147,8 @@ struct RenderOutput {
...
 fn render_mode_mip(sizef: vec3<f32>, nsteps: i32, start_coord: vec3<f32>, step_coord: vec3<f32>) -> RenderOutput {
     // Ideas for improvement:
-    // * We could textureLoad() the 27 voxels surrounding the initial location
-    //   and sample from that in the refinement step. Less texture loads and we
-    //   could do linear interpolation also for formats like i16.
-    // * Create helper textures at a lower resolution (e.g. min, max) so we can
-    //   skip along the ray much faster. By taking smaller steps where needed,
-    //   it will be both faster and more accurate.
+    // * use lighting from the scene lighting system to illuminate the isosurface
+    // * surface position refinement could potentially be sped up
```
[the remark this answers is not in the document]
kevinyamauchi: Oops, sorry about that. Those are indeed unintentional. I will revert.

pygfx/materials/_volume.py
```diff
@@ -134,4 +134,32 @@ class VolumeMipMaterial(VolumeRayMaterial):
 
 
 class VolumeIsoMaterial(VolumeRayMaterial):
-    """A material rendering a volume using isosurface rendering."""
+    """A material rendering a volume using isosurface rendering.
+
+    Parameters
+    ----------
+    isosurface_threshold : float
```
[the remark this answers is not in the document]
kevinyamauchi: Sounds good!

kevinyamauchi:
Thanks for the feedback, @almarklein !

> I prefer to have the render functions together in the same file. But we have to fix that error regarding the uniform. So maybe we can rename render_mode_{{ mode }}() to simply raycast() or something. And then use this trick:

I will give this a go.

> What constants do you mean?

Sorry for being unclear. I meant the parameters for the lighting:
https://github.com/pygfx/pygfx/blob/ef138442baf2b5f304a6980b1d530d85425ffbeb/pygfx/renderers/wgpu/wgsl/light_phong_simple.wgsl#L15-L17

As well as the step sizes for the ray marching.

I could imagine  users wanting control over all of those parameters and they are currently hard coded.

almarklein:
I think its ok if the lighting parameters are hard-coded; when we move to scene lights, the user has full control over the lights.

We could include the step size for ray marching as a prop on the material.

pygfx/materials/_volume.py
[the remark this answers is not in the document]
kevinyamauchi: Done!

pygfx/materials/_volume.py
```diff
@@ -134,4 +135,52 @@ class VolumeMipMaterial(VolumeRayMaterial):
...
+        This color is added to the final color and is unaffected by lighting.
+        The alpha channel of this color is ignored.
+        """
+        return Color(self.uniform_buffer.data["emissive_color"])
+
+    @emissive.setter
+    def emissive(self, color):
+        color = Color(color)
+        self.uniform_buffer.data["emissive_color"] = color
+        self.uniform_buffer.update_range(0, 1)
```
kevinyamauchi, line 186: Do we want to expose the emissive color on the material? I just did so because the phong implementation in `light_phong_simple.wgsl` requires it. it seems nice to have, but I don't have a strong preference.

kevinyamauchi:
Thanks again for the feedback, @almarklein !

I have updated the PR with the following changes:
- unified the iso and mip raycasting shaders into a single file
- switched to the lighting function in `light_phong_simple.wgsl`
- exposed the step sizes, emisssive color, and shininess on the material

Are there any tests you would like me to add? Or is the testing in `test_materials.py` sufficient?

On a related note, I think I've done something that is causing the Docs CI to fail, but I couldn't tell by looking at the logs. 

Thanks!

examples/feature_demo/volume_iso_render.py
```diff
@@ -0,0 +1,41 @@
+"""
+Isosurface Volume Rendering
+==================
```
panxinmiao, line 3: >  I think I've done something that is causing the Docs CI to fail, but I couldn't tell by looking at the logs.

The line length here needs to be the same to the title.
```suggestion
Isosurface Volume Rendering
===========================
```

pygfx/materials/_volume.py
Korijn: Yes, why not!

Korijn (approved):
@almarklein you are also good with this PR?

examples/feature_demo/volume_iso_render.py
kevinyamauchi: Thanks! All green :)

almarklein (approved):
Nice work! 🚀

PR #774, opened 2024-06-02
Provide options to better control font look

hmaarrfk:
The general issues arise when redering text are the following:

- The glyph ref size of 48 is a little too small for some narrow fonts. - Proposal, we can provide a parameter to override it at the Geometry Level.
- The outline width always "adds" thickness to the font, given the low resolution of the above, this doesn't walways look good. - Proposal, provide an "inner-outline-width" parameter.

I demonstrate this with a font that is slightly narrower than Noto Sans (Honestly, it took us a very long time to verify that the font was different since they are so similar).

Let me know what you think.

Best,

48 ref_size (present default), outline 0.3
image

128 ref_size, outline 0.3
image

128 ref_size, outline 0.3, inner outline 0.05
image

Adjust
```python
python examples/feature_demo/text_align.py "f0 128" --ref-size 128 --outline-thickness 0.3 --inner-outline-thickness 0.05
```
as needed.

I've marked this as a draft since, at the very least, I would have to remove the `Source Sans 3` specific code before merging.


pygfx/utils/text/_shaper.py
```diff
@@ -17,7 +17,7 @@
 REF_GLYPH_SIZE = 48  # 48px == 64pt
```
hmaarrfk, line 17: It might just be a good idea to remove this variable to avoid any kind of confusion.

hmaarrfk:
Hmm, its basically impossible to get the same "look" between 128 and 48

since for 48 glyph size, the small number of pixels literally makes the font look very different.

(turned on debug mode to look at the atlas values)
image

almarklein:
How should I inspect these screenshots? Is it the outline of the "f"  that bumps into the "0" (orange), or is it the "0"  itself becoming more narrow (magenta)?

image

hmaarrfk:
The screenshots here are just showing how the different options behave. 

I was on a hunt to find the source of the orange problem you outlined in our application. 

I created these different options that may be of use however they did not resolve my original problem. 

The orange problem you highlighted is resolved with some kind of two pass rendering. 

I believe the purple is an artifact of my misalignment. Except for the case where we specify an inner outline (the last case)
128 ref_size, outline 0.3, inner outline 0.05

almarklein:
What if (in the shader) the depth of the outline is pushed back a bit?

Maybe something like this:
```wgsl
// Move text closer to camera, since its oftern overlaid on something. The text is moved closer than the outline.
out.depth = varyings.position.z - 1e-6 * (2.0 - outline);
```

I figured that with a 24bit depth buffer, the step should be about `1/2**24 == 5.96e-08`, but that shows no difference, so I used a larger (but still very small) number (`1e-6`).

hmaarrfk:
> I figured that with a 24bit depth buffer, the step should be about 1/2**24 == 5.96e-08, but that shows no difference, so I used a larger (but still very small) number (1e-6).

is `z` a normalized number? It may be that 100 + 5.96E-8 is 100 due to rounding errors. At this stage in the shader pipeline, is "-z" always "closer"?

Did you actually code the above???? or just wrote it from memory. 

> // Move text closer to camera, since its oftern overlaid on something. The text is moved closer than the outline.
> out.depth = varyings.position.z - 1e-6 * (2.0 - outline);

its beautiful!
image

I know i'm zooming in alot, but everybody in our company thinks that "f0 looks a little weird" so my zooming in was just to make the problem more obvious.

almarklein:
> is z a normalized number?

Yes, the `varyings.position`  is in NDC, with `z` in 0..1.

> It may be that 100 + 5.96E-8 is 100 due to rounding errors. 

Yeah, exactly. There's the f32 in the shader to account for, as well as the 24 bit of the int depth buffer. But the depth buffer is not linear ... which complicates things. I'm hoping to understand what a reasonable smallest value is for the offset. We're likely not the first in search for the minimal depth bump ...

> At this stage in the shader pipeline, is "-z" always "closer"?

Yes 0 and 1 represent the near and far planes, respectively.



Did you actually code the above???? or just wrote it from memory.

hmaarrfk:
Adding f0 to the `hello_text.py` and stopping the rotation from happening, i can create the following artifacts:
image

which are kinda informative...

While 6E-8 helped when the "en-face" case was considered, it continue to as the user rotated in 3D. 1E-6 did not.

https://github.com/pygfx/pygfx/assets/90008/0d025287-bdc7-4ff6-9dd2-9b46b04d51f9

pygfx/geometries/_text.py
```diff
@@ -964,6 +975,15 @@ def text_align_last(self, align):
         self._text_align_last = align
         self.apply_layout()
 
+    @property
+    def ref_glyph_size(self):
+        return self._ref_glyph_size
+
+    @ref_glyph_size.setter
+    def ref_glyph_size(self, value):
+        self._ref_glyph_size = value
```
Korijn, line 984: should this be reactive via _store?

pygfx/renderers/wgpu/shaders/textshader.py
```diff
@@ -25,7 +24,7 @@ def __init__(self, wobject):
         material = wobject.material
         self["screen_space"] = geometry.screen_space
         self["aa"] = material.aa
-        self["REF_GLYPH_SIZE"] = REF_GLYPH_SIZE
+        self["REF_GLYPH_SIZE"] = wobject.geometry.ref_glyph_size
```
Korijn, line 27: is this properly reactive @almarklein ?

hmaarrfk:
fyi, i think i can narrow the scope of this entire PR to a single line that almar suggested.

Most of the additions were me learning about the problem in giving myself new toggle to play with so that I could intelligently reproduce.

hmaarrfk:
Closing in favor of https://github.com/pygfx/pygfx/pull/776

I've written the note that is important in the meta issue tracking font improvements.

Thank you both!

PR #776, opened 2024-06-04
Move outline behind text itself.

hmaarrfk:
Without this PR:
Screenshot from 2024-06-04 12-28-12

Because the effect was obvious without `f0` i've simply removed it from the final text.


hmaarrfk:
it might be our delta isn't large enough, because now in world space the text is behind the line. will have to investigate. 
image

hmaarrfk:
i'm not sure why the tests are failing, i downloaded and copied over the PNG...

almarklein:
Our depth buffer is actually 32 bit float:

https://github.com/pygfx/pygfx/blob/217d90e86ec6cdb6386e80e83c56e595df9847fb/pygfx/renderers/wgpu/engine/blender.py#L521-L527

Then you'd expect that the smallest step would be `np.finfo(np.float32).eps`  which is `1.192e-07`.

I looked into this a bit deeper by modifying `hello_text.py`: giving the "hello world"  text an outline_thickness of 0.5, disabling the rotation, and positioning the text at (0, 0, 0.5), i.e. exactly on the plane.

When setting the offset to `2e-7` the text is correctly on top, when viewed from the front. But when rotating the camera, z-fighting is seen. When I increase to `3e-7`, the fighting is mostly gone, but easy to observe again when looking at an angle up close:

image

The further I increase the offset, the harder it becomes to see the z-fighting at an angle. Let's use `1e-6` (1 millionth) 😄

hmaarrfk:
i swear i had 1e-6 in there not 6e-8....


Thanks for the investigation.

hmaarrfk:
I'm really at a loss... I tried downloading the artifacts 3 times.

I do feel like it could be a 1 bit value difference but i can't really find it.

Korijn:
Here are the diffs:

validate_text_outline-alpha
validate_text_outline-rgb

pygfx/renderers/wgpu/wgsl/text.wgsl
```diff
@@ -227,6 +227,16 @@ fn fs_main(varyings: Varyings) -> FragmentOutput {
...
 
+    // Move text closer to camera, since its often overlaid on something.
+    // The text is moved closer than the outline so that the outline of one character
+    // does not hide the text of an other.
+    // The depth buffer is a 32 bit float.
+    // Setting the incremented depth to be less than 1e-6 caused
+    // z-fighting to occur leading to artifacts.
+    // See: https://github.com/pygfx/pygfx/pull/776#issuecomment-2149374275
+    // And: https://github.com/pygfx/pygfx/pull/774#issuecomment-2147458827
+    out.depth = varyings.position.z - 1e-6 * (2.0 - outline);
```
almarklein, line 238: I suspect that at the edge of the outline (where it moves from 0-1) there is z-fighting going on. This (I think) would solve that.
```suggestion
    out.depth = varyings.position.z - seslect(1e-6 , 2e-6 , outline == 0.0);
```

hmaarrfk: This didn't really help. I might have to sleep on this. feel free to push.

hmaarrfk:
maybe we just use 0.4 instead of 0.5 as outline thickness and call it a day.

almarklein:
> use 0.4 instead of 0.5 as outline thickness 

if that does the trick ...

hmaarrfk:
0.4 still "shows" the problem without the "proposed fix"
image

almarklein (approved):
🚀

hmaarrfk:
For future onlookers, can help adjust `hello_text.py` to make it easier to control and use as a playground.

```diff
diff --git a/examples/introductory/hello_text.py b/examples/introductory/hello_text.py
index 020b764..d4591e4 100644
--- a/examples/introductory/hello_text.py
+++ b/examples/introductory/hello_text.py
@@ -26,7 +26,12 @@ scene.add(plane)
 # Create two texts, one on each side of the plane
 text1 = gfx.Text(
     gfx.TextGeometry("Hello world", font_size=2.8),
-    gfx.TextMaterial(color="#ddd"),
+    gfx.TextMaterial(
+        color="#ddd",
+        outline_color="#000",
+        # Choose a very thick outline to ensure the effect is noticeable
+        outline_thickness=0.50,
+    ),
 )
 text1.local.position = (0, 0, 0.55)
 plane.add(text1)
@@ -63,6 +68,7 @@ t_last = perf_counter()


 def before_render():
+    return
     global t_last
     t_new = perf_counter()
     dt = t_new - t_last
```

hmaarrfk:
I rebased to get the changes from main about the CIs. 

Are you all OK with rebasing? Do you want me to just squash all of this into 1-2 commits? code, validation, screenshot diffs? Or should i just leave it as is.

almarklein:
We squash-merge PR's into `main`, so that each pr is one commit in the main branch. That means we don't have to care about rebasing/squashing commits in branches of prs themselves.

Korijn (approved):

hmaarrfk:
> We squash-merge PR's into main, so that each pr is one commit in the main branch. That means we don't have to care about rebasing/squashing commits in branches of prs themselves.

awesome, will save me time ;)

hmaarrfk:
i guess this doesn't work with `blend_mode='ordered1'`. :( maybe i should port by alpha values to work with `ordered2`....

PR #778, opened 2024-06-05
More flexible linking controllers to cameras

### LINKED ISSUE #635: Dissociating x and y scrolling
Hi, 
Thank you for this fantastic tool. I was testing fastplotlib with @kushalkolar. We were wondering if it's possible to dissociate panning on the x and y controllers separately. For example, if I want to scroll horizontally without moving the y-axis.

almarklein:
Closes #635

This provides a relatively simple solution to the problem outlined in #635, omitting the need for meta-controllers or two controllers modifying one camera.

* [x] When adding a camera to a controller, a filter can be provided, so that only a subset of the camera state is applied for that camera.
* [x] Allow setting `x`, `y` and `z` camera states individually, instead of always setting `position`.
* [x] Add two examples.

Some context: there are multiple levels at which a form of linking could have been implemented:
* Incoming events. These could be fed into multiple controllers. A problem is that we'd have to mimic that event took place in the appropriate viewport, otherwise it would not be accepted. yuk.
* At the controller action level. This is where we talk about e.g. pan, azimuth, elevation, zoom. This seems like a great level on first sight. However, this breaks down when two controllers are linked that have a different interaction model. Also, over time, the cameras may "drift" due to round-off errors.
* At the level of camera state (what we did now). One huge advantage is that camera state is (partially) copied from one camera to another, so they won't "drift". This also fits the best in the current controller implementation and keeps things modular.

### API changes

* `PerspectiveCamera.set_state()` allows setting `x`, `y`, and `z`.
* `Controller.add_camera()`  has new args ` include_state` and `exclude_state`.


Korijn (approved):

BalzaniEdoardo:
Thanks Almar! Good job!

Addressing the camera drift is a very good point. I like this solution, I will try it out and if I am not able to make it work, I'll reach out again and ask you!

BalzaniEdoardo:
Hi Almar! 
We just check out the camera sync example. It does exactly what we needed, very neat!

almarklein:
Great! Thanks for the feedback!

PR #782, opened 2024-06-07
Add text overlays to the transparency examples

hmaarrfk:
I really wasn't able to get an internal feeling of what each one did.

I took some of our overlay code and simplified it to make it more relatable.

I didn't want the overlay code to be "too much" so I kept the default font settings as much as I could.

https://github.com/pygfx/pygfx/assets/90008/38a562bd-e3ed-4661-91a1-5b70e740ec7a


hmaarrfk:
(incidentally, you can see why text with outline with `ordered1` blending looked so bad for us.)

almarklein (approved):
Nice!

Korijn (approved):
Transparency is an incredibly deep topic in graphics. It will always be challenging, even for the experts. :)

PR #783, opened 2024-06-08
Make a logo for overlay with a mesh

hmaarrfk:
image

I wanted to learn about meshes, and so I figured I would help Almar with a suggestion I made offline.

You guys killed me with your gradient. I got all the coordinates by writing down the numbers from the svg using inkscape. You even had one for the blue....

I don't really like it since I wish it was pixel perfect. I also want to make it easier to use in the examples. currently, I still feel it adds too many lines of code in the example.

I wish I was able to make it a point material, and then we could just place it anywhere, and ensure that it is pixel perfect on the GPU. All gradients included. That would require a custom shader, but it would be pretty cool!

### DIFF
--- a/examples/feature_demo/image_overlay.py
+++ b/examples/feature_demo/image_overlay.py
@@ -60,7 +60,12 @@ camera.local.scale_y = -1
 camera.zoom = 1.2
 
 controller = gfx.PanZoomController(camera, register_events=renderer)
+logo, logo_camera = gfx.make_logo_scene_camera()
+
+def render():
+    renderer.render(scene, camera, flush=False)
+    renderer.render(logo, logo_camera)
 
 if __name__ == "__main__":
-    canvas.request_draw(lambda: renderer.render(scene, camera))
+    canvas.request_draw(lambda: render())
     run()
--- a/examples/feature_demo/image_overlay.py
+++ b/examples/feature_demo/image_overlay.py
@@ -60,12 +60,7 @@ camera.local.scale_y = -1
 camera.zoom = 1.2
 
 controller = gfx.PanZoomController(camera, register_events=renderer)
-logo, logo_camera = gfx.make_logo_scene_camera()
-
-def render():
-    renderer.render(scene, camera, flush=False)
-    renderer.render(logo, logo_camera)
 
 if __name__ == "__main__":
-    canvas.request_draw(lambda: render())
+    canvas.request_draw(lambda: gfx.render_with_logo(renderer, scene, camera))
     run()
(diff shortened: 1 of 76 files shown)


examples/feature_demo/image_overlay.py
```diff
@@ -62,5 +62,5 @@
 controller = gfx.PanZoomController(camera, register_events=renderer)
 
 if __name__ == "__main__":
-    canvas.request_draw(lambda: renderer.render(scene, camera))
+    canvas.request_draw(lambda: gfx.render_with_logo(renderer, scene, camera))
```
hmaarrfk, line 65: alright, this change i'm happy with. Should be "easy" to add to examples and whatnot without disruption.

pygfx/helpers/_logo.py
```diff
@@ -0,0 +1,137 @@
...
+    # Seems like the coordinates were about 100 in size
+    triangle.local.scale = 1 / 100, 1 / 100, 1
+    blue_lines.local.scale = 1 / 100, 1 / 100, 1
+
+    logo = Group()
+    logo.add(triangle, blue_lines)
+    logo.local.scale = scale, -scale, 1
+    logo.local.position = (scale / 20, scale, 0)
+
+    camera = ScreenCoordsCamera()
```
hmaarrfk, line 129: The logo is placed at the bottom right since it is "convenient". We don't need to move it when the window is rescaled.

Korijn:
I guess you could render it as a sprite via a texture. But obviously your approach is going to look crisp, no matter what. Cool stuff!

hmaarrfk:
One challenge I faced with rendering logos with transparency as images is that it really depends on the blend mode to not hide the stuff behind the transparent region. I'm hoping that with a mesh this is less of a concern.

hmaarrfk:
The chosen colors mean that it looks great on the many example backgrounds without any outline.
image

hmaarrfk:
The one thing i'm worried about is that the new function
```
gfx.render_with_logo
```

removes the call

```
renderer.render()
```
or
```
renderer.flush()
```

from the examples. For this reason, I didn't want to add this to the tutorial examples. But maybe there is a way to do it well.

examples/feature_demo/volume_slice3.py
```diff
@@ -24,7 +24,7 @@
 
 vol = gfx.Volume(
     gfx.Geometry(grid=voldata),
-    gfx.VolumeSliceMaterial(clim=(0, 2000), plane=(0, 0, -1, index)),
+    gfx.VolumeSliceMaterial(clim=(0, 2000), plane=(0, 0, -1, index), pick_write=True),
```
hmaarrfk, line 27: This was a distinct unrelated big.

PR #784, opened 2024-06-10
Add minip volume rendering

kevinyamauchi:
As discussed in #86 , this PR adds minimum intensity projection rendering. 

https://github.com/pygfx/pygfx/assets/1120672/b289208c-035d-4f5d-b551-743613fd1d82

^ example script: `/examples/feature_demo/volume_minip_render.py`

### DIFF
--- a/pygfx/renderers/wgpu/wgsl/volume_ray.wgsl
+++ b/pygfx/renderers/wgpu/wgsl/volume_ray.wgsl
@@ -214,7 +214,7 @@ $$ if mode == 'mip'
     }
 
 $$ elif mode == 'minip'
-    // raycasting function for MIP rendering.
+    // raycasting function for MINimum Intensity Projection rendering.
     fn raycast(sizef: vec3<f32>, nsteps: i32, start_coord: vec3<f32>, step_coord: vec3<f32>) -> RenderOutput {
         // Ideas for improvement:
         // * We could textureLoad() the 27 voxels surrounding the initial location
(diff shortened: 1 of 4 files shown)


kevinyamauchi:
I believe the basic functionality is working and I would appreciate feedback on the following :
- what do we want to call this rendering mode? As a starting point I chose `minip`, but I am open to other suggestions.
- there is some code duplicated between the minip a mip `raycast()` functions. do we want to refactor so there is less duplication (e.g., the coloring and picking code)? I think the functions are short enough that it is fine as is, but I am fine with refactoring if you prefer.

Thanks!

pygfx/renderers/wgpu/wgsl/volume_ray.wgsl
```diff
@@ -213,6 +213,75 @@ $$ if mode == 'mip'
         return out;
     }
 
+$$ elif mode == 'minip'
+    // raycasting function for MIP rendering.
```
almarklein, line 217:
```suggestion
    // raycasting function for MINimum Intensity Projection rendering.
```

almarklein (commented):
>  what do we want to call this rendering mode? As a starting point I chose minip, but I am open to other suggestions.

The name "minip" seems to be widely used. It's what vispy uses and e.g. https://en.wikipedia.org/wiki/Minimum_intensity_projection. So let's just use that.

> I think the functions are short enough that it is fine as is

I agree.

almarklein (approved):

Korijn (approved):

almarklein:
Thanks! 🙏

PR #787, opened 2024-06-14
Add Ruler object

almarklein:
Relates to #739 

### Introduction

This PR adds a new world object that shows a ruler. The primary purpose is to function as a base component for axii, but it can also be used to measure stuff interactively in a scene.

### The idea

Use a `gfx.Line` to draw the line, a `gfx.Points` to draw the tickmarks, and a set of `gfx.Text`s to draw the tick texts (I want to eventually use a single text object for this). These objects are wrapped in a new `gfx.Ruler` world object. That object is like a group object, and provides the API.

### What makes the API hard

We can establish a few steps that the ruler must perform, in order.

* User must provide start_pos, end_pos, start_value.
* The user must provide ticks, or `None` for automatic ticks.
* The camera must be used to determine the size on screen.
* From this the distance between ticks must be calculated.
* The sub-objects (line, points, text objects) must be updated to show line, ticks, and tick labels.

On top of that we want to let the user know the calculated tick-distance, and the values for the visible ticks, so that gridlines can be rendered in the right place.

In a previous version, I more or less implemented each step as a method, and let the user call things in correct order. But this was rather cumbersome. Instead, there's now a single  `update()` method, and it returns a dict with information that the user might want to use.

### API

The API is up for discussion. Ideas welcome. ~~Right now I used relatively simple methods. One needs to at least two calls to "apply" the ruler. This leads to somewhat more verbose code, but makes it more modular too. For example, the result of `calculate_tick_step()` can be used to set the step size in the grid!~~

Quick example:
```py
ruler = gfx.Ruler(**any_props)
ruler.start_pos = ...
ruler.end_pos = ...
ruler.other_props = ...

def animate():
    rulery.update(camera, canvas.get_logical_size())
    # Or, when you want the tick values:
    #ruler_stats = rulery.update(camera, canvas.get_logical_size())
```

### Notes on the "visible range"

The ruler has logic to determine a good distance between ticks so that ticks are not too far apart or too close. In other words, the tick density on screen remains more or less constant. But this means that the density of ticks in world space increases as one zooms in. So when zooming in with a factor 10, when there would normally be 20 ticks visible, there would be 200 on the total line. And possibly much more with a perspective camera. 

Therefore, the rulers detects the part of the line that is visible on screen, and only renders ticks that are in that range. Implementing this took some time to get right.

### Tasks

* [x] Add `anchor_offset` to `TextGeometry`, so the text does not hug the line.
* [x] Basic implementation.
* [x] Re-apply ticks as you zoom in and out. 
* [x] Discuss and flesh out the API.
* [x] Example use as a measuring tape.
* [x] Example use in a plot.
* [x] Hide tickmarks at beginning and end of the line.
* [x] Optimize tick step selection.
* [x] Implement exponential notation and umKM units.
* [x] Automatically determine text anchor from the line orientation.
* [x] Better tick spacing for perspective cameras.

### Tasks for later

* (try to) implement a `TextGeometry` that can show multiple single-line string at different positions in the scene. Apart from a performance benefit (less world objects), this also has an API advantage, as users can then change the material to change the appearance of the ticks.
* Implement a points material to draw tickmarks (using simple points for now).

### Some screenshots

image

image


kushalkolar:
This is amazing! :smile: :partying_face: I played with it a bit, I wonder if we could use resize and controller or camera callbacks instead of the animation function to set any change in ticks etc. I'll try this out in a few days, currently travelling/conferencing.

almarklein:
> I wonder if we could use resize and controller or camera callbacks instead of the animation function to set any change in ticks etc.

Our camera's and controllers don't have callbacks. This discussion regarding updates/animations relates to #495 too

pygfx/objects/_ruler.py
```diff
@@ -0,0 +1,489 @@
...
+        """The points object that shows the ruler's tickmarks."""
+        return self._points
+
+    # Note: text should also be here eventually
+
+    # -- Main properties
+
+    @property
+    def start_pos(self):
+        """The start posision of the ruler as a 3-tuple or 3-element array.
```
Korijn, line 75: in world space I'm assuming?

pygfx/objects/_ruler.py
```diff
@@ -0,0 +1,489 @@
...
+        """The format to display the tick values."""
+        return self._tick_format
+
+    @tick_format.setter
+    def tick_format(self, format):
+        self._tick_format = str(format)
+
+    @property
+    def tick_side(self):
+        """Whether the ticks are on the left or right of the line.
```
Korijn, line 161: what about top and bottom?

pygfx/objects/_ruler.py
```diff
@@ -0,0 +1,489 @@
...
+        # Get corresponding screen coordinates.
+        # Fall back to full line when the selected region is empty,
+        # so that we can still calculate the step size, because calling code
+        # may still need it, e.g. to configure a grid.
+        ndc_t1 = ndc1 * (1 - t1) + ndc2 * t1
+        ndc_t2 = ndc1 * (1 - t2) + ndc2 * t2
+        ndc_t1_2d = ndc_t1[:2] / ndc_t1[3]
+        ndc_t2_2d = ndc_t2[:2] / ndc_t2[3]
+        screen_t1 = ndc_t1_2d * half_canvas_size
+        screen_t2 = ndc_t2_2d * half_canvas_size
```
Korijn, line 267: if ndc1 and ndc2 are rows in a single matrix, then you should be able to use broadcasting to have half of the lines of code here right?

pygfx/objects/_ruler.py
```diff
@@ -0,0 +1,489 @@
...
+            tick_values = list(ticks.keys())
+            if abs(tick_values[0] - start_value) < 0.5 * tick_auto_step:
+                self._text_object_pool[1].geometry.set_text("")
+            if abs(tick_values[-1] - end_value) < 0.5 * tick_auto_step:
+                self._text_object_pool[index - 1].geometry.set_text("")
+
+        # Apply
+        self.clear()
+        self.add(self._line, self._points)
+        self.add(*self._text_object_pool)
```
Korijn, line 489: does this mean the same objects are added over and over again?

pygfx/objects/_ruler.py
```diff
@@ -0,0 +1,489 @@
...
+            self.points.geometry.sizes.update_range()
+        else:
+            # Allocate new buffers
+            new_n_slots = max(min_n_slots, int(n_positions * 1.2))
+            positions = np.zeros((new_n_slots, 3), np.float32)
+            sizes = np.zeros((new_n_slots,), np.float32)
+            self.points.geometry.positions = Buffer(positions)
+            self.points.geometry.sizes = Buffer(sizes)
+
+        def define_text(pos, text):
```
Korijn, line 434: does the pool ever shrink?

pygfx/utils/compgeo.py
```diff
@@ -0,0 +1,150 @@
...
+    # previous. Otherwise (1, 2, 3) it is 50%.
+    #
+    #  |------|------|------|------|
+    #  0      1      2      3      4
+    #
+    # AK: I spent some time optimizing this code. Using numpy arrays
+    # and np.argmin is not faster. What does help is re-using the sample
+    # positions. With 5 samples per level, we can re-use 2 or 3 of the
+    # samples at each step. By using tuples, copying these into the new
+    # slot is efficient.
```
Korijn, line 64: it's a long shot, but perhaps the built in [bisect](https://docs.python.org/3/library/bisect.html) module could be leveraged here as well? not sure, I didn't read the code too carefully

Korijn (dismissed):
I don't see any reason to block this PR from merging, my comments are questions and suggestions :)

pygfx/objects/_ruler.py
almarklein: Model space, actually. Will update the docs to make this explicit.

pygfx/objects/_ruler.py
almarklein: See the line below: "Imagine standing on the start position, with the line in front of you." The actual positioning depends on the angle, and is such that there are no jumps "over" the line. 

It's a bit like "starboad" and "port side" on ships 😄

pygfx/objects/_ruler.py
almarklein: Nice, will do!

pygfx/objects/_ruler.py
almarklein: Yes, if `n_positions` is smaller than `2 * n_slots`, and larger than `min_n_slots`.

pygfx/objects/_ruler.py
almarklein: Good point, we can limit adding/removing stuff to when the pool is resized. Will do.

pygfx/objects/_ruler.py
almarklein: No wait, the pool is managed differently. There's ``self._text_object_pool[index + 1 :] = []`` some lines below. I will have a look at whether this can be combined better with the "pool" of lines and points.

almarklein: Done

pygfx/objects/_ruler.py
almarklein: But I understand the confusion. I wonder if we can come up with a better term.

pygfx/utils/compgeo.py
almarklein: Thanks for the suggestion! Looking into it ...

kushalkolar:
Taking a look at this again, I can say that it's now very performant! :smile: . I'd like to take a few days to try out some things to see if there are any minor additional things that I may need.

It's easily possible to make the ruler follow the camera by modifying the animation function, probably makes sense to have this as an option in fastplotlib rather than here I guess :thinking: . This type of functionality is common in matplotlib, matlab etc.

https://github.com/pygfx/pygfx/assets/9403332/0f53ef9a-ee73-4d48-a121-6692cae53590

```python
"""
Line plot
=========

Show a line plot, with axii and a grid.
"""

# sphinx_gallery_pygfx_test = 'run'

import numpy as np
import pygfx as gfx
from pylinalg import vec_transform, vec_unproject


canvas = WgpuCanvas()

scene = gfx.Scene()

background = gfx.Background.from_color("#000")

grid = gfx.Grid(
    None,
    gfx.GridMaterial(
        major_step=1,
        minor_step=0,
        thickness_space="screen",
        major_thickness=2,
        minor_thickness=0.5,
        infinite=True,
    ),
    orientation="xy",
)
grid.local.z = -1001

rulerx = gfx.Ruler(tick_side="right")
rulery = gfx.Ruler(tick_side="left", min_tick_distance=40)

x = np.linspace(20, 980, 200, dtype=np.float32)
y = np.sin(x / 30) * 4
positions = np.column_stack([x, y, np.zeros_like(x)])

line = gfx.Line(
    gfx.LineMaterial(thickness=4.0, color="#aaf"),
)
scene.add(background, grid, rulerx, rulery, line)

camera = gfx.OrthographicCamera(maintain_aspect=False)
camera.show_rect(0, 1000, -5, 5)

def animate():
    vs = renderer.logical_size
    # place the ruler close to the left and bottom edges of the viewport
    xscreen, yscreen = 0.1 * vs[0], 0.9 * vs[1]

    pos_ndc = (xscreen / vs[0] * 2 - 1, -(yscreen / vs[1] * 2 - 1), 0)

    pos_ndc += vec_transform(camera.world.position, camera.camera_matrix)
    xworld, yworld, _ = vec_unproject(pos_ndc[:2], camera.camera_matrix)

    rulerx.start_pos = camera.world.x - camera.width / 2, yworld, -1000
    rulerx.end_pos = camera.world.x + camera.width / 2, yworld, -1000
    rulerx.start_value = rulerx.start_pos[0]
    statsx = rulerx.update(camera, canvas.get_logical_size())

    rulery.start_pos = xworld, camera.world.y - camera.height / 2, -1000
    rulery.end_pos = xworld, camera.world.y + camera.height / 2, -1000
    rulery.start_value = rulery.start_pos[1]
    statsy = rulery.update(camera, canvas.get_logical_size())

    major_step_x, major_step_y = statsx["tick_step"], statsy["tick_step"]
    grid.material.major_step = major_step_x, major_step_y
    grid.material.minor_step = 0.2 * major_step_x, 0.2 * major_step_y

    # print(statsx)

if __name__ == "__main__":
    canvas.request_draw(animate)
    run()
```

kushalkolar:
Oh idea! Mapping the world space to the underlying dataunits. For example, the x-axis is often time and integer steps on the x-axis may be video frames, but the user actually wants to display timepoints. What's the best way to do this? One idea is allowing a user function to be passed that maps from world space to the data space. This space mapping function could be used in `Ruler._get_ticks_uniform()`?

pygfx/utils/compgeo.py
Korijn: Awesome!

pygfx/utils/compgeo.py
```diff
@@ -145,3 +141,24 @@ def value_from_t(t):
         t = samples_t[1] + dt * t_step
 
     return t
+
+
+# We use our own bisect functions. Otherwise, to use bisect.bisect with descending
+# values, we need the key param, which adds complexity, but also does not work pre py10.
```
Korijn, line 147: 3.10 I guess?

pygfx/utils/compgeo.py
```diff
@@ -145,3 +141,24 @@ def value_from_t(t):
...
+            return i
+    return i + 1
+
+
+def bisect_desc(values, ref):
+    """Like bisect(), but the values are assumed in descnding order."""
+    for i, v in enumerate(values):
+        if v <= ref:
+            return i
+    return i + 1
```
Korijn, line 164: Bisect builtin is supposed to be very fast because it uses binary search and is implemented in C I think. It's surprising to see that your naive for loop performs better? I can only imagine it to be true if you happen to be testing on examples where the lists are very small or the bisect value is always an early value.

pygfx/utils/compgeo.py
almarklein: Now I see why it looked weird 😅

pygfx/utils/compgeo.py
almarklein: Our lists are always just 3 items.

We don't have a big list to find the insertion point form but more of a description of a line of which we need to find the reference point. We call bisect repeatedly as part of the binary search.

almarklein:
@kushalkolar Yeah the performance is much better than the first iteration. Nice suggestions! I added them to #739. For now I just want to get this merged. The mapping for the ticks should be an easy addition though!

pygfx/utils/compgeo.py
almarklein: So I guess it's not so much the bisect function itself that helped improve this code, but more the idea of bisection :)

Korijn (approved):

Issue #48, opened 2020-07-08
Improvements to point material


almarklein:
* [ ] Depth attenuation
* [ ] Perhaps more materials for different shapes

almarklein:
Also check out https://observablehq.com/@rreusser/selecting-the-right-opacity-for-2d-point-clouds

Korijn:
By depth attenuation, the goal is to be able to set the radius of points in world space.

This would visually resemble a 3D sphere in the scene.

almarklein:
Added an item for per-vertex marker shape. See https://github.com/fastplotlib/fastplotlib/pull/549.

Some thoughts on this:

The current markers are shaped programmatically based on the point coords. What marker (i.e. code) to use it determined using templating. Making this dynamic could be done using a big multi-if statement, which possibly hurts performance. An alternative could be an array of function objects, but I'm not sure whether wgsl supports this.

Benchmarking reported in #772 suggests that using sdf's instead of computed markers may be more performant. And if we put all our builtin markers in a single texture, we can easily switch between markers! I think this would be the best approach, because it also allows us to create shapes that are not possible with the computed approach.

hmaarrfk:
> that are not possible with the computed approach.

iq would say bogus!  https://youtu.be/8--5LwHRhjk?list=PL0EpikNmjs2CYUMePMGh3IjjP4tQlYqji
Now jokes are done.

> that using sdf's instead of computed markers may be more performant

I'm having a hard time understanding the chain of thought. Which is more performance, "SDF" or "texture lookup"? I think SDF is quite cool since it is "infinite" resolution, but the utility of that advantage is rather limited. It would be good to retain this feature.

I think "avoiding branches" is likely not super critical. All fragment shaders are taking the "same branch" for a given glyph, so even if you put things in a huge if statement, I don't think it would be "disastrous".

However, I do have to say that one of the slowest examples to run is the `validate_point_markers.py` which I've always attributed to compilation of all the different shaders.

almarklein:
>  https://youtu.be/8--5LwHRhjk?list=PL0EpikNmjs2CYUMePMGh3IjjP4tQlYqji

Wow! That's really cool, and inspiring!

wulfdewolf:
Hi, we use pygfx to visualise neural activity data, which often consists of plotting 'spikes' (events in time).
For this we typically use short thin vertical lines (e.g. |). 
The bitmaps/text glyphs would allow us to do that using PointsMarkerMaterial.

[...]

PR #789, opened 2024-06-16
Custom point marker SDF enabling logo insertion

hmaarrfk:
So this creates a few markers to make the pixel perfect logo ;)

image


hmaarrfk:
Why is scikit-image 0.22.0 getting found? I feel like this "size changed numpy error" got fixed in 0.23.X and should not there.

pygfx/renderers/wgpu/wgsl/points.wgsl
```diff
@@ -313,6 +313,75 @@ fn fs_main(varyings: Varyings) -> FragmentOutput {
...
+    let r3_out_blue = -r3_out - inner_offset;
+    let r3_in_blue = r3_out_blue + pygfx_width;
+    let r3_blue = max(
+        max(r1_out, r2_in),
+        max(r3_out_blue, -r3_in_blue)
+    );
+
+    let inner_triangle = min(r1_blue, min(r2_blue, r3_blue));
+
+    return vec2<f32>(outer_edge_triangle, inner_triangle) * size;
```
hmaarrfk, line 380: I found it strange that your original intersections and unions were written in "scale aware" coordinates". 

It was easier for me to write it in these "normalized units" and only think of the scale at the very end.

pygfx/renderers/wgpu/wgsl/points.wgsl
```diff
@@ -313,6 +313,75 @@ fn fs_main(varyings: Varyings) -> FragmentOutput {
...
+    let v3 = normalize(vec2<f32>(triangle_x, triangle_x));
+    let r3_out = dot(coord_1 - vec2(1, one_minus_triangle_x), v3);
+    let r3_in  = r3_out + pygfx_width;
+    let r3 = max(r3_out, -r3_in);
+
+    // return max(max(r1, r2), r3);
+    let outer_edge_triangle = max(
+        max(r1_out, max(r2_out, r3_out)),
+        -max(r1_in, max(r2_in, r3_in))
+    );
```
hmaarrfk, line 352: After writing these, I was curious if you were consider ever writting "intersection", "union", "difference" as macros.

It would have made it make more sense to me as writing, then.

Eventually, I got used to it.

hmaarrfk:
I think the tests are failing because numpy 2.0 just got released:

```
Collecting numpy (from pygfx==0.2.0)
  Downloading numpy-2.0.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (60 kB)
```

Honestly, I had asked that numpy not release it for python 3.9 but they didn't listen.

The issue is that scikit-image didn't release 0.23.X for python 3.9, the next release 0.24.X support python 3.9 to alleviate this issue for others.

It should come out imminently,

pygfx/helpers/_logo.py
```diff
@@ -0,0 +1,46 @@
...
+        PointsMarkerMaterial(
+            size=size,
+            color="#FFE64B",
+            marker="pygfx_outer",
+            edge_color="#000",
+            edge_width=edge_width,
+        ),
+    )
+    logo = Group()
+    logo.add(logo_inner, line_outer)
```
hmaarrfk, line 32: I'm honestly not so sure why the logo is getting rendered in the "correct" order. Is there some way we can "guarantee" it?

Korijn:
I'm glad you like the new logo so much! :)

hmaarrfk:
> I'm glad you like the new logo so much! :) I think your version looks better than the one I put together...

the svg one? 

I wish i had the patience to write fully custom shader to allow the color gradient, i'm just trying to make "easy" for now.

I learned alot from Rougier's paper so it was a worthwhile exercise for me to learn about graphics.

hmaarrfk:
I like the svg one alot, and I just tried to copy it as much as I could (I didn't get the gradients right).

I somewhat also wish I could get the "ribbon wrapping effect" through the depth buffer. That would be cool allow a transparency (I think of the blue as a ribbon that wraps around the yellow triangle).

pygfx/renderers/wgpu/wgsl/points.wgsl
almarklein: The code for most markers was copied from Rougier's paper and adjusted for wgsl. TBH I have not attempted creating my own marker yet. I guess that makes you officially the expert on this code ;)

almarklein:
Really cool to fly through a cloud of logos, and having a way to draw the logo like this. TBH I'm not a fan of the the `render_with_logo()`. I think adding a logo to the screenshots in the gallery would be cool, but changing the API for it is a bit too much. We can overlay the API by e.g. patching the renderer in the `docs/conf.py` or simply adding a bitmap to each image.

hmaarrfk:
> We can overlay the API by e.g. patching the renderer in the docs/conf.py or simply adding a bitmap to each image.

This is probably a better solution. I also didn't like the change in API especially since these examples have been so instrumental in my learning.

pygfx/renderers/wgpu/wgsl/points.wgsl
```diff
@@ -464,6 +465,9 @@ fn get_signed_distance_to_shape_edge(coord_in: vec2<f32>, size: f32) -> f32 {
         let r4 = length(coord-c1)-size/5;
         return max( min(r1,max(max(r2,r3),-coord.y)), -r4);
 
+    $$ elif shape == 'custom'
+        {{ custom_sdf }}
```
hmaarrfk, line 469: I think it would be good to put all of this in a clean function that doesn't have access to all the niceties of variables defined in `get_signed_distance_to_shape_edge`.

hmaarrfk: it could be that the signature is simply that of `get_signed_distance_to_shape_edge`

examples/feature_demo/fly_controller_logo.py
```diff
@@ -0,0 +1,155 @@
...
+    );
+
+    let r3_out_blue = -r3_out - inner_offset;
+    let r3_in_blue = r3_out_blue + pygfx_width;
+    let r3_blue = max(
+        max(r1_out, r2_in),
+        max(r3_out_blue, -r3_in_blue)
+    );
+
+    return min(r1_blue, min(r2_blue, r3_blue)) * size;
```
hmaarrfk, line 74: If this strategy was chosen, it was a little strange to me that the "sdf" had to be multiplied by the size. this would be important to document in `custom_sdf`.

hmaarrfk:
To minimizing leaking the internals of the SDF function, i'm suggesting that this https://github.com/pygfx/pygfx/pull/822 gets merged first.

examples/feature_demo/fly_controller_logo.py
```diff
@@ -0,0 +1,196 @@
+"""
```
hmaarrfk, line 1: maybe this example should just be renamed to something else. I don't like how I've highjacked the `fly_controller.py` name making that one harder to press tab onto.

I wanted to show:

1. Customizing of points with custom SDF.
2. Overlaying of your own custom logo onto your scene.

examples/validation/validate_points_markers.py
```diff
@@ -38,6 +38,67 @@
 scene = gfx.Scene()
 scene.add(gfx.Background.from_color("#bbb", "#777"))
 
+pygfx_sdf = """
+    let m_sqrt_3 = 1.7320508075688772;
+
+    // The coords below assume OpenGL coords, but y is flipped, so we flip the coords.
+    // Its also shifted so that the center of the triangle is at the origin.
+    var coord_for_sdf = vec2<f32>(coord.x, -coord.y) / size + vec2<f32>(0.5);
```
hmaarrfk, line 46: There is always a bit of SDF black magic.... i tried to remove the "minus" sign, but i just get black now.

I'll see what I can do about this, but it feels like an optimization nobody asked for so I'm not so worried.

examples/feature_demo/fly_controller_logo.py
hmaarrfk: i've renamed the logo

pygfx/renderers/wgpu/wgsl/points.wgsl
hmaarrfk: done in https://github.com/pygfx/pygfx/pull/822

hmaarrfk:
I feel like this would be the first time you are allowing users to inject code directly in the WGSL.

I understand if you might need time to consider how this is done and documented.

looking forward to helping build out customization options like this.

pygfx/renderers/wgpu/wgsl/points.wgsl
Korijn: it depends on how good the compiler is. there's usually a chain of them. in this case the first compilation is handled by naga (WGSL -> shader code), and the second compilation is performed by the GPU and its drivers. In those steps it is possible that the macros simply get compiled away and there will be no performance difference. but it is highly dependent on the versions of naga and drivers, and the specific hardware capabilities. so all in all, using macros may impact performance but it also may not, depending on the environment in which the shader code runs.

pygfx/renderers/wgpu/wgsl/points.wgsl
Korijn: feel free to resolve these discussion threads of you feel they are "done". they'll only block the PR and perhaps unintentionally create the impression you're still working on it.

examples/feature_demo/fly_controller_logo.py
Korijn: do you want to document it or not? feel free to drive this thread to closure yourself by making a decision

pygfx/renderers/wgpu/wgsl/points.wgsl
hmaarrfk: Interesting.  Thanks!

pygfx/utils/enums.py
```diff
@@ -151,6 +151,7 @@ class MarkerShape(Enum):
     spade = None  #: ♠
     club = None  #: ♣
     pin = None  #: 📍
+    custom = None  # Custom shape allowing users to provide their own SDF function
```
Korijn, line 154: I think you also need to add `custom` to the docstring in `materials/_points.py` in the setter for marker

pygfx/materials/_points.py
```diff
@@ -342,6 +351,19 @@ def marker(self, name):
             )
         self._store.marker = resolved_name
 
+    @property
+    def custom_sdf(self):
+        """The SDF code for the marker shape when the marker is set to custom.
+
+        Negative values are inside the shape, positive values are outside the
+        shape.
```
hmaarrfk, line 359:
```suggestion
        shape.

         The SDF's takes in two parameters `coords: vec2<f32>` and `size: f32`.
        The first is a WGSL coordinate and `size` is the overall size of
        the texture. The returned value should be the signed distance from
        any edge of the shape. Distances (positive and negative) that are 
        less than half the `edge_width` in absolute terms will be colored
        with the `edge_color`. Other negative distances will be colored by
        `color`.
```

Korijn (changes requested):
> I feel like this would be the first time you are allowing users to inject code directly in the WGSL.
> I understand if you might need time to consider how this is done and documented.

It looks great to me! I had 1 minor docstring comment. Otherwise, I think all you need to do is get all the open comment threads resolved and then we can merge this.

pygfx/utils/enums.py
hmaarrfk: added. please see new docstrings.

examples/feature_demo/fly_controller_logo.py
hmaarrfk: yeah, i think its useful. took me 1 month to feel comfortable with SDFs

pygfx/materials/_points.py
```diff
@@ -287,7 +296,10 @@ def marker(self):
         * Matplotlib compatible characters: "osD+x^v<>".
         * Unicode symbols: "●○■♦♥♠♣✳▲▼◀▶".
         * Emojis: "❤️♠️♣️♦️💎💍✳️📍".
+        * A string containing the value "custom"
 
+        In the that the "custom" marker is used, the WGSL
+        code defined by `custom_sdf` will be used.
```
almarklein, line 302:
```suggestion
        * A string containing the value "custom". In this case, the WGSL
          code defined by ``custom_sdf`` will be used.
```

almarklein (approved):
Nice, just one other small docstring change.

pygfx/materials/_points.py
```diff
@@ -244,11 +244,20 @@ class PointsMarkerMaterial(PointsMaterial):
...
     )
 
-    def __init__(self, *, marker="circle", edge_width=1, edge_color="black", **kwargs):
+    def __init__(
+        self,
+        *,
+        marker="circle",
+        edge_width=1,
+        edge_color="black",
+        custom_sdf="return -size;",
```
Korijn, line 253: did you accidentally commit this?

almarklein: It also took me a while before I figured out this is a piece of wgsl. Maybe better to default to None to avoid confusion, and in the shader do `self["custom_sdf"] = material.custom_sdf or "return -size;"`

hmaarrfk: I definitely snuck this in late in the review. I agree with almars suggestion 

I just wanted to make something "basic" to give a display. 

The scariest thing on the GPU is when you see nothing.

hmaarrfk: Maybe 


min(abs(coord))

hmaarrfk: Min(abs(coord)) - size

hmaarrfk:
Ok i agree that "None" is a better default.

Korijn (approved):

hmaarrfk:
now its time for me to write an SDF for my own logo!!!

Korijn:
Sorry for the long delay on the review. I am switching jobs and it's limiting my headspace for this project. I hope to get back into it with less stress in september.

hmaarrfk:
Oh don't be sorry at all. 

I totally understand things and we are trying to figure out how to help as well! 

We've patched your code plenty already and getting things merged I'm just trying to give back where I can. 

I know you are all trying to wrap up the core infrastructure so I'm trying to focus my comments on the engine and less this cosmetic stuff!

hmaarrfk:

PR #795, opened 2024-06-19
Buffer chunking and more

### LINKED ISSUE #19: Improvements to Buffer

### LINKED ISSUE #600: Improve buffer/texture update_range

### LINKED ISSUE #714: Textures from arrays with little- or big-endian byte order
Hello!

While working on #707 I had a look at `Texture` because I ran into issues with image volumes that were little- or big-endian encode (e.g. `>u4`) - which seems not uncommon for e.g. microscopy formats:

```python
>>> import pygfx as gfx
>>> import numpy as np
>>> dummy = np.zeros((10, 10, 10), dtype='>u4')
>>> tex = gfx.Texture(dummy, dim=3)
TypeError: Cannot convert '>I' to wgpu format. You should provide data with a different dtype.
```

It looks like `get_item_format_from_memoryview()` is the initial gatekeeper as it complains because only native (`=`) [byte-orders](https://numpy.org/doc/stable/reference/generated/numpy.dtype.byteorder.html) are in the mapping. I tried circumventing it but then `to_texture_format()` complains when trying to run `gfx.show()`.

At the end of the day I was just wondering if that was a hard constraint (i.e. it actually matters to the GPU) or whether it would be possible to be more forgiving with regard to endian-ness?

### LINKED ISSUE #812: Crashing pygfx with zero length arrays
Show how you can crash rust with a 0-length array
    
I'm really sorry for opening so many issues but this one seems intellectually interesting,
    
It seems that a zero length position array (which can occur easily if you try to procedurally generate points and hit an edge case) can cause a rust panic.
    
I modified the example code to show how it can occur in a reproducible way.
     
Backtrace included.
    
Not sure if there is a good way to avoid these errors at a pygfx/wgpu level.
    
```
$ RUST_BACKTRACE=1 python examples/introductory/points_basic.py
Detected skylake derivative running on mesa i915. Clears to srgb textures will use manual shader clears.
thread '<unnamed>' panicked at src/conv.rs:1193:30:
invalid size
stack backtrace:
   0: rust_begin_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:72:14
   2: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   3: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   4: wgpuDeviceCreateBindGroup
   5: ffi_call_unix64
   6: ffi_call_int
   7: cdata_call
   8: _PyObject_Call
             at /usr/local/src/conda/python-3.12.4/Objects/call.c:367:19
   9: PyCFunction_Call
             at /usr/local/src/conda/python-3.12.4/Objects/call.c:387:12
  10: _PyEval_EvalFrameDefault
             at /home/conda/feedstock_root/build_artifacts/python-split_1718618246744/work/build-static/Python/bytecodes.c:3262:26
  11: _PyEval_EvalFrame
             at /usr/local/src/conda/python-3.12.4/Include/internal/pycore_ceval.h:89:16
  12: _PyEval_Vector
             at /usr/local/src/conda/python-3.12.4/Python/ceval.c:1683:12
  13: _PyFunction_Vectorcall
             at /usr/local/src/conda/python-3.12.4/Objects/call.c:419:16
  14: _PyObject_VectorcallTstate
             at /usr/local/src/conda/python-3.12.4/Include/internal/pycore_call.h:92:11
  15: method_vectorcall
             at /usr/local/src/conda/python-3.12.4/Objects/classobject.c:91:18
  16: <unknown>
  17: _ZN7QWidget5eventEP6QEvent
  18: _ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent
  19: _ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent
  20: _ZN14QWidgetPrivate14sendPaintEventERK7QRegion
  21: _ZN14QWidgetPrivate10drawWidgetEP12QPaintDeviceRK7QRegionRK6QPoint6QFlagsINS_14DrawWidgetFlagEEP8QPainterP21QWidgetRepaintManager
  22: _ZN14QWidgetPrivate13paintOnScreenERK7QRegion
  23: _ZN13QWidgetWindow5eventEP6QEvent
  24: _ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent
  25: _ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent
  26: _ZN22QGuiApplicationPrivate18processExposeEventEPN29QWindowSystemInterfacePrivate11ExposeEventE
  27: _ZN22QWindowSystemInterface22sendWindowSystemEventsE6QFlagsIN10QEventLoop17ProcessEventsFlagEE
  28: _ZL17xcbSourceDispatchP8_GSourcePFiPvES1_
  29: g_main_dispatch
  30: g_main_context_iterate_unlocked.constprop.0
  31: g_main_context_iteration
  32: _ZN20QEventDispatcherGlib13processEventsE6QFlagsIN10QEventLoop17ProcessEventsFlagEE
  33: _ZN10QEventLoop4execE6QFlagsINS_17ProcessEventsFlagEE
  34: _ZN16QCoreApplication4execEv
  35: <unknown>
  36: cfunction_vectorcall_NOARGS
             at /usr/local/src/conda/python-3.12.4/Objects/methodobject.c:481:24
  37: _PyObject_VectorcallTstate
             at /usr/local/src/conda/python-3.12.4/Include/internal/pycore_call.h:92:11
  38: PyObject_Vectorcall
             at /usr/local/src/conda/python-3.12.4/Objects/call.c:325:12
  39: _PyEval_EvalFrameDefault
             at /home/conda/feedstock_root/build_artifacts/python-split_1718618246744/work/build-static/Python/bytecodes.c:2714:19
  40: PyEval_EvalCode
             at /usr/local/src/conda/python-3.12.4/Python/ceval.c:578:21
  41: run_eval_code_obj
             at /usr/local/src/conda/python-3.12.4/Python/pythonrun.c:1722
  42: run_mod
             at /usr/local/src/conda/python-3.12.4/Python/pythonrun.c:1743
  43: pyrun_file
             at /usr/local/src/conda/python-3.12.4/Python/pythonrun.c:1643
  44: _PyRun_SimpleFileObject
             at /usr/local/src/conda/python-3.12.4/Python/pythonrun.c:433
  45: _PyRun_AnyFileObject
             at /usr/local/src/conda/python-3.12.4/Python/pythonrun.c:78
  46: pymain_run_file_obj
             at /usr/local/src/conda/python-3.12.4/Modules/main.c:360
  47: pymain_run_file
             at /usr/local/src/conda/python-3.12.4/Modules/main.c:379
  48: pymain_run_python
             at /usr/local/src/conda/python-3.12.4/Modules/main.c:629
  49: Py_RunMain
             at /usr/local/src/conda/python-3.12.4/Modules/main.c:709
  50: Py_BytesMain
             at /usr/local/src/conda/python-3.12.4/Modules/main.c:763:12
  51: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  52: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:360:3
  53: <unknown>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
fatal runtime error: failed to initiate panic, error 5
Aborted (core dumped)
```

See https://github.com/hmaarrfk/pygfx/pull/7/files for diff

```patch
diff --git a/examples/introductory/points_basic.py b/examples/introductory/points_basic.py
index 96eb8de21..06f87a157 100644
--- a/examples/introductory/points_basic.py
+++ b/examples/introductory/points_basic.py
@@ -21,8 +21,11 @@
 scene = gfx.Scene()
 
 positions = np.random.normal(0, 0.5, (100, 3)).astype(np.float32)
-sizes = np.random.rand(100).astype(np.float32) * 50
-colors = np.random.rand(100, 4).astype(np.float32)
+
+# But slice it beyond its length to test a (0, 3) array
+positions = positions[100:]
+sizes = np.random.rand(positions.shape[0]).astype(np.float32) * 50
+colors = np.random.rand(positions.shape[0], 4).astype(np.float32)
 geometry = gfx.Geometry(positions=positions, sizes=sizes, colors=colors)
 
 material = gfx.PointsMaterial(color_mode="vertex", size_mode="vertex")

```

almarklein:
Closes #19, closes #600, closes #812, closes #714.

## Tasks

* [x] Implement chunked buffer uploads.
* [x] Implement chunked texture uploads.
* [x] Unit tests for buffer.
* [x] Unit tests for texture.
* [x] Perform benchmarks to compare performance.
* [x] See if we can reduce chunksize to 1/32th of size?
* [x] Check if max chunk size degrades performance by increased Python looping.
* [x] Implement  `update_indices()`.
* [x] Remove ambiguity between NHW and HWC (e.g. using explicit size).
* [x] Refactor how rgb textures are handled.
* [x] Docs.
* [x] Support for big endian. 

Follow up:
* #828

## Benchmarks buffer

Before:
```
      upload_buffer_full_naive (20x) - cpu: 21.61
  upload_buffer_full_optimized (20x) - cpu: 21.69
    upload_buffer_full_noncont (20x) - cpu: 95.19
            upload_buffer_half (20x) - cpu: 10.25
    upload_buffer_two_quarters (20x) - cpu: 19.99
            upload_100_buffers (20x) - cpu: 21.16
```
After:
```
      upload_buffer_full_naive (20x) - cpu: 21.45
  upload_buffer_full_optimized (20x) - cpu: 19.52
    upload_buffer_full_noncont (20x) - cpu: 52.75
            upload_buffer_half (20x) - cpu: 10.03
    upload_buffer_two_quarters (20x) - cpu: 10.21
            upload_100_buffers (20x) - cpu: 23.25
```

Observations:
* The new `set_data()` method avoids a datacopy, though this only makes it a wee bit faster.
* Performance for non-contiguous data is less than for contiguous data, as expected, but it has improved nearly 2x!
* The chunking makes that updating the first and last quarter is about as fast as updating half.
* Updating many buffers seems a wee bit slower. Will look into that.

## Benchmarks texture

*will update as things progress*

Before:
```
       upload_tex1d_full_naive (20x) - cpu:  0.11 ms
   upload_tex1d_full_optimized (20x) - cpu:  0.12 ms
     upload_tex1d_full_noncont (20x) - cpu:  0.31 ms
           upload_tex1d_half_x (20x) - cpu:  0.12 ms
   upload_tex1d_two_quarters_x (20x) - cpu:  0.14 ms
                      divider1 (20x) - cpu:  0.00 ms
       upload_tex2d_full_naive (20x) - cpu: 12.26 ms
   upload_tex2d_full_optimized (20x) - cpu: 13.37 ms
     upload_tex2d_full_noncont (20x) - cpu:171.32 ms
           upload_tex2d_half_x (20x) - cpu: 10.75 ms
           upload_tex2d_half_y (20x) - cpu:  6.35 ms
   upload_tex2d_two_quarters_x (20x) - cpu: 12.39 ms
   upload_tex2d_two_quarters_y (20x) - cpu:  6.39 ms
                      divider2 (20x) - cpu:  0.00 ms
       upload_tex3d_full_naive (20x) - cpu: 27.25 ms
   upload_tex3d_full_optimized (20x) - cpu: 25.88 ms
     upload_tex3d_full_noncont (20x) - cpu:131.90 ms
           upload_tex3d_half_x (20x) - cpu: 20.47 ms
           upload_tex3d_half_y (20x) - cpu: 19.09 ms
           upload_tex3d_half_z (20x) - cpu: 12.77 ms
   upload_tex3d_two_quarters_x (20x) - cpu: 38.93 ms
   upload_tex3d_two_quarters_y (20x) - cpu: 19.63 ms
   upload_tex3d_two_quarters_z (20x) - cpu: 13.27 ms
```

After:
```
         upload_tex1d_full_naive (20x) - cpu:  0.12 ms
   upload_tex1d_full_optimized (20x) - cpu:  0.10 ms
     upload_tex1d_full_noncont (20x) - cpu:  0.19 ms
        upload_tex1d_quarter_x (20x) - cpu:  0.10 ms
     upload_tex1d_two_eights_x (20x) - cpu:  0.14 ms
                      divider1 (20x) - cpu:  0.00 ms
       upload_tex2d_full_naive (20x) - cpu: 13.92 ms
   upload_tex2d_full_optimized (20x) - cpu: 13.40 ms
     upload_tex2d_full_noncont (20x) - cpu: 95.69 ms
        upload_tex2d_quarter_x (20x) - cpu: 14.40 ms
        upload_tex2d_quarter_y (20x) - cpu:  3.64 ms
     upload_tex2d_two_eights_x (20x) - cpu: 13.97 ms
     upload_tex2d_two_eights_y (20x) - cpu:  3.20 ms
    upload_tex2d_four_eights_x (20x) - cpu: 13.38 ms
    upload_tex2d_four_eights_y (20x) - cpu:  6.51 ms
  upload_tex2d_chunk_stripes_x (20x) - cpu: 12.83 ms
  upload_tex2d_chunk_stripes_y (20x) - cpu:  7.22 ms
          upload_tex2d_random8 (20x) - cpu:  0.48 ms
         upload_tex2d_random16 (20x) - cpu:  0.98 ms
         upload_tex2d_random32 (20x) - cpu:  1.58 ms
         upload_tex2d_random64 (20x) - cpu:  3.18 ms
        upload_tex2d_random128 (20x) - cpu:  6.22 ms
        upload_tex2d_random256 (20x) - cpu: 11.78 ms
        upload_tex2d_random512 (20x) - cpu: 12.92 ms
       upload_tex2d_random1024 (20x) - cpu: 12.24 ms
       upload_tex2d_random2048 (20x) - cpu: 11.90 ms
     upload_tex2d_100_textures (20x) - cpu: 19.03 ms
                      divider2 (20x) - cpu:  0.00 ms
       upload_tex3d_full_naive (20x) - cpu: 29.33 ms
   upload_tex3d_full_optimized (20x) - cpu: 27.38 ms
     upload_tex3d_full_noncont (20x) - cpu: 81.43 ms
        upload_tex3d_quarter_x (20x) - cpu: 25.59 ms
        upload_tex3d_quarter_y (20x) - cpu: 26.64 ms
        upload_tex3d_quarter_z (20x) - cpu:  6.69 ms
     upload_tex3d_two_eights_x (20x) - cpu: 25.09 ms
     upload_tex3d_two_eights_y (20x) - cpu: 26.88 ms
     upload_tex3d_two_eights_z (20x) - cpu:  7.01 ms
    upload_tex3d_four_eights_x (20x) - cpu: 25.50 ms
    upload_tex3d_four_eights_y (20x) - cpu: 24.74 ms
    upload_tex3d_four_eights_z (20x) - cpu: 14.78 ms
  upload_tex3d_chunk_stripes_x (20x) - cpu: 25.70 ms
  upload_tex3d_chunk_stripes_y (20x) - cpu: 25.10 ms
  upload_tex3d_chunk_stripes_z (20x) - cpu: 14.07 ms
          upload_tex3d_random8 (20x) - cpu:  0.62 ms
         upload_tex3d_random16 (20x) - cpu:  0.97 ms
         upload_tex3d_random32 (20x) - cpu:  1.98 ms
         upload_tex3d_random64 (20x) - cpu:  3.64 ms
        upload_tex3d_random128 (20x) - cpu:  7.27 ms
        upload_tex3d_random256 (20x) - cpu: 18.46 ms
        upload_tex3d_random512 (20x) - cpu: 24.61 ms
       upload_tex3d_random1024 (20x) - cpu: 24.29 ms
       upload_tex3d_random2048 (20x) - cpu: 24.25 ms
```

Observations:
* The new `set_data()` method avoids a datacopy, though this only makes it a wee bit faster.
* Performance for non-contiguous data has improved a lot!
* Other than that, uploads are generally slower :/  work to do.

## API changes

* New `buffer.set_data()` and `texture.set_data()` methods.
* New `buffer.update_full()` and `texture.update_full()` methods.
* New `buffer.update_indices()` and `texture.update_indices()` methods.
* New `force_contiguous` arg for `Buffer()` and `Texture()` for cases where data is updated often.
* New `chuck_size` arg for `Buffer()` and `Texture()` to override the chunk size for syncing data.
* Can set `nitems` when creating a `Buffer` to override the interpreted number of items from the shape of the array.
* Can set `size` when creating a `Texture` to override the interpreted size from the shape of the array.
* The `.mem` attribute of `Buffer` and `Texture` is renamed to `.view` and is an `np.ndarray` instead of a `memoryview`. 
* It is no longer possible to create zero-sized buffers or textures (wgpu crashes when attempting to use one).
* Buffers and textures support data in big endian format.


almarklein:
Status: all the chunking stuff is implemented and tested. Next up is doing some benchmarking so I can tweak the implementation and chunk parameters for performance.

panxinmiao:
When users explicitly specify the size of a Texture, can we disregard the shape of the data, treat it solely as binary data, and use the user-specified size as the Texture size (with appropriate checks)? I believe this approach can offer better flexibility for dynamic texture data updates.
﻿
Additionally, I think it is crucial for users to explicitly specify the size when initializing a Texture. Of course, if users only provide data without specifying the size, our API can attempt to infer the size from the data's shape (using the current `size_from_data()` method). However, this inference is not always accurate. We should issue a warning and advise users to avoid relying on this method because, fundamentally, we cannot determine the texture's size attribute solely from the shape.
﻿
If `len(shape) == 4`, it is most likely in NHWC layout, where the size is (W, H) and the number of channels is C, which is straightforward.
﻿
If `len(shape) == 3`, theoretically, it could be NHW (n grayscale images) or HWC (a single bitmap). We cannot determine this definitively. We can only assume it is HWC (as is the case most of the time) and issue a warning.
﻿
Moreover, not only can N and C be 1, but H and W can also be 1. In such cases, if any dimension of 1 is collapsed, the situation becomes even more complicated.

almarklein:
The `dim` arg is meant to remove this ambiguity. It can indeed fix it for channels, but I realize now that it cannot really disambiguate for `N`, nor any collapsed dimensions. I think the latter is not that much of a problem, because the user can create a view that has singleton dimensions. But the former is trickier.  Thanks for bringing this up!

panxinmiao:
I think that when users explicitly specify the size of a Texture, we should disregard other conditions and adhere to the user-specified size, adjusting the data to fit this size. This approach can provide a fallback solution for most edge cases. 
Only when the user does not specify a size should we attempt to infer the size automatically, issuing a warning if there is any ambiguity.

almarklein:
> I think that when users explicitly specify the size of a Texture, we should disregard other conditions and adhere to the user-specified size, adjusting the data to fit this size.

Done! Also changed the internal memoryview to a ndarray instead, which is reshaped to the texture size, which simplifies the code in a few places.

> Even, we can force users to specify a size when create texture.

Note that `dim` is mandatory, and together with the shape, we can derive the texture size pretty well, if we simply assume its not a stack. This makes all stack-less use-case simpler.

almarklein:
Created benchmarks for texture. For now, the chunking eum ... negatively affects performance. I'll try to find out whether this is just due to chunking too much, or due to the overhead of calculating the chunks.

almarklein:
Pfew ... 😌

hmaarrfk:
What command do i run on my linux + gpu machines t report these benchmarking numbers that you have at the top?

almarklein:
Run this: https://github.com/pygfx/pygfx-benchmarks/blob/main/benchmarks/bm_texture.py

Korijn (approved):

PR #800, opened 2024-06-20
Refactor mesh shading code

panxinmiao:
To enhance the reusability of the code, facilitate the expansion of the current Mesh shading properties, and allow for the addition of more shading models in the future, it is necessary to refactor the current Mesh shading code. 

The main objectives of this refactor are:

1. Extract the Punctual light-related lighting logic for reuse across different shading models.
2. Use pointers for passing lights to avoid extensive parameter copying.
3. Temporarily move other Material-related shading properties and logic to the main file of the Mesh fragment shader. In the future, consider extracting more granular reusable code segments and using these segments to construct shading logic for each Mesh material (utilizing separate WGSL files as the framework for the material shading logic).


almarklein (approved):
IIUC none of the logic has changed; just moving code around, right?

panxinmiao:
> IIUC none of the logic has changed; just moving code around, right?

Yes, there are no any changes to the shading logic.

almarklein:
That's the validation examples doing what they're for 😉

pygfx/renderers/wgpu/wgsl/light_pbr.wgsl
```diff
@@ -100,83 +100,42 @@ fn computeMultiscattering(normal: vec3<f32>, view_dir: vec3<f32>, specular_color
 }
 
 fn RE_IndirectSpecular_Physical(radiance: vec3<f32>, irradiance: vec3<f32>,
-        geometry: GeometricContext, material: PhysicalMaterial, reflected_light: ReflectedLight) -> ReflectedLight{
+        geometry: GeometricContext, material: PhysicalMaterial, reflected_light: ptr<function, ReflectedLight>){
```
Korijn, line 103: I had to look up the pointer type definition syntax to understand this, I thought it was some kind of to-from mapping (function that returns a type). But instead, `function` means the data is stored in the function's address space. Is that correct?

Korijn (approved):

Korijn:
@almarklein please hit the merge button if you are also OK with these changes!

pygfx/renderers/wgpu/wgsl/light_pbr.wgsl
panxinmiao: Yes, it is a local variable, using memory locations in the "function" address space.

See:  https://www.w3.org/TR/WGSL/#ref-ptr-types

almarklein:
Thanks @panxinmiao !

PR #801, opened 2024-06-25
Improve spelling for render_mask in blender.py

hmaarrfk:
Not sure how you feel about enums in 2024, but in case you wanted this refactor.


almarklein (approved):
Nice! The `RendeMask` enum was implemented later, which is why some code still uses ints :)

Korijn (approved):

Issue #661, opened 2024-02-27
About documenting properties


almarklein:
Our materials (and a few other classes) have a lot of (settable) properties. We have made a point to make it possible to set each property at object instantiation, e.g. `LineMaterial(dash_pattern=[0, 3], color='red')`.

This means there are two places to document each property:
* In the docstring of the property itself.
* In the `Parameters` section of the docstring of the class.

I don't think we want these to be a copy of each-other. But what would be a good rule of thumb to follow?

My thoughts:
* A short one-line summary in both cases.
* Detailed explanation in the property docs.
* For enums, list the possible values in both cases.
* For somewhat complex values (e.g. `dash_pattern`) refer to property docs. Maybe provide small example.
* Default value in the parameter docs.

almarklein:
Rethinking this: Let's just make a point to add each property to the constructor, with its default value. But in the class docs we simply say that *properties values can be passed as keyword arguments*.

PR #803, opened 2024-06-28
Add MeshToonMaterial

panxinmiao:
This PR introduces `MeshToonMaterial` and implements toon shading.

image

### DIFF
--- a/pygfx/materials/_mesh.py
+++ b/pygfx/materials/_mesh.py
@@ -506,8 +506,113 @@ class MeshPhongMaterial(MeshBasicMaterial):
     # TODO: more advanced mproperties, Unified with "MeshStandardMaterial".
 
 
-# todo: MeshToonMaterial(MeshBasicMaterial):
-# A cartoon-style mesh material.
+class MeshToonMaterial(MeshBasicMaterial):
+    """
+    A material implementing toon shading.
+
+    Parameters
+    ----------
+    emissive : Color
+        The emissive (light) color of the mesh. This color is added to the final
+        color and is unaffected by lighting. The alpha channel of this color is
+        ignored.
+
+    gradient_map : Texture
+        Gradient map for toon shading. The gradient map sampler method is always 'nearest'.
+
+    kwargs : Any
+        Additional kwargs will be passed to the :class:`base class
+        <pygfx.MeshBasicMaterial>`.
+    """
+
+    uniform_type = dict(
+        MeshBasicMaterial.uniform_type,
+        emissive_color="4xf4",
+        normal_scale="2xf4",
+        emissive_intensity="f4",
+    )
+
+    def __init__(self, emissive="#000", gradient_map=None, **kwargs):
+        super().__init__(**kwargs)
+
+        self.emissive = emissive
+        self.emissive_map = None
+        self.emissive_intensity = 1.0
+
+        self.normal_map = None
+        self.normal_scale = (1, 1)
+
+        self.gradient_map = gradient_map
+
+    @property
+    def emissive(self):
+        """The emissive (light) color of the mesh.
+        This color is added to the final color and is unaffected by lighting.
+        The alpha channel of this color is ignored.
+        """
+        return Color(self.uniform_buffer.data["emissive_color"])
+
+    @emissive.setter
+    def emissive(self, color):
+        color = Color(color)
+        self.uniform_buffer.data["emissive_color"] = color
+        self.uniform_buffer.update_range(0, 1)
+
+    @property
+    def emissive_intensity(self):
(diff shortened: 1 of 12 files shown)


panxinmiao:
Ready for review now :)

pygfx/renderers/wgpu/shaders/meshshader.py
```diff
@@ -179,6 +188,33 @@ def get_bindings(self, wobject, shared):
         sampler = GfxSampler(material.map_interpolation, "repeat")
         texturing = "texture/auto"
 
+        # todo, check uv is present and has the right shape. (introduce uv channel for texture)
+        # specular map configs, for basic and phong materials
+        if getattr(material, "specular_map", None) and not isinstance(
+            material, MeshStandardMaterial
+        ):
```
almarklein, line 195: Is this todo still relevant?

Also maybe add a comment on why this `if` is not entered for `MeshStandardMaterial`.

pygfx/renderers/wgpu/shaders/meshshader.py
```diff
@@ -143,7 +150,9 @@ def get_bindings(self, wobject, shared):
             bindings.append(Binding("s_colors", rbuffer, geometry.colors, "VERTEX"))
         elif self["color_mode"] in ("vertex_map", "face_map"):
             bindings.extend(
-                self.define_texcoords_and_colormap(
+                # todo: check uv is present and has the right shape. (introduce uv channel for texture)
+                # todo: unify the logic with other maps, introduce uv channel and texture sampler
```
almarklein, line 154: Is this todo still relevant? Looks like the texcoords are handled correctly, and it is checked for compatibility with the map in `define_colormap()`.

pygfx/materials/_mesh.py
```diff
@@ -506,8 +506,113 @@ def shininess(self, value):
...
+    """
+    A material implementing toon shading.
+
+    Parameters
+    ----------
+    emissive : Color
+        The emissive (light) color of the mesh. This color is added to the final
+        color and is unaffected by lighting. The alpha channel of this color is
+        ignored.
+
```
almarklein, line 519: Let's also include `normal_map`?

almarklein (commented):
Nice work!

pygfx/renderers/wgpu/shaders/meshshader.py
panxinmiao: My idea is to introduce the `uv_channel` attribute in textures first (to identify different texcoord sets), and based on `uv_channel`, determine if the corresponding texcoord exists and matches the shape of this texture.

Once these steps are completed, the logic for the `colormap` should align with that of other textures. This is just a record of the plan.

pygfx/renderers/wgpu/shaders/meshshader.py
panxinmiao: > Also maybe add a comment on why this `if` is not entered for `MeshStandardMaterial`.

Ah this is redundant code (overlooked during cherrypicking from #587). 😅 


We can introduce the `specular_map` property for `MeshBasicMaterial` and `MeshPhongMaterial`. The lighting model of `MeshPhongMaterial` requires the `specular_strength` parameter. Additionally, the shading logic for `env_map` in `MeshBasicMaterial` and `MeshPhongMaterial` also requires the `specular_strength` parameter. Currently, `specular_strength` is hardcoded to 1.0 but can be read from `specular_map`.

However, the lighting model of `MeshStandardMaterial` does not require `specular_strength`, and its `env_map` uses IBL (Image-Based Lighting) technology, thus does not need the `specular_strength` parameter.

 I will remove this code in this PR, and I plan to add support for `Specular_map` to `MeshBasicMaterial` and `MeshPhongMaterial` in another PR.

pygfx/materials/_mesh.py
panxinmiao: I've noticed that the constructor for `MeshStandardMaterial `also lacks parameters for `normal_map` (and some other properties aren't listed in the Material constructor's parameters either). It seems like there isn't a standardized approach here.

For advanced materials, there may be many optional properties (which might be added gradually). Should all these optional properties be listed in the constructor method? Perhaps the constructor parameters should only include the most crucial properties of the current material (those that users need to focus on and significantly affect the material's appearance)?

almarklein: Good point. The approach is that all properties are settable in the constructor. So the other missing cases should (at some point) be fixed.

As for the docs, I created #661 earlier. I think we should stop duplicating the docstrings. I added a comment in that issue.

For now, in this PR, maybe just add the most important once, and we can correct it all when we implement #661

Korijn (approved):

PR #804, opened 2024-06-29
Sdf clipping planes

hmaarrfk:
This implements clippings planes using SDFs which my understanding is that they are more efficient than if statements on GPUs.

Pretty hard to benchmark though...


almarklein:
Nice!

> Pretty hard to benchmark though...

Let's try though! E.g. https://github.com/pygfx/pygfx-benchmarks/blob/main/benchmarks/bm_points.py

almarklein (approved):
Happy that the benchmarks show a positive effect! https://github.com/pygfx/pygfx-benchmarks/pull/1

almarklein:
And happy that the work I put in the benchmark mechanics is also paying off in decisions like this!

Korijn (approved):

PR #807, opened 2024-07-07
Add the name property to more objects

hmaarrfk:
I found this while trying to troubleshoot some code of mine.

### DIFF
--- a/pygfx/objects/_more.py
+++ b/pygfx/objects/_more.py
@@ -24,8 +24,7 @@ class Group(WorldObject):
     """
 
     def __init__(self, *, visible=True, name=""):
-        super().__init__(visible=visible)
-        self.name = name
+        super().__init__(visible=visible, name=name)
 
 
 class Scene(Group):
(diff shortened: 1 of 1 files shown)


pygfx/objects/_more.py
```diff
@@ -35,8 +37,8 @@ class Scene(Group):
 
     """
 
-    def __init__(self):
-        super().__init__()
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
hmaarrfk, line 41: this sneaks in the ability to start the scene off as "invisible" too. Not sure if this is a welcome change or not.

pygfx/objects/_more.py
```diff
@@ -335,6 +337,8 @@ class Text(WorldObject):
         recommended to let the renderer decide, using "auto".
     position : Vector
         The position of the object in the world. Default (0, 0, 0).
+    name : str
+        The name of the text object for inspection and debugging.
```
hmaarrfk, line 341: I tried to differentiate the "name" string, from the "text string".

pygfx/objects/_more.py
almarklein: Looks reasonable to me!

almarklein (approved):

Korijn (approved):

PR #809, opened 2024-07-07
Question: Demonstrate edge cases in line dashing

hmaarrfk:
It seems that if you use a pixel_ratio of 1, and a line thickness of 1 the dashing miraculously disappears???

I haven't spent too much time investigating this, but I was curious as to why my dashing went away when I set the pixel_ratio to 1 (was trying to see if I could speedup my application that way)

is this a bug or expected behavior?

Screenshot from 2024-07-07 16-22-54

### DIFF
--- a/examples/validation/validate_line_dash.py
+++ b/examples/validation/validate_line_dash.py
@@ -15,8 +15,11 @@ from wgpu.gui.auto import WgpuCanvas, run
 import pygfx as gfx
 
 
-canvas = WgpuCanvas(size=(1000, 1000))
+pixel_ratio = 2.
+canvas = WgpuCanvas(size=(1000, 1000), title=f"Dashing {pixel_ratio=:}")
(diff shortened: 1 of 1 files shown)


almarklein:
If the line thickness is just a bit larger, it will have dashing, although it will all look a bit "crumbly".

When the lines is one pysical pixel wide (or less), this line kicks in:

https://github.com/pygfx/pygfx/blob/cba5dd1d0c017c414c92497a6efd8b7d2f9a9fec/pygfx/renderers/wgpu/wgsl/line.wgsl#L791-L800

And ... there's no variable here that makes it affected by dashing. This is a bug.

almarklein:
Fixed in #811

hmaarrfk:
awesome!

hmaarrfk:
Here a screenshot of the result.

Thanks Almar!

image

PR #811, opened 2024-07-08
Fixed that dashing was not applied to thin lines

almarklein:
See #809


Korijn:
maybe add a validation example?

almarklein:
Ready

Korijn (approved):

PR #814, opened 2024-07-11
unproject to get range of screen space to set rulers

kushalkolar:
This modifies the ruler example to unproject from screen to world space to determine the range of the current world space. The current example does not account for camera zoom. I guess that by unprojecting we should be able to cover all cases?

For example if we use the [current ruler example](https://github.com/pygfx/pygfx/blob/792d6ed36788f52776e159ed0d0729563805e0d2/examples/feature_demo/line_plot.py) and just add a `camera.zoom = 0.5` at the end we get:

image


almarklein (approved):
Nice!

Korijn (approved):

PR #817, opened 2024-07-11
Allow a minimum number of fonts to always be cache

hmaarrfk:
I feel like we know we want "at least 1 font". (which is already the case), but I think that users want to load many other fonts (italics and maybe a few more at least).

We use the PyGFX Stats, so that creates "at least 1 font more than our application's font" and the continuous eviction creates random lags.

I also have some experience creating caches, so I spent a bit of time refactoring things. I hope that is "appreciated".

I feel like "20MB of fonts" for a graphics intensive python application is "fine".

Before this, if I would wait 10 seconds, I would get the following evictions:

```
Removing /usr/share/fonts/truetype/noto/NotoSans-Regular.ttf from cache
Removing /usr/share/fonts/truetype/noto/NotoSans-Regular.ttf from cache
```

With the following changes, I no longer get it.

I would say that an LRU cache is likely enough, but the LRU cache might not work very well if the user is flipping through many many fonts quickly


hmaarrfk:
maybe i should have not done the refactor, now i have to "fix" the tests ...

pygfx/utils/text/_shaper.py
```diff
@@ -104,16 +152,8 @@ def shape_text_hb(text, font_filename, direction=None):
...
-    if cached:
-        blob, face, font = cached
-    else:  # Cache miss
-        blob = uharfbuzz.Blob.from_file_path(font_filename)
-        face = uharfbuzz.Face(blob)
-        font = uharfbuzz.Font(face)
-        font.scale = ref_size, ref_size
-        CACHE_HB.set(font_filename, (blob, face, font))
+    # Load font, maybe from the cache
+    blob, face, font = CACHE_HB[font_filename]
```
hmaarrfk, line 156: Generally, I really like the `lru_cache` from python's STL, but it doesn't give enough tools for inspection of the cache and tuning.

So I've always resorted to building my own custom ones that I could invalidate based on other logic (such as time in this case).

hmaarrfk: One aspect of the lru_cache is that you don't think about the cache getter and setter, it just "does it for you".

hmaarrfk:
On main

https://github.com/user-attachments/assets/a340c50a-ee01-4113-951f-b0f4a8e02770

<details><summary> print statements </summary>

```
diff --git a/pygfx/utils/_dirs.py b/pygfx/utils/_dirs.py
index 73e201f..d61c592 100644
--- a/pygfx/utils/_dirs.py
+++ b/pygfx/utils/_dirs.py
@@ -22,7 +22,9 @@ def get_resources_dir():
     with context as path:
         pass
     # Return the dir. We assume that the data files are on a normal dir on the fs.
-    return str(path.parent)
+    directory = str(path.parent)
+    print(directory)
+    return directory
 
 
 def get_cache_dir():
diff --git a/pygfx/utils/text/_shaper.py b/pygfx/utils/text/_shaper.py
index 418ac74..f5e99a3 100644
--- a/pygfx/utils/text/_shaper.py
+++ b/pygfx/utils/text/_shaper.py
@@ -84,6 +84,7 @@ class TemporalCache:
             reverse=True,
         )
         for key, lt in to_remove[self._minimum_font_files :]:
+            print(f"Evicting {key}")
             self._lifetimes.pop(key, None)
             self._cache.pop(key, None)

```
</details>


https://github.com/user-attachments/assets/8dba591b-4be5-4a18-a4f2-663fdd79a44b

pygfx/utils/text/_shaper.py
```diff
@@ -73,8 +78,12 @@ def check_lifetimes(self):
         # it slower already for 100 items, which is way more realistic
         # font count. So let's keep things simple :)
         pretty_old = time.time() - self._ref_lifetime
-        to_remove = {key for key, lt in self._lifetimes.items() if lt < pretty_old}
-        for key in to_remove:
+        to_remove = sorted(
+            ((key, lt) for key, lt in self._lifetimes.items() if lt < pretty_old),
+            key=lambda x: x[1],
+        )
+        items_to_remove = max(len(self._cache) - self._minimum_items, 0)
```
almarklein, line 85: I'm not sure I follow the logic here. The `._minimum_items` applies to the subset of items that are older than `pretty_old`.  So I think this should be something like:

```py
        n_pretty_new = len(self._cache) - len(to_remove)
        items_to_remove = max(n_pretty_new - self._minimum_items, 0)
```

pygfx/utils/text/_shaper.py
```diff
@@ -41,29 +41,34 @@ def shape_text(text, font_filename, direction=None):
...
+    def __getitem__(self, key):
+        """Gets the object corresponding to the given key.
+
+        Will reset the item's lifetime.
         Getting triggers the lifetimes of all items to be checked.
         """
-        try:
+        if key in self._cache:
             res = self._cache[key]
-        except KeyError:
```
almarklein, line 56: From what I understand, the `try` avoids a double lookup for cache hits. Or is there another reason for preferring `if` here?

pygfx/utils/text/_shaper.py
almarklein: Yeah, this is nice!

pygfx/utils/text/_shaper.py
hmaarrfk: Centralizing the one way to fill the cache with valid data. 

I mean. You use your cache in 3 palaces. So it isn't so hard. But generally you want to define the function once and just forget about it

hmaarrfk: i'm happy to add back the try except. let me do that now.

pygfx/utils/text/_shaper.py
hmaarrfk: I don't think it works out.

consider:

* len(self._cache) = 10
* 9 items are "too old"
* minimum_items = 5
* We thus want at the end to have 5 items in our cache limited by `minimum_items`

thus
* n_pretty_new = 10 - 9 = 1
* items_to_remove = max(1 - 5, 0) = max(-4, 0)
* so you remove nothing???

I renamed the varaible to `max_items_to_remove`

The slicing operator effectively does

```
items_to_remove = min(max_items_to_remove, len(pretty_old))
```

pygfx/utils/text/_shaper.py
```diff
@@ -73,8 +78,13 @@ def check_lifetimes(self):
...
         pretty_old = time.time() - self._ref_lifetime
-        to_remove = {key for key, lt in self._lifetimes.items() if lt < pretty_old}
-        for key in to_remove:
+        to_remove = sorted(
+            ((key, lt) for key, lt in self._lifetimes.items() if lt < pretty_old),
+            key=lambda x: x[1],
+        )
+        max_items_to_remove = max(len(self._cache) - self._minimum_items, 0)
+
+        for key, lt in to_remove[:max_items_to_remove]:
```
almarklein, line 87:
```suggestion
        old_items = sorted(
            ((key, lt) for key, lt in self._lifetimes.items() if lt < pretty_old),
            key=lambda x: x[1],
        )
        max_items_to_remove = max(len(self._cache) - self._minimum_items, 0)

        for key, lt in old_items[:max_items_to_remove]:
```

pygfx/utils/text/_shaper.py
almarklein: My brain was thinking in the wrong direction. Now I get it. The `max_items_to_remove` is simply how many items can we remove, let's removed this many items, starting with the oldest. The renaming helps. I applied another rename to help further.

almarklein (approved):

hmaarrfk:
Thank you!

PR #818, opened 2024-07-11
Make hte resources path more resilient to changing python environment

hmaarrfk:
Ok this is quite an edge case but consider this:

1. An early adopter developer trying to test your code.
2. They install some supported version.
3. They need to test some example on the main branch. (such as your cool new ruler)
4. They run `pip install -e . -vv` to using the main branch in editable mode
5. The font caching engine then tries to re-open a a Noto-font file but can't resolve the resource correctly.

I feel like using `__file__` is likely easy enough but maybe it isn't comptaible with some pyinstaller stuff?

Trying to test on the CIs.

### DIFF
--- a/examples/feature_demo/multiple_fonts.py
+++ b/examples/feature_demo/multiple_fonts.py
@@ -1,0 +1,63 @@
+"""
+Multiple Fonts
+==============
+
+Example demonstrating the capabilities of text to be rendered in
+multiple different fonts.
+"""
+
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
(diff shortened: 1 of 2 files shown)


almarklein:
I think the use-case you outline should work. I also use `pip install -e .`. In what way does it fail?

hmaarrfk:
Because I already have an other window of my application open at the same time ;) do the existing application breaks. 

As I said. Very bad bad behavior from my part ;)

hmaarrfk:
Let me try to take a screen cast at some point today

hmaarrfk:
I'm not really sure what was breaking in my shader code before. I can't recreate right at this moment.

I'll have to circle back to this once i have a recreator and a fix.

Issue #799, opened 2024-06-19
Image with the size set by the material


hmaarrfk:
In an offline conversation with Almar he had asked me if it would be beneficial to have an image object who's size in world space was determined by a property different than the geometry grid. 

I think I've found our main usecase:

1. Generally, we might have a surface, be-it an image, that we might be interested in replacing the texture with an other one.
2. However, the size of the texture may change quite dynamically with the usage of the system. Think of "preview mode" in a biomedical engineering context and "record mode". One may have much lower pixel count, but you still want the illusion that you are capture the same region of interest in "physical space".
3. We had been historically swapping the geometry size, then appropriately swapping the scale.
4. The problem is that the fact of changing the "scale" of the image object, may change the location of any children that you want to place along the texture itself.

Our solution has been to:

1. Always embed the image in a larger group. This is somewhat tiring and the abstract nature is hard communicate across our team (even I get confused, and I mostly recreated the hierarchy). The higher level group stays "constant" in shape, making it easier to add labels and whatnot
2. Constantly changing the scale of the image object is easy to forget. The typical workflow consists of
    - Change the image geometry.
    - Change the scale.
    - It would be preffarable to "just change the geometry's grid"

To implement this "scale invariant image" my general strategy has been:

1. Add a property to the image material  called `shape_mode` (the name could be better, suggestions appreciated). It can change between `geometry` and `material`.
2. Add a property to the image material called `image_shape` that sets the extent of the image shape (shape = yx, size = xy).
3. Then for the Image world object, switch between grabbing the geometry `bounding_box` to that of the material.

```python
class MyImage(gfx.Image):
    def get_bounding_box(self):
        if self.material.shape_mode == "material":
            image_shape = self.material.image_shape
            return np.asarray([
                [0, 0, 0],
                [image_shape[1], image_shape[0], 0]
            ], dtype='float64')
        else:
            return super().get_bounding_box()
```
5. The shader gets modified with:

```wgsl
            $$ if shape_mode == "geometry"
                let pos = vec2<f32>(size.xy);
            $$ else
                let pos = vec2<f32>(
                    f32(u_material.image_shape[1]),
                    f32(u_material.image_shape[0])
                );
            $$ endif
            geo.positions = array<vec3<f32>,4>(
                vec3<f32>(pos.x, 0., 0.0),
                vec3<f32>(pos.x, pos.y, 0.0),
                vec3<f32>(0., 0., 0.0),
                vec3<f32>(0., pos.y, 0.0),
[...]

PR #820, opened 2024-07-13
Example to demo differences between the image shader and mesh

hmaarrfk:
Closes: https://github.com/pygfx/pygfx/issues/799


Korijn:
I wonder if a grid would be useful here to show the 0.5 pixel offset?

hmaarrfk:
I didn't want to be too negative about the grid object, but I currently find it quite jarring.

Moving around in the example i can generate the following displays:

```
$ python examples/feature_demo/line_plot.py
```

image
image
image
image

It makes me really lose track of what the grid is about, and so I was hoping to take a look at it eventually before using it in a real example.

My attempt at providing a visual guide is with a red line at the top.

examples/feature_demo/image_scaling.py
```diff
@@ -0,0 +1,85 @@
+"""
+Displaying an image in a scene
+==============================
+
+Displaying an image can be done one of two different ways based the details
```
almarklein, line 5:
```suggestion
Displaying an image can be done in two different ways based the details
```

examples/feature_demo/image_scaling.py
```diff
@@ -0,0 +1,85 @@
...
+is the important information. By default, nearest neighbor interpolation
+is used to sample the texture.
+
+A gfx.Mesh with a plane geometry can be used to display the texture with
+spanning from the one edge of the geometry to an other.
+The pixel corners at integer coordinates. This is useful when one wants
+to stretch the provided data to fill a specific area. Be default bilinear
+interpolation is used to sample the texture.
+
+The example provide allows users zoom in and out of the scene to study the
```
almarklein, line 22:
```suggestion
The example allows users zoom in and out of the scene to study the
```

almarklein (commented):
Just 2 typos. Otherwise looks like a nice addition to the gallery!

hmaarrfk:
> Just 2 typos. Otherwise looks like a nice addition to the gallery!


ps. you can just push in these case to avoid you having to come back to this PR. your choice.

Korijn (approved):

PR #822, opened 2024-07-14
Cleanup the point shader to enable people to better customize it

hmaarrfk:
* m_sqrt_2 has been added as a const in std -> `SQTR_2`
* m_pi has been replaced by the existing constant in std -> `PI`
* half_size was used in two locations. I replaced them with `size * 0.5` directly.
* coord was used to invert the y axis. I manually inverted the y axis in all the SDFs
* input parameter has been rename `coord` instead of `coord_in` to make it easier to use.

This should result in 0 changes to any tests, but make it easier to add a custom template in as is being suggested in https://github.com/pygfx/pygfx/pull/789


almarklein (approved):
Thanks!

Korijn (approved):

hmaarrfk:

PR #823, opened 2024-07-21
Provide tools to debug SDFs

hmaarrfk:
I kept using shadertoy to debug my personal SDFs. It gets annoying to translate between GSL languages.

I'll keep working on this later, fun weekend math project. But i'm leaving it here in case anybody is tempted in using to make pixel perfect SDFs.

The biggest issues you see with the current SDFs is when you try to make the edge width much much thicker than the point.
You start to see the issues that IQ warns you about.

He would prefer a more exact definition of the heart.
https://iquilezles.org/articles/distfunctions2d/

The zoom debugging tool also shows why the pin looks cut off, and may motivate the need to ~expand the triangle vertices to accomodate for the edge thickness overall~ (my debugging shader is currently wrong and it seems that the painting does extend further than I thought. tbd).

<details><summary>old video demo</summary>

https://github.com/user-attachments/assets/8675181e-ad8a-4ce8-8909-42b15d756a22
</details>

https://github.com/user-attachments/assets/f6304f66-1289-40a0-918b-f037ac4617a6


hmaarrfk:
ps you can also see artifacts from fonts. we'll get to a pixel perfect font shader on day ;)

Korijn:
> I kept using shadertoy to debug my personal SDFs. It gets annoying to translate between GSL languages.

https://github.com/pygfx/shadertoy

You can use the same shader language in this shadertoy you know! :)

Korijn:
> ps you can also see artifacts from fonts. we'll get to a pixel perfect font shader on day ;)

There will always be limits to resolution as long as we use masks or sdfs. You'd have to render directly from curve data to get truly perfect results, but that's not nearly as performant.

Right now, pygfx renders using single channel sdfs as generated by freetype. You can squeeze more information into an sdf using multiple channels, but the implementation complexity goes up an order of magnitude. See here: https://github.com/Chlumsky/msdfgen

hmaarrfk:
> You can use the same shader language in this shadertoy you know! :)

Thats awesome! I saw you were able to recreate his iconic 3D SDFs
https://github.com/pygfx/shadertoy/blob/main/examples/shadertoy_glsl_sdf.py
I realized that I had cloned this repo locally but didn't really understand how to write shaders.

> You can squeeze more information into an sdf using multiple channels, but the implementation complexity goes up an order of magnitude.

Thanks for the reference on the text. Very interesting!

I think some of my biggest challenges on the text right now is that the font doesn't look correctly spaced out (some characters look too close together). But since I understand your code I was going to debug it first.

Some background
In short I got an SDF for our personal logo to be written, but I'm trying for the stretch goal of making it exact.
IQ's notes indicate that when you try to draw contours around the SDF and your starting SDF isn't exact, then you start to get very strange looking results. in fact, this is what I saw when I tried to turn on the "edge" and the outline showed similar artifacts as the heart does. For this, and other reasons, I had to change my personal strategy for our logo. Ultimately, its pretty hard to get parametric shapes to look good no matter what....

https://github.com/user-attachments/assets/a7ba1c9d-4dc8-42b7-855a-ad08d3bbc743

It doesn't escape me that likely nobody cares about 1-2 pixels, but ultimately this mental exercise is helping me understand some visualization tricks you and IQ have implemented in various shaders. Maybe one day I'll understand the tricks behind pristine grid....

hmaarrfk:
What are people's gut feelings on this "feature"

almarklein (approved):
> What are people's gut feelings on this "feature"

I think its a relatively small change that allows tweaking the sdfs *much* better, so thanks!

Korijn (approved):

PR #824, opened 2024-07-21
Allow users to customize which edge of the SDF is painted

hmaarrfk:
So I definitely spent way too much time on this, but the mods were easy and I wanted to explore what would happen as I painted different parts of my SDF.

In fact, some references https://iquilezles.org/articles/interiordistance/ discuss that SDFs are not always correct on the interior so it may be helpful to provide ways for users to paint their SDFs only on the outside. Getting a mathematically correct SDF definitely adds beauty, but it is pretty challenging.

I've expanded the validation example to show some of the difference.

You can play with with the edge with as well.

https://github.com/user-attachments/assets/1709764c-3567-4fd4-92de-2874c5044471

P.S. You should likely tell me if these PRs are welcome, happy to just maintain my own shaders and classes based on your infrastructure. I do think they are helpful in general and will be helpful to the science community at large.


pygfx/renderers/wgpu/shaders/pointsshader.py
```diff
@@ -57,6 +57,16 @@ def __init__(self, wobject):
...
 
+        edge_mode = str(material.edge_mode)
+        if edge_mode == "inner":
+            self["edge_mode"] = "inner"
+        elif edge_mode == "outer":
+            self["edge_mode"] = "outer"
+        # elif edge_mode == "centered":
+        # elif edge_mode == "auto":
+        else:
+            self["edge_mode"] = "centered"
```
Korijn, line 68: Why not just make the default centered? It doesn't seem like auto actually does anything clever?

hmaarrfk: i was following the color mode pattern, happy to change this, i got bored of the "auto" too by the end.

hmaarrfk:
hmmm. text seems to be difficult to get to render consistently on the CIs.... the error seems to be associated with the text....

Korijn:
Make sure to pull the screenshots from the special CI job "Screenshots / Regenerate"

hmaarrfk:
> Make sure to pull the screenshots from the special CI job "Screenshots / Regenerate" (here for your latest build, under "Artifacts": https://github.com/pygfx/pygfx/actions/runs/10031557567)

I did, but my local git thinks that the replaced file is byte for byte identical.

Korijn:
There's no randomness here. If the fonts are different... are both CI jobs using the same exact version of freetype-py (and all other packages for that matter)? You can check the logs.

hmaarrfk:
I was just about to say that I think there is something different in Freetype or the font since it seems to have to do with the "a", but honestly its really strange since there are many other "a"s in there...
https://github.com/user-attachments/assets/92e84a9c-0299-4257-8756-ce7fec39d110

hmaarrfk:
I don't know, the logs aren't that different..... it is possible 

It could be that it is clashing between system fonts and some pypi shipped font.
Locally I get:
```
get_hb_face: font_filename: /usr/share/fonts/truetype/noto/NotoSans-Regular.ttf
get_ft_face: font_filename: /usr/share/fonts/truetype/noto/NotoSans-Regular.ttf
```

hmaarrfk:
Could there be something different in font choice?

image

hmaarrfk:
could it be that the font atlas is getting corrupted?

.github/workflows/ci.yml
```diff
@@ -146,11 +146,20 @@ jobs:
...
         python -c "from examples.tests.testutils import wgpu_backend; print(wgpu_backend)"
+    # We make sure to run the compare validation examples section first
+    # Since it seems that there is some dependency on the order in which
+    # the tests are run. It is unclear what is causing it but this order now
+    # matches the generation of the screenshots which seems to help stabilize
+    # the tests.
+    # https://github.com/pygfx/pygfx/pull/824
+    - name: Compare validation examples
+      run: |
+        pytest -s -v examples -k test_examples_compare
```
hmaarrfk, line 157: I feel like you are so thorough @Korijn that I don't need to point you to this change but this is what resolved the issue for me......

I'm not sure if you have any ideas...

Korijn: Well, firstly you overlooked duplicating this part of the pipeline step:

```
      env:
        PYGFX_EXPECT_LAVAPIPE: true
```

Secondly, if the order matters, it's usually related to caching. Maybe the font manager evicts the font and reloads it, and then the SDFs generated ny freetype-py are different the second time around? Could be that there is some randomness in the SDF generation algorithm as a performance optimization.

Korijn:
>  Screencast.from.2024-07-21.18-33-48.mp4

Only those few pixels in that one `a` are wrong?

You could use numpy to subtract one image from the other and see "how wrong" they are. Sometimes it says something.

Korijn:
This is really odd though. Why is that super local group of pixels wrong? Are they still wrong if you only render the background? Or if we only render the markers? Or only the text (without background)? Etc. we have to isolate the problem.

.github/workflows/ci.yml
hmaarrfk: This is only used in one actual test that checks that asserts that the lava pipe pipeline is used. 

I'm happy to add it here and the regeneration worlflow

hmaarrfk:
> we have to isolate the problem.

I'm generally supportive of this idea, but I thought you had acknowledged that weird things happen with the lava pipe pipeline. 

We had similar font strangeness in the past when we looked at z fighting. Given the construction it should have been 100% repeatable (we were comparing a static scene) but it wasn't. 


Debasing might have deleted some trials, but I had compared trying to turn off antialiasing but it didn't resolve the problem either. 

If you really want to get to the bottom of it, i can try to look into running locally with docker

hmaarrfk:
In the latest tests, i show that if you run *just the comparison examples* then it passes, but as soon as you run then in the same call as the other tests things start to fail.

almarklein:
My guess is that this is either a lavapipe glitch, or a corruption in the atlas. Looks like both cases are rare and you were fortunate to somehow trigger it with these changes 😄 

What if you change some of the text in that example, would be interesting to see what happens.

This is an opportunity to investigate what might be a (rare) bug. But I also understand if you have better things to do 😅

hmaarrfk:
> This is an opportunity to investigate what might be a (rare) bug. But I also understand if you have better things to do 😅

I agree..... but this might need things like a bump in Ubuntu...... or moving to more pacakges from conda. I would mostly like to keep this distinct from this very separate feature which focuses on SDFs.

I would like to suggest:
1. Merging  commits 1- 4
2. Pull request with 3e9d3309e2d8176a755dd26c5386424faaca9428 demonstrating the bug and giving us a playground to explore the bug.

We have overall just struggled with fonts (beyond just with PyGFX, mostly with Qt) ourselves opting to move to SVGs in many cases to resolve these discrepancies. Beyond just fonts, with SDFs, I've found that if you are at an edge `<` and `<=` can make a huge difference in the rendering results. Something like that might be happening here.

Troubleshooting this will have to wait for when I have time to dive into it on my end. My deep dive into SDFs this weekend got me to create these different edge painting modes and the debugging mode in https://github.com/pygfx/pygfx/pull/823

hmaarrfk:
I revived https://github.com/pygfx/pygfx/pull/719 and added a commit to disable system fonts. 

Locally at least, now the following works on Ubuntu 24.04

```
pytest  examples -k test_examples_compare
```

If i add a print statement in the font caching getters I get the expecte font loaded from the pygfx directory.

hmaarrfk:
With #827 folded into here, I can get

```
PYGFX_WGPU_ADAPTER_NAME=llvmpipe pytest  examples -k test_examples_compare
```
to pass and
```
PYGFX_WGPU_ADAPTER_NAME=llvmpipe pytest  examples
```
to fail. But the failure locally looks much more like what i report in the detail in https://github.com/pygfx/pygfx/pull/827#issue-2424096745

But in truth, it is 2 pixels off:
```python
In [1]: import imageio as iio
In [2]: original = iio.imread('examples/screenshots/validate_points_markers.png')
In [3]: rendered = iio.imread('examples/screenshots/diffs/validate_points_markers.png')
In [4]: np.max(np.abs((original.astype('float') - rendered.astype('float'))))
2.0
```

hmaarrfk:
And updating hte code above with:
```
In [7]: ci = iio.imread('/home/mark/Downloads/screenshot-diffs/validate_points_markers.png')
In [8]: np.min((original.astype('float') - ci.astype('float')))
0.0
In [10]: np.max((original.astype('float') - ci.astype('float')))
2.0
```
so for 2 pixel counts, on code that is heavily dependent on AA, good enough????

hmaarrfk:
And now this is officially weird:
```
PYGFX_WGPU_ADAPTER_NAME=llvmpipe pytest  examples -x  -k points_markers --count=3
```
fails....

hmaarrfk:
I'm not sure if there anything you want to really improve on this, I think it is pretty solid barring the 2 pixel difference that can happen...

```
 E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=1
E           
E           Mismatched elements: 3 / 4800000 (6.25e-05%)
E           Max absolute difference among violations: 2
E           Max relative difference among violations: 0.01785714
E            ACTUAL: array([[[119, 119, 119, 255],
E                   [119, 119, 119, 255],
E                   [119, 119, 119, 255],...
E            DESIRED: array([[[119, 119, 119, 255],
E                   [119, 119, 119, 255],
E                   [119, 119, 119, 255],...
```

I don't think that it is possible to diagnose an issue due to LLVMPIPE without the ability to iterate quickly, I've made suggested improvements in https://github.com/pygfx/pygfx/pull/827 to help with that. But it may be something that we should look to do in a separate merge request.

For one, I see many 1 pixel differences in gradients. Perhaps there is something else that is there, but I don't think it should block this improvement...

Korijn:
Well, long story short, we can't merge this while CI is failing :)

hmaarrfk:
> Well, long story short, we can't merge this while CI is failing :)

this is OK, but now its obvious that I'm not seriously changing the tests but rather increasing `atol = 1` to `atol = 2`.

almarklein:
I'm fine with merging this. We plan to revisit mage comparison (https://github.com/pygfx/image-comparison) anyway.

hmaarrfk:
why can't i re-open my own PR? i was just trying to restart the CIs. I still think that this is very useful for SDF design.

Korijn:
Are we ready to merge here?

hmaarrfk:
yes from my end.

Korijn (approved):

hmaarrfk:
great thanks!

almarklein:
I had to think of this again when the image comparison failed in #902. I had an idea what could explain the different results. Writing this down so we can look into it another time.

I think what's going on is that when the sdf is sampled, it is not fully contained within that sdf's patch. Either pixels from the neighbouring glyph affect the sampling, or a different result may happen depending on whether the glyph is at the edge of the atlas or not.

This could explain why certain glyphs look slightly different depending on whether other tests were run first; the atlas will be pre-populated differently, so the glyph ends up in a different slot.

Would be good to implememt a proper test for this, and/or limit texcoords when a glyph is sampled (similar as for the u and v plane for yuv420p in #873).

PR #825, opened 2024-07-22
Fix for upcoming deprecation in rtd

almarklein:
Read The Docs is [making some changes](https://about.readthedocs.com/blog/2024/07/addons-by-default/) that also affects our docs. This PR applies the changes proposed by rtd.

Further, I switched the domain from pygfx.com to pygfx.org, since that seems more common for open source projects.

Using `docs.pygfx.org`. I could not get the root `pygfx.org` to work. Perhaps that's for the better, because we can leave the root for a fancy landing page someday.

I set-up redirects on a server of myself for pygfx.com and pygfx.org, to point to docs.pygfx.org

## Eventual situation

*will update as DNS resolves, may take a day or so*

* [x] My server redirects pygfx.com -> pygfx.org
* [x] My server redirects pygfx.org -> docs.pygfx.org
* [x] docs.pygfx.org works at RTD
* [x] RTD redirects pygfx.readthedocs.org to our canonical domain
* [x] Remove pygfx.com from rtd domains once it resolves to my server.


almarklein:
Will pick this up later today or tomorrow. RTD is still validating certificates and DNS takes time to propagate globally.

almarklein:
Ok, docs.pygfx.org is now working, but I broke pygfx.com.

almarklein:
Pfew. All domain names seem to be resolved.

PR #826, opened 2024-07-22
Fix pointer down interaction in stent demo volume_render1

hmaarrfk:
is it correct that you intended that the user clicks on the volume not the slice?

That is my understanding.


almarklein (approved):
> is it correct that you intended that the user clicks on the volume not the slice?

Yes, looks like that's the intention  :)

Korijn (approved):

Korijn (approved):

PR #829, opened 2024-07-23
Fix infgrid artifacts for ortho camera

### LINKED ISSUE #821: Unsolicited Feedback on the Grid Shader
I was playing around with the grid example:

and wanted to structure the feedback i had
https://github.com/pygfx/pygfx/pull/820#issuecomment-2227026954

I'm not sure what is the source for many of these, but I've found that the lines generated from the examples
```
examples/feature_demo/line_plot.py
```

Provide very strange results
image

Notice how the grid just randomly disappears.

I wish i could say that the artifacts go away with "thicker lines"
```
diff --git a/examples/feature_demo/line_plot.py b/examples/feature_demo/line_plot.py
index 45dd22f..a91ebd1 100644
--- a/examples/feature_demo/line_plot.py
+++ b/examples/feature_demo/line_plot.py
@@ -26,8 +26,8 @@ grid = gfx.Grid(
         major_step=1,
         minor_step=0,
         thickness_space="screen",
-        major_thickness=2,
-        minor_thickness=0.5,
+        major_thickness=5,
+        minor_thickness=4,
         infinite=True,
     ),
     orientation="xy",
@@ -64,7 +64,7 @@ def animate():
     rulery.start_value = rulery.start_pos[1]
     statsy = rulery.update(camera, canvas.get_logical_size())

-    major_step_x, major_step_y = statsx["tick_step"], statsy["tick_step"]
+    major_step_x, major_step_y = 2 * statsx["tick_step"], 2 * statsy["tick_step"]
     grid.material.major_step = major_step_x, major_step_y
     grid.material.minor_step = 0.2 * major_step_x, 0.2 * major_step_y
```
image

Zooming into the center, you see some very strange things at what appears to be some kind of symmetry point.
image

I'm hoping that this stirs some ideas as to why this might be happening. Seems like a rounding error might be causing some of this.

I don't think it has to do with the pixel_ratio since when I set a pixel_ratio of 1 it still occurs:
image

almarklein:
Fixes #821

Confirmed on Mac and Windows Intel.

### DIFF
--- a/pygfx/renderers/wgpu/wgsl/grid.wgsl
+++ b/pygfx/renderers/wgpu/wgsl/grid.wgsl
@@ -33,14 +33,10 @@ fn vs_main(in: VertexInput) -> Varyings {
         // Get point on the plane closest to the origin (handy for debugging)
         //let pos_origin = (d / length(abc)) * abc;
 
-        // Get distance between camera and the plane
+        // Get position of camera projected on the plane
         let cam_pos = vec3<f32>(u_stdinfo.cam_transform_inv[3].xyz);
         let cam_k = (dot(abc, cam_pos) + d) / length(abc);
         let cam_pos_on_grid = cam_pos - cam_k * abc;
-        let distance_cam_to_grid = abs(cam_k); // == distance(cam_pos, cam_pos_on_grid);
-        // let distance_cam_to_grid = abs(p0.y - cam_pos.y);  // debug: only works for xz plane.
-
-
 
         // The closer the camera is to the plane, the less space you need to
         // make the horizon look really far away. Scaling too hard will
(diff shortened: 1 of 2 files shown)


hmaarrfk:
beautiful. it used to take me 0.1 seconds to break it, now my 10 seconds of scrolling are without any obvious shader artifacts!!!

Korijn (approved):
I wish I understood why this fixes the issue!

pygfx/renderers/wgpu/wgsl/grid.wgsl
```diff
@@ -64,9 +66,21 @@ fn vs_main(in: VertexInput) -> Varyings {
         //  | / /           \\|
         //  x-----------------x
 
+        // Get scale multiplier for near quad.
+        // With an orthographic camera, we use the scale factors for x and y (camera width and height).
+        // For perspective cameras we use the distance to the plane and assume an aspect ratio of 1.
+        var near_multiplier = vec2<f32>(1.0);
+        if is_orthographic() {
```
hmaarrfk, line 73: to improve understanding, maybe this if statement should be around the whole computation for `distance_cam_to_grid`.

I understand that it won't help performance, but it may help the understanding in saying that:

"Our computation for the distance between the camera to the grid is only valid when the camera isn't an orthographic projection"

My understanding is that the nature of the orthographic projection makes the camera matrix illposed in certain situations.

almarklein:
I updated the comments to make it more clear, and moved the distance calcualtion into the if.

I found that the artifacts show up because of the big difference in scale between the x and y axis. The solution is basically to scale the grid in the same way. While writing the comments I realized that the solution does make the assumption that the grid is aligned with the camera. But I think this is a safe assumption for cases where you have a non-unit aspect ratio.

Korijn (approved):

PR #831, opened 2024-07-26
Improve support for custom tick formatting

almarklein:
* [x] Can now set `tick_format` to a function that formats the tick.
* [x] Fix that `tick_format` was not used when a sequence of `ticks` are given.
* [x] When  `.ticks` is a dict, the values that are float also pass via `tick_format`.
* [x] All ruler props can be passed via the constructor. 
* [x] Added validation example.


almarklein:
cc @clewis7 @kushalkolar

kushalkolar:
This looks really nice! :smile: . The tick formatting should cover all reasonable use cases!

Korijn (approved):

PR #834, opened 2024-08-01
Multitext and text refactoring

almarklein:
## Introduction

The primary purpose of this work is to support multi-text (multiple pieces of text each with their own position). To implement this, quite a bit of refactoring was needed, and I improved several things along the way.

## The idea

Text is divided into multiple blocks, typically one per line/paragraph, similar
to how Qt splits text in blocks in its editor component. This makes it easy to
edit one block, and then merely shift the other blocks to update the layout. In
the multi-text scenario, these blocks can be individually controlled and
positioned by the user.

Each text block again divides its text into multiple items. Each item is a unit
of text that is moved as a whole during layout. Each word typically becomes one
item. Each item is created from one or more text pieces which each can have a
different format (bold, italic, size).

We already had `TextItem`, this PR uses it in nearly the same way, but adds support for multiple different formatting within a single item.

## Most notable changes

Added:

* The `MultiTextGeometry` adds API to create `TextBlock` instances.
* The new `TextBlock` class provides user-API to control and position individual text blocks.
* More markdown formatting: headers and bullets.
* Added `TextGeometry.paragraph_spacing`.
* Added support for `TextGeometry.max_width` (was just a stub until now).
* Added a few validation examples.

Changes:

* The `TextGeometry` can update text more efficiently, re-using `TextBlock` and `TextItem` objects.
* The `MultiTextGeometry.position` buffer can also be used directly to position text blocks.
* The `TextGeometry` props `anchor` and `text_align` are now enums and may be checked more strictly.
* Better handling of text direction (e.g. for Arabic text).
* Support for vertical text. Full layout support, except for `text_align`.
* `TextGeometry.set_text()` and `.set_markdown()` no longer have a `family`, `style` and `weight` args, use the geometry's family prop, and markdown formatting instead.
* `text_align='start'` means 'right'  for Arabic text, but `text_align='left'` is actually left.
 

## Todo

* [x] Update tests and write a few new ones.
* [x] I can trigger an error with the `line_plot.py` example related to allocating glyphs.
* [x] Check rotating text in multi-text scenario.

## After this

* Optimizations for setting text, to e.g. make the stats widget more performant.
  * Optimize hot paths using caching and/or skips for simple cases.
  * Combine buffers, e.g. glyph position and sizes.
  * There are reserved bit in the atlas index that we can use, e.g. upside down when y-axis is flipped?
* Use multi-text for the ruler.
* Allow `TextBlock` to override some of the geometry's props. Many of them can technically be overriden relatively easily, but let's do this as we need it (I expect to overload `anchor` for the ruler).


kushalkolar:
:rocket: I just came across another usecase with something I'm doing right now, heatmap annotation like seaborn:

image

almarklein:
> I just came across another usecase 

Thanks, it helps to have some concrete examples!

hmaarrfk:
we would love something like this for:
https://www.youtube.com/shorts/AHyU4qgAEwA

In the demo, we have about 40,000 labels
image

on a powerful GPU, its fine, but we are always looking for more efficiently ways to generate it in "real time" and make it accessible for weaker machines.

Having so many nodes in the graph can definitely slow things down.

With this kind of idea, we could start to collate the groups of labels that have a similar transformation into a single set.

(the images in the video snippet can move relative to each other in a pretty dynamic way).

Korijn:
Would multi text also make it easier to do complex text layout like paragraphs, and even pages etc?

almarklein:
Paragraphs are already supported, so some extend, and don't need the extra `posisions` thing, because all positioning is relative. I'm not sure what pages would have to look like ...

almarklein:
I thought I quickly finish this pr before moving on with #954 et al. I forgot how complex the text logic was 😉 

I first figured that the text-item concept  that we already used could be used so that the individual text items can be positioned using `geometry.positions`. This would solve the use-case of labels. But I eventually went with a slightly more sophisticated approach that would better support more advanced use-cases like making an editor component or things like that.

The geometry has its text spread over multiple `TextBlock` objects. These are quite similar to Qt's `TextBlock`, as in they'd normally represent one line/paragraph of text. Each can thus apply its own layout. These text blocks can then either be positioned manually (via `geometry.positions`), or automatically by the `geometry` to simply put them beneath each-other.

That way, when you have a big chunk of text, we can make updates to individual lines/paragraphs, layout only needs to re-run for that that block, and the other blocks only need to be shifted up/down a bit. With some extra work, we can even support per-word edits quite efficiently.

Most of this is in place now. I'm now reanimating the layout logic. And then I need to address the many todo's I left in the code and create a cool multi-text example.

Korijn:
Maybe the Stats object is a nice test case, since it has static text but constantly updating numbers.

examples/feature_demo/text_multiple_grid.py
```diff
@@ -0,0 +1,67 @@
...
+text = gfx.Text(
+    gfx.MultiTextGeometry(
+        anchor="top-left", anchor_offset=0.5, screen_space=False, font_size=3
+    ),
+    gfx.TextMaterial(color="#fff", outline_thickness=0.2),
+)
+scene.add(text)
+
+
+def peudo_random(*seeds):
```
claydugo, line 30:
```suggestion
def pseudo_random(*seeds):
```

claydugo:
The `text_multiple_grid` example is really great!

almarklein:
@panxinmiao can I please get your feedback on rendering of Chinese text?

This PR also improves support for vertical text. I took some Chinese text from the web (a poem, I think). This is how it renders left-to-right, which is what I understand to be the 'modern' style:

image

And top-to bottom, which I understand to be the more historic style:

image

Does this look correct to you?

panxinmiao:
Amazing, The style perfectly correct, and the visual result looks absolutely great!

Ps: It's a very famous classical Chinese poem (that nearly every child can recite in China), expressing homesickness. :)

almarklein:
> Out of curiosity, may I ask about the changes to the screenshots. I didn't expect any changes to things like the point markers test?

The changed screenshots are due to improvements to the the calculations of the bounding box and handling of whitespace.

hmaarrfk:
Nice, did you ever create that helper function to sample an atlas correctly?

I'm thinking of https://github.com/pygfx/pygfx/issues/905#issue-2742425834

almarklein:
> did you ever create that helper function to sample an atlas correctly?

not yet

almarklein:
@Korijn This pr is ready from my end.

pygfx/objects/_text.py
```diff
@@ -0,0 +1,38 @@
...
+        # When rendering in screen space, the world transform is used
+        # to establish the point in the scene where the text is placed.
+        # The only part of the local transform that is used is the
+        # position. Therefore, we also keep a transform containing the
+        # local rotation and scale, so that these can be applied to the
+        # text in screen coordinates.
+        # Note that this applies to the whole text, all text blocks rotate around
+        # the text-object origin. To rotate text blocks around their own origin,
+        # we should probably implement TextBlock.angle.
+        matrix = la.mat_compose((0, 0, 0), self.local.rotation, self.local.scale)
```
Korijn, line 37: Can't you just take `self.local.matrix[:3, :3]`?

pygfx/geometries/_base.py
```diff
@@ -205,7 +205,7 @@ def get_bounding_sphere(self) -> NDArray[np.float32] | None:
                         )
                     else:
                         bsphere = np.array(
-                            [center[0], center[1], center[1], radius], np.float32
+                            [center[0], center[1], center[2], radius], np.float32
```
Korijn, line 208: What 😂

pygfx/objects/_text.py
almarklein: Good point! Can even do with the 2x2 matrix, ignoring rotation in z.

almarklein:
Short term planning:

* Get this pr approved.
* Do a release of pygfx.
* Merge this PR.
* Do text performance improvements.
* Atlas offset thingy.
* Release again.
* Blending ... view etc.

Korijn (approved):

hmaarrfk:
> Short term planning:
> 
> * Get this pr approved.
> * Do a release of pygfx.
> * Merge this PR.
> * Do text performance improvements.
> * Atlas offset thingy.
> * Release again.
> * Blending ... view etc.

Thanks for the mini roadmap!

almarklein:
I actually meant to do a pygfx release before merging this 😅 , but not a big deal; this PR is surprisingly backwards compatible.

hmaarrfk:
> this PR is surprisingly backwards compatible.

lets see how it withstands to my search and replace text ;)

Korijn:
> I actually meant to do a pygfx release before merging this 😅 , but not a big deal; this PR is surprisingly backwards compatible.

Sorry, I read your list but I still missed it somehow...

almarklein:
> You can just revert the merge if you need to.

nah, it's alright

PR #835, opened 2024-08-01
Bump version

almarklein:
Let's do a release before holidays/Euroscipy. There's some cool stuff since the last release, like the ruler, new buffer texture api, custom SDF markers, fix for dashing in thin lines ...

Maybe we can get #803 in before merging this and doing the release.

PR's to merge before this:
* #803
* #847
* #848

### DIFF
--- a/pygfx/__init__.py
+++ b/pygfx/__init__.py
@@ -26,7 +26,7 @@ import pylinalg
 
 del pylinalg
 
-__version__ = "0.3.0"
+__version__ = "0.4.0"
 version_info = tuple(map(int, __version__.split(".")))
 
 __wgpu_version_range__ = "0.15.1", "0.17.0"


pygfx/__init__.py
```diff
@@ -26,7 +26,7 @@
 
 del pylinalg
 
-__version__ = "0.3.0"
+__version__ = "0.4.0"
 version_info = tuple(map(int, __version__.split(".")))
 
 __wgpu_version_range__ = "0.15.1", "0.17.0"
```
kushalkolar, line 32: will wgpu need to be pinned because of #816

almarklein: Yes, good point!  I'm on holiday now. I may or may not find the time to fix this :)

Korijn (approved):

PR #837, opened 2024-08-11
A more perfect SDF for the pin

hmaarrfk:
The old SDF had artifacts at the intersection of the circle and the lower triangle.

https://github.com/user-attachments/assets/552a846a-d927-4400-b7f5-5db617e01c49


hmaarrfk:
not too big of a fan of triangular shape but i'll play around with an other shape later.

almarklein (dismissed):
> not too big of a fan of triangular shape but i'll play around with an other shape later.

It looks better than before, so I'm fine with merging as-is :)

hmaarrfk:
I would prefer we merge https://github.com/pygfx/pygfx/pull/824 first, and come back to this in a bit. IQ has a few other perfect SDF functions that we can use to improve on this.

hmaarrfk:
Improving the look beyond what i have now is definitely a 1-2 day effort i don't want to put in now.

almarklein (approved):
This is already a huge improvement. Thanks!

Korijn (approved):

PR #842, opened 2024-08-27
Fix pick_write not really being dynamic

### LINKED ISSUE #840: Can't set pick_write=True after first rendering pass
Hi! This is related to issue #720 and PR #722 which made `pick_write` dynamic. 

In my use case, changing `pick_write` dynamically works like a charm as long as the material is initialised with `pick_write=True`, i.e. I can repeatedly toggle between `True` and `False` to determine whether the object should be pickable. However, when a material starts off with `pick_write=False`, setting it to `pick_write=True` afterwards doesn't have any effect.

While putting together a stand-alone reproducer, I realised that this issue only appears when trying to set `pick_write=True` after the first rendering pass. For background: I'm running my app interactively (i.e. non-blocking) from an ipython session and I was trying set `pick_write=True` while working with the scene.

```python
import pygfx as gfx

canvas = WgpuCanvas()
scene = gfx.Scene()

cube = gfx.Mesh(
    gfx.MeshBasicMaterial(pick_write=False),  # <- initialize with pick_write=False
)
scene.add(cube)

def on_hover(event):
   print("Hovering over object:", event.pick_info["world_object"])

scene.add_event_handler(on_hover, "pointer_move")

camera = gfx.PerspectiveCamera(70, 16 / 9)
camera.show_object(scene)

controller = gfx.TrackballController(camera, register_events=renderer)

def animate():    
    canvas.request_draw()
    cube.material.pick_write = True  # <- setting pick_write=True after the first renderer.render() has no effect
    
if __name__ == "__main__":    
    canvas.request_draw(animate)
    run()
```

Thanks!

almarklein:
Fixes #840

The problem was that the shaders remained intact and where never updated when the `pick_write` prop is changed.

### DIFF
--- a/pygfx/renderers/wgpu/engine/pipeline.py
+++ b/pygfx/renderers/wgpu/engine/pipeline.py
@@ -335,6 +335,7 @@ class PipelineContainer:
             with wobject.tracker.track_usage("reset"):
                 self.wobject_info["pick_write"] = wobject.material.pick_write
             changed.update(("bindings", "pipeline_info", "render_info"))
+            self.wgpu_shaders = {}
 
         if "bindings" in changed:
             self.shader.unlock_hash()


Korijn (approved):

hmaarrfk:
so fast!

PR #850, opened 2024-09-12
Enhance and fix materials

panxinmiao:
This PR fixes and enhances support for mesh materials with the following updates:

- Added `specular_map` support for both `MeshBasicMaterial` and `MeshPhongMaterial`.
- Added support for the `emissive_map`, `emissive_intensity`, `normal_map`, and `normal_scale` properties in `MeshPhongMaterial`.
- Fixed a bug where the `emissive_intensity` property was not work.
- Add a new example `earth.py` that demonstrates the full capabilities of `MeshPhongMaterial`.


almarklein (approved):
Nice work, as always! 🚀

Korijn (approved):
Amazing!!

PR #852, opened 2024-09-14
Fix binding layout

panxinmiao:
After updating to the latest version of `wgpu-py`, I encountered an error in one of my use cases related to a mismatch between ResourceBinding and the pipeline layout:

```
Shader global ResourceBinding { group: 0, binding: 0 } is not available in the pipeline layout
Buffer structure size 48, added to one element of an unbound array, if it's the last field, ended up greater than the given min_binding_size, which is 16
```

Upon look into it, I found that the `min_binding_size` was not properly set when creating the bind group layout.

This issue likely existed for a long time, but previous versions of `wgpu` did not report the error. It’s possible that the `min_binding_size` parameter did not take effect or was properly validated in earlier versions.

### DIFF
--- a/pygfx/renderers/wgpu/engine/binding.py
+++ b/pygfx/renderers/wgpu/engine/binding.py
@@ -96,7 +96,7 @@ class Binding:
                 "buffer": {
                     "type": getattr(wgpu.BufferBindingType, subtype),
                     "has_dynamic_offset": False,
-                    "min_binding_size": resource.itemsize,
+                    "min_binding_size": resource.nbytes,
                 },
             }
         elif self.type.startswith("sampler/"):


panxinmiao:
Uh, a strange CI error. 
The logs show that there was an issue with `validate_culling` during rendering, but the screenshot seems to have been generated correctly. And, I'm unable to reproduce the error locally. 🤔

hmaarrfk:
(Very strange error and now I feel responsible because I touched that code last)

hmaarrfk:
Try to restart the ci. If there is no option close and open the pr

almarklein (approved):
Let's fix this and bump a bugfix release.

Korijn (approved):

PR #857, opened 2024-09-19
Use vstack and not row_stack

hmaarrfk:
/home/mark/git/pygfx/pygfx/objects/_ruler.py:259: DeprecationWarning: `row_stack` alias is deprecated. Use `np.vstack` directly.
    np.row_stack([self._start_pos, self._end_pos]),

image


Korijn (approved):

almarklein:
Thanks!

PR #859, opened 2024-09-24
Use pyproject.toml and ruff

almarklein:
* [x] Use `pyproject.toml` instead of `setup.py`.
* [x] Use `ruff` instead of flake8 and black.
* [x] Required version ranges for wgpu and pylinalg are defined in `pyproject.toml` but still checked at runtime in `__init__.py`.
* [x] Fixed linting errors (hooray for `ruff check --fix`).
* [x] Update CI.
* [x] Check for references to flake8 and black. 
* [x] Merge with main before applying many changes below.
* [x] Enable more ruff lint rules.
* [x] Actually apply ruff formatting (let's do this after a first round of review). Ruff claims to be almost compatible with black, but there are differences (in 38 files).


Korijn:
If you want ruff format to be compatible with black you need to adopt the prescribed configuration as well

almarklein:
There are subtle differences, but quite a few of 'm: https://docs.astral.sh/ruff/formatter/black/ 
The formatter does not seem to allow much configuration. But I guess its ok. We only need to get used to use ruff instead of black.

I also have pending changes to apply ruff in wgpu-py. I'll make a pr for those before I finish this pr.

Korijn:
Not having much configuration is the charm :) no options = less discussions!

Korijn:
All makes sense!

pygfx/__init__.py
```diff
@@ -23,35 +23,61 @@
...
+    except ImportError:
+        return {}
+    # Load versions
+    this_dir = os.path.dirname(os.path.abspath(__file__))
+    pyproject_file = os.path.join(this_dir, "..", "pyproject.toml")
+    if os.path.isfile(pyproject_file):
+        with open(pyproject_file, "rb") as fp:
+            dependencies = tomllib.load(fp)['project']['dependencies']
+    else:
+        dependencies = importlib.metadata.requires("pygfx")
```
Korijn, line 46: I don't fully understand why you need to first attempt to read the toml file? Isn't importlib.metadata.requires enough?

almarklein: Good question.

We have version ranges for wgpu-py and pylinalg, for which we know pygfx works. Pip respects them, so for end-users we don't really have to do this check. This check is therefore mosly for people who use bleeding edge versions of wgpu and/or pygfx. It's a nice warning in case you've updated one, but not the other.

The problem with `importlib.metadata.requires("pygfx")` is that it returns the values as they were the last time you did `pip install -e .`. So basically worthless for this scenario.

Having written this down, I think we can simply only perform the check when there's a pyproject file, bc that's a good indication pygfx is installed in editable/dev mode.

almarklein:
> Not having much configuration is the charm :) no options = less discussions!

I agree.

I checked on both wgpu-py and pygfx, and although applying `ruff format` the first time triggers quite a lot of (small) changes, then running `black` again just makes 3 small changes. These cases can actually be written differently so that black and ruff don't fight over them.

My only concern is how this can possibly disrupts people's workflow, e.g. because they have the black formatter active in VS Code, so when they save a file, VSCode may change the formatting *the wrong way*. With what I said above, it looks like this does not have to be a problem.

tests/geometries/test_text_geometry.py
```diff
@@ -15,39 +15,39 @@ def test_text_geometry_text():
...
+    assert geo.sizes.nitems == 6
 
     # If setting smaller text, buffer size is oversized
     geo.set_text("x")
-    geo.positions.nitems == 6
-    geo.indices.nitems == 6
-    geo.sizes.nitems == 6
+    assert geo.positions.nitems == 6
+    assert geo.indices.nitems == 6
+    assert geo.sizes.nitems == 6
```
almarklein, line 50: omg

conftest.py
```diff
@@ -1,9 +1,14 @@
 """Global configuration for pytest"""
 
+import os
 import numpy as np
 import pytest
 
 
+# System fonts can interfere with screenshots that expect a particular font to be used
+os.environ["PYGFX_DISABLE_SYSTEM_FONTS"] = "1"
```
almarklein, line 9: Headsup @hmaarrfk, I set this env var in `conftest.py`. It was in `pytest.ini`, but from what I can tell, this never worked unless you have `pytest-env` installed.

hmaarrfk: ah you are correct. sorry about that. I guess that is pulled in in my dev envs.

I understand why you would go this route instead of adding pytest-env.

I forgot what the motivation was for me using that pytest-env variable. It might have been to switch between Qt backends in my QtGUI applicaiton. Sometimes code runs at import time, so you can't be sure hte line you added gets set in time.

tests/utils/test_text_atlas.py
```diff
@@ -77,7 +77,7 @@ def test_atlas_resize():
     prev_array = atlas._array
 
     # We can fit 9 regions of 8x8 in it. So 4 more.
-    for i in range(4):
+    for _i in range(4):
```
hmaarrfk, line 80: hmm i didn't know this trick, i always struggled with `_` since it didn't give any info as to what the variable was.

almarklein: I just did what ruff suggested 🤷

Korijn (approved):

PR #861, opened 2024-10-03
Get a longer version string on dev installs

### LINKED ISSUE #696: Would you consider miniver for version management?

almarklein:
Closes #696

This idea was discussed in https://github.com/pygfx/pygfx/issues/696, and recently @hmaarrfk nerd-sniped me again in https://github.com/hmaarrfk/pygfx/pull/9.

I still come to the conclusion that the source not having a hard-coded version number can easily become a source of problems (PyInstaller being one example).

So what I did in this PR: Still a hard-coded `__version__`, but when the code is in a git-repo (i.e. a dev build), the extra info from git is added to the version (number of commits since the release, git hash, dirty status).

Also added a tiny CLI, so one can do `python -m pygfx version`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,11 +116,12 @@ jobs:
     - name: Install package and dev dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -U -e .
+        pip install -U .
         pip install pytest psutil pyinstaller
     - name: Unit tests
       run: |
-        pytest -v pygfx/__pyinstaller
+        pushd $HOME
+        pytest -v --pyargs pygfx.__pyinstaller
 
   test-examples-build:
     name: Test examples ${{ matrix.pyversion }}
@@ -149,6 +150,7 @@ jobs:
         pip install -e .[examples]
     - name: Show wgpu backend
       run: |
+        python -c "import pygfx; print(pygfx.__version__)"
         python -c "from examples.tests.testutils import adapter; print(adapter.info)"
     - name: Test examples
       env:
@@ -188,6 +190,7 @@ jobs:
           rm -rf ./pygfx
           pushd $HOME
           pip install $GITHUB_WORKSPACE/dist/*.tar.gz
+          python -c "import pygfx; print(pygfx.__version__)"
           popd
           # don't run tests, we just want to know if the sdist can be installed
           pip uninstall -y pygfx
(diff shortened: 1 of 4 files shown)


hmaarrfk:
I did mean to get you excited about reducing dev churn! Glad it worked. However, I purposefully didn't erase my history to give you a glimpse of the amount of frustration I had. Personally, I was just curious if it was truely broken with pyinstaller....

Regarding:

> PyInstaller being one example

The original source of the issue is that:
1. Calling `pip install . -vv` will "install things in your site-packages directory".
2. Calling pyinstaller from the same directory will preferentially find `pygfx` in the current directory.

With miniver `pygfx` from the `site-packages` directory has the "frozen version", but `pygfx` from the local directory had the dynamic version. This kind of confusion between `site-packages` and the current working dir is pretty common whenver you do code gen during the setup phase.

The "fix" was to move away from the `pygfx` project during the pyinstaller phase:
```
pushd ${HOME}
pytest --pyargs -vx pygfx.__pyinstaller
```
https://github.com/hmaarrfk/pygfx/pull/9/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR136

to ensure that the "fully installed version of the pygfx" code gets picked.

almarklein:
> I did mean to get you excited about reducing dev churn! 

Well, this implementation still requires manually bumping the version number. So the only benefit, really, is the availability of a more precise version.

That's also why I don't think the pyinstaller thing is a problem, but I'll apply the `pushd ${HOME}` trick to make sure .

hmaarrfk:
thank you for working on this! I think it will help me continue to write compatibility shims as you introduce breaking changes ;)

Korijn (dismissed):

hmaarrfk:
Just a note for others, this now yields:

```
0.5.0.post3+ge8a9580
```

Korijn (approved):

PR #862, opened 2024-10-10
Do not mix stdout and stderr in version detection

hmaarrfk:
Consider the strange strange case that a user wants to start their app through Steam.

And steam provides a 32 bit LDPRELOAD in 2024.

So the output is actually

> CompletedProcess(args=['git', 'describe', '--long', '--always', '--tags', '--dirty', '--first-parent'], returncode=0, stdout=b"ERROR: ld.so: object '/home/mark/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.\nv0.5.0-3-g94e5939-dirty\n")

Which then wrecks any access to the version variable.

Korinj knew that these kinds of bugs would happen...

### DIFF
--- a/pygfx/_version.py
+++ b/pygfx/_version.py
@@ -71,7 +71,7 @@ def get_version_info_from_git():
         "--first-parent",
     ]
     p = subprocess.run(
-        command, cwd=repo_dir, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        command, cwd=repo_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE
     )
 
     # Parse the result into parts
--- a/pygfx/_version.py
+++ b/pygfx/_version.py
@@ -70,20 +70,33 @@ def get_version_info_from_git():
         "--dirty",
         "--first-parent",
     ]
-    p = subprocess.run(
-        command, cwd=repo_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-    )
+    try:
+        p = subprocess.run(
+            command, cwd=repo_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+    except Exception as e:
+        logger.warning("Could not get pygfx version: " + str(e))
+        p = None
 
     # Parse the result into parts
-    output = p.stdout.decode(errors="ignore")
-    if p.returncode:
-        logger.warning("Could not get pygfx version: " + output)
+    if p is None:
         parts = (None, None, "unknown")
     else:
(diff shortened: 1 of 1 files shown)


Korijn (approved):
> Korinj knew that these kinds of bugs would happen...

😂

almarklein:
Probably good to then also show the stderr in the warning message?

hmaarrfk:
Ah in the case of failure. Sure.

hmaarrfk:
Will do

almarklein:
> Consider the strange strange case that a user wants to start their app through Steam.

I know you're a power-user, but this is next-level shit.

hmaarrfk:
```
In [1]: from pygfx._version import repo_dir
Could not get pygfx version.

stdout:

stderr: error: unknown option `foo'
usage: git describe [--all] [--tags] [--contains] [--abbrev=<n>] [<commit-ish>...]
   or: git describe [--all] [--tags] [--contains] [--abbrev=<n>] --dirty[=<mark>]
   or: git describe <blob>

    --[no-]contains       find the tag that comes after the commit
    --[no-]debug          debug search strategy on stderr
    --[no-]all            use any ref
    --[no-]tags           use any tag, even unannotated
    --[no-]long           always use long format
    --[no-]first-parent   only follow first parent
    --[no-]abbrev[=<n>]   use <n> digits to display object names
    --[no-]exact-match    only output exact matches
    --[no-]candidates <n> consider <n> most recent tags (default: 10)
    --[no-]match <pattern>
                          only consider tags matching <pattern>
    --[no-]exclude <pattern>
                          do not consider tags matching <pattern>
    --[no-]always         show abbreviated commit object as fallback
    --[no-]dirty[=<mark>] append <mark> on dirty working tree (default: "-dirty")
    --[no-]broken[=<mark>]
                          append <mark> on broken working tree (default: "-broken")

```

Please feel free to suggest and edit the formatting


Also handling in the case that you have errors in the subprocess command:
```
In [1]: import pygfx
Could not get pygfx version: [Errno 2] No such file or directory: 'gitthatdoesneixt'

In [2]:
```

(of course, i had to simulate the errors by editing the source....)

hmaarrfk:
> > Consider the strange strange case that a user wants to start their app through Steam.
> 
> I know you're a power-user, but this is next-level shit.

I honestly couldn’t really believe the failure mode myself. 

I have been using git describe tricks for 6 years plus…..

almarklein (approved):

Korijn (approved):

PR #867, opened 2024-10-16
Add labelled image grid example

claydugo:
gridded

I am hoping this example is useful moving forward to demonstrate some of the different challenges myself and colleagues have encountered in our unique image display use case.

### DIFF
--- a/examples/feature_demo/labelled_image_grid.py
+++ b/examples/feature_demo/labelled_image_grid.py
@@ -5,6 +5,9 @@ Labelled Image Grid
 This example demonstrates how to create a grid of images with labels that change color when hovered over.
 """
 
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
 import random
 
 import imageio.v3 as iio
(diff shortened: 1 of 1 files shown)


examples/feature_demo/labelled_image_grid.py
```diff
@@ -0,0 +1,101 @@
...
+
+for position in np.ndindex(grid_shape):
+    img = random.choice(images)
+    add_image(img, position)
+
+camera = gfx.PerspectiveCamera(70)
+camera.show_object(scene)
+camera.local.scale_y = -1
+
+def update_text_visibility():
```
claydugo, line 87: This portion can go entirely for this example but is an interaction I have been wanting to build out. 

In its current form it will not draw the labels on scene creation because the camera height starts > 3000. 

Then you will experience a noticeable lag spike as you zoom to "visible" region.

claydugo: > Then you will experience a noticeable lag spike as you zoom to "visible" region.

https://github.com/pygfx/pygfx/pull/834 would fix this. I can re-implement with multi-text once its done.

Korijn: I guess this also relates to camera frustum culling.

Korijn (approved):

Korijn:
Cool example!

PR #873, opened 2024-10-22
an example of how YUV colorspace can be implemented

hmaarrfk:
The cool thing is that the GPU code doesn't change for "subsampled" colorspaces.

This also introduces how one can sample from multiple grids to create color at a single point.

cc: @kingbedjed


hmaarrfk:
i admit my creativity is really lacking for this example.

examples/feature_demo/image_plus_points_yuv.py
```diff
@@ -0,0 +1,107 @@
...
+im_v = image_yuv420[height + u_height :]
+# Rehsape can cause "implicit" copies of the data, so we assign the shape which causes the data to never get copied
+im_u.shape = height // 2, width // 2
+im_v.shape = height // 2, width // 2
+
+image = gfx.Image(
+    gfx.Geometry(
+        grid=gfx.Texture(im_y, dim=2),
+        grid_u=gfx.Texture(im_u, dim=2),
+        grid_v=gfx.Texture(im_v, dim=2),
```
hmaarrfk, line 72: I wonder if there is a planar texture that could hold both `u` and `v` together. Often they can be returned as:

* 1 contiguous buffer in memory yuv all contiguous.
* 2 contiguous buffers in memory, y only, then "uv" together.
* 3 contiguous buffers.

It would be nice in the first 2 cases if we could take advantage of this to reduce the number of uploads to the GPU. I think "case 2" might be more common than one thinks.

almarklein: I was wondering about this too.

almarklein:
Would be nice if we could stream from an actual video file, so we can also comparethe performance of a shader-based approach with just reading frames as rgb.

Here's a little code that gives easy access to a yuv420p video:
```py

import os

import numpy as np
import imageio
import av

def get_video_pix_fmt(filename):
    container = av.open(filename)
    return container.streams.video[0].codec_context.pix_fmt

def convert_to_yuv420(filename1, filename2):
    input_container = av.open(filename1)
    input_stream = input_container.streams.video[0]

    output_container = av.open(filename2, mode='w')
    output_stream = output_container.add_stream('libx264', rate =input_stream.base_rate)
    output_stream.width = input_stream.width
    output_stream.height = input_stream.height
    output_stream.pix_fmt = 'yuv420p'

    for frame in input_container.decode(video=0):
        frame = frame.reformat(width=input_stream.width, height=input_stream.height, format='yuv420p')
        packet = output_stream.encode(frame)
        output_container.mux(packet)

    output_container.mux(output_stream.encode(None))
    input_container.close()
    output_container.close()

# Get a video file (a standard video from imageio)
filename = imageio.core.Request("imageio:cockatoo.mp4", "r?").filename

if get_video_pix_fmt(filename) != "yuv420p":
    filename_yuv420 = filename.replace(".mp4", "_yuv420.mp4")
    if not os.path.isfile(filename_yuv420):
        print(f"Converting video from {get_video_pix_fmt(filename)} to yuv420p ... ", end="")
        convert_to_yuv420(filename, filename_yuv420)
        print("done")
    filename = filename_yuv420

assert get_video_pix_fmt(filename) == "yuv420p"

```

almarklein:
I made an attempt to load the data into an texture with two layers. Basically this:
```
   ________
  |         |
  |  y      |
  |_________|
  | u  |    |
  |---------|
  |_v__|____|
```
So half of the second layer is empty, but in total the texture is still half the size of an rgba texture (rgb textures don't really exist).

The 'problem' with this approach is that the image must be padded, and the u and v images are copied in non-contiguous blocks:

```py
uv_shape = height //2, width // 2
image_2layer = np.zeros((2, height, width), np.uint8)
image_2layer[0,] = image_yuv420[:height]
image_2layer[1, :height//2, :width//2] = image_yuv420[height : height + u_height].reshape(uv_shape)
image_2layer[1, height//2:, :width//2] = image_yuv420[height + u_height :].reshape(uv_shape)
```

I've got this working (image is rendered in color) locally.

----

It would also be possible to keep the data as-is, with u and v "stretched":
```
   ________
  |         |
  |  y      |
  |         |
  |_________|
  |_u_______|
  |_v_______|
  |         |
  |_________|
```

Then the u and v must be sampled using `textureLoad`, i.e. without interpolation. But maybe that looks good enough.

In this approach, we could skip any padding and data copies, by creating a Texture with no representation on the GPU, and then uploading the data in two chunks (one for y, one for uv). This would require some changes to the `Texture` implementation, but I see no technical hurdles.

hmaarrfk:
just for your nerdy pleasure, not all yuv returned from libav is "contiguous" when they come out of ffmpeg. PyAV invokes a copy to return them to you.
https://github.com/PyAV-Org/PyAV/pull/1393#issuecomment-2098276096

However, you can play with the memory allocator to make sure they are.... we haven't gone that far yet.
 
Ps feel free to push to my branch changes to the shader, i won't mind ;)

hmaarrfk:
> padding

repeatedly padding is pretty disastrous for "4k" or "larger" video.

I think this format that I have focused on is called "YUV420_I420". I am under the impression that it is one of the "native" formats that comes our of both software accelerated and hardware accelerated decoders.


I wonder if we can:

* Internally model things as a single in a 1D the buffer to the GPU.
* Split it into separate Y, U, and V views, each a 2D view of a portion of the array.
* pass it to the shader.

All to avoid a copy

hmaarrfk:
If you intend to merge this I can try to create videos with both limited range and full range YUV. I think it is important to have that option, mostly for documenting what the difference between the two is.

almarklein:
> I can try to create videos with both limited range and full range YUV. I think it is important to have that option, mostly for documenting what the difference between the two is.

Sounds like a good idea; I am curious to what it means :)

almarklein:
I did some benchmarks:

| saved-format | read-format | FPS |
|-|-|-|
| yuv444p (cockatoo.mp4) |  rgb24 | 85 | 
| yuv444p (cockatoo.mp4) |  rgba | 135 | 
| yuv444p (cockatoo.mp4) |  yuv420p | 165 | 
| yuv420p (cockatoo_yuv420.mp4) |  rgb24 | 100 | 
| yuv420p (cockatoo_yuv420.mp4) |  rgba | 220 | 
| yuv420p (cockatoo_yuv420.mp4) |  yuv420p | 250 | 

Interesting observations:

* Switching from rgb to rgba already gives a huge performance benefit!
* The performance difference between rgb and yuv420 is massive when the video is stored as yuv420p!
* I imagine that at these frame-rates, we're also measuring overhead, so the real benefit is likely higher than what we see here.

pygfx/geometries/_base.py
```diff
@@ -142,7 +142,7 @@ def get_bounding_box(self):
             if self._aabb_rev == self.grid.rev:
                 return self._aabb
             # account for multi-channel image data
-            grid_shape = self.grid.data.shape[: self.grid.dim]
+            grid_shape = tuple(reversed(self.grid.size[: self.grid.dim]))
```
almarklein, line 145: This fixes the bounding box logic so it also works with textures that don't own data.

almarklein:
BTW: for those who, like me, wondered what `p` means in `yuv420p`: it means planar. So we can also have `rgbp`, which would have the colors in planes instead of pixel-tuples. Isn't that how tiff stores images?

hmaarrfk:
> Isn't that how tiff stores images?

typically this is how the multichannel scientific images are stored. 

I’ve never seen planar planar RGB

hmaarrfk:
Do you mind listing the resolution next to your benchmark numbers?

almarklein:
> I love that I got you hooked on this performance deep dive.

Hehe. This turned out to be very interesting and insightful.

almarklein:
This is ready from my end.

BTW: this PR also makes it easier to use multi-layer textures for other things, like #882.

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,176 @@
+"""
+Video yuv
+=========
+
+This example demonstrates how to show video data in a very efficient way,
+eliminating as much data-copies as we can.
```
hmaarrfk, line 6:
```suggestion
eliminating as much data-copies and color conversion operations as we can.
```

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,176 @@
...
+
+Frames in yuv420p format are half the size of RGB frames (and even less
+compared to RGBA) so the upload itself is cheaper, and memory
+consumption on the CPU and GPU are reduced.
+
+Benchmarks
+----------
+
+This script can be done to benchmark for different formats. We've found
+that using yuv420p data is certainly the fastest. Using rgba is a bit
```
hmaarrfk, line 35: Ultimately the most important aspect is that the data storage format be aligned with the data display format. 

For example, if the video is encoded as RGBA (very rare in many consumer formats) it may be an advantage to operate the shader with RGBA. However, most video is encoded as YUV420 with limited color range. As such we found that ….

almarklein: Yeah, I understand. Though I was surprised by the relatively good performance of converting to rgba using av. It looks like the rgb-to-rgba conversation we do in pygfx is a tighter bottleneck.

hmaarrfk: One of the big big challenges in all these libraries is assessing the single core performance (which for me is all I care about since at some point my problem is embarrassingly parallelizable) so it is possible to ffmpeg is using multiple threads to do its work. (This is Something I have tuned in the past) 

I mostly left this note here so I can help expand on this motivation a little bit when I get to a computer.

hmaarrfk:
As for full range and limited color range, the first few sections of this wiki article explain it well

https://trac.ffmpeg.org/wiki/colorspace

It should be something that we can read in from the metadata and apply the appropriate math in the shader. 

Not that ffmpeg bused to call full range yuvj420p but they recently deprecated it. I think we should follow their new best practices and use a separate property to specify the range for yuv.

examples/feature_demo/video_yuv.py
almarklein: Ah, so if I understand what you're saying correctly, the relatively good performance of letting ffmpeg convert to rgba for us, is likely because ffmpeg uses multithreading, and that "trick" loses its benefit when you want to load multiple frames in parallel.

hmaarrfk: Multiple frames from multiple different video streams. Correct.

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,176 @@
...
+----
+
+Frames in yuv420p format are half the size of RGB frames (and even less
+compared to RGBA) so the upload itself is cheaper, and memory
+consumption on the CPU and GPU are reduced.
+
+Benchmarks
+----------
+
+This script can be done to benchmark for different formats. We've found
```
hmaarrfk, line 34:
```suggestion
This script can be done to benchmark for different formats. We've performed
benchmarks against videos encoded with yuv420p and we've found that
```

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,176 @@
...
+that using yuv420p data is certainly the fastest. Using rgba is a bit
+slower, although it consumes more memory, so its disadvantage may
+increase for heavy workloads. Reading the data as RGB has terrible
+performance. This is unfortunate, because it's the format that many
```
hmaarrfk, line 38:
```suggestion
performance. We believe that this is caused by the fact that RGB images must
be converted to RGBA pior to uploading them to the GPU.
This is unfortunate, because it's the format that many
```

pygfx/renderers/wgpu/wgsl/image_common.wgsl
```diff
@@ -2,10 +2,57 @@
...
+    return vec4<f32>(r, g, b, 1.);
+}
 
 fn sample_im(texcoord: vec2<f32>, sizef: vec2<f32>) -> vec4<f32> {
     $$ if img_format == 'f32'
-        return textureSample(t_img, s_img, texcoord.xy);
+        $$ if colorspace == 'yuv420p'
+            let y = textureSample(t_img, s_img, texcoord.xy, 0).x;
+            let u = textureSample(t_img, s_img, texcoord.xy / 2.0, 1).x;
+            let v = textureSample(t_img, s_img, texcoord.xy / 2.0 + vec2<f32>(0.5, 0.0), 1).x;
```
hmaarrfk, line 35: I'm not 100% convinced you avoid a copy when the source is contiguous and the destination is not contiguous.

Your destination buffer here implies that you are using:

```
   ________
  |         |
  |  y      |
  |_________|
  | u  |    |
  |---------|
  |_v__|____|
```
Which is not super optimal.

almarklein: The v is actually to the right of the v (I changed that relatively recently, but it does not really matter). In any case, what you say is true: in the destination texture, the u and v are not contiguous. I don't think this is a problem, it does not result in an extra copy or anything like that, it's copying the data into the texture anyway, and the API for doing so does not suggest that strides hurt performance.

It would be possible to copy u and v to a contiguous part of the texture, but then the interpolation would be all wrong.

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,267 @@
...
+        tex.send_data((0, 0, 0), y)
+        tex.send_data((0, 0, 1), u)
+        tex.send_data((w // 2, 0, 1), v)
+    elif FORMAT == "yuv444p":
+        y = data[0]
+        u = data[1]
+        v = data[2]
+        tex.send_data((0, 0, 0), y)
+        tex.send_data((0, 0, 1), u)
+        tex.send_data((0, 0, 2), v)
```
almarklein, line 237: I think we can even just do `tex.send_data((0, 0, 0), data)` here.

pygfx/resources/_texture.py
```diff
@@ -253,10 +265,33 @@ def usage(self):
...
+        * "physical": the colors are in the physical / linear space, where lighting
+          calculations can be applied. The shader convers sRGB colors to physical in
+          the shader before doing color computations.
+        * "yuv420p": A common video format. The data is represented as 3 planes (y, u, and v).
+          The y represents intensity, and is at full resolution. The u and v planes are a
+          quarter of the size. The planes must be stored in two layers of the texture,
+          with the u and v plane next to each-other in top half the second layer.
+        * "yuv444p": A lesser common video format. The data is represented as 3 planes
+          (y, u, and v) similar to yuv420p however the u and v planes are stored
+          at full resolution.
```
almarklein, line 282: Nice to have support for yuv444p too, makes a lot of sense, it being a simpler variant of yuv420p.

pygfx/resources/_texture.py
```diff
@@ -42,7 +42,9 @@ class Texture(Resource):
         automatically determined from the data.
     colorspace : str
         If this data is used as color, it is interpreted to be in this
-        colorspace. Can be "srgb" or "physical". Default "srgb".
+        colorspace. Can be "srgb", "physical", "yuv420p", or "yuv444p". Default "srgb".
+    colorrange : str
+        For YUV textures, this is either "limited", or "full". See... TODO BY MARK
```
almarklein, line 47: I'm half-wondering if this should just be a suffix to the colorspace, e.g. "yuv420p-full"

examples/feature_demo/video_yuv.py
hmaarrfk: I thought this would work but it didn’t. 

Gave me garbled results.

pygfx/renderers/wgpu/wgsl/image_common.wgsl
hmaarrfk: Btw. You are interpolating between u and v at their boundary as is. So it isn’t entirely correct as is right now.

pygfx/resources/_texture.py
hmaarrfk: yeah, i wanted to mostly compare the performance of difference color conversions.

pygfx/resources/_texture.py
hmaarrfk: I just feel like we would be making the whole `yuvj420p` vs `yuv420p` mistake that took ffmpeg like 4 years to undo recently.

They now explicitely have the pixel format and the color range as two separate parameters.

hmaarrfk:
> BTW: this PR also makes it easier to use multi-layer textures for other things, like #882.

@almarklein perhaps we can split off the important parts for #882 separately from this?

I think there are still a few design concerns that have to do with:

1. Correctness of interpolations.
2. A few performance benchmarks accross differrent OS/hardware combinations.

I generally think a YUV object is really important base object to build out.

Personally, we have found it difficult to find a "frame by frame" video player that allows you to "zoom in and out" of videos. I think that would be a cool small application to build out with PyGFX as a demo (one day you know).

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,267 @@
...
+
+# Setup the rest of the viz
+
+CanvasClass = OffscreenCanvas if OFFSCREEN else WgpuCanvas
+canvas = CanvasClass(size=(w // 2, h // 2), max_fps=999, vsync=False)
+renderer = gfx.renderers.WgpuRenderer(canvas, show_fps=True)
+scene = gfx.Scene()
+scene.add(im)
+camera = gfx.OrthographicCamera(w, h)
+camera.local.position = w // 2, h // 2, 0
```
hmaarrfk, line 211: I take it the idea here was to ensure that we render every. single. pixel. of. the. video.

How can I make sure the internal texture is the right size?

For example, your M1 uses fractional scaling, and so does my modern laptop.

hmaarrfk:
Ok. Here is a radical idea….

Can each of the arrays have multiple MIP levels, making it so that we can write to MIP levels 0 in the case of native resolution subtextures. Such as Y. 

Then we can write to MIP levels 1 for substances textures.

pygfx/renderers/wgpu/wgsl/image_common.wgsl
almarklein: Good point, we may want to tweak the texcoords to clip the last half-pixel.

examples/feature_demo/video_yuv.py
almarklein: That was not even intentional. The default canvas size 640x480, I changed that 640x360 to give it the same aspect ratio as the video. The fact that, with a pixel_ratio of 2, this is exactly the same as the video resolution was a coincidence 😅 

BTW on a screen that is not hidpi, the internal texture will be twice the size, so the "every-single-pixel-is-rendered applies to any screen (except perhaps when someone has a weird OS pixel_ratio). In any case, I don't think it matters much; the rendering is not the bottleneck.

almarklein:
> @almarklein perhaps we can split off the important parts for https://github.com/pygfx/pygfx/issues/882 separately from this?

To be honest I'd rather wrap up this PR, and improve on it in future PR's. I think that if we only merge the multi-layer texture stuf, and someone would contribute something with that, we'd get into merge conflicts because we both touch the image wgsl. So I think its better to have the base version for yuv support in place, and then add to it where necessary (e.g. fix the interpolation).

> Can each of the arrays have multiple MIP levels, making it so that we can write to MIP levels 0 in the case of native resolution subtextures. Such as Y. [...] Then we can write to MIP levels 1 for substances textures.

I think that can work, using [textureSampleLevel()](https://www.w3.org/TR/WGSL/#texturesamplelevel) in the shader. The downside is that we than create 3 layers, times 1.25, so it allocates nearly twice the amount of memory on the GPU, and over half of it is not used.

The advantages are that it avoid that u and v then are discontiguous, and that the interpolation at their boundary. I'm not sure it's worth it, because we can fix the interpolation and I doubt that the contiguousness affects performance.

hmaarrfk:
> To be honest I'd rather wrap up this PR

Alright, please take a look at the latest commit. It provides an option to use 3 different grids instead of a single one. I'm really interested in benchmarking the different options but I won't really have time to do so for a few (working) days at least.

> I think that can work, using [textureSampleLevel()](https://www.w3.org/TR/WGSL/#texturesamplelevel) in the shader. The downside is that we than create 3 layers, times 1.25, so it allocates nearly twice the amount of memory on the GPU, and over half of it is not used.

Are you sure this is the case, or are the "3 layers" only created "if they are used". So that would be "3 layers  * 1.25 in virtual memory", but then in resident GPU memory, only "3 * 1 + 2 * 0.25" would be used? However, this is definitely more effort than I have today (and in the coming weeks) to explore so I won't be able to act on much. 

Thank you for the reference to the function that one would use!

Regarding the contiguous source + destination. To my knowledge, PCIe transfers MUST be done on aligned memory, both on the HOST (CPU) and the TARGET (GPU). This is often NOT a requirement for shared memory integrated GPUs (intel or apple M1-M4).

Ultimately, there aren't many good ways to write heuristics that will work in all cases. Here I use cupy to transfer some data FROM the GPU to the CPU (the other way seems to happen asyncronously and I seem to not be able to get it to do it in a blocking fashion to complete my timing correctly).

image

This is the main finding one should take:  above 1MB transfer size, it seems that we can saturate the bandwidth of PCIe gen3. Why does this matter, well a 1080p stream is 2Megapixels. so the Y component is 2MB, and the UV are each 0.5MB. if you make 3 BIG transfers, then you are pretty close to your max theoretical speed on PCIe Gen3. But if you transfer UV in 540 different transfers (even if each are aligned), then you are getting really close to 1/5th of the performance. I don't think this is python limited, but maybe.

In either case, these kinds of performance numbers are often the ones we see, keeping things in "1MB chunks"  is good for image transfers. Smaller than that you are hurting modern hardware.

almarklein:
> To my knowledge, PCIe transfers MUST be done on aligned memory, both on the HOST (CPU) and the TARGET (GPU).

Ah, I had no idea ... nice to see the option added, so we can actually benchmark it. I wonder how much it further improves performance. Though I guess to benchmark it well, you'd need to stream multiple video's :)

hmaarrfk:
> To my knowledge, PCIe transfers MUST be done on aligned memory, both on the HOST (CPU) and the TARGET (GPU).

BTW, for reference, apple basically removed the NVME controller from their SSDs likely to get around these kinds of limitations. Most SSDs have a controller on them (see https://www.tomshardware.com/news/samsung-launches-ssds-based-on-5nm-ssd-controller)  whereas MAC seems to have completely removed them. Their new SSDs just have the storage chips on them https://youtu.be/cJPXLE9uPr8?t=996 The controller was just an other ARM processor, and thus all your system is doing is:

1. Packetizing things into a format that PCIe can send.
2. The arm processor receives it and puts it into a format those storage chips can hold.....

Why have a middle man...


> Though I guess to benchmark it well, you'd need to stream multiple video's :)

I'm trying to see if I can get meaningful measurements to appear on some GPU enabled systems I have. But right now, they are on the other side of the world and I did alot of work to disable the GPU from getting used by PyGFX by default lol....

hmaarrfk:
Oh man, I'm so sad, I can't see any difference on a GPU powered device:

<details><summary>my sad sad benchmark on an NVIDIA GPU + offscreen + large 4k video</summary>

```
$ python examples/feature_demo/video_yuv.py --offscreen --three-grid-yuv --filename bbb_sunflower_2160p_60fps_normal.mp4
pygfx version from git (0.4.1) and __version__ (0.6.0) don't match.
Reading video in yuv420p format: 198.53 frames per second
No windowing system present. Using surfaceless platform
No config found!
No config found!
Max vertex attribute stride unknown. Assuming it is 2048
Time to first draw: 0.64
FPS: 89.1
FPS: 92.7
FPS: 84.4
FPS: 65.6
FPS: 60.4
FPS: 56.1
FPS: 52.4
FPS: 52.2
FPS: 52.3

██ pygfx_adapter_info:

      vendor:  NVIDIA
architecture:
      device:  NVIDIA GeForce RTX 4090
 description:  550.107.02
   vendor_id:  4.31K
   device_id:  9.86K
adapter_type:  DiscreteGPU
backend_type:  Vulka
```


```
$ python examples/feature_demo/video_yuv.py --offscreen --no-three-grid-yuv --filename bbb_sunflower_2160p_60fps_normal.mp4
pygfx version from git (0.4.1) and __version__ (0.6.0) don't match.
Reading video in yuv420p format: 199.73 frames per second
No windowing system present. Using surfaceless platform
No config found!
No config found!
Max vertex attribute stride unknown. Assuming it is 2048
Time to first draw: 0.64
FPS: 86.3
FPS: 91.2
FPS: 76.9
FPS: 65.3
FPS: 74.9
FPS: 63.2
FPS: 51.4
FPS: 52.1
FPS: 51.5

██ pygfx_adapter_info:

      vendor:  NVIDIA
architecture:
      device:  NVIDIA GeForce RTX 4090
 description:  550.107.02
   vendor_id:  4.31K
   device_id:  9.86K
adapter_type:  DiscreteGPU
backend_type:  Vulkan
```
</summary>

almarklein:
> I can't see any difference on a GPU powered device

You can also see this as good news 🤷

hmaarrfk:
Some tests:

Doing this

```patch
diff --git a/examples/feature_demo/video_yuv.py b/examples/feature_demo/video_yuv.py
index 2e02017..d36e2fd 100644
--- a/examples/feature_demo/video_yuv.py
+++ b/examples/feature_demo/video_yuv.py
@@ -348,6 +348,8 @@ def animate():
         rgba[:, :, :3] = data
         tex.send_data((0, 0, 0), rgba)

+
+def animate():
     with stats:
         renderer.render(scene, camera, flush=False)
     stats.render(
```
<details><summary>Do not load data, do not send data, just animate</summary>

```
git/pygfx$ python examples/feature_demo/video_yuv.py --offscreen --three-grid-yuv --filename ../../bbb_sunflower_2160p_60fps_normal.mp4
pygfx version from git (0.4.1) and __version__ (0.6.0) don't match.
Reading video from ../../bbb_sunflower_2160p_60fps_normal.mp4
Format: yuv420p
Color range: limited
Offscreen: True
Three grid YUV: True
Reading video in yuv420p format: 200.45 frames per second
No windowing system present. Using surfaceless platform
No config found!
No config found!
Max vertex attribute stride unknown. Assuming it is 2048
Time to first draw: 0.63
FPS: 208.6
FPS: 214.7
FPS: 210.8
FPS: 160.9
FPS: 147.3
FPS: 146.4
FPS: 145.7
FPS: 145.3
FPS: 144.7

██ pygfx_adapter_info:

      vendor:  NVIDIA
architecture:
      device:  NVIDIA GeForce RTX 4090
 description:  550.107.02
   vendor_id:  4.31K
   device_id:  9.86K
adapter_type:  DiscreteGPU
backend_type:  Vulkan
```
</details>

| Experimental Condition | FPS |
|------------------------------------|------|
| Just render the same old scene | 200 -> 144 fps |
| Just render the same, but keep transfering data | 180 fps (flat????) | 
| Just render the same old, do not transfer data, load next frame | 93 fps -> 50 fps | 

Basically, I guess the slowdowns are at the interplay between reading in the next frame which seems to be causing the frame rate to drop very fast.

examples/feature_demo/video_yuv.py
hmaarrfk: ok, well i made it so.

pygfx/renderers/wgpu/wgsl/image_common.wgsl
hmaarrfk: i'm not sure you'll be able to do it for all the different interpolation modes. How would you get around bicubic???? Is that a hardware interpolation mode?

In either case, with the 3 grid mode, I'm not sure it is strictly necessary to do it for the "3D grid" mode.

examples/feature_demo/image_plus_points_yuv.py
hmaarrfk: I guess we have both now and we can experiment.

examples/feature_demo/video_yuv.py
```diff
@@ -0,0 +1,368 @@
...
+# The script will then render to an offscreen canvas for a few seconds instead.
+
+# Set the filename of the file. We use imageio to get a stock video.
+# TODO: when https://github.com/imageio/imageio/pull/1103 is merged and a new release is out, we can use the yuv420 video instead.
+# Local: FILENAME = "/Users/almar/dev/py/imageio-binaries/images/cockatoo_yuv420.mp4"  # yuv420p
+# FILENAME = imageio.core.Request("imageio:cockatoo.mp4", "r?").filename  # yuv444p
+
+# Set the format to read the frames in. If this matches the video's storage format,
+# no conversion will have to be done. But it also affects how the data is handled in Pygfx.
+# Set to "rgb24", "rgba", or "yuv420p"
```
almarklein, line 89: These comments can be deleted, as they are part of the argparse stuff now.

pygfx/renderers/wgpu/wgsl/image_common.wgsl
almarklein: > i feel like you have a good idea on how to do this. Can i leave it to you?

Yeah, I can have a look.

> How would you get around bicubic???? Is that a hardware interpolation mode?

Anything beyond linear interpolation (including cubic) must be implemented in the shader.

examples/feature_demo/image_plus_points_yuv.py
almarklein: awesome! This example turned into a video-optimization-benchmark 😄

almarklein:
@hmaarrfk ready?

hmaarrfk:
Ready on my end

almarklein:
ping @Korijn 😄

pygfx/renderers/wgpu/wgsl/image_common.wgsl
```diff
@@ -2,10 +2,70 @@
...
 fn sample_im(texcoord: vec2<f32>, sizef: vec2<f32>) -> vec4<f32> {
     $$ if img_format == 'f32'
-        return textureSample(t_img, s_img, texcoord.xy);
+        $$ if colorspace == 'yuv420p'
+            $$ if three_grid_yuv
+            let y = textureSample(t_img, s_img, texcoord.xy).x;
+            let u = textureSample(t_u_img, s_img, texcoord.xy).x;
+            let v = textureSample(t_v_img, s_img, texcoord.xy).x;
+            $$ else
+            let txy = clamp(texcoord.xy / 2.0, 0.5 / sizef, 0.5 - 0.5 / sizef);
```
hmaarrfk, line 40:
```suggestion
            // In this implementation we share a single 2D texture between U and V
            // We must therefore take care to not sample at the edge where
            // the texture will be poorly interpolated. See
            // https://github.com/pygfx/pygfx/pull/873#issuecomment-2516613301
            let txy = clamp(texcoord.xy / 2.0, 0.5 / sizef, 0.5 - 0.5 / sizef);
```

Korijn (approved):

PR #878, opened 2024-11-04
Adjust to new wgpu-py

almarklein:
* [x] Drop python 3.8, test for 3.13.
* [x] Use `_sync()` methods.
* [x] Don't set stencil attributes when using a depth-only texture.  
* [x] Bump wgpu version (yet to be released).


almarklein:
This is ready too. Let's do a release when this and #884 are merged.

Korijn (approved):

PR #879, opened 2024-11-04
Implement texture/buffer send_data()

almarklein:
This enables writing to a texture without the texture having data on the CPU. This can remove the need to create data-copies, as in #873.

* [x] Add `texture.send_data()`.
* [x] Add `buffer.send_data()`.
* [x] Test in the form of a validation example.
* [x] Example.


hmaarrfk:
So if I understand this correct today, if I have a large texture (1GB) it likely lives in 3 places:

1. A decompressed state on the CPU.
2. A wgpu CPU representation
3. As transmitted from WGPU to the "GPU" whether integrated or discrete.

Today, One copies the data from 1->2, then asynchronously, the data gets copied from 2->3.

This would allow one to go from 1->3, skipping 2?

almarklein:
> [...] today, if I have a large texture (1GB) it likely lives in 3 places:

Not necessarily. If your data is already one complete chunk, then `1 == 2`. But if, for whatever reason, in order to create 2, you'd have to allocate an array and copy 1 into it, that's a waste that can be a problem if performance matters a lot. 

> This would allow one to go from 1->3, skipping 2?

Exactly. Another advantage could be if you want to preserve memory, so the gc can clean 1 and 2.

almarklein:
@Korijn any thoughts on this before I move this further?

pygfx/resources/_texture.py
```diff
@@ -258,6 +262,55 @@ def generate_mipmaps(self):
         """Whether to automatically generate mipmaps when uploading to the GPU."""
         return self._generate_mipmaps
 
+    def send_data(self, offset, data):
+        """Method to directly sync chunks of data to the GPU texture.
+
+        This provides a lower-level data-upload approach, intended for
+        power-users who want to avoid data copies. Can only be used when
+        the texture has no local data. Requires the COPY_DIST usage.
```
Korijn, line 270: Typo?

Korijn (approved):
Looks good to me. I think if this is a performance critical code path, perhaps all the input validation is over kill. If it is, you could add a kwarg to skip validation and just accept the upload chunk directly.

almarklein:
rdy

Korijn (approved):

Issue #641, opened 2024-02-06
How do I create a pygfx geometry with texture coordinates from an open3d triangle mesh?


cansik:
I use [open3d](https://github.com/isl-org/Open3D) a lot to manipulate 3d models and point clouds. Open3d's built-in 3d engine is ok for a lot of things, but pygfx is much easier to use. That's why I like to use pygfx to render the 3d models, and to efficiently transfer the open3d [triangle mesh](https://www.open3d.org/docs/release/python_api/open3d.geometry.TriangleMesh.html) to pygfx geometry, I've implemented the following method. The method is based on [pygfx.geometries.geometry_from_trimesh](https://pygfx.readthedocs.io/en/latest/_autosummary/geometry/pygfx.geometries.geometry_from_trimesh.html).

```python
def load_from_open3d(path: str) -> gfx.Geometry:
    # load open3d mesh as model and extract first mesh
    o3d_model: rendering.TriangleMeshModel = io.read_triangle_model(path)
    o3d_mesh_info: rendering.TriangleMeshModel.MeshInfo = o3d_model.meshes[0]
    o3d_mesh: geometry.TriangleMesh = o3d_mesh_info.mesh

    # create geometry
    triangle_uvs = np.array(o3d_mesh.triangle_uvs, dtype=np.float32)
    triangles = np.array(o3d_mesh.triangles, dtype=np.uint32)

    vertex_normals = np.array(o3d_mesh.vertex_normals, dtype=np.float32)
    vertex_colors = np.array(o3d_mesh.vertex_colors, dtype=np.float32)
    vertices = np.array(o3d_mesh.vertices, dtype=np.float32)

    # translate uv coordinates from opengl to wgpu (uv.y = 1 - uv.y)
    triangle_uvs_wgpu = (triangle_uvs * np.array([1, -1]) + np.array([0, 1])).astype(np.float32)

    # create pygfx geometry
    gfx_geometry = gfx.Geometry(indices=triangles, positions=vertices,
                                normals=vertex_normals, texcoords=triangle_uvs_wgpu)

    return gfx_geometry
```

Take a look at the Minimal Reproducible Example (MRE): [cansik/gfx-open3d-example](https://github.com/cansik/gfx-open3d-example)

#### Problem Description
Loading the geometry works, however, when adding a texture to the model, the texture coordinates are completely off (see Figure 1).

[...]

PR #885, opened 2024-11-14
Support for loading Open3D triangle meshes, materials and pointclouds into pygfx

cansik:
This pull request (PR) adds support for loading [Open3D](https://github.com/isl-org/Open3D) geometries and materials into pygfx, as discussed in #641. Open3D provides a comprehensive suite of tools for loading, processing, and saving 3D geometry data. This PR introduces compatibility features similar to the existing integration for Trimesh.

image

The implementation supports loading triangle meshes, materials, and point clouds from Open3D objects. Additionally, I have added two new examples to the gallery to demonstrate these capabilities.

### Key Features and Methods:

- `pygfx.geometries._compat`
  - `geometry_from_open3d(x)` - Is exposed through `pygfx.geometries` and calls one of the other two methods.
  - `geometry_from_open3d_triangle_mesh(x)`
  - `geometry_from_open3d_point_cloud(x)`

- `pygfx.materials._compat`
  - ~`texture_from_numpy_array(image_array, dim=2, **kwargs)`~
  - `material_from_open3d(x)` - Is exposed through `pygfx.materials`

- `pygfx.utils.load`
  - `meshes_from_open3d(mesh_model, materials=True)`


cansik:
There is currently an issue, that open3d is only built for python 3.11 and the gallery is built on 3.12. How should this be fixed? Removing it from the gallery completely?

setup.py
```diff
@@ -37,6 +37,7 @@
         "trimesh",
         "httpx",
         "gltflib",
+        "open3d",
```
almarklein, line 40: We did some refactoring a few weeks ago, replacing `setup.py` with a `pyproject.toml`. I'm not sure how familar you are with git, but this is what I'd do:

* Undo this change here and commit it.
* Checkout and pull the latest main.
* Then `git merge main` in your branch.
* Add the "open3d" dependency in the right place again.

pygfx/materials/_compat.py
```diff
@@ -144,3 +144,132 @@ def material_from_trimesh(x):
...
+        raise ValueError("Input array must be a 2D or 3D numpy array")
+
+    # Normalize and adjust channels if needed
+    if image_array.ndim == 2:
+        # Assume single-channel grayscale image
+        buffer_shape = (*image_array.shape, 1)
+        image_array = image_array.reshape(buffer_shape)
+
+    # Create a memoryview of the data for the texture
+    m = memoryview(image_array)
```
almarklein, line 192: This is not necessary, you can pass a numpy array to a `gfx.Texture`.

pygfx/materials/_compat.py
```diff
@@ -144,3 +144,132 @@ def material_from_trimesh(x):
         raise NotImplementedError(f"Conversion of {type(x)} is not supported.")
 
     return gfx_material
+
+
+def texture_from_numpy_array(image_array, dim=2, **kwargs):
```
almarklein, line 149: I don't get the use for this function. It does not seem to add anything that `gfx.Texture(image_array, dim=2)` already does, except converting to an ndarray when the input is an open3d image. I think this conversion is best solved in another way.

pygfx/materials/_compat.py
```diff
@@ -144,3 +144,132 @@ def material_from_trimesh(x):
...
+    -------
+    converted : Material
+        The converted pygfx material.
+
+    """
+
+    import open3d as o3d
+    from pygfx.materials import MeshPhongMaterial, MeshStandardMaterial
+    from pygfx import Color
+    import numpy as np
```
almarklein, line 216: Makes sense to put the `open3d` import here, so its only loaded when needed, but the other imports can be at the top of the module.

almarklein (changes requested):
Nice work! I think the structure makes sense. Made a few comments.

> There is currently an issue, that open3d is only built for python 3.11 and the gallery is built on 3.12

We can either disable running the examples that need open3d, or we can change the python version for the doc builds
[here](https://github.com/pygfx/pygfx/blob/e3834df7f8c9e230d6723838aed90207e0df0958/.readthedocs.yaml#L9) and [here](https://github.com/pygfx/pygfx/blob/e3834df7f8c9e230d6723838aed90207e0df0958/.github/workflows/ci.yml#L51).

I'm ok with the latter. I guess it makes sense that some of the dependencies used in the examples may not run on the latest Python version.

setup.py
cansik: I missed to fetch from upstream before working on the integration, thanks for the hint. I adjusted it and added open3d to the dev requirements.

pygfx/materials/_compat.py
cansik: Good point, I removed the method and directly create the texture from the numpy array.

pygfx/materials/_compat.py
cansik: Has been refactored and removed.

pygfx/materials/_compat.py
cansik: Yes, some of the imports are anyways already imported. I have refactored it.

cansik:
@almarklein  I have updated code as requested.

In my opinion I would leave it at version 3.12 at the moment. Open3D currently doesn't have any great examples that need to be built anyway, and the 3.12 version should be released soon. But if you want to downgrade the CI Python version, I can of course do it.

almarklein:
> In my opinion I would leave it at version 3.12 at the moment.

Ok, let's do that, and enable the example when Open3D works on 3.12

pyproject.toml
```diff
@@ -49,7 +49,7 @@ docs = [
     "imgui-bundle>=1.2.1",
 ]
 tests = ["pytest", "psutil", "trimesh", "httpx", "gltflib", "imageio"]
-dev = ["pygfx[lint,tests,examples,docs]"]
+dev = ["pygfx[lint,tests,examples,docs]", "open3d"]
```
almarklein, line 52: Can you please add it to the list of example deps instead? (The `dev` deps include the `examples` deps).

pygfx/utils/load.py
```diff
@@ -562,3 +564,54 @@ def _volume_from_voxelgrid(vxl, cmap=None, clim="data"):
...
+    -------
+    results : list of gfx.Mesh
+        A list of gfx.Mesh objects converted from the input mesh model.
+
+    Raises
+    ------
+    ValueError
+        If the input is not an Open3D TriangleMeshModel.
+    """
+    import open3d
```
almarklein, line 590: Would be nice to apply the same `find_spec` trick as we do for `trimesh`, so that when open3d is not installed, ppl get a friendlier error message.
``py
if not find_spec("open3d"):
        raise ImportError(
            "The `open3d` library is required for this function: pip install open3d"
        )
```

File: pygfx/__init__.py [hunk 7]
```diff
@@ -16,7 +16,7 @@
 from .renderers import *
 
 from .utils.color import Color
-from .utils.load import load_mesh, load_meshes, load_scene
+from .utils.load import load_mesh, load_meshes, load_scene, meshes_from_open3d
```
almarklein, line 19: To make the API somewhat consistent with regards to the functions privided by `pygfx.utils.load`, I think it makes sense if this'd be `load_meshed_from_open3d(path)` instead.

The geometry and material compat functions are fine, but the public function in `load.py` (the ones we import here, anyway) all work with a path. Does that make sense?

pyproject.toml
cansik: But won't it fail then on 3.12?

pygfx/__init__.py
cansik: `meshes_from_open3d` does not take a`Path` as argument, but a `TriangleMeshModel`, which is a combination of materials and meshes. So it's neither a geometry helper nor a material helper.

I can understand your concern and it was not clear to me where to put the method. I could create a `load_meshed_from_open3d(path)` which is using the current method, but I don't really see the benefit (other than having one more method). Trimesh is already doing a good job in `load_meshes`. The method `meshes_from_open3d(trianglemeshmodel)` is used to convert an already in-memory `TriangleMeshModel` into a `List[gfx.Mesh]`.

But I may not have the full picture here. My intention is to make it possible to use Open3D to load and process 3D models, because it's really good at that. But when it comes to visualisation, Open3D can be a real pain if you want to customise the rendering or the UI. That's where pygfx does a much better job, with a clear API and a lot of freedom. What I did not intend to do was to "replace" trimesh or gltflib completely.

Of course, I am open to other ideas and I could extend the integration to cover more cases (like loading from file and so on).

pygfx/utils/load.py
cansik: I have added the spec check for all methods that use open3d.

pyproject.toml
almarklein: Oh right ... 🤔  Perhaps its better to not specify it as a dependency. If someone runs the example, that person will have to manually install open3d for the time being.

pygfx/__init__.py
almarklein: > My intention is to make it possible to use Open3D to load and process 3D models, [...]

Right, and then you need something that takes the in-memory model, not a filename. Gotcha.

The thing I'm a bit worried about is that our API w.r.t. these load/conversion methods feels a bit inconsistent. There are multiple libs that can read meshes, and if we want to play nice with more of these, in a way that does not get messy, we need a better plan. Something to look into in the future ...

I think that for now, we can cancel this little change `from .utils.load import ... meshes_from_open3d`. This means that in the examples you'd have to do `gfx.load.meshes_from_open3d`. This is a bit less pretty, but when the time comes that we implement a better load/convert API, we can probably leave `meshes_from_open3d` intact, while if it would be in the `gfx` namespace, may want to deprecate it. 

In other words, I care a lot about keeping the `gfx` namespace clean, but I don't mind a lot if `gfx.load` gets multiple somewhat inconsistent functions 😄

almarklein: Maybe @Korijn has an opinion about this too.

cansik: Yes, let's remove it from the global import to keep the load API consistent. I've already changed the code.
Is there any idea of moving the IO part for gfx into it's own library in the future?

pyproject.toml
cansik: I've added the same `find_spec` to the open3d examples for now and removed the dependency from pygfx completely.

pygfx/__init__.py
Korijn: The important parts are:

* Ensure open3d imports are optional
* The compatibility layer between open3d objects and pygfx objects is what matters most
* Direct loading functions are for convenience
* Consistency between different libraries is a pipe dream

almarklein: > Is there any idea of moving the IO part for gfx into it's own library in the future?

I don't think there's much use. We might as well pick one of these libs and award it with "the official pygfx mesh io lib" 😉. And as Korijn said, abstracting over all io libraries is not feasible.

almarklein (approved):

pygfx/__init__.py
almarklein: I made an issue to discuss how to move forward with io: https://github.com/pygfx/pygfx/issues/888

cansik: Sounds good to me. I think the PR now fulfils the requirements.

cansik:
@Korijn I just wanted to check in and see if you've had a chance to look at the code yet?

Korijn (approved):
LGTM! Thanks!

Issue #343, opened 2022-10-13
Improvements to Color


almarklein:
Colorspaces
* Support for HSLuv, a colorspace with uniform brightness https://www.hsluv.org/

CSS:
* Support "rgb(100%, 50%, 0%)"
* Support "hsl(...)" and "hsla(...)"

Other:
* Color.lerp(color, t) linear interpolate towards other color.
* Color.lerpHSL(color, t) same as lerp but interpolate in HSL / HSLuv space.
* Color.lighter(factor) and Color.darker(factor)

PR #887, opened 2024-11-16
implement suggested improvements listed in issue343

zetyquickly:
Issue https://github.com/pygfx/pygfx/issues/343

1. Added support for HSLuv (with new dependency). With unified interface in range [0, 1] like others.
2. CSS string support for rgb(100%, 50%, 0%) was maintained
3. Added support for hsl(XYZdeg, XY%, YZ%), hsla(...), hsv(...), hsva(...), hsluv(...), hsluva(...)
4. Added `lerp(self, target, t)` for RGB and `lerp_in_hue(self, target, t, colorspace)` for all other hue-based colorspaces
5. Added `lighter(self, factor, colorspace)` and `darker(self, factor, colorspace)`, implemented only in HSL and HSLuv because they have lightness components

### DIFF
--- a/pygfx/utils/color.py
+++ b/pygfx/utils/color.py
@@ -423,7 +423,7 @@ class Color:
         return h / 360.0, s / 100.0, light / 100.0, self.a
 
     def lerp(self, target, t):
-        """Linear interpolate from source color towards target color with factor t.
+        """Linear interpolate from source color towards target color with factor t in RGBA space.
 
         Parameters
         ----------
@@ -443,7 +443,7 @@ class Color:
         a = self.a + (target.a - self.a) * t
         return Color(r, g, b, a)
 
-    def lerp_in_hue(self, target, t, colorspace="hsv"):
+    def lerp_in_hue(self, target, t, colorspace="hsluv"):
         """Linear interpolate from source color towards target color with factor t in specified colorspace.
 
         Parameters
@@ -453,7 +453,7 @@ class Color:
         t : float
             Interpolation factor between 0 and 1
         colorspace : str
-            The colorspace to interpolate in. One of "hsl", "hsluv", "hsv"
+            The colorspace to interpolate in. One of "hsl", "hsluv", "hsv". Default is "hsluv"
 
         Returns
         -------
@@ -490,7 +490,7 @@ class Color:
 
         return to_rgba(h, s, light, a)
 
-    def lighter(self, factor=0.5, colorspace="hsv"):
+    def lighter(self, factor=0.5, colorspace="hsluv"):
         """Make the color lighter by the given factor.
 
         Parameters
@@ -498,7 +498,7 @@ class Color:
         factor : float
             Factor to lighten by, between 0 and 1. Default is 0.5
         colorspace : str
-            The colorspace to use, one of "hsl", "hsluv", or "hsv". Default is "hsl"
+            The colorspace to use, one of "hsl", "hsluv", or "hsv". Default is "hsluv"
 
         Returns
         -------
@@ -525,7 +525,7 @@ class Color:
             v = v + (1 - v) * factor
             return Color.from_hsv(h, s, v, self.a)
 
-    def darker(self, factor=0.5, colorspace="hsv"):
+    def darker(self, factor=0.5, colorspace="hsluv"):
         """Make the color darker by the given factor.
 
         Parameters
@@ -533,7 +533,7 @@ class Color:
         factor : float
             Factor to darken by, between 0 and 1. Default is 0.5
         colorspace : str
-            The colorspace to use, one of "hsl", "hsluv", or "hsv". Default is "hsl"
+            The colorspace to use, one of "hsl", "hsluv", or "hsv". Default is "hsluv"
 
         Returns
         -------
--- a/pygfx/utils/color.py
+++ b/pygfx/utils/color.py
@@ -2,10 +2,20 @@
 
 import ctypes
 import colorsys
-import hsluv
-
 
 F4 = ctypes.c_float * 4
+
+# Try to import hsluv once at module level
+try:
+    import hsluv
+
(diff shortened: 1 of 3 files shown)


pygfx/utils/color.py
```diff
@@ -320,10 +339,32 @@ def from_hsv(cls, hue, saturation, value):
...
+        """Create a Color object from an a color in the HSV (a.k.a. HSB) colorspace.
+
+        HSV stands for hue, saturation, value (aka brightness). The hue
+        component indicates the color tone. Values go from red (0) to
+        green (0.333) to blue (0.666) and back to red (1). The
+        satutation indicates vividness, with 0 meaning gray and 1
+        meaning the primary color. The value/brightness indicates goes
+        from 0 (black) to 1 (white).
+
+        The alpha channel is optional and defaults to 1.
```
almarklein, line 353: I think we should either just add an optional `alpha` param to `from_hsv()`, or add `from_hsva` with a mandatory alpha channel. Otherwise it feels like there's too much overlap.

I'm leaning towards having a single `from_x` method (with an optional alpha), for each colorspace. And for each `to_x` method also have a `to_xa` method.

pygfx/utils/color.py
```diff
@@ -335,11 +376,190 @@ def from_hsl(cls, hue, saturation, lightness):
...
+            else:
+                h1 += 1.0
+        h = (h1 + (h2 - h1) * t) % 1.0
+        s = s1 + (s2 - s1) * t
+        light = l1 + (l2 - l1) * t
+        a = self.a + (target.a - self.a) * t
+
+        return to_rgba(h, s, light, a)
+
+    def lighter(self, factor=0.5, colorspace="hsl"):
```
almarklein, line 509: Since hsluv is more human friendly, maybe that should be the default? Or is there a compatibility of expectation that makes hsl a better default?

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "freetype-py",
     "uharfbuzz",
     "jinja2",
+    "hsluv==5.0.0",
```
almarklein, line 27: We're rather careful about adding dependencies. I Checked this dependency: https://github.com/hsluv/hsluv-python: It's a single-module pure Python lib without further dependencies. IMO fine to add this for this purpose.

almarklein (commented):
Nice work! Thanks! 🚀

pygfx/utils/color.py
Korijn: HSL is very commonly used I believe

pyproject.toml
Korijn: We can make it optional? Don't need it unless you use all these color functions.

zetyquickly: can make it optional and raise an error in case it's being used

pygfx/utils/color.py
zetyquickly: I think the more important question should we implement lighter in RGB where we interpolate to "white" and to "black" when it's "darker"? or just stick to the Hue based colorspaces with Lightness components only

pygfx/utils/color.py
zetyquickly: will do

pyproject.toml
almarklein: We can also vendor the 300-line module?

pygfx/utils/color.py
almarklein: Qt does the conversion in hsv: https://doc.qt.io/qt-6/qcolor.html#lighter Let's do that too.

zetyquickly: If we do lightening in HSLuv by default, then we won't need to adjust the saturation, so the updated color won't look oversaturated.

But that requires the HSLuv to be a non-optional dependency

zetyquickly:
# Comparing Color Lightening Methods

Original color: `#33691E` (green)

## Lighter by 0.5

| Colorspace | Result |
|------------|---------|
| HSL | HSL lighter 0.5 |
| HSLuv | HSLuv lighter 0.5 |
| HSV | HSV lighter 0.5 |

## Lighter by 0.9

| Colorspace | Result |
|------------|---------|
| HSL | HSL lighter 0.9 |
| HSLuv | HSLuv lighter 0.9 |
| HSV | HSV lighter 0.9 |


HSL is the worst, dunno what's better: HSLuv or HSV only based on one example

pyproject.toml
zetyquickly: made it optional

pygfx/utils/color.py
zetyquickly: done

pyproject.toml
Korijn: Agreed. We can eventually come back on this decision anyway. Let's add it as a dep (and not vendor)

zetyquickly: updated. made hsluv default for interpolation and ;ighter/darker

zetyquickly: updated. now it's a non-optional dependency

pygfx/utils/color.py
zetyquickly: updated. made hsluv a default option for interpolation and lighter/darker

pyproject.toml
Korijn, line 27: Can you explain why you pinned it to 5.0.0 exactly? I see 5.0.4 is available for example...

zetyquickly: there was no particular reason

almarklein: Let's pin to "hsluv==5".

Korijn: Let's use ~=5 to be a little more flexible

zetyquickly: updated with a range to be consistent with previous libs in toml

almarklein:
Thanks @zetyquickly !

Issue #882, opened 2024-11-13
Multi-channel images


almarklein:
## Intro

It would be cool to support multi-channel images in pygfx, as it's a common use-case in certain scientific fields like microscopy. The idea is that you'd have multiple image of the same object, but taken with a different modality / imaging technique. See e.g. the images here: https://github.com/imageio/imageio-binaries/pull/25

## Combining semantics

We'd need some way to determine how the pixel values should be combined into the final rgba value. But what makes sense? 

It seems reasonable to be able to set the `clim` of each channel. Further it makes sense for each value to have an associated color or colormap. Then they can be combined by e.g. averaging the result.

Other options could be additive composition.  

## State of the art

In Napari, using `add_image()`  or `view_image()`  and setting the `channel_axis` arg, goes into multi-channel mode. You can then specify a list for most props (such as `contrast_limits`, `color_map` and `blending`). Under the hood, it creates a Vispy image object for each channel, and dispatches the props to them. So it's basically our option A, but with a friendlier API.

Other examples?

### Option A: scene compositing

An approach that can be applied already is to simply create multiple images, put them on top of each-other, and let the pygfx blending mechanism combine them.

An advantage is that it's easy to manage each channel separately, e.g. setting contrast limits and colormap.

A problem is that the manner in which the channels can be combined is constrained to the supported blend modes. To sortof average them, you could probably use `weighted` blending and set the alpha to 0.95, but that feels a bit hacky.

Maybe we can introduce a blend mode that averages all incoming fragments. That looks like a reasonable solution, but if the multi-channel image is in a scene with other objects, the blending for these objects would be weird.

### Option B: multi-layer textures

Assuming that the channels are all the same resolution, it makes sense to use texture layers: in wgpu textures, you can fit multiple layers of 2d images into a single texture, basically using the dimension that is normally used for `z`  in a 3D texture. It uses `wgpu.TextureViewDimension.d2_array`.

An advantage is that you avoid having `n_channels` times as much images, which is easier to manage, but may also help performance if you have a lot of images.

The challenge here is how to set the properties of the individual channels. Maybe `material.clim` and `material.map` can be tuples? 

### Option C: Grouping

Just a brainfart for now, but what if you create separate `Image`  objects, like in A, and put them as children on a common parent, that would do *something* to combine their result. Maybe this groups composites into its own result with a special blend mode, before its rendered into the real render target together with the rest of the scene.

cc @Korijn @hmaarrfk @claydugo @kushalkolar

[...]

PR #889, opened 2024-11-21
Introduce TextureMap

panxinmiao:
Add more robust support for texture in MeshMaterial to align with glTF specification requirements.

The goal is to enable pygfx users to easily load and display standard glTF models "out of the box" without needing to delve into model details or write processing code for each individual model.

This PR is part of that effort.

----

### API changes

* `ImageBasicMaterial.map_interpolation` is removed in favor of `TextureMap.mag_filter` and `.min_filter`.
* `LineMaterial.map_interpolation` is removed in favor of `TextureMap.mag_filter` and `.min_filter`.
* `PointsMaterial.map_interpolation` is removed in favor of `TextureMap.mag_filter` and `.min_filter`.
*  `VolumeBasicMaterial.map_interpolation` is removed in favor of `TextureMap.mag_filter` and `.min_filter`.


panxinmiao:
The CI build failed due to issues with the new version of imgui_bundle.
See : https://github.com/pygfx/wgpu-py/pull/645

almarklein:
This is really nice! I'm only not sure about the API. We need a way for the glTF loader to let the render know what sampler properties to use for a texture.

We don't expose something like a sampler object. This is partly because these things are sometimes determined by the material, so creating a sampler object would be an abstraction leak, I think.

Adding sampler properties to the texture also does not feel right, because a texture may never be used via a sampler.

I'm thinking that we should:
* Put the sampler properties together, in a separate object, maybe it can just be a dict.
* Attach this to the texture via a prop called something like `.sampler_hints` or `.sampler_info` or `.sampler_properties`.
* Then the shader can use these to set the sampling. And in cases where it makes sense, they can be ignored/overloaded by e.g. a material property.

I think we don't need these "hints" to be responsive. The idea is to set them before the texture's first use. If a dynamic property is needed, that should be implemented via a material. 

@korijn, @panxinmiao do you think this makese sense?

To give a bit of idea of what this *could* look like in the glTF loader:
```py
tex = gfx.Texture(...)

sampler_props = {
    "mag_filter": "nearest",
    ...
}
tex.sampler_hints.update(sampler_props) 
```

almarklein:
> But you can reuse one texture with different samplers right?

Yes. And you also load pixel values without a sampler. I think we want to offer a way for the thing that creates a texture to provide "intended sampling", and then it's up to the shader to honor that or not. E.g. for the fancy textures for mesh (pbr) rendering these sampling values will be used, but someone can do something like this:
``py
image = gfx.Image(
    gfx.Geometry(grid=mesh.material.normal_map),
    gfx.ImageBasicMaterial(interpolation="linear"),
)
```
and that will use the same texture, but ignore the sampling hints.

My suggestion using a `texture.sampling_hints` dict is one idea to do this. I'm open to other methods :) But the point is to associate these sampling values with the texture without them "dictating" the sampling. Does that make sense?

**Comment by panxinmiao** [comment]
> * Put the sampler properties together, in a separate object, maybe it can just be a dict.
> * Attach this to the texture via a prop called something like `.sampler_hints` or `.sampler_info` or `.sampler_properties`.
> * Then the shader can use these to set the sampling. And in cases where it makes sense, they can be ignored/overloaded by e.g. a material property.

For the current abstraction of textures in Pygfx, I think this approach makes sense. Let's do it this way first. :)

However, I feel like maybe we're missing a "**Mapping**" abstraction object, or perhaps we've confused the concepts of **texture** and **mapping**.

- A **texture** is the image or pattern itself. It is typically a 2D image containing pixel data, represented as a matrix, and serves as a static resource. The **Texture** object is an abstraction of this "resource".

- On the other hand, **mapping** is the process of applying a texture to a 3D model. The most common mapping method is UV mapping. The **Mapping** object should abstract this "process", including how the texture is mapped onto the model's surface.

In our material system, the various "maps" we use should actually be **Mapping** objects, not **Textures**. These should contain both the "resource" itself and the "method" by which that resource is applied.

**Comment by almarklein** [comment]
@panxinmiao good point! It makes a lot more sense for these sampling/mapping props to be associated with the material!

My first guess was:
```py
gfx.Mesh(
    ...
    gfx.MeshMaterial(map=some_texture, map_sampling=map_sampling_object)
)
```
But I also really like your idea:
```py
gfx.Mesh(
    ...
    gfx.MeshMaterial(map=some_texture)  # default sampling
)

gfx.Mesh(
    ...
    gfx.MeshMaterial(map=gfx.Map(some_texture, mag_filter="nearest", ...))
)
```

@Korijn what do you think?

Korijn:
The mapping seems like a very elegant solution! It fits well with the Material abstraction.

panxinmiao:
The `TextureMap` has been introduced. 
Various map properties in the Material class have been updated to use `TextureMap` instead of `Texture`. For backward compatibility, when a `Texture` is set in a Material's map, it will automatically be converted to a `TextureMap`.

The only uncertainty I have is regarding the `colorspace` property. It could either be placed in the `Texture` or the `TextureMap`. I tend to favor placing it in the Texture, as it describes the data of the Texture, and it's unlikely that different TextureMaps would share the same Texture but with different `colorspace` settings.

pygfx/renderers/wgpu/shader/base.py
```diff
@@ -233,16 +239,22 @@ def define_colormap(self, texture, texcoords, interpolation="linear"):
...
         In the WGSL the colormap can be sampled using ``sample_colormap()``.
         Returns a list of bindings.
         """
 
-        self["colormap_interpolation"] = interpolation
-        sampler = GfxSampler(interpolation, "clamp")
-        texture_view = self._define_colormap_texture(texture)
+        self["colormap_interpolation"] = (
+            map.mag_filter
+        )  # todo: remove colormap_interpolation?
```
panxinmiao, line 251: Currently, the colormap has some special sampler logic compared to other texturemaps. If I understand correctly, this is to enable the colormap to support different data formats? 
Is there any specific consideration for using `textureLoad` instead of `textureSample` for sampling certain colormap textures in WGSL?

almarklein:
> The only uncertainty I have is regarding the colorspace property. It could either be placed in the Texture or the TextureMap. I tend to favor placing it in the Texture, as it describes the data of the Texture,

I think I agree with you; the colorspace is also a more intrinsic property of the data, rather than a choice on how to represent the data. Plus it's used when the texture is used on a `geometry.grid`, in which case we don't have a map.

pygfx/renderers/wgpu/shader/base.py
panxinmiao: > The use of `textureLoad` is to sample from textures that have a format for which we cannot create a sampler. This is determined by `colormap_format` being `f32 / u32 / i32`.

`textureSample` can also be used with `u32` and `i32` texture formats as long as `GPUTextureSampleType` is appropriately set to `uint` or `sint`. However, for integer-format textures, interpolation and filtering are not supported, which aligns with the behavior of `textureLoad`. That said, this method provides a unified API with floating-point textures, making the code simpler and more convenient. 

Additionally, if interpolation and filtering operations are required for integer-format textures, it is recommended to preprocess the texture by converting it into a floating-point format, as filtered sample values are inherently floating-point.

pygfx/resources/_texturemap.py
```diff
@@ -0,0 +1,109 @@
...
+    Parameters
+    ----------
+    texture : Texture
+        The texture to use for this map.
+    channel : int, optional
+        The uv channel of the texture to use. Default is 0.
+    mag_filter : str, optional
+        The magnification filter to use. Default is "linear".
+    min_filter : str, optional
+        The minification filter to use. Default is "linear".
```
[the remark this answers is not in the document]
panxinmiao: I’m open to this idea but would prefer to retain flexibility for advanced users who need to set separate values.

I see two possible approaches:

- Add a `filter` parameter

Use this value for both properties if min_filter and mag_filter are not explicitly specified, as follows:

```python
self._min_filter = min_filter or filter
self._mag_filter = mag_filter or filter
```

- Single `filter` attribute

Accept a tuple when users want to assign different values to the two properties:

```python
map = TextureMap(filter=('LINEAR', 'NEAREST'))
```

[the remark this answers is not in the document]
almarklein: I like the first option. That way it feels natural for both types of users:
```py
# scientists
TextureMap(filter='linear')
# GPU dev
TextureMap(min_filter='linear', mag_filter='linear')
```

pygfx/renderers/wgpu/shader/base.py
panxinmiao: If you agree, I plan to attempt a refactor of the colormap-related logic in a separate PR, with the following goals:

- Unify the code logic for all `TextureMap` types, so that `colormap` behaves consistently with all other maps and no longer requires special handling.

- Add support for setting `alpha_mode` ("opaque", "mask", "blend") and `alpha_test` (`alpha_cut_off`) to allow us to pass the test cases available [here](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/AlphaBlendModeTest/README.md).

- Clean up the code related to Vertex Color mode to ensure compatibility with the test cases provided [here](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/VertexColorTest/README.md).

almarklein: Sounds good!

pygfx/resources/_texturemap.py
```diff
@@ -0,0 +1,119 @@
...
+    Parameters
+    ----------
+    texture : Texture
+        The texture to use for this map.
+    channel : int, optional
+        The uv channel of the texture to use.
+        e.g. with `channel=7`, it will use the `geometry.texcoords7`.
+        Default is 0, which uses the `geometry.texcoords`.
+    filter : str, optional
+        The filter to use for both magnification and minification if not being set separately.
```
almarklein, line 19:
```suggestion
        The filter to use for magnification, minification znd mipmap, if not being set separately.
```

pygfx/resources/_texturemap.py
```diff
@@ -0,0 +1,119 @@
...
+    """A texture map.
+
+    TextureMap is used to define how a texture should be used in a material.
+    including which uv channel to use, filtering, and wrapping
+
+    Parameters
+    ----------
+    texture : Texture
+        The texture to use for this map.
+    channel : int, optional
```
almarklein, line 14: I'm a bit worried about confusion with color channels (e.g. #882). Would it make sense to rename this to something like `coords` or `coord_index`? Or is `channel`  a common name in this context?

Idea: maybe we could do `coords="texcoords7"`, i.e. it represents the full name on the geometry to use. That would actually be more explicit and simpler.

pygfx/resources/_texture.py
```diff
@@ -221,6 +221,9 @@ def __init__(
             self._chunk_mask = np.ones(shape, bool)
             self._chunk_list = None
 
+        # self._channel = channel
+        # self._sampler_hints = {}
+
```
almarklein, line 226: Some old commented code here, and down below.

almarklein (commented):
Nice work, a few small suggestions.

pygfx/resources/_texturemap.py
panxinmiao: **UV channel** is a specialized term in computer graphics, representing different coordinate sets of a model. To avoid ambiguity, we can rename the attribute as `uv_channel` .

For the parameter, I believe using an index is preferable to using the geometry attribute name (like texcoords7). There are two reasons for this: first, using an index follows common conventions and habits, and second, from the perspective of user code simplicity, `TextureMap(uv_channel=7)` is much more convenient than `TextureMap(uv_channel="texcoords7")` :)

almarklein: `uv_channel`  sounds good!

almarklein (approved):

Korijn (approved):

pygfx/renderers/wgpu/shaders/imageshader.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, wobject):
             self["img_nchannels"] = 3
         elif material.map is not None:
             self["use_colormap"] = True
-            self["colorspace"] = material.map.colorspace
+            self["colorspace"] = material.map.texture.colorspace
```
hmaarrfk, line 73: Would it be possible to add a property to `map` that is defined as:


```
@property
def colorspace(self):
    return self.texture.colorspace
```
so as to avoid this needed change in downstream users?

Perhaps even with a "warning" that tells users to use `map.texture.colorspace` instead of `map.colorspace`.

Mostly a thin transition layer.

I'm happy to make the PR.

hmaarrfk: hmm the changes to the interface of `define_img_colormap` are more breaking for myself, so i've created a more complicated shims that work for me. 

I don't want push the focus toward these backward compatibility concerns too much.

PR #894, opened 2024-12-03
Auto update skeleton and skeleton_helper

panxinmiao:
This is an optional callback that gets executed immediately before a `WorldObject` is rendered. It can be very useful in specific scenarios, allowing users to dynamically update the settings or state of a particular world object before each frame is rendered.

It’s important to note that this hook is primarily intended for updating **the state or behavior of the object itself** or adding specific logic for certain objects in the scene. However, it is **not suitable** for general per-frame updates to internal GPU objects, such as updating the vertex attributes of geometry or properties in material, as these resources may be shared across multiple WorldObject instances and generally need to be updated only once per frame.

Similarly, the automatic updating of `Skeleton` faces a comparable challenge. While placing such updates in the `on_before_renderer` method of the `SkinnedMesh` object is not ideal from a design perspective, it serves as a temporary solution to relieve users from the burden of explicitly managing skeleton updates in their application logic. This is especially helpful when dealing with complex scenes (e.g., loading from glTF files), where users would otherwise need to locate all SkinnedMesh instances in the scene and manually trigger skeleton updates—a process that is both tedious and undermines out-of-the-box usability.

### DIFF
--- a/examples/feature_demo/animation_mixer.py
+++ b/examples/feature_demo/animation_mixer.py
@@ -336,9 +336,6 @@ def animate():
 
     animation_mixer.update(dt)
 
-    model_obj.children[0].skeleton.update()
-    skeleton_helper.update()
-
     with stats:
         renderer.render(scene, camera, flush=False)
 
--- a/examples/feature_demo/gltf_animations.py
+++ b/examples/feature_demo/gltf_animations.py
@@ -178,9 +178,6 @@ def animate():
     dt = clock.get_delta()
     mixer.update(dt)
 
-    model_obj.children[0].skeleton.update()
-    skeleton_helper.update()
-
     with stats:
         renderer.render(scene, camera, flush=False)
     stats.render()
--- a/examples/feature_demo/gltf_viewer.py
+++ b/examples/feature_demo/gltf_viewer.py
@@ -245,22 +245,9 @@ gui_renderer.set_gui(draw_imgui)
 load_remote_model(0)
 
 
-def update_skeleton(obj):
-    if isinstance(obj, gfx.SkinnedMesh):
-        obj.skeleton.update()
-
-
 def animate():
     dt = clock.get_delta()
     mixer.update(dt)
-
-    # Update the skeleton
-    # todo: this is a temporary workaround, we should update the skeleton automatically
-    # see: https://github.com/pygfx/pygfx/pull/894
-    if skeleton_helper and skeleton_helper.visible:
-        skeleton_helper.update()
-    if model_obj:
-        model_obj.traverse(update_skeleton)
 
     with stats:
         renderer.render(scene, camera, flush=False)
--- a/examples/feature_demo/skinned_mesh.py
+++ b/examples/feature_demo/skinned_mesh.py
@@ -113,9 +113,6 @@ def animate():
         rotation_y = math.sin(t) * 2 / len(mesh.skeleton.bones)
         bone.local.rotation = la.quat_from_euler((0, rotation_y, 0))
 
-    mesh.skeleton.update()
-    skeleton_helper.update()
-
     renderer.render(scene, camera)
     canvas.request_draw()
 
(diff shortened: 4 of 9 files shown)


almarklein:
I'm close to starting the update-propagation work in pygfx (#495). We should fit this use-case in that work.

Korijn:
> I'm close to starting the update-propagation work in pygfx (#495). We should fit this use-case in that work.

I assume this means we hold off on merging this PR?

almarklein:
Yes, this use-case should be covered with the new events / viewport etc. We can keep this open, as a reminder for this use-case.

Korijn:
With #950 this situation has already changed quite a bit, but it's still not ideal.

almarklein:
I think we already have it, and it's called `update_uniform_buffers`. In https://github.com/pygfx/pygfx/pull/955 I make that more explicit.

I actually ran into this because I'm revisiting #834 and need a hook like this too 😄

Korijn:
Does that mean we can close this PR?

panxinmiao:
> I think it should be adjusted to implement _update_object() for some skeletons / skins.

I have updated the code, but I think there are some issues that need to be reiterated.

Currently, the roles of _update_object() and on_before_render() are essentially the same, both serving for automatic updates of objects. However, as I mentioned earlier:

> It’s important to note that this hook is primarily intended for updating the state or behavior of the object itself or adding specific logic for certain objects in the scene. However, it is not suitable for general per-frame updates to internal GPU objects, such as updating the vertex attributes of geometry or properties in material, as these resources may be shared across multiple WorldObject instances and generally need to be updated only once per frame.

theoretically, the automatic updating of the Skeleton is not suitable to be handled here.

Let me try to explain it more clearly.

If we need to update the state of the SkinnedMesh itself, placing it in the SkinnedMesh's _update_object() method is appropriate. However, in this case, we are updating the Skeleton object, which is an internal object, not a renderable World Object. Therefore, the Skeleton object itself is not suitable for the _update_object() method. Placing its update in a similar method of the SkinnedMesh is a necessary compromise.

The example in skeleton_bind_detached.py illustrates this well. The Skeleton object is shared by three SkinnedMesh instances, and under the current approach, the Skeleton object is updated three times per frame during rendering, which is completely unnecessary. If an internal object is shared by more renderable world objects, the problem becomes even more pronounced.

Consider a use case where we have a dynamic Material, whose color needs to be automatically updated every frame, and this Material is shared by thousands of renderable world objects. Clearly, this kind of automatic update is not suitable for the _update_object() method of the world object.

The current automatic update mechanism using _update_object() has some limitations, as it is not suitable for internal objects (non-renderable world objects) that need to be updated every frame. We may need a more low-level automatic update mechanism to manage all internal objects that require updates on every frame.

For automatic updates of skeletons, this can currently be used as a temporary solution.

almarklein:
Good point. Would it be possible to maintain a flag, so that it will only do actual work on the first call?

panxinmiao:
> Good point. Would it be possible to maintain a flag, so that it will only do actual work on the first call?

I’ve considered this idea before, but it would require the Renderer (or another component) to track the frame count and timestamps from the start of rendering, and pass them as parameters to the _update_object() callback. Currently, I can think of situations where some objects need to be updated every frame (based on the frame number), while others need to be updated according to the timestamp (for example, if we’re implementing a dynamic water surface material with ripple effects, the rendering of the water surface would need to be updated based on the timestamp).

However, even with this approach, if an internal object is shared by many world objects, we’d need to repeatedly check every frame whether the object needs updating (by comparing its last update frame number or timestamp with the provided parameters), which doesn’t seem ideal. My suggestion is that, at the start of rendering, the Renderer (or some other component) could collect all the internal objects that need updating and manage their updates in a unified manner every frame.

almarklein:
I think there's two things to consider. One is animations. The `_update_object` is not suitable for that. The canvas will get an animate event for this, and maybe the `before_draw()` event from the new `rendercanvas` can also be used for this. If necessary, a flag can be set when an animation-step is done, so that it can be checked in `_update_object` to actually set/sync some low-level stuff.

Issue #874, opened 2024-10-25
Upcoming API breakage


almarklein:
*The implementation of #495 and #482 are coming closer. In this issue I plan to keep track of the state, and what it means for downstream code. How and when to best transition, etc. I will update this top post as we move along.*

## Current status

* In wgpu-py, we're making progress on the lower-level mechanics related to the update propagation work. 
* We're moving `wgpu.gui` to a new library [rendercanvas](https://github.com/pygfx/rendercanvas), where most of this work will happen. The `wgpu.gui` will remain for the time being, and deprecated at a later stage. For more info see https://github.com/pygfx/wgpu-py/issues/627
* We'll make sure to have a release with the current API before we start breaking stuff.
* We'll try to chop up the changes in smaller pieces that can be released separately.

almarklein:
Not directly, but I think it will make it easier to solve that problem.

PR #902, opened 2024-12-13
Implement a dither blend mode

almarklein:


hmaarrfk:
I'll be impressed if you can tame GPU randomness.


xref: https://github.com/pygfx/pygfx/issues/874
xref: https://github.com/pygfx/pygfx/issues/874

Korijn:
Is the dithering stable from frame to frame if the camera is not adjusted? That's the effect on Nintendo Switch games for example. I think they use a hash or some other derivative of the camera matrix as the source of randomness to achieve this

almarklein:
> Is the dithering stable from frame to frame if the camera is not adjusted? 

The seed consists of 4 parts: the xy ndc coord, the z-coord, and the object id. If you only use the first 2, you indeed get that nintendo vibe where the noise seems "stuck to the screen". The problem with that is that if two transparent objects overlap, they get the same pattern, so they don't "blend" well. To this end I added the object id. Actually, with only the object id you'd still get the same effect. But if you have an object that has multiple faces behind each-other, they'd again get the same pattern. I added the depth for this.

I realized just now that the object-id introduces a random element 😆, so I need to re-think the seeds to make the noise deterministic 🤔

almarklein:
> Is the dithering stable from frame to frame if the camera is not adjusted? 

Expanding a bit:

The random function does not really produce a random number by itself, but is really more a hash of the input seed. So you want one of the seeds to vary with the screen coords. If you only do this you get that Nintendo effect. So the camera transform is not even needed.

To prevent multiple objects that are behind each-other, from getting the same pattern, you need more seeds.

I initially used the object-id, but that breaks our image testing. I looked briefly into the concept of a scene-id (e.g. using overridable constants, but in the end, using the depth as a seed covers most cases. The only thing is that we lose the Nintendo-vibe, but I am not really sure whether its desirable anyway, since it also looks a bit like your screen is dirty.

almarklein:
This is good from my end. This can be expanded upon later, e.g. there's techniques to apply a certain filter to reduce the noise.

Korijn (approved):

Issue #557, opened 2023-06-25
Enhancing Texture Support: Color Space and Mapping Mode 


panxinmiao:
I intend to add more texture mapping support in my free time. However, before doing that, there are two issues that need to be discussed.

The first issue is one that has been mentioned before. Currently, our shaders assume that all textures of the same material are in the same color space (determined solely by the color space of the material's map property, typically sRGB), and they are uniformly converted to the physical space within the shader. However, this assumption is not always accurate. In reality, different textures within the same material may have different colorspaces. This is particularly relevant for HDR textures, as they are generally in the physical space. For example, when the material contains both the color map in the SRGB space and the HDR environment map in the physical space, we are currently unable to handle it.

My suggestion is to let each texture manage and handle its own color space, rather than handling it uniformly within the material shader. It would be convenient to utilize the built-in texture format management strategy of wgpu (xxxx-srgb).

The second issue is regarding the mapping mode of textures, which should also be managed by the textures themselves, rather than by the material. Currently, only environment map involves different mapping modes, so we added the "env_mapping_mode" attribute to the material to manage different mapping modes for environment map. However, as we expand our support for various texture functionalities, this approach becomes less desirable.

A better approach would be to add a "mapping" attribute to the Texture object, which determines the "texcoord" used when sampling the texture with the "textureSample" function in the shader (by default, it directly uses the UV attributes, but there are also some mapping modes that need to be calculated, such as the current "CUBE REFLECTION"). In light of this, I believe we need to carefully design shader templates to provide the capability of assembling shader code related to texture mapping.

panxinmiao:
> So I'm thinking about use-cases that are either scientific or special custom visualizations. But valid cases nonetheless.
> 
Is it feasible for users to directly set the "Texture.colorspace" of the corresponding resource to physical space for these scientific visualization use cases? In this way, wgpu will not automatically convert the color space of the texture, leaving it entirely up to the shader to control, which should be specific to the use case for these shader programs.

In fact, there is another issue here that may not be so obvious. For PBR rendering, sampling the texture first and then converting it to physical space may not be so accurate (although it may not be a big deal in most cases), because the interpolation is done in sRGB space during the sampling, while we expect the interpolation to be done in physical space.

> So I would be ok with introducing a Texture.colormap value that represents "use a wgpu xxxx-srgb texture for this".

Perhaps this is a viable solution.

> Would these include mapping-mode support for more of the maps, or more mapping modes to chose from?

[...]

PR #906, opened 2024-12-16
Implement colorspace tex-srgb

almarklein:
Ref #557

This, I think, solves the remaining point in #557. In my view, there are valid cases for handling srgb both by conversion in the shader, as by using an `-srgb` texture format. This implements the latter.


almarklein:
Interesting to see that the glTF-compliant method produces less pretty results (see the bands in the new validation screenshot).

Korijn:
@panxinmiao

Korijn (approved):

panxinmiao:
> Interesting to see that the glTF-compliant method produces less pretty results (see the bands in the new validation screenshot).

In fact, in this use case, the "banding" effect in the image is actually correct. This is because the color data is in the sRGB space, while np.linspace(0, 255, 6) performs interpolation in linear space. The gradient between the generated colors is not linear because the sRGB space itself is non-linear. In sRGB space, the gradient between colors is uneven.

This use case precisely demonstrates the error of interpolating directly in the sRGB space. To generate smooth gradient colors in sRGB space, the data should first be converted to linear space, then interpolated, and finally converted back to sRGB space.



```python

import numpy as np

# Generate interpolation in linear space
lin_rgb = np.linspace(0, 1, 6)  # Linear interpolation from 0 to 1
# Convert linear RGB back to sRGB space
srgb = np.clip(lin_rgb ** (1/2.2), 0, 1) 

# Scale the interpolation result back to the range of 0-255
srgb_255 = np.clip(srgb * 255, 0, 255).astype(np.uint8)
```

panxinmiao:
If the image data is in the "smooth" sRGB space (i.e., the data is non-linear), and interpolation is performed before sampling as it is currently done, I expect that it would actually result in "banding."

almarklein:
@panxinmiao I don't entirely agree 😄   But I think we do agree that both approaches have valid use-cases, and this PR tries to make it so that we have can both. Do you have any comments on the implementation or is this good to go from your end?

----

In the mean time we can discuss this band effect.

I meant to create a gradient in sRGB space, so `np.linspace()`  interpolates in the sRGB space, simply because I state that the data is sRGB. I started with 100 samples, in which case the band looks smooth in all cases. Then I used less samples and see how the results differ. Although downsampled, it's still sRGB space. It's the same effect of using a lower resolution texture map.

The shader is sampling from source data that is in sRGB space, so it makes perfect sense that the result looks better if the interpolation happens in the sRGB space.

If the image would represent "physical" data, then the "source" is in linear space, so missing values are best reconstructed by interpolation in the linear space.

panxinmiao:
I understand your point. In this use case, you aim to create an image that appears visually smooth in the sRGB space, which requires performing linear interpolation directly in the sRGB space. However, the corresponding physical color values would then be nonlinear. Following the "tex-srgb" workflow (i.e., decoding first and then performing interpolation in the physical space), banding artifacts will occur, which is expected. However, this is a special use case (working directly in the sRGB space), whereas modern rendering engines and related shader and lighting algorithms are typically designed for linear workflows and operate in linear space.

To address your requirement, we only need to set both the input and output textures to non-sRGB formats (although we currently don’t support setting the output texture format). I understand that you may consider the data to "represent" sRGB space values, but this is not critical. For the program, data is just data, and its meaning is defined by the user. Texture formats simply dictate how data is processed and do not necessarily align with its actual significance. By configuring this way, we essentially instruct the program to "process the data directly without decoding into physical space." Since no space conversion is applied, you can think of the shader algorithms as effectively operating in the sRGB space.

panxinmiao:
>  Do you have any comments on the implementation or is this good to go from your end?

I think the current approach is acceptable, at least it can solve some of the current needs. However, it is evident that this will increase the complexity of shaders and code.

almarklein:
In #910 you mention that "this issue still needs further resolution". I was under the impression that the gltf loader can simply set `colorspace="tex-srgb"` for things to work as glTF intended. What part is still missing?

panxinmiao:
> I was under the impression that the gltf loader can simply set `colorspace="tex-srgb"` for things to work as glTF intended. What part is still missing?

This is exactly what I mean.

Additionally, the logic related to texture colorspaces needs to be adjusted. My idea is to standardize the logic for all textures and ensure that each texture’s colorspace property is truly effective (rather than just the colormap’s colorspace being considered).

For example, for an env_map, typically in the sRGB space, when loading from glTF, we simply specify its colorspace as "tex-srgb" in the glTF loader. However, for environment map textures created by the user, they can choose "tex-srgb", "srgb", or even "physical" (any of these cases are possible). We need to handle these different settings. When the user specifies "srgb", the shader must be able to handle the space conversion (even though this may not be entirely correct or fully compliant with the glTF specification). When "tex-srgb" or "physical" is specified, no conversion should be performed.

The increase in shader code complexity that I mentioned earlier refers to this issue.

almarklein:
I'm not sure what the problem is. The colorspace attribute is honored in (AFAICT) all cases where the texture is interpreted as color. 

And yes, this has made the code a bit more complex, but that's the price we pay (and one that I'm happy to pay) for supporting the different use-cases related to srgb.

panxinmiao:
> I'm not sure what the problem is. 

What I meant in #910 is just to note down the tasks that are still pending:

1. Setting the correct format for each texture in the glTF loader.
2. Handling the colorspace of each texture in the shaders.

I plan to complete these tasks in a single PR. Simply noting that the issue with that test case has not been fully resolved before that PR :)

Issue #907, opened 2024-12-16
Feature request: Infinite resolution screenshot


hmaarrfk:
It comes up from time to time, to be able to render a given scene at a resolution much much higher than what is supported by webgpu. The limit here seems to be between 8k x 8x to 16k x 16k.

This can be useful in the case that one crafts a beautiful montage (lets take for example `examples/validation/validate_points_markers.py`) and we want to really showcase the full details on a mega high resolution printout (say a mural).

With the ability to design things that are "pixel perfect"  it may be desirable to really showcase the exact rendering details of the pygfx pipeline
![image](https://github.com/user-attachments/assets/1946ac69-9732-463b-a453-c3acf0c6f27c)

along with all of the imperfections:
![image](https://github.com/user-attachments/assets/03f61456-cc67-446c-9e29-e59fdb9cb752)

The idea would be that pygfx could simulate what a 100k x 100k (10Gigapixel offscreen canvas) canvas might look like, and issue the repeated rendering commands to webgpu in order to render the whole thing.

The few concerns that come to mind:
1. Maintaining frustrum culling on GPU.
2. Dealing with "pixel perfect" interpolation at the seam of tiles. While not entirely related this is reminescent of https://github.com/pygfx/pygfx/issues/905
3. The ability to define the rendering canvas dimensions
4. The ability to provide an output buffer.
5. The ability to specify if we want `rgba` or `rgb`  output buffer.

Initial ideas that came up in a conversation with almar is that we could play around with the `rect` feature of
the
```
renderer.render
```
calls and "stitch" them together.

Korijn:
I mean... Just define your scene and render it in tiles? That's how it's always done I think.

hmaarrfk:
> I mean... Just define your scene and render it in tiles? That's how it's always done I think.

Yeah as I was writing this "feature request" I started drafting a demo implementation.

I think there are a few parameters to get right which makes this tricky, and potentially worthy of an example:

1. Getting the camera parameters right. plus, minus, and the right orientation.
2. Getting the camera "orientated" correct in 3D space. (I probably should get better at internalizing quaternions)
   * I feel like this is different for orthographic cameras and perspective cameras.
3. Getting shaders that work on NDC coordinates to work corrrectly. In taking the example I showed above, I hit the fact that the background shader is written directly in NDC, so it is effectively "tiled" as you render.
[...]

PR #909, opened 2024-12-17
Support taking hirez screenshots

almarklein:
This is a stab at  #907

This adds an arg to `renderer.render()` so you can "zoom" onto a region of the screen. The zoom happens in NDC space and is applied via the camera's projection transform. This means that it works for both orthographic and perspective cameras. The logical size is adjusted appropriately so that things maintain their intended size.

The resulting hirez screenshot is always an integer factor larger than the canvas. In terms of physical pixels, they should line up perfectly. Logical pixels are float, so this solution should work regardless of the canvas resolution.

* [x] Added `tile` arg to `renderer.render()`.
* [x] Added example.
* [x] Text and line thickness etc. expressed in logical pixels keeps its proper size.
* [x] Use [setViewOffset](https://threejs.org/docs/#api/en/cameras/OrthographicCamera.setViewOffset) api?
* [x] Document how this relates to the `pixel_ratio`.
* [x] Fix text rendering aa width for non-defailt pixel ratios.
* [x] Fix that backgrounds repeat. 
* [x] Export as rgb or rgba.
* [ ] Stream tiles to disk to support really massive screenshots?  -> later
* [x] Make this work for our other cameras.


almarklein:
image

Korijn:
In three.js there is something called setViewOffset in OrthographicCamera as a convenience mechanism. https://threejs.org/docs/#api/en/cameras/OrthographicCamera.setViewOffset

Is it interesting for our use case?

almarklein:
> Is it interesting for our use case? (Unsure)

That API offers some interesting additional use-cases 🤔

almarklein:
20000x24000. Notice how the edge crosses through markers and text and it works fine:

image

examples/other/hirez_screenshot.py
```diff
@@ -0,0 +1,232 @@
...
+        gfx.TextGeometry(marker, anchor="middle-right", font_size=1),
+        gfx.TextMaterial("#000"),
+    )
+    text.local.y = -y
+    text.local.x = 0
+    scene.add(text)
+
+
+camera = gfx.OrthographicCamera()
+camera.show_object(scene, scale=0.7)
```
hmaarrfk, line 201:
```suggestion
camera.show_object(scene, scale=0.7)

camera_state = camera.get_state()

# To help visualize the rendering ability, we draw two boxes
# One entirely contained within the FOV of the main image, the other
# entirely outside of the FOV of the main image.
# By disabling AA, we should be able to use this to validate that we don't
# render more than we should.
line_thickness = 1
box_width = camera_state['width']
box_height = camera_state['height']
logical_size = canvas.get_logical_size()
if logical_size[0] > logical_size[1]:
    box_width *= logical_size[0] / logical_size[1]
else:
    box_height *= logical_size[1] / logical_size[0]
inner_lines_geometry = gfx.box_geometry(
    width=box_width - line_thickness,
    height=box_height - line_thickness,
)


inner_lines_geometry.positions.data[..., 0] += camera_state['position'][0]
inner_lines_geometry.positions.data[..., 1] += camera_state['position'][1]

outer_lines_geometry = gfx.Geometry(positions=[
    [0, 0, 0],
    [box_width + line_thickness, 0, 0],
    [box_width + line_thickness, box_height + line_thickness, 0],
    [0, box_height + line_thickness, 0],
    [0, 0, 0],
])
outer_lines_geometry.positions.data[..., 0] -= (box_width + line_thickness) / 2
outer_lines_geometry.positions.data[..., 1] -= (box_height + line_thickness) / 2

outer_lines_geometry.positions.data[..., 0] += camera_state['position'][0]
outer_lines_geometry.positions.data[..., 1] += camera_state['position'][1]
scene.add(
    gfx.Line(
        inner_lines_geometry,
        gfx.LineMaterial(
            color="red",
            thickness=line_thickness,
            thickness_space="world",
            aa=False
        ),
    )
)
scene.add(
    gfx.Line(
        outer_lines_geometry,
        gfx.LineMaterial(
            color="blue",
            thickness=line_thickness,
            thickness_space="world",
            aa=False
        ),
    )
)
```

hmaarrfk: I think you should be able to get non integer scaling (i believe in you!)

In my naive implementation this clearly helps visualize the issues:
image

almarklein:
That `camera.set_view_offset()` is pretty nice. It allows for other interesting use-cases, and makes it possible to control the logical size of the super-screenshot (i.e. scaling either pixel_ratio or logical size).

almarklein:
I also found that the aa-band of text became very large when I used `set_view_offset()` to apply a smaller logical size. Turns out the shader was basically hard-coded for a pixel_ratio of 2. I fixed this as well.

hmaarrfk:
PS. I think there are use cases for "NDC" based shaders beyond the background:

1. Vertex Shader: Creating effects where the edge of screen "pushes" against various vertex shaders warping them. https://github.com/hmaarrfk/pygfx/pull/8
2. Fragment Shader: Creating effects that mix in the NDC coordinates with the vertex `varying`  information.

Thus it would be good if whatever solution you have in mind, would enable us to continue to write these shader in a structured manner!

I've personally just "disabled" the two effects aboves when creating megalarge screenshots. but it would be great if we didn't have to ;)

Just trying to share different stories about various usecases!

almarklein:
Background now works.

> I think there are use cases for "NDC" based shaders beyond the background:

You can now do:
```wgsl
let virtual_ndc_pos = ndc_pos.xy * u_stdinfo.ndc_offset.xy + u_stdinfo.ndc_offset.zw;
```

image

almarklein:
A confirmed with a POC that writing to PNG in a streaming fashion is possible, but will get back to this another day.

Korijn:
I think in this case it would be good to add a validation example?

hmaarrfk:
This is really cool! 

You can get non-integer scaling by adjusting your loop at the end by slicing out the final border images!

I'm working to adjust my shaders to support the new offset information shared.

examples/other/hirez_screenshot.py
```diff
@@ -0,0 +1,262 @@
...
+            full_size[1],
+            ix * tile_size[0],
+            iy * tile_size[1],
+            tile_size[0],
+            tile_size[1],
+        )
+        im = np.asarray(canvas.draw())
+        # im = im[:,:,:3]  # rgba -> rgb
+        row.append(im)
+    rows.append(np.column_stack(row))
```
hmaarrfk, line 252:
```suggestion
    rows.append(row)
```

examples/other/hirez_screenshot.py
```diff
@@ -0,0 +1,262 @@
...
+
+# Safe full image
+full_im = np.vstack(rows)
```
hmaarrfk, line 255:
```suggestion
full_im = np.block(rows)
```

For very very large images (which is the subject for in this PR), your are incuring 2 copies of all the renders"

1. for the column stack
2. The second for the vstack

By using modern (last 6 years) call to `np.block` it will do a single copy.
```
a = np.full((5, 4), fill_value=0)
b = np.full((5, 4), fill_value=1)
c = np.full((2, 4), fill_value=2)
d = np.full((2, 4), fill_value=3)

np.block([[a, b], [c, d]])
```

outputs
```
array([[0, 0, 0, 0, 1, 1, 1, 1],
       [0, 0, 0, 0, 1, 1, 1, 1],
       [0, 0, 0, 0, 1, 1, 1, 1],
       [0, 0, 0, 0, 1, 1, 1, 1],
       [0, 0, 0, 0, 1, 1, 1, 1],
       [2, 2, 2, 2, 3, 3, 3, 3],
       [2, 2, 2, 2, 3, 3, 3, 3]])
```
https://github.com/numpy/numpy/pull/11971

hmaarrfk: also ,the example i have shows that it will even support non integer upscaling since the last "rows /columns" can be of different size, so long as they create this block-wise array.

almarklein: Nice, I did not know about `np.block` yet!

almarklein: FYI `np.block` concatenates over the last dimensions, so I had to nest the tiles in a list to make it work as intended.

Korijn:
Is this ready for review and merge?

almarklein:
Yep!

examples/other/hirez_screenshot.py
```diff
@@ -0,0 +1,262 @@
...
+# The scale factor determines how many tiles are created to build the
+# final screenshot (upscale_factor**2). Using a higher value allows
+# creating screenshots above the wgpu texture size limit.
+upscale_factor = 4
+
+# Determine the canvas size, and thereby the resolution of each tile.
+# The tile_size and full_size that we calculate below are both in logical pixels.
+canvas_size = 1200, 1000
+
+if True:
```
Korijn, line 31: Could you put this bool in a variable so there's a name for the flag? It documents the logic a bit more clearly.

examples/screenshots/validate_view_offset.png
Korijn, line 1: The screenshot looks a bit weird, I would expect the grey lines to be "overlaid" over the image, but actually they are really "in between" the four quadrants of the image. Is that intentional?

almarklein: Yes, that's intentional, the quadrants are "pulled apart" a bit, to make it clear that they are separately rendered.

examples/other/hirez_screenshot.py
almarklein: done

Korijn (approved):

PR #914, opened 2024-12-25
Improve Transform performance part 2

Korijn:
Follow up to #911

Changes:
* Remove performance workarounds in animation mixer and text object
* Bumpy pylinalg to v0.5 to pull in `mat_compose` performance improvement https://github.com/pygfx/pylinalg/pull/92
  * This already yields an approximate 15-20fps improvement for the `skinning_animation.py` example on my machine

Plans:
* Refactor transform to be based on position/scale/rotation and recompute matrix lazily


Korijn:
Looks like the updated pylinalg mat_compose does not support scalar values for the scaling argument. I'll have to fix that upstream and will make sure to add a test case.

PR #916, opened 2024-12-29
Base AffineTransform state on position/scale/rotation instead of matrix

Korijn:
Changes:

* Base AffineTransform state on position/scale/rotation instead of matrix
* All ndarrays are now read only
* Reuse all local state ndarrays instead of constantly replacing them
* Remove performance workarounds to allow for performance testing

This yields really impressive performance increases (roughly 300-400%) for examples where transforms are updated every frame.
And we still have all our world space transform getters and setters with lazy evaluation & caching.
No need to call "update()" functions manually in pygfx (like in Three.js).
Best of both worlds! I'm very pleased with this outcome.

Future work (separate PR):
* I could add an option to AffineTransform to toggle between the matrix or position/scale/rotation as the source of truth. There are scenarios where that is preferred. See #920

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -4,7 +4,7 @@ from time import perf_counter_ns
 import weakref
 import functools
 
-from typing import Tuple, Union
+from typing import Tuple
 
 
 PRECISION_EPSILON = 1e-7
@@ -141,13 +141,19 @@ class AffineBase:
 
     last_modified: int
 
-    def __init__(self, *, reference_up=(0, 1, 0), is_camera_space=False):
+    def __init__(self, /, *, reference_up=(0, 1, 0), is_camera_space=False):
         self.update_callbacks = {}
+
         self.is_camera_space = int(is_camera_space)
-        self._scaling_signs = np.asarray([1, 1, 1], dtype=float)
 
         self._reference_up_provider = None
-        self._reference_up = la.vec_normalize(reference_up)
+        self._reference_up = la.vec_normalize(reference_up, dtype=float)
+        self._reference_up_view = self._reference_up.view()
+        self._reference_up_view.flags.writeable = False
+
+        self._scaling_signs = np.asarray([1, 1, 1], dtype=float)
+        self._scaling_signs_view = self._scaling_signs.view()
+        self._scaling_signs_view.flags.writeable = False
 
     def _set_reference_up_provider(self, provider: "RecursiveTransform"):
         self._reference_up_provider = provider
@@ -157,14 +163,14 @@ class AffineBase:
         """The zero-tilt reference vector used for the direction setters."""
         if self._reference_up_provider:
             return self._reference_up_provider._own_reference_up
-        return self._reference_up
+        return self._reference_up_view
 
     @reference_up.setter
     def reference_up(self, value):
         if self._reference_up_provider:
             self._reference_up_provider._own_reference_up = value
         else:
-            self._reference_up = la.vec_normalize(value)
+            self._reference_up[:] = la.vec_normalize(value)
 
     @property
     def matrix(self):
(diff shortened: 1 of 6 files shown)


Korijn:
@panxinmiao with these changes I am getting ~130 FPS on my machine for the Skinned Mesh example:

image

Is this "good enough" in terms of performance? :)

pygfx/controllers/_orbit.py
```diff
@@ -97,7 +97,8 @@ def _update_rotate(self, delta):
         elif new_elevation > bounds[1]:
             delta_elevation = bounds[1] - elevation
 
-        r_azimuth = la.quat_from_axis_angle(up, -delta_azimuth)
+        # up.copy() is a temporary workaround for https://github.com/pygfx/pylinalg/issues/95
+        r_azimuth = la.quat_from_axis_angle(up.copy(), -delta_azimuth)
```
Korijn, line 101: I will create a new pylinalg release to fix this issue

pygfx/utils/transform.py
```diff
@@ -558,61 +563,132 @@ class AffineTransform(AffineBase):
...
-        return view_array
+        return self._matrix
 
     @matrix.setter
     def matrix(self, value):
-        self.untracked_matrix[:] = value
+        position, rotation, scale = la.mat_decompose(value)
+        self._position[:] = position
+        self._rotation[:] = rotation
+        self._scale[:] = scale
```
Korijn, line 680: Set scale signs

pygfx/controllers/_orbit.py
Korijn: Done, see https://github.com/pygfx/pylinalg/pull/96

pygfx/utils/transform.py
Korijn: Done

Korijn:
I see I've managed to break something related to axis flips:

image

Korijn:
This is ready to merge from my end. So happy to reach this state, we now really have the public API that I initially envisioned when we started working on the new Transform API.

pygfx/utils/transform.py
```diff
@@ -558,61 +549,115 @@ class AffineTransform(AffineBase):
...
+        self._scale_view = self._scale.view()
+        self._scale_view.flags.writeable = False
 
     def flag_update(self):
         self.last_modified = perf_counter_ns()
         super().flag_update()
 
+    @property
+    def position(self) -> np.ndarray:
+        return self._position_view
```
almarklein, line 580: Since we re-implement these (`position`, `rotation`, `scane`) here, would it make sense to move their implementation in `AffineBase` (the one doing `return self._decomposed[0]`) to `RecursiveTransform`?

almarklein (commented):
Super nice work!

pygfx/utils/transform.py
Korijn: I have thought about that and actually went back and forth a few times on that thought. But I think the elegance of the architecture is kept intact this way. The idea of a generalized base transform that can derive anything off a matrix is not valid anymore if I go that route.

Also, it would make the follow-up PR much harder to implement. :) See #920

pygfx/utils/transform.py
```diff
@@ -690,10 +735,7 @@ def __init__(
         self.own = None
         self.last_modified = perf_counter_ns()
 
-        if isinstance(matrix, AffineBase):
-            self.own = matrix
-        else:
-            self.own = AffineTransform(matrix, is_camera_space=is_camera_space)
+        self.own = base
```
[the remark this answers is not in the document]
Korijn: I guess I didn't really respond to your concern. There is only one place where transforms are instantiated in the codebase, and it's not really a public API element. So I don't think it's that important to type check. I also want to keep this code lean for performance.

pygfx/utils/transform.py
almarklein: Makes sense!

PR #922, opened 2025-01-03
Add transform validation example

almarklein:


examples/validation/validate_transform.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def create__boxes(ttype1, value1, ttype2, value2, ttype3, value3, *, state_basis):
+    box_a = gfx.Mesh(geometry, material_a)
+    box_b = gfx.Group()
+    box_c = gfx.Mesh(geometry, material_b)
+
+    for box in (box_a, box_b, box_c):
+        box.local.state_basis = state_basis
+        box.world.state_basis = state_basis
```
almarklein, line 33: This has not effect now, but it should cover the alternative code paths of  #920

Korijn: RecursiveTransform (box.world) has no state_basis property though

almarklein: Good point. Fixed.

almarklein:
For the record, this example will trigger both `state_basis` options when #920 is included.

Korijn:
I've included this in #920

PR #924, opened 2025-01-06
Improve TextureMap docs

### LINKED ISSUE #913: kwargs for `wrap` in `TextureMap`?
Hi! Catching up with #889 , and I saw [this example](https://docs.pygfx.org/latest/_gallery/validation/validate_image_colormap.html) and wondering about the various options for `warp` which seem to correspond to the wgpu [`AddressMode` enums](https://wgpu.rs/doc/wgpu/enum.AddressMode.html) . I don't see `"clamp"`, is that an alias for `"clamp-to-border"`? Both seem to produce the same effect if I slightly modify the example:

For scientific visualizations using colormaps for images I'm assuming that `clamp-to-border` is usually what people want? 

```python
import numpy as np
import pygfx as gfx


renderer = gfx.renderers.WgpuRenderer(WgpuCanvas(size=(600, 600)))

im = np.repeat(np.linspace(0, 2, 100).reshape(1, -1), 48, 0).astype(np.float32)
colormap = gfx.Texture(np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]], np.float32), dim=1)

image = gfx.Image(
    gfx.Geometry(grid=gfx.Texture(im, dim=2)),
    gfx.ImageBasicMaterial(
        clim=(0, 1),
        map=gfx.TextureMap(
            colormap,
            filter="linear",
            wrap="clamp",
        ),
    ),
)
image.local.y = 52

scene = gfx.Scene()
scene.add(image)

camera.show_rect(-4, 103, -4, 103)

renderer.request_draw(lambda: renderer.render(scene, camera))


if __name__ == "__main__":
    run()
```

image

Thanks!

almarklein:
Mention possible values for `filter` and `wrap` in docstring.


almarklein:
cc @panxinmiao

almarklein:
Oh sorry, I had not seen #918 yet 😆  @panxinmiao I assume "repeat" is a more conventional default for meshes?

panxinmiao:
> I assume "repeat" is a more conventional default for meshes?

Yes, the default mode is more often set to 'repeat'. The glTF specification also stipulates that when the wrap mode of a texture is not specified, 'repeat' is used by default.

almarklein:
I reverted all changes except the improvements to the docs. Ready.

Korijn (approved):

PR #933, opened 2025-01-14
Fix the issue where stopping an AnimationAction did not restore the target to its initial state.

### LINKED ISSUE #932: Deactivating AnimationAction Does Not Reset the Target to Its Initial State
I noticed that after PR #916 , stopping an animation no longer correctly resets the target to its initial state.

Before PR #916 :
Video_2025-01-13_173347

After PR #916:
Video_2025-01-13_172932

It seems that in certain cases, after modifying position, rotation, or scaling, the matrix is not properly updated. 
I’ve tried tracing the code but haven’t pinpointed the exact cause yet.

Additional clarification:

- `AnimationAction.pause()` pauses the timeline of the current AnimationAction, but the AnimationAction will continue to affect the target.

- `AnimationAction.stop()` stops the animation entirely. The AnimationAction should no longer influence the target, and if there are no other active AnimationActions, the target should revert to its initial state.

panxinmiao:
close #932.

### DIFF
--- a/pygfx/animation/animation_mixer.py
+++ b/pygfx/animation/animation_mixer.py
@@ -55,7 +55,7 @@ class AnimationMixer(RootEventHandler):
         for t in action._clip.tracks:
             key = (t.target, t.path)
             if key not in self.__property_ori_cache:
-                ori_value = self._get_path_value(t.target, t.path)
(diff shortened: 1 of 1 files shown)


Korijn (approved):

almarklein:
Thanks for the quick fix!

PR #939, opened 2025-01-15
Associate pipeline-container also with Material

### LINKED ISSUE #937: Proposal: associate pipeline-container with material
## Intro

In #912 the pipeline-containers (the objects containing the low-level wgpu objects to draw a wobject) are associated with both the wobject and the renderstate (using the new `WeakAssociativeContainer`). There is a pipeline-container for every drawn combination of  wobject and renderstate. And when either the wobject or renderstate object is deleted, so is the pipeline-container and all the wgpu objects stored on it.

The behaviour prior to #912  is very similar, but implemented differently. 

## Proposal

My proposal is to *also* associate the pipeline-container with the material (the `WeakAssociativeContainer` can do that). This would mean that there would be a pipeline-container for every combination of wobject, material, and renderstate. And that it is deleted when either of these is.

## Consequences

In the current situation, when changing a material, the pipeline container is updated, and changed as little as possible. If the material is compatible (i.e. results in the same wgsl after templating), the change is very efficient. 

In the new situation, changing a material results in a whole new pipeline-container object. However, since we cache shader modules for re-use, creating the new pipeline-container will not require a wgsl recompile.

The major advantage is that users can hold on to a material and use it later, which will then basically be free, because the pipeline-container can be used unchanged. It gives users some control over the lifetime of the wgpu objects in a pretty intuitive way.

## Use cases

Let's consider two use-cases.

Use case 1, swap out material for a new one:
```py
def a function_called_on_eg_mouse_hover():
    wobject.material = NewMaterial(...)
```

Use case 2, switch between two existing materials:
```py
material1 = AMaterial()
material2 = AnotherMaterial()
def a function_called_on_eg_mouse_hover():
    wobject.material = material2 if wobject.material is material1 else material1
```

|  | usecase 1, compatible material | usecase 1, incompatible material | usecase 2, compatible material | usecase 2, incompatible material |
| -- | -- | -- | -- | -- |
| current situation | update bindings | shader recompile | update bindings | shader recompile |
| with proposal | new pipeline-container, but re-use shader | new pipeline container, new shader compile | nothing | nothing |

almarklein:
Closes #937 

The idea:

* Add `material`  to the key of `PIPELINE_CONTAINER_GROUPS`.
* Use a tracker per pipeline-container-group instead of per wobject.

Todo:

* [x] Basic implementation seems to be working: no pipeline updates at all when switching to previously used materials.
* [x] Update tests.
* [x] ~I don't think `wobject.geometry` needs to be stored on `_store` either.~  yes it does :)
* [x] Refactor trackable a bit, renaming `root` to `tracker`. No logical changes.


examples/feature_demo/multi_select.py
```diff
@@ -92,7 +96,19 @@ def random_rotation():
     )
 
 
+initialized = False
+
+
 def animate():
+    global initialized
+
+    # Render with both materials, to initialize them.
```
Korijn, line 105: Would be awesome if there was some way to do thus without rendering a frame... Like some games have a initial loading screen where all future shaders and pipelines are precompiled.

almarklein: I made a quick start, but this quickly becomes bigger than fits the scope of this PR. I offloaded the idea into an issue: https://github.com/pygfx/pygfx/issues/941

Korijn (approved):

PR #942, opened 2025-01-17
Fix ior & specular not work correctly

panxinmiao:
Ior and specular did not take effect correctly in MeshPhysicalMaterial:
image

Fixed:
image


panxinmiao:
Addition:
To view the “SpecularTest” case with "gltf_viewer.py", #940 is required, as it implements `GLTFMaterialsIorExtension` and `GLTFMaterialsSpecularExtension` in the glTF loader, which loads the corresponding data from the glTF model.

Korijn (approved):

PR #943, opened 2025-01-18
Add `package_name` as an argument for `load_wgsl`

m-agour:
Hello pygfx team,

We wanted to load a custom wgsl shader file but since `load_wgsl` function has `package_name` directing to pygfx and we want to use as many pygfx functions as we can, can `package_name` be an argument so that we are able to use it directly?

Thanks

### DIFF
--- a/pygfx/renderers/wgpu/wgsl/__init__.py
+++ b/pygfx/renderers/wgpu/wgsl/__init__.py
@@ -8,10 +8,9 @@ import importlib.resources
 
 
 @functools.lru_cache(maxsize=None)
-def load_wgsl(shader_name):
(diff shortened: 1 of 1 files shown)


almarklein (approved):
Sounds like a good idea!

Have you also seen `register_wgsl_loader()`  in `templating.py`?

Korijn (approved):

m-agour:
> Sounds like a good idea!
> 
> Have you also seen `register_wgsl_loader()` in `templating.py`?

I haven’t seen it yet, but I’ll look into it. Thanks for pointing it out! Also thanks for the quick merge!

PR #944, opened 2025-01-19
Optimize last_modified tracking

Korijn:
With these micro optimizations (based on cprofile reports) I was able to boost the FPS on the skinning animation example from 145 to 170 on my machine.

Changes:

* Simplified world object transform buffer update tracking to a boolean instead of a WeakKeyDictionary
* Minimized function call overhead for flag_update() since it is being called ridiculously often (it's the main bottleneck in this example)

To get this example to run even faster, the following problems would need to be tackled, as these are the bottlenecks:

* Make flag_update() much cheaper to do or avoid the need for it completely somehow
* Make Stats.__exit__() much more efficient (right now it recomputes all text rendering info on every frame, it should just update maybe a few characters/TextItems instead)
* Make Interpolation of transform attributes more efficient (perhaps precompute?)

I'm not going to pursue any of these challenges right now

### DIFF
--- a/pygfx/renderers/wgpu/engine/renderer.py
+++ b/pygfx/renderers/wgpu/engine/renderer.py
@@ -489,9 +489,9 @@ class WgpuRenderer(RootEventHandler, Renderer):
 
         # Update transform uniform buffers
         for wobject in wobject_list:
-            transform = WorldObject.transform_updates.pop(wobject, None)
-            if transform:
-                wobject._update_uniform_buffers(transform)
+            if wobject._uniform_buffers_dirty:
+                wobject._update_uniform_buffers()
+                wobject._uniform_buffers_dirty = False
 
         # Prepare the shared object
         self._shared.pre_render_hook()
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -132,8 +132,6 @@ class WorldObject(EventTarget, Trackable):
         id="i4",
     )
 
-    transform_updates = weakref.WeakKeyDictionary()
-
     def __init__(
         self,
         geometry=None,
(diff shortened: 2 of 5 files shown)


panxinmiao:
I observed that in the cprofile reports too, the consumption of flag_update() is disproportionately high. This is a method call that occurs very frequently, and it would be great if there were a way to completely eliminate this overhead. This is also the primary additional performance cost compared to the temporary solution in #746.

> Make Interpolation of transform attributes more efficient (perhaps precompute?)

Precomputing the interpolation of transforms doesn't seem feasible because we only know the interpolation time points at runtime, making precomputation impossible. The main time cost in interpolation comes from retrieving the interpolation interval in the animation track data, which has already been heavily optimized in #756.

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -489,9 +489,9 @@ def render(
 
         # Update transform uniform buffers
         for wobject in wobject_list:
-            transform = WorldObject.transform_updates.pop(wobject, None)
-            if transform:
-                wobject._update_uniform_buffers(transform)
+            if wobject._uniform_buffers_dirty:
+                wobject._update_uniform_buffers()
+                wobject._uniform_buffers_dirty = False
```
almarklein, line 494: What would the performance be if `transform_updates` would be a `WeakSet` of world objects, and this'd be:
```py
for wobject in WorldObject.transform_updates:
    wobject._update_uniform_buffers()
```

A disadvantage is that it updates all dirty transforms, not only of the wobjects that are going to be rendered just now. But an advantage is that you don't have to iterate over all wobjects.

Korijn: Well, that's almost exactly what it was! This is faster.

Korijn:
I am going to revise this later

pygfx/renderers/wgpu/engine/renderer.py
almarklein: I was triggered by this because I'm thinking how to improve performance for may objects, renderbundles being one solution. But we can revisit this when the time comes. There may be better ways to handle this "caching". So let's just stick with what you have now.

Korijn: Yeah, I understand, but this loop is not even visible in the profiling data. So it's just not worth improving further right now. flag_update is like 40% of the total runtime.

PR #948, opened 2025-01-22
Make `bind_matrix` a read-only property, and clean up `np.linalg.inv`.

panxinmiao:
See title.

### DIFF
--- a/pygfx/objects/_skins.py
+++ b/pygfx/objects/_skins.py
@@ -77,18 +77,16 @@ class Skeleton:
         """Generate the bone_inverses array if not provided in the constructor."""
         self.bone_inverses.clear()
         for bone in self.bones:
-            self.bone_inverses.append(np.linalg.inv(bone.world.matrix))
+            self.bone_inverses.append(bone.world.inverse_matrix)
 
     def pose(self):
         """Reset the skeleton to the binding-time pose."""
         for i, bone in enumerate(self.bones):
-            bone.world.matrix = np.linalg.inv(self.bone_inverses[i])
+            bone.world.matrix = la.mat_inverse(self.bone_inverses[i])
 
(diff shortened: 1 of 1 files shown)


Korijn (approved):

Korijn:
Now that #950 was merged, can you rebase this? Then we can review & merge.

pygfx/objects/_skins.py
```diff
@@ -77,18 +77,16 @@ def calculate_inverses(self):
...
         for i, bone in enumerate(self.bones):
-            bone.world.matrix = np.linalg.inv(self.bone_inverses[i])
+            bone.world.matrix = la.mat_inverse(self.bone_inverses[i])
 
         for bone in self.bones:
             if bone.parent and isinstance(bone.parent, Bone):
-                bone.local.matrix = (
-                    np.linalg.inv(bone.parent.world.matrix) @ bone.world.matrix
-                )
+                bone.local.matrix = bone.parent.world.inverse_matrix @ bone.world.matrix
```
Korijn, line 89: Is it perhaps a good idea to set state_basis to matrix for this bone object? Not sure which attribute is updated more often?

panxinmiao: For animation/interpolation, using position/scale/rotation is undoubtedly more efficient. 

The pose() method restores the skeletal pose at the time of binding and serves as a low-frequency API for user interaction. Since bone_inverses already stores the world matrices of the bones at the time of binding, the "local.matrix" at the bingding time can be directly derived from this. There is no need to separately cache or store the bound ”local.matrix“, ”position“, ”scale“, or ”rotation“ data.

almarklein (approved):

Korijn (approved):

Issue #947, opened 2025-01-21
Completely disabling Anti Aliasing (AA)???


hmaarrfk:
I was moving over some code of mine to PyGFX (from numpy to PyGFX) and i spent some time realizing that something was still enabled in terms of AA within my pygfx pipeline.

The example tries to show an image within the PYGFX pipeline, then tries to save the screenshot. Pretty simple right????

The problem is that all the edges seem to be washed out.

It can be "obvious" if you uncomment the lines that make the image a red square.

Even in the screenshot I took as an example below, you can clearly tell that the PyGFX generated stuff is washed out.

```python
import numpy as np
import imageio.v3 as iio
from wgpu.gui.offscreen import WgpuManualOffscreenCanvas as OffscreenCanvas
import pygfx as gfx

astronaut = iio.imread("imageio:astronaut.png")
# Don't make the aspect ratio square so we can
# really test things
astronaut = astronaut[:200]

# A basic test case???
# astronaut[...] = 0
# astronaut[50:150, 128:384, 0] = 200

scene = gfx.Scene()
image = gfx.Image(
    gfx.Geometry(grid=gfx.Texture(astronaut, dim=2)),
    gfx.ImageBasicMaterial(clim=(0, 255)),
)
# Off by half???
image.local.position = (0.5, -0.5, 0)
# image.local.position = astronaut.shape[1] / 2, - astronaut.shape[0] / 2, 0
image.local.scale = (1, -1, 1)
scene.add(image)

canvas = OffscreenCanvas(
    # size in xy
    size=(astronaut.shape[1], astronaut.shape[0]),
    # Try to disable AA
    pixel_ratio=1,
)
renderer = gfx.renderers.WgpuRenderer(canvas)

camera = gfx.OrthographicCamera(astronaut.shape[1], astronaut.shape[0])
camera.set_state({
    'position': [astronaut.shape[1] / 2, -astronaut.shape[0] / 2, 100],
    'width': astronaut.shape[1],
    'height': astronaut.shape[0],
})

canvas.request_draw(lambda: renderer.render(scene, camera))
img = np.asarray(canvas.draw())[..., :3]  # ignore alpha

difference = np.abs(img.astype('float32') - astronaut.astype('float32')).astype('uint8')
iio.imwrite('screenshot_expected.png', astronaut)
iio.imwrite('screenshot_result.png', img)
iio.imwrite('screenshot_difference.png', difference)

from numpy.testing import assert_array_equal

[...]

Issue #929, opened 2025-01-10
Confusion about size_space


FlynnOConnell:
I work with @kushalkolar and I'm trying to implement size_space as a tool to scale scatter-points/lines on an image when users zoom in or out. I looked at the [Points drawn with different size_space](https://docs.pygfx.org/stable/_gallery/validation/validate_points_size_space.html) example but I'm still confused. Especially about what "model" space is. Thanks :)

FlynnOConnell:
Got it. That does indeed clarify things, thank you!

PR #949, opened 2025-01-22
Tweak some docstrings

almarklein:
* The `renderer.pixel_ratio` and `.pixel_filter` ref #947
* The `Coordspace` enum, ref #929


Korijn:
Maybe it's worthwhile to explicitly have a short section in the guide on "Anti-aliasing" that describes which options (across the public API) are related, and how to completely shut it off.

Korijn (approved):

PR #950, opened 2025-01-22
Optimize last_modified tracking

Korijn:
🚀 Another massive performance boost! The skinning animation example runs at 180 fps on my machine now.

* Callback mechanism and weakrefs are completely removed from the Transform API, of course this implies refactoring in a few areas, notably lighting, but actually it is a big improvement because now light buffers are only updated just before rendering, instead of on every transform update
  * Note: I didn't look much further into the lights than needed to make the tests pass
  * Note 2: ⚠️ This is technically a breaking change since `on_update` is removed from the Transform API
* `last_modified` has its own specialized implementation in `AffineTransform`, `RecursiveTransform` and `Camera`, to be as simple as it can be in each case
* `flag_update` no longer immediately propagates, instead propagation happens lazily when a cache attribute is accessed in `RecursiveTransform.last_modified` and `Camera.last_modified` (and **not** in `AffineTransform`)
* The `cache` decorator only accesses `last_modified` once instead of three times
* WorldObjects have a `_world_last_modified` flag which is used to track if the buffers need an update when a frame is rendered

⏭️ **Next up:** I am working on another improvement to eliminate redundant propagation of last_modified, but I will do that in a separate pull request.

🥳 **Cool detail:** Some pretty cool news, with this PR, finally, the Transform API is no longer at the top of the cProfile report!

image

> ncalls: Total number of calls to the function. If there are two numbers, that means the function recursed and the first is the total number of calls and the second is the number of primitive (non-recursive) calls.

So `RecursiveTransform.last_modified` is called ~460k times in one full animation run, and 4m times if you include recursive calls! 😱 It's definitely worthwhile to keep optimizing this, I think 😁 it may be pygfx' hottest codepath!

### DIFF
--- a/pygfx/cameras/_base.py
+++ b/pygfx/cameras/_base.py
@@ -4,7 +4,7 @@ import numpy as np
 import pylinalg as la
 
 from ..objects._base import WorldObject
-from ..utils.transform import cached, callback
+from ..utils.transform import cached
 
 
 class Camera(WorldObject):
@@ -27,16 +27,17 @@ class Camera(WorldObject):
 
     def __init__(self):
         super().__init__()
-        self.last_modified = perf_counter_ns()
+        self._last_modified = perf_counter_ns()
 
         self._view_size = 1.0, 1.0
         self._view_offset = None
 
-        self.world.on_update(self.flag_update)
+    def flag_update(self):
+        self._last_modified = perf_counter_ns()
 
-    @callback
-    def flag_update(self, *args, **kwargs):
-        self.last_modified = perf_counter_ns()
+    @property
+    def last_modified(self) -> int:
+        return max(self._last_modified, self.world.last_modified)
 
     def set_view_size(self, width, height):
         """Sets the logical size of the target. Set by the renderer; you should typically not use this."""
--- a/pygfx/helpers/_lights.py
+++ b/pygfx/helpers/_lights.py
@@ -3,7 +3,6 @@ import math
 import numpy as np
 
 from ..objects import Light
-from ..utils.transform import AffineBase, callback
 
 from .. import (
     sphere_geometry,
@@ -42,10 +41,9 @@ class PointLightHelper(Mesh):
         material = MeshBasicMaterial(color="#fff")
         super().__init__(geometry, material)
 
-        self.world.on_update(self._update)
+    def _update_uniform_buffers(self):
+        super()._update_uniform_buffers()
 
-    @callback
-    def _update(self, transform: AffineBase) -> None:
         if self._color is None and isinstance(self.parent, Light):
             color = self.parent.color
             if color != self.material.color:
@@ -87,8 +85,6 @@ class DirectionalLightHelper(Line):
         self.ray_length = ray_length
         self.show_shadow_extent = show_shadow_extent
 
-        self.world.on_update(self._update)
-
     @property
     def ray_length(self):
         """The length of the arrows indicating light rays."""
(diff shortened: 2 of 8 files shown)


Korijn:
Ready to merge on my end!

almarklein (approved):
Epic work 😄 🚀

panxinmiao:
Great! 🚀 

I am a firm believer in the approach of "traversing the scene graph before rendering and actively updating the scene's Transform Matrix."  😄

Korijn:
> Great! 🚀
> 
> I am a firm believer in the approach of "traversing the scene graph before rendering and actively updating the scene's Transform Matrix." 😄

I agree, I think that's where we are now 😁

panxinmiao:
> ⏭️ **Next up:** I am working on another improvement to eliminate redundant propagation of last_modified, but I will do that in a separate pull request.

Nice to hear that, 😄 

I have always had an idea:

During each frame's render (in the renderer's render() method), when we traverse the scene graph to update the world matrices of nodes, is there a way to completely bypass the finer-grained propagation mechanism of RecursiveTransform based on flag_update()?

In this case, we only need to simply start from the root node (the scene object) and recursively update the world matrix for each child node.

You might think that for static scenes, updating the world matrix here is redundant. However, for nearly any dynamic scene, this approach is undoubtedly faster than the one that requires tracking the update propagation state of the world transform.

This method incurs almost no additional overhead—it's just matrix multiplication. Furthermore, since the update starts from the root node and proceeds downward, the number of matrix multiplications is fixed. For each node in the scene, matrix multiplication occurs only once, and 4x4 matrix multiplication is extremely fast.

I even suspect that the cost of performing a single 4x4 matrix multiplication might be smaller than the extra performance overhead caused by RecursiveTransform automatically tracking "whether the world matrix needs to be updated." If that’s the case, even for static scenes, where the world matrix doesn’t need updating, the act of "checking whether the world matrix needs to be updated" might itself introduce enough performance overhead to counterbalance the cost of matrix multiplication.

I will try to do some testing and verification when I have time.

panxinmiao:
> You might think that for static scenes, updating the world matrix here is redundant (Unnecessary 4x4 matrix multiplication). However, for nearly any dynamic scene, this approach is undoubtedly faster than the one that requires tracking the update propagation state of the world transform.

In addition, if the user determines that it is a static scene (or if the user wants to fully control the update timing of the Transform matrix), we can also provide an option to disable automatic updates of the world matrix.

Korijn:
I respectfully disagree. In this PR, the overhead of flag_update has been eliminated entirely. Flag_update **does not propagate anymore**.

It's now working exactly as you describe, only when a frame is rendered are world matrices computed by traversing the scene graph.

And I have one more PR to go to make it even more efficient.

panxinmiao:
> It's now working exactly as you describe, only when a frame is rendered are world matrices computed by traversing the scene graph.
> 

Yes, I’m aware of that. 

However, right now, when traversing the scene graph before rendering, we first need to determine whether the world matrix of each object needs to be updated, and only update those that actually need it. The problem is, checking whether the world matrix needs updating isn’t as simple or lightweight as it seems (because you have to check if the transformations of all its ancestor nodes have changed). I suspect that the cost of this check might not be smaller than the overhead of simply performing a single 4x4 matrix multiplication.

Korijn:
I see. Well, I think that you have predicted my next move exactly, so I'll ask you to wait for the next PR to come.

I will also measure the difference with your proposal!

almarklein:
Maybe I'm stating the obvious here, but IIUC the plan is that during the scene-graph traversal, you check the flag (without any propagation) and when it is dirty, you update the matrix, and from that of all its children (and their children etc). So you kind of have the best of both worlds.

Korijn:
> Maybe I'm stating the obvious here, but IIUC the plan is that during the scene-graph traversal, you check the flag (without any propagation) and when it is dirty, you update the matrix, and from that of all its children (and their children etc). So you kind of have the best of both worlds.

That's the plan.

PR #953, opened 2025-01-23
Refactor to have a single scene traverse in the renderer

almarklein:
Refactor the renderer a bit so it has an explicit place where the scene is traversed. The idea is that this is done exactly once for each draw, and that different kinds of info is collected during the traversal, as to minimize looping over large scene graphs.

In the current implementation the lights are also collected in this pass. 

Relates to #950 too.


almarklein:
For context, Korijn and I discussed caching in general and for the transforms in specific, and we concluded that having a single traversal (instead of more than one) will help performance, and provides a place where we can check for different update/caching stuff.

Korijn (approved):

PR #890, opened 2024-11-22
Lint Ruff 0.8.0 and bump imgui-bundle version to match wgpu-py

claydugo:
Blocked CI in https://github.com/pygfx/pygfx/pull/876

PR #956, opened 2025-01-26
add python min to imgui pin

tlambert03:
This pins `imgui-bundle>=1.6.0; python_version >= '3.10'` (adding the min python version) in pyproject.

`uv sync` is a pretty great way to simply get up and running, and if you try to run `uv sync --extra dev` here, you get an error:

```
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (python_full_version == '3.9.*'):
  ╰─▶ Because the requested Python version (>=3.9) does not satisfy Python>=3.10 and imgui-bundle>=1.6.0 depends on Python>=3.10, we can conclude
      that imgui-bundle>=1.6.0 cannot be used.
      And because only the following versions of imgui-bundle are available:
          imgui-bundle<=1.6.0
          imgui-bundle==1.6.1
          imgui-bundle==1.6.2
      we can conclude that imgui-bundle>=1.6.0 cannot be used.
      And because pygfx[dev] depends on imgui-bundle>=1.6.0 and your project requires pygfx[dev], we can conclude that your project's requirements
      are unsatisfiable.

      hint: The `requires-python` value (>=3.9) includes Python versions that are not supported by your dependencies (e.g., imgui-bundle>=1.6.0
      only supports >=3.10). Consider using a more restrictive `requires-python` value (like >=3.10).
```

It does actually make sense... imgui-bundle itself would never work in 3.9, so might as well? 🤷‍♂️

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ examples = [
     "scikit-image",
     "trimesh <4.6",
     "gltflib",
-    "imgui-bundle>=1.6.0; python_version >= '3.11'",
+    "imgui-bundle>=1.6.0; python_version >= '3.10'",
 ]
 docs = [
     "sphinx>7.2",
@@ -48,7 +48,7 @@ docs = [
     "scikit-image",
     "trimesh <4.6",
     "gltflib",
-    "imgui-bundle>=1.6.0; python_version >= '3.11'",
+    "imgui-bundle>=1.6.0; python_version >= '3.10'",
 ]
 tests = ["pytest", "psutil", "trimesh <4.6", "httpx", "gltflib", "imageio"]
 dev = ["pygfx[lint,tests,examples,docs]"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ examples = [
     "scikit-image",
     "trimesh <4.6",
     "gltflib",
-    "imgui-bundle>=1.6.0",
+    "imgui-bundle>=1.6.0; python_version >= '3.11'",
 ]
 docs = [
     "sphinx>7.2",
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Your setting yourself up for failure if you start to install all of every packages example dependencies. 

But honestly, why not revert https://github.com/pygfx/pygfx/pull/890#discussion_r1899023260

I fixed the problem in wgpu to ensure that everything was good even with the older version

tlambert03:
> Your setting yourself up for failure 

i'm plenty capable of dealing with issues as they arise, thanks.  environments are easily disposable and i'm well aware that i can always install the bare minimum should anything be strange.    In any case, it's just `pip install -e .[dev]`, which is pretty standard practice no?

> But honestly, why not revert https://github.com/pygfx/pygfx/pull/890#discussion_r1899023260

that's fine too.  i just opened this to let you know you had an unsatisfiable dep.  do with it what you like.  including close

ivoflipse:
Perhaps the examples should leverage PEP 723 so you an easily run them using `uv`
https://docs.astral.sh/uv/guides/scripts/

hmaarrfk:
> i'm plenty capable of dealing with issues as they arise, thanks

I didn't mean anything as an insult.

Sorry if it came out that way.

See: https://github.com/pygfx/pygfx/pull/958 and https://github.com/pygfx/wgpu-py/pull/649

Korijn:
> that's fine too. i just opened this to let you know you had an unsatisfiable dep. do with it what you like. including close

Much appreciated, that kind of thing is not supposed to go unnoticed. I will adjust CI to cover the minimum supported python version, that should already improve our capacity to detect such problems going forward.

> Perhaps the examples should leverage PEP 723 so you an easily run them using `uv`
> [https://docs.astral.sh/uv/guides/scripts](https://docs.astral.sh/uv/guides/scripts/)

I have to read up on that, thanks for the reference!

PR #958, opened 2025-01-27
decrease minimum version of imgui-bundle

hmaarrfk:
See https://github.com/pygfx/wgpu-py/pull/649

Closes: https://github.com/pygfx/pygfx/pull/956

cc: @tlambert03


Korijn:
I prefer this change over #956, as it restores consistency. I also pushed a commit to ensure we test the examples on the minimum supported python version going forward.

tlambert03:
Yep, definitely better

Issue #951, opened 2025-01-23
RuntimeError: Request adapter failed (1)


m-agour:
Hello Pygfx team. 

I am getting this error:

```
libEGL warning: MESA-LOADER: failed to open zink: /usr/lib/dri/zink_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)

libEGL warning: MESA-LOADER: failed to open swrast: /usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)

enumerate_adapters: Initialization of an object has failed
Traceback (most recent call last):
  File "/home/agour/try.py", line 32, in <module>
    disp.show(cube)
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/pygfx/utils/show.py", line 168, in show
    self.renderer = WgpuRenderer(self.canvas)
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/pygfx/renderers/wgpu/engine/renderer.py", line 150, in __init__
    self._shared = get_shared()
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/pygfx/renderers/wgpu/engine/shared.py", line 311, in get_shared
    Shared()
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/pygfx/renderers/wgpu/engine/shared.py", line 77, in __init__
    self._adapter = wgpu.gpu.request_adapter_sync(
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/wgpu/_classes.py", line 106, in request_adapter_sync
    return gpu.request_adapter_sync(
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_api.py", line 376, in request_adapter_sync
    return awaitable.sync_wait()
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_helpers.py", line 283, in sync_wait
    return self._finish()
  File "/home/agour/miniconda3/envs/fury/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_helpers.py", line 268, in _finish
[...]

PR #959, opened 2025-01-27
Add  more details on drivers in install instructions

almarklein:
Ref https://github.com/pygfx/pygfx/issues/951

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
     - name: Install llvmpipe and lavapipe for offscreen canvas
       run: |
         sudo apt-get update -y -qq
-        sudo apt-get install --no-install-recommends -y libegl1-mesa-dev libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -143,7 +143,7 @@ jobs:
     - name: Install llvmpipe and lavapipe for offscreen canvas
       run: |
         sudo apt-get update -y -qq
-        sudo apt-get install --no-install-recommends -y libegl1-mesa-dev libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
```
almarklein, line 146: We were using a mix of commands with/without this flag. I removed it in all.

Korijn: Depending on what you install, it can save a lot of time to include the flag. But in our case I don't think any recommended extras are being installed, so it's fine.

almarklein: Thanks for checking. If it would, we should probably update this line everywhere to be consistent :)

Korijn (approved):

PR #960, opened 2025-02-04
Add typing to `objects._base`

tlambert03:
asked @almarklein recently if he'd be open to some "gentle" typing in the library.  I would find it super useful as a user.

I'll take it one file at a time for now, so as to get a sense for what typing constructs feel acceptable for the devs working on this codebase.  Feel free to object to anything and it can be removed

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import random
 import weakref
 import threading
-from typing import List, Tuple
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterator, List, Tuple
 import pylinalg as la
 from time import perf_counter_ns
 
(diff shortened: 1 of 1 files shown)


Korijn (approved):

pygfx/objects/_base.py
```diff
@@ -1,7 +1,9 @@
+from __future__ import annotations
```
almarklein, line 1: Are there Python versions that we support that need this?

tlambert03: No, but in this file it lets you remove the quotes from the self-referencing hints for WorldObject. It’s something I’ve largely gotten in the habit of (unless I know I need runtime type inspection).  If it’s something you’d prefer to use only when absolutely necessary though, I can do that too

tlambert03: Oh wait, actually probably yes: I think I used the pipe instead of Union, which is 3.10+

almarklein: > it lets you remove the quotes from the self-referencing hints for WorldObject

Ah, right. Now I remember we actually use it in wgpu-py too. Precisely to avoid quotes.

PR #961, opened 2025-02-04
Add typing to Geometry and Material

tlambert03:
an extra close look would be good for Geometry, i think i got it accurate without going too nuts, but can relax to `Any` if desired

### DIFF
--- a/pygfx/geometries/_base.py
+++ b/pygfx/geometries/_base.py
@@ -1,8 +1,16 @@
+from __future__ import annotations
+
+import collections
+from typing import TYPE_CHECKING, Iterable
 import numpy as np
 import pylinalg as la
 
 from ..utils.trackable import Trackable
 from ..resources import Resource, Buffer, Texture
+
+if TYPE_CHECKING:
+    import collections.abc
+    import numpy.typing as npt
 
 
 class Geometry(Trackable):
@@ -27,7 +35,7 @@ class Geometry(Trackable):
 
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs: Resource | npt.ArrayLike | collections.abc.Buffer):
         super().__init__()
 
         self._aabb = None
@@ -46,6 +54,7 @@ class Geometry(Trackable):
                     val = np.array(val, dtype=dtype)
                 # Create texture or buffer
                 if name == "grid":
+                    val = np.asanyarray(val)
                     dim = val.ndim
                     if dim > 2 and val.shape[-1] <= 4:
                         dim -= 1  # last array dim is probably (a subset of) rgba
@@ -84,24 +93,24 @@ class Geometry(Trackable):
             # Store
             setattr(self, name, resource)
 
-    def __setattr__(self, key, new_value):
+    def __setattr__(self, key: str, new_value: Resource) -> None:
         if not key.startswith(("_", "morph_")):
             if isinstance(new_value, Trackable) or key in self._store:
                 return setattr(self._store, key, new_value)
         object.__setattr__(self, key, new_value)
 
-    def __getattribute__(self, key):
+    def __getattribute__(self, key: str) -> Resource:
         if not key.startswith(("_", "morph_")):
             if key in self._store:
                 return getattr(self._store, key)
         return object.__getattribute__(self, key)
 
-    def __dir__(self):
+    def __dir__(self) -> Iterable[str]:
         x = object.__dir__(self)
         x.extend(dict.keys(self._store))
         return x
 
-    def get_bounding_box(self):
+    def get_bounding_box(self) -> npt.NDArray[np.float32] | None:
         """Compute the axis-aligned bounding box.
 
         Computes the aabb based on either positions or the shape of the grid
@@ -159,7 +168,7 @@ class Geometry(Trackable):
         else:
             return None
 
-    def get_bounding_sphere(self):
+    def get_bounding_sphere(self) -> npt.NDArray[np.float32] | None:
         """Compute a bounding sphere.
 
         Uses the geometry's axis-aligned bounding box, to estimate a sphere
--- a/pygfx/geometries/_base.py
+++ b/pygfx/geometries/_base.py
@@ -10,7 +10,7 @@ from ..resources import Resource, Buffer, Texture
 
 if TYPE_CHECKING:
     import collections.abc
-    import numpy.typing as npt
+    from numpy.typing import NDArray, ArrayLike
 
 
 class Geometry(Trackable):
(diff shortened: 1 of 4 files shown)


pygfx/geometries/_base.py
```diff
@@ -84,24 +93,24 @@ def __init__(self, **kwargs):
...
         return object.__getattribute__(self, key)
 
-    def __dir__(self):
+    def __dir__(self) -> Iterable[str]:
         x = object.__dir__(self)
         x.extend(dict.keys(self._store))
         return x
 
-    def get_bounding_box(self):
+    def get_bounding_box(self) -> npt.NDArray[np.float32] | None:
```
almarklein, line 113: If we `from numpy.typing import NDArray, ...` at the top, we can make this a bit more neat, I think?

```suggestion
    def get_bounding_box(self) ->NDArray[np.float32] | None:
```
Looking at the rendered docs:
image

almarklein:
RTD renders docs for pr's, so the results can be inspected, which is handy for pr's like this: https://pygfx--961.org.readthedocs.build/961/_autosummary/geometry/pygfx.geometries.Geometry.html#pygfx.geometries.Geometry

almarklein:
>  i think i got it accurate without going too nuts

I agree :)

almarklein:
I am playing with this a bit in VSCode. Interestingly VSCode already gives type hints by deriving expected types from the code 😮, I never realized this. I think adding type annotations helps makes these  hints more precise/clear.

With things like this:
```py
ABCDTuple: TypeAlias = tuple[float, float, float, float]
```

The name `ABCDTuple` is all that VSCode and the docs show, from what I can tell. So important to use clear names (current name looks fine by me, just noting this for myself and other reviewers).

almarklein:
>  I went alias here because the type is longer, but that's definitely something I could switch back. Your call

I like the alias!

pygfx/geometries/_base.py
```diff
@@ -1,9 +1,17 @@
...
+from typing import TYPE_CHECKING, Iterable
 import numpy as np
 import pylinalg as la
 
 from ..utils.trackable import Trackable
 from ..resources import Resource, Buffer, Texture
 
+if TYPE_CHECKING:
+    import collections.abc
+    from numpy.typing import NDArray, ArrayLike
```
Korijn, line 13: Just out of curiosity, why not put all typing specific imports in this if block?

```py
from __future__ import annotations

import collections
from typing import Iterable
```

pygfx/geometries/_base.py
```diff
@@ -116,12 +123,12 @@ def get_bounding_box(self):
             no finite positions.
 
         """
-        if hasattr(self, "positions"):
-            if self._aabb_rev == self.positions.rev:
+        if isinstance(positions := getattr(self, "positions", None), Buffer):
```
tlambert03, line 126: want to draw your attention here, since this will affect runtime behavior (but is more strictly accurate).

One thing I've generally struggled with when using pygfx is having a good sense for what the attributes are.  having special attribute names that *may* be present on a Geometry, and must have a specific type when present, definitely makes for difficulty not only in typing, but in interpretability. 

Just flagging that for now, curious whether it's a pattern you're fully settled and happy with for reasons of flexibility/performance, or are open for refactors

pygfx/geometries/_base.py
tlambert03: can do if you'd like.  usually it comes from IDE autoimport and once typing is imported (for TYPE_CHECKING), things like ruff have no reason to put additional attributes from that module behind the guard

pygfx/geometries/_base.py
tlambert03: done

pygfx/geometries/_base.py
Korijn: Doesn't isinstance raise an error when you pass None though?

tlambert03: No, it’s fine to ask whether None is an instance of Buffer (it’s not an error).  issubclass however will raise an error of the first argument is not a type.

That said, I have no problem going back.  The main difference here is that you would get a type error on main if positions isn’t a buffer, while here it would just do nothing (silently!). It’s fine to leave the type error and remove this guard if you prefer

Korijn: > having special attribute names that _may_ be present on a Geometry, and must have a specific type when present, definitely makes for difficulty not only in typing, but in interpretability

Would you mind sharing if you think there's any (specific) element(s) of the docs that we should improve in this regard?

pygfx/geometries/_base.py
hmaarrfk: Type hints are cool!

one other consideration is "import time". For libraries like numpy, that is fine.

But when you get to "heavy libraries" like tensorflow, or pytorch, having them be imported at runtime really slows you down.

I go through this periodically in my application, where the import time swells, then eventually i audit the code, and find that something like tensorflow / pytorch just increase the import time to 3-4 seconds.

For example, if one had a pygfx application:
```
pygfx --help
```
would be expected to be VERY fast, like 30 ms fast!

tlambert03: not sure I'm following the comment.  typing needn't import heavy libraries if they're only used for type hints.  was there a suggestion/request/caution here?  or just a reminder to guard heavy imports behind `TYPE_CHECKING`?  (since this project already uses ruff, we can use [ruff-TC](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) to ensure that

hmaarrfk: > or just a reminder to guard heavy imports behind TYPE_CHECKING?

this.

tlambert03: added TC rule to ruff linter

pygfx/geometries/_base.py
almarklein: I understand the "fuzzyness" of the geometry object. It may be worth discussing the idea of producing a standardized list of attributes that a geometry commonly has, and their types and shape. We could even make real properties for them? Though it should always be possible for custom objects to use new geometry attributes. Let's discuss in a separate issue.

> Many attributes of Trackables have no "go-to-definition" ability

I was not aware that. But this we can fix with type annotations, correct?

almarklein: FWIW somethings I like to think of the geometry as just a dict with a bunch of buffers and textures 🤷

Korijn (approved):

pygfx/geometries/_base.py
tlambert03: sounds good.

python can definitely express "if a key exists, it has this type" with TypedDict.  Having the attribute access as a shortcut makes things a bit harder.  But there are a few things we can do.  I have another branch working on Resource and will open a new PR to discuss those things

PR #963, opened 2025-02-04
🚀 RecursiveTransform optimization

Korijn:
Comparing main against this branch on my machine we go from 160 FPS to 180 FPS on the skinning animation example.

**Changes:**

Clean up:
* Sprinkled some more type hints in `transform.py`
* Renamed `_reference_up_provider` to `_wrapper` and moved it to `AffineTransform` since it is only used by that subclass
* Constructor arguments `base` and `parent` on `RecursiveTransform` require specific subclasses of `AffineBase` now, it didn't make sense anymore, and it's simply required if we want to go ahead with further performance boosting as done in this PR

The meat of this PR that enables the speed boost:
* `RecursiveTransform` now has a `children` list, maintained by `WorldObject` (just like it maintains the `parent` field)
* Using the `children` list, recursive cache invalidation is moved back to `RecursiveTransform.flag_update` (so at write time, not at read time) and I've used cProfile to micro optimize this function, I'm very sure this is the most performant implementation we can get
* `RecursiveTransform.parent` can be `None` now, avoiding a redundant matrix multiplication in the tree (it used to be a dummy `AffineTransform` instance for the root)

**Performance details:**

This is faster because we now avoid lots of redundant cache checks, and the hot path (`flag_update`) is _much_ more performant.

If you look at `RecursiveTransform.flag_update` closely you'll see an if statement that looks redundant, because the for loop would not iterate even once on an empty list, but it makes a 10 FPS difference. 🤡 I guess there's a lot of overhead involved in iteration, even if the list is empty.

I also looked at implementations with a stack (both list and deque based) so we don't have to recurse, but that's about 1.5x slower than using recursion.

**Next steps:**

Since I think I've squeezed all the performance I can out of the scene graph mechanics (whew! 😅), the next PR will focus on reusing the ndarrays in the cache, instead of constantly reallocating them.

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -722,28 +722,22 @@ class RecursiveTransform(AffineBase):
         is_camera_space=False,
     ) -> None:
         super().__init__(reference_up=reference_up, is_camera_space=is_camera_space)
-        self._last_modified = perf_counter_ns()
-
-        self._parent = None
-        self.own = None
-
+        self.last_modified = perf_counter_ns()
+        self._parent = parent
+        self._children = []
         self.own = base
         self.own._set_reference_up_provider(self)
 
-        if parent is None:
-            self._parent = AffineTransform()
-        else:
-            self._parent = parent
-
     def flag_update(self):
         """Signal that this transform has updated."""
-        self._last_modified = perf_counter_ns()
-
-    @property
-    def last_modified(self) -> int:
-        return max(
-            self._last_modified, self.own.last_modified, self._parent.last_modified
-        )
+        # TODO: micro-optimize
+        # use deque maybe?
+        stack = [self]
+        while stack:
+            current = stack.pop()
+            # TODO: use perf_counter_ns()?
+            current.last_modified = self.last_modified
+            stack.extend(current._children)
 
     @cached
     def __own_reference_up(self) -> np.ndarray:
@@ -774,10 +768,7 @@ class RecursiveTransform(AffineBase):
 
     @parent.setter
     def parent(self, value):
-        if value is None:
-            self._parent = AffineTransform()
-        else:
-            self._parent = value
+        self._parent = value
         self.flag_update()
 
     @cached
@@ -798,6 +789,7 @@ class RecursiveTransform(AffineBase):
     @matrix.setter
     def matrix(self, value):
         self.own.matrix = self._parent.inverse_matrix @ value
+        self.flag_update()
 
     def __matmul__(self, other) -> Union["RecursiveTransform", np.ndarray]:
         if isinstance(other, AffineBase):
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -738,9 +738,8 @@ class RecursiveTransform(AffineBase):
     def flag_update(self):
         """Signal that this transform has updated."""
         self.last_modified = perf_counter_ns()
-        if (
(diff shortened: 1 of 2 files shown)


pygfx/utils/transform.py
```diff
@@ -491,17 +484,38 @@ def __init__(
...
+        if self._wrapper:
+            return self._wrapper._parent_reference_up
+        return self._reference_up_view
+
+    @reference_up.setter
+    def reference_up(self, value):
+        if self._wrapper:
+            self._wrapper._parent_reference_up = value
+        else:
+            self._reference_up[:] = la.vec_normalize(value)
```
Korijn, line 510: I guess this should call the base implementation like we've done for a few other of these setters

pygfx/utils/transform.py
```diff
@@ -714,80 +721,76 @@ class RecursiveTransform(AffineBase):
...
-    def _matrix(self):
-        mat = self._parent.matrix @ self.own.matrix
-        mat.flags.writeable = False
-        return mat
+    def _matrix(self) -> np.ndarray:
+        if self._parent:
+            mat = self._parent.matrix @ self.own.matrix
+            mat.flags.writeable = False
+            return mat
+        return self.own.matrix
```
Korijn, line 790: I could move this to the property getter actually and skip the cache descriptor overhead

pygfx/utils/transform.py
Korijn: Done

Korijn:
@almarklein ready for review!

almarklein:
So in my own words, the transform now shadows the scene graph structure, so that invalidation (setting `last_modified`) is quick even though it recurses. It avoid having to go to the world-object to traverse the graph, which is slower but also tricky due to double-references.

pygfx/utils/transform.py
```diff
@@ -491,17 +484,38 @@ def __init__(
...
+        self._wrapper: RecursiveTransform = None
+
     def flag_update(self):
         """Signal that this transform has updated."""
         self.last_modified = perf_counter_ns()
+        if self._wrapper:
+            self._wrapper.flag_update()
+
+    def _set_wrapper(self, wrapper: RecursiveTransform):
+        self._wrapper = wrapper
```
almarklein, line 496: This does mean that the local and world transform reference each-other. Is there a sensible place/time where `self._wrapper` can be set to None again? Maybe when the world object's `__del__`.

For clarity / others reading along: things referencing each-other are cleaned up by Python's cg, but it takes longer.

Korijn: > This does mean that the local and world transform reference each-other.

Yeah, they already did. I just moved this method from AffineBase to AffineTransform. Because of all of these links we have already, I plan to change (in a later PR) WorldObject.parent from a weakref back to a plain reference, because it's not really making a difference.

> Is there a sensible place/time where self._wrapper can be set to None again? Maybe when the world object's __del__.

With every PR I've opened recently, I've been embracing more and more the fact that the Transform API is not really being used outside of the WorldObject class, and it's fine if we specialize it for that use case. We could set it to None in `WorldObject.__del__` if you like, but there would be many more things to "detach" really.

almarklein: > Is it OK if I do this in a separate PR?

Absolutely. Just wanted to point this out.

almarklein (approved):

PR #965, opened 2025-02-05
⚡ Micro optimization

Korijn:
* Addresses [this comment](https://github.com/pygfx/pygfx/pull/963#discussion_r1942551469) by @almarklein by breaking the reference cycle between `AffineTransform` and `RecursiveTransform` when a `WorldObject` is deleted
* Micro-optimizes `RecursiveTransform.flag_update` a little bit more with the walrus operator; instead of looking up an attribute on `self` multiple times, it's done once and after that the variable is a local which is much faster to access

Timing per call (time spent inside function and all subfunctions)
* main: 4.454e-06
* this branch: 4.016e-06

It makes every call to `RecursiveTransform.flag_update` 10% faster and there are _so many calls_ so it's totally worth it :)

Note: I haven't tried `__slots__` yet, that is probably worthwhile now that I've completed all the major refactoring. Will do that in a next PR

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -208,6 +208,9 @@ class WorldObject(EventTarget, Trackable):
 
     def __del__(self):
         id_provider.release_id(self, self.id)
+        self.local._set_wrapper(
+            None
+        )  # break the circular reference so GC has it a little easier
 
     @property
     def up(self) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


almarklein (approved):
🚀

Korijn:
I tried `__slots__` and surprisingly it didn't make any difference, the measurements are almost identical. I realized the main bottleneck in flag_update is iteration and function calls, not attribute access.

almarklein:
I wonder if by this time you should consider benchmarking on multiple Python versions 🤔

PR #966, opened 2025-02-06
Provide a path for end users to register their own blenders

hmaarrfk:
The ability to register blenders would remove a big 'non sactioned' customizations I have for PyGFX.

It would also allow more end user blender modes.

I know the blenders/rendering pipeline is being refactored. But this would really help in the meantime (after release 0.8.0 of course).

Thanks for considering.

### DIFF
--- a/pygfx/renderers/wgpu/engine/__init__.py
+++ b/pygfx/renderers/wgpu/engine/__init__.py
@@ -96,6 +96,6 @@ objects in a scene, its likely that many of these have the same material.
 def _register_blenders():
     # triggers an import which will allow the blenders to register
     # themselves
-    from .blender import WgpuRenderer
+    from .blender import WgpuRenderer  # noqa F401
 
 _register_blenders()
--- a/pygfx/renderers/wgpu/engine/__init__.py
+++ b/pygfx/renderers/wgpu/engine/__init__.py
@@ -92,12 +92,3 @@ objects in a scene, its likely that many of these have the same material.
 
 (Figures created with https://asciiflow.com/)
 """
-
-
-def _register_blenders():
-    # triggers an import which will allow the blenders to register
-    # themselves
-    from .blender import WgpuRenderer  # noqa F401
-
-
-_register_blenders()
--- a/pygfx/renderers/wgpu/engine/__init__.py
+++ b/pygfx/renderers/wgpu/engine/__init__.py
@@ -92,3 +92,10 @@ objects in a scene, its likely that many of these have the same material.
 
 (Figures created with https://asciiflow.com/)
 """
+
+def _register_blenders():
+    # triggers an import which will allow the blenders to register
+    # themselves
+    from .blender import WgpuRenderer
+
+_register_blenders()
--- a/pygfx/renderers/wgpu/engine/__init__.py
+++ b/pygfx/renderers/wgpu/engine/__init__.py
@@ -93,9 +93,11 @@ objects in a scene, its likely that many of these have the same material.
 (Figures created with https://asciiflow.com/)
 """
 
+
 def _register_blenders():
     # triggers an import which will allow the blenders to register
     # themselves
     from .blender import WgpuRenderer  # noqa F401
+
 
 _register_blenders()
--- a/pygfx/renderers/wgpu/engine/blender.py
+++ b/pygfx/renderers/wgpu/engine/blender.py
@@ -9,6 +9,7 @@ import wgpu  # only for flags/enums
 from ....utils.enums import RenderMask
 from .flusher import create_full_quad_pipeline
 from .shared import get_shared
+from .renderer import WgpuRenderer
 
 # Notes:
 # - The user code provides color as-is in rgba.
@@ -689,7 +690,7 @@ class BaseFragmentBlender:
         """Overload this to setup the specific combiner-pass."""
         return None
 
-
+@WgpuRenderer.register_blend_mode("opaque")
 class OpaqueFragmentBlender(BaseFragmentBlender):
     """A fragment blender that pretends that all surfaces are opaque,
     even if they're not.
@@ -698,6 +699,7 @@ class OpaqueFragmentBlender(BaseFragmentBlender):
     passes = [FullOpaquePass()]
 
 
+@WgpuRenderer.register_blend_mode("dither")
 class DitherFragmentBlender(BaseFragmentBlender):
     """A fragment blender that pretends that all surfaces are opaque,
     even if they're not.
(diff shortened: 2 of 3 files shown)


pygfx/renderers/wgpu/engine/__init__.py
```diff
@@ -92,3 +92,12 @@
...
 """
+
+
+def _register_blenders():
+    # triggers an import which will allow the blenders to register
+    # themselves
+    from .blender import WgpuRenderer  # noqa F401
+
+
+_register_blenders()
```
almarklein, line 103: I don't understand the purpose of this.

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -321,6 +321,19 @@ def blend_mode(self):
         """
         return self._blend_mode
 
+    @staticmethod
+    def register_blend_mode(name, blender=None):
+        """Register a new blender for usage with rendering pipelines."""
```
almarklein, line 326: Let's document that this is experimental. Until now the blender was an implementation detail, this makes it a bit public. So let's make it use-at-your-own risk.

almarklein (commented):
I think this can work as an experimental feature, and we can refine it (i.e. document what a blender should look like) after the upcoming refactoring.

pygfx/renderers/wgpu/engine/__init__.py
hmaarrfk: As you said. Most users don’t know what the blender code and will never import it. 

I need something to import the blender code otherwise none of the blenders get registered

almarklein: Maybe just `from . import blender   # noqa - import to register blenders` in renderer.py?

hmaarrfk: let me try that. this introduces circular dependencies which likes to break every other version of python or something like that.

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -321,6 +322,28 @@ def blend_mode(self):
         """
         return self._blend_mode
 
+    @staticmethod
+    def register_blend_mode(name, blender_class=None):
+        """Register a new blender for usage with rendering pipelines.
+
+        Note that Blender classes are highly experimental and their inteface
+        is expected to change rapidly from pygfx version 0.7.0 to
```
hmaarrfk, line 330:
```suggestion
        is expected to change rapidly from pygfx version 0.7.0 to
```

pygfx/renderers/wgpu/engine/renderer.py
hmaarrfk: > So let's make it use-at-your-own risk.

Do you want to use the convention of `_register_blend_mode` or just add enough comments?

hmaarrfk: ok i tried to add maximum caution

pygfx/renderers/wgpu/engine/blender.py
```diff
@@ -535,6 +536,9 @@ class BaseFragmentBlender:
     Each renderer has one blender object.
     """
 
+    # Each blender should define a unique name so that it may be
+    # correctly registered in the renderer.
+    name = None
```
hmaarrfk, line 541: Sorry for the "major" refactor here but lmk what you think. I think it makes it cleaner to define the name of the blender as part of the blender class not "the name of the blender as the renderer believes it to be".

It does create the possibility of clashes downstream, but those issues tend to resolve themselves by the consuming libraries.

hmaarrfk:
No matter what, "simplifying" the code always causes the number of lines to increase 
image

sorry.

pygfx/renderers/wgpu/engine/blender.py
almarklein: I think this makes sense. As a side-effect, the decorators can also omit the `()` so the implementation of `_register_blend_mode` can be simpler too :)

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -321,33 +322,61 @@ def blend_mode(self):
...
-            "weighted_depth": blender_module.WeightedDepthFragmentBlender,
-            "weighted_plus": blender_module.WeightedPlusFragmentBlender,
-        }
-        if value not in m:
-            raise ValueError(
-                f"Unknown blend_mode '{value}', use any of {set(m.keys())}"
-            )
+
+        b = self._blenders_available.get(value)
+        if blender is None:
```
almarklein, line 374:
```suggestion
        blender = self._blenders_available.get(value, None)
        if blender is None:
```

almarklein: And the `b()`below should be `blender()`

hmaarrfk: This clashes with `from . import blender` in ruff.

hmaarrfk: ok i worked around it in a different way

pygfx/renderers/wgpu/engine/blender.py
hmaarrfk: is this the syntax you had in mind?

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -321,33 +322,53 @@ def blend_mode(self):
...
-            "weighted": blender_module.WeightedFragmentBlender,
-            "weighted_depth": blender_module.WeightedDepthFragmentBlender,
-            "weighted_plus": blender_module.WeightedPlusFragmentBlender,
-        }
-        if value not in m:
-            raise ValueError(
-                f"Unknown blend_mode '{value}', use any of {set(m.keys())}"
-            )
+
+        blender = self._blenders_available.get(value)
```
almarklein, line 365: I would have sworn that
```py
{}.get("foo")
```
would raise a `KeyError` ... turns out it does not. I've been writing *a lot* of `.get(.., None)` because of this assumption 😄  Weird.

almarklein (approved):
Nice!

pygfx/renderers/wgpu/engine/renderer.py
hmaarrfk: Pop raises an error. Which is a little annoying in terms of behavior difference…

almarklein: Right, that's what has confused me!

Korijn (approved):

hmaarrfk:
Thanks a bunch!

PR #968, opened 2025-02-06
Improve z fighting control for text

hmaarrfk:
I think this is "more correct" right @almarklein ? I was looking at this while debugging my stuff.

It wasn't the source of the problem, but I think it is an improvement.

LMK.


hmaarrfk:
The new errors are from:
image

The screencast studies the error.
1. I take the "new" screenshot and use contrast adjustment help visually highlight the differences
2. I scroll up and down along the "l" to show that it is consistent next to the "e".
3. I take the "old" screen, apply similar contrast adjustment
4. I scroll up and down and you can see the color change when it gets closer to the "e"

https://github.com/user-attachments/assets/6ce44cf9-0ff7-4deb-af85-01b5354d91de

In the old screenshots, the colors along the "l" would vary from `0x0e05015` to `0x090311` (color picker used in GIMP)

hmaarrfk:
For small font size, the outline really helps get the text to "pop" which is why I've been spending so much time on this outline...

almarklein:
Hard to understand the implications from the code alone. I think your screencast is convincing, so let's add it :)

almarklein (approved):

almarklein:
I really appreciate the meticulous attention to detail.

pygfx/renderers/wgpu/wgsl/text.wgsl
```diff
@@ -237,7 +237,8 @@ fn fs_main(varyings: Varyings) -> FragmentOutput {
     // z-fighting to occur leading to artifacts.
     // See: https://github.com/pygfx/pygfx/pull/776#issuecomment-2149374275
     // And: https://github.com/pygfx/pygfx/pull/774#issuecomment-2147458827
-    out.depth = varyings.position.z - 1e-6 * (2.0 - outline);
+    // float32 precision is about 5-6 digits
+    out.depth = varyings.position.z * (0.99998 + 1e-5 * outline);
```
Korijn, line 241: I'm not sure if your intention is to try and take the smallest possible step in float32, but you can directly do that like this:

```
uint asInt = asuint(myFloat);
asInt++;
float next = asfloat(asInt);
```

(pseudocode, I don't actually know how to do this in WGSL)

ignore this comment if it's not what you're trying to do :)

hmaarrfk: I am trying to do that. But I’ve been around floats long enough to know that this isn’t very smart to do since GPUs sometimes don’t adhere to IEEE for floats and as the float gets really small or really big, the standard way floats work isn’t really guaranteed. 

Finally. We might want to increase or decrease the step size in the future.

hmaarrfk:
```suggestion
    // The math below is equivalent to (1 - 1e-5 * (2 - outline))
    out.depth = varyings.position.z * (0.99998 + 1e-5 * outline);
```

Korijn: Yeah. Sounds good to me!

Korijn (approved):

hmaarrfk:
> Hard to understand the implications from the code alone. I think your screencast is convincing, so let's add it :)

Missed this one, but I anticipate more tweaking from this to come. 

Ultimately the issue is that all these edge cases.

I did two things here:

1. Increased the absolute shift. from +-1E-6 to +-1E-5
2. Made it a "relative" shift by multiplying instead of adding.

PR #969, opened 2025-02-06
⚡⚡ Micro-optimized flag_update again

Korijn:
Per call timing for `flag_update` is now 1.963e-06. It was 4.016e-06 before. So 2x as fast! The trick was to make the function do less function calls and maximize its use of locals.

I also tried `__slots__` but it didn't affect performance. I did rename a bunch of stuff in the process, since the double underscore prefixes were triggering name mangling and that was not my intention.

### DIFF
--- a/pygfx/utils/transform.py
+++ b/pygfx/utils/transform.py
@@ -20,12 +20,14 @@ class cached:  # noqa: N801
 
     """
 
+    __slots__ = ("compute_fn", "name")
+
     def __init__(self, compute_fn=None) -> None:
         self.compute_fn = compute_fn
         self.name = None
 
     def __set_name__(self, clazz, name) -> None:
-        self.name = f"_{name}_cache"
+        self.name = f"cache_{name}_cache"
 
     def __get__(self, instance: AffineBase, clazz=None) -> Any:
         if instance is None:
@@ -488,9 +490,12 @@ class AffineTransform(AffineBase):
 
     def flag_update(self):
         """Signal that this transform has updated."""
-        self.last_modified = perf_counter_ns()
+        # note: this function has been heavily micro-optimized
+        # please don't modify it carelessly
+        last_modified = perf_counter_ns()
+        self.last_modified = last_modified
         if wrapper := self._wrapper:
-            wrapper.flag_update()
+            wrapper.flag_update(last_modified)
 
     def _set_wrapper(self, wrapper: RecursiveTransform):
         self._wrapper = wrapper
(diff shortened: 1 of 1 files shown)


pygfx/utils/transform.py
```diff
@@ -735,16 +740,22 @@ def __init__(
...
+        if last_modified is None:
+            last_modified = perf_counter_ns()
+        self.last_modified = last_modified
         if children := self.children:
             for child in children:
-                child.flag_update()
+                if child.children:
+                    child.flag_update(last_modified)
+                else:
+                    child.last_modified = last_modified
```
almarklein, line 755: With this, if the child has children, the child's `last_modified` is not set. Is that intentional?

almarklein: From what I see, the idea is to traverse faster by hopping two generation at a time instead of one, is that correct?

pygfx/utils/transform.py
```diff
@@ -20,12 +20,14 @@ class cached:  # noqa: N801
...
 
+    __slots__ = ("compute_fn", "name")
+
     def __init__(self, compute_fn=None) -> None:
         self.compute_fn = compute_fn
         self.name = None
 
     def __set_name__(self, clazz, name) -> None:
-        self.name = f"_{name}_cache"
+        self.name = f"cache_{name}_cache"
```
almarklein, line 30: Is this intentional (the name "cache" occuring twice)?

Korijn: * Yes, it doesn't really matter what it is, as long as it's unlikely to clash with attribute names actually want to use.
* My problem with the original name was that it led to double underscores if `name` happened to be prefixed with a single underscore, and that in turn triggered name mangling. That's why I put some words in the prefix here.

pygfx/utils/transform.py
Korijn: * If the child has children, we recurse into the child's flag_update. That starts by setting self.last_modified, and then iterates over its children again. So we're not skipping anything.
* The idea is to avoid recursing into the leaf nodes, and just setting last_modified on them directly. Leaf nodes are usually the majority of the nodes in the scene graph so it makes an enormous difference.

almarklein: Right, with hopping I meant the recursion step. We recurse less, but handle two generation at each recursion.

I see how this should set `last_modified` on all descendants when this is the second recursion. But if this is the initial call, the `child` does not get its  `last_modified` set.

Korijn: > We recurse less, but handle two generation at each recursion.

But we don't! We only do this at the leaf level.

almarklein: Aaah! I see what I missed now. I somehow understood that `if child.children` it was going to call `flag_update` on *those* children, which is not what's happening. 

Sorry for the confusion!

PR #972, opened 2025-02-08
💾 Cache reuse

Korijn:
* Reuse cached ndarrays

### DIFF
--- a/pygfx/cameras/_base.py
+++ b/pygfx/cameras/_base.py
@@ -108,7 +108,7 @@ class Camera(WorldObject):
         self._view_offset = None
         self.flag_update()
 
-    def _update_projection_matrix(self) -> np.ndarray:
+    def _update_projection_matrix(self, cache) -> np.ndarray:
         raise NotImplementedError()
 
     def get_state(self):
@@ -124,9 +124,14 @@ class Camera(WorldObject):
         return self.world.inverse_matrix
 
     @cached
-    def projection_matrix(self) -> np.ndarray:
-        base = self._update_projection_matrix()
-        if self._view_offset is None:
+    def projection_matrix(self, cache) -> np.ndarray:
+        if cache is not None:
+            cache.flags.writeable = True
+
+        only_base = self._view_offset is None
+        base = self._update_projection_matrix(cache if only_base else None)
+        if only_base:
+            base.flags.writeable = False
             return base
 
         view_offset = self._view_offset
@@ -143,17 +148,27 @@ class Camera(WorldObject):
                 [0.0, 0.0, 1.0, 0.0],
                 [0.0, 0.0, 0.0, 1.0],
             ],
-            np.float32,
+            dtype=np.float32,
         )
-        return ndc_matrix @ base
+        out = np.matmul(ndc_matrix, base, out=cache)
+        out.flags.writeable = False
+        return out
 
     @cached
-    def projection_matrix_inverse(self) -> np.ndarray:
-        return la.mat_inverse(self.projection_matrix)
+    def projection_matrix_inverse(self, cache) -> np.ndarray:
+        if cache is not None:
+            cache.flags.writeable = True
+        out = la.mat_inverse(self.projection_matrix, out=cache)
+        out.flags.writeable = False
+        return out
 
     @cached
-    def camera_matrix(self) -> np.ndarray:
-        return self.projection_matrix @ self.view_matrix
+    def camera_matrix(self, cache) -> np.ndarray:
+        if cache is not None:
+            cache.flags.writeable = True
+        out = np.matmul(self.projection_matrix, self.view_matrix, out=cache)
+        out.flags.writeable = False
+        return out
 
 
 class NDCCamera(Camera):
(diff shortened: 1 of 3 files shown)


Korijn:
It's working now, not sure yet if it makes things faster, or just less memory hungry only.

hmaarrfk:
Slightly slower if the benchmark is to be believed:

`g29ea9a1` == `main`

```
(dev) ✔ ~/git/pygfx-benchmarks [add_benchmark_for_transforms|✚ 2…2]
mark@carbon $ (cd /home/mark/git//pygfx && git describe --tags) && python benchmarks/bm_transforms.py
v0.7.0-29-g29ea9a1
    bm_transforms_100_children (20x) - cpu:  0.01 ms
   bm_transforms_1000_children (20x) - cpu:  0.04 ms
bm_transforms_1000_children_with_50_grandchildren (20x) - cpu:  2.92 ms
(dev) ✔ ~/git/pygfx-benchmarks [add_benchmark_for_transforms|✚ 2…2]
mark@carbon $ (cd /home/mark/git//pygfx && git describe --tags) && python benchmarks/bm_transforms.py
v0.7.0-35-ged2e1da
    bm_transforms_100_children (20x) - cpu:  0.01 ms
   bm_transforms_1000_children (20x) - cpu:  0.04 ms
bm_transforms_1000_children_with_50_grandchildren (20x) - cpu:  3.00 ms
(dev) ✔ ~/git/pygfx-benchmarks [add_benchmark_for_transforms|✚ 2…2]
mark@carbon $ (cd /home/mark/git//pygfx && git describe --tags) && python benchmarks/bm_transforms.py
v0.7.0-29-g29ea9a1
    bm_transforms_100_children (20x) - cpu:  0.01 ms
   bm_transforms_1000_children (20x) - cpu:  0.04 ms
bm_transforms_1000_children_with_50_grandchildren (20x) - cpu:  2.95 ms
(dev) ✔ ~/git/pygfx-benchmarks [add_benchmark_for_transforms|✚ 2…2]
mark@carbon $ (cd /home/mark/git//pygfx && git describe --tags) && python benchmarks/bm_transforms.py
v0.7.0-35-ged2e1da
    bm_transforms_100_children (20x) - cpu:  0.01 ms
   bm_transforms_1000_children (20x) - cpu:  0.04 ms
bm_transforms_1000_children_with_50_grandchildren (20x) - cpu:  3.09 ms
```

linux, python 3.12.8, `Intel(R) Core(TM) Ultra 7 165U`

Korijn:
I think I will split this branch into a few pieces then.

Korijn:
* The reason it got slower was a mistake on my part, I corrected it in my last push to this branch, see c745903
* Let's merge #973 first, afterwards this PR will be purely about cache reuse so we can measure the difference more easily

Korijn:
I've redone the changes on this branch so it's independent from #973 (it won't be free of conflicts). But this way we can test and benchmark the changes independently.

Korijn:
I've compared this branch against main, it seems like it performs ever so slightly less, but if I analyze the cProfile reports and compare them, I can't tell the difference. I think it's safe to merge this. There are less calls to `np.empty` and the like, which is what we were aiming for.

On my machine, with both this branch and #973 incorporated, I get 200 FPS on the skinning animation example. 🥳 All the hard work really paid off!

I think with these two PRs I can consider the Transform API to be done, and acceptable, finally. It's such a core element of pygfx, I want it to be great when we get to v1.0.0.

hmaarrfk:
> I think it's safe to merge this. 

But the added variable `cache` makes this very difficult to read / maintain...

> There are less calls to np.empty and the like, which is what we were aiming for. Less pressure on the garbage collector.

I think this might be a false hope. numpy has a policy to cache small arrays https://github.com/numpy/numpy/issues/18358#issuecomment-774709255
i wish i knew where the source was...

hmaarrfk:
I think this is their allocation code:
https://github.com/numpy/numpy/blob/097e18e3a194023a10bf78b00f2c6d939ee0a858/numpy/_core/src/multiarray/alloc.c#L37

I'm trying to read it, but it seems that the variable called `NBUCKET` is actually the size of the array to cache in certain cases

https://github.com/numpy/numpy/blob/097e18e3a194023a10bf78b00f2c6d939ee0a858/numpy/_core/src/multiarray/alloc.c#L172

Korijn:
> But the added variable `cache` makes this very difficult to read / maintain...

Can you explain what about this bothers you particularly, so I can think of an improvement?

hmaarrfk:
> Can you explain what about this bothers you particularly, so I can think of an improvement?

Maybe it will be clearer when the other merge request goes in. 

However, without any performance improvement , I feel this falls into the realm of premature optimization. 

Additional if statements cause new errors, untested combinations of behavior. New requirements on code refactoring make this all more complicated. The general premature optimization concerns.

almarklein:
Yeah, I agree the `cache` argument adds some complexity. Not a lot, but hard to justify if we cannot observe a benefit.

I think I do spot a few places where we set `.flags.writeable = False` which we may want to keep?

almarklein:
> It doesn't seem to bring any significant benefits.

I think this means we've reached the point where we've made it as fast as it can be 🚀

Korijn:
> I think I do spot a few places where we set `.flags.writeable = False` which we may want to keep?

Yeah I'll port those over to a new PR.

PR #978, opened 2025-02-11
Components setter & animation optimization

Korijn:
This makes it possible to set all three transform components in a single property assignment, to avoid calling `flag_update` 3x more than necessary.

Work in progress.


Korijn:
This is ready to merge.

almarklein (approved):

PR #980, opened 2025-02-13
Add Iridescence support to MeshPhysicalMaterial

panxinmiao:
This PR introduces support for iridescence in MeshPhysicalMaterial.

It enables rendering effects where hue varies depending on the viewing and illumination angles, similar to phenomena observed in soap bubbles, oil films, and the wings of many insects.

Related glTF extension:  [KHR_materials_iridescence](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_iridescence/README.md).

image

### DIFF
--- a/examples/feature_demo/gltf_viewer.py
+++ b/examples/feature_demo/gltf_viewer.py
@@ -27,9 +27,7 @@ renderer = gfx.WgpuRenderer(canvas)
 
 scene = gfx.Scene()
 
-dl = gfx.DirectionalLight()
-dl.local.position = (-3, 10, 10)
-scene.add(gfx.AmbientLight(intensity=0.1), dl)
+scene.add(gfx.AmbientLight(intensity=0.1))
 camera = gfx.PerspectiveCamera(45, 1280 / 720)
 
 gfx.OrbitController(camera, register_events=renderer)
@@ -60,7 +58,7 @@ response.raise_for_status()
 model_list: list = response.json()
 
 # filter out models having "core" in tags
-model_list = [m for m in model_list if "core" in m.get("tags", [])]
+# model_list = [m for m in model_list if "core" in m.get("tags", [])]
 
 mixer = gfx.AnimationMixer()
 
--- a/examples/feature_demo/gltf_viewer.py
+++ b/examples/feature_demo/gltf_viewer.py
@@ -82,6 +82,8 @@ background = gfx.Background(None, gfx.BackgroundSkyboxMaterial(map=env_tex))
 background.visible = False
 scene.add(background)
 
+scene.add(gfx.Background.from_color((0.1, 0.1, 0.1, 1)))
+
 
 def add_env_map(obj):
     if isinstance(obj, gfx.Mesh) and isinstance(obj.material, gfx.MeshStandardMaterial):
--- a/examples/feature_demo/gltf_viewer.py
+++ b/examples/feature_demo/gltf_viewer.py
@@ -58,7 +58,7 @@ response.raise_for_status()
 model_list: list = response.json()
 
 # filter out models having "core" in tags
-# model_list = [m for m in model_list if "core" in m.get("tags", [])]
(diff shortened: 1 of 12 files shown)


kushalkolar:
This looks very cool! 😄

examples/feature_demo/gltf_viewer.py
```diff
@@ -60,7 +58,7 @@
 model_list: list = response.json()
 
 # filter out models having "core" in tags
-model_list = [m for m in model_list if "core" in m.get("tags", [])]
+# model_list = [m for m in model_list if "core" in m.get("tags", [])]
```
almarklein, line 61: Comment or remove?

examples/feature_demo/gltf_viewer.py
```diff
@@ -84,6 +82,8 @@
 background.visible = False
 scene.add(background)
 
+scene.add(gfx.Background.from_color((0.1, 0.1, 0.1, 1)))
```
almarklein, line 85: Can a scene have two backgrounds?

almarklein (commented):
LGTM except the example seems to need a bit of cleanup.

examples/feature_demo/gltf_viewer.py
panxinmiao: Ah, I removed the tag filter to display all models in the glTF asset library for easier testing of models related to Iridescence.

Currently, many glTF extensions are not yet implemented, so for now, let's keep only the "core" tag. 
I think once we support more advanced rendering features in the glTF extensions, we will remove this restriction finally.

examples/feature_demo/gltf_viewer.py
panxinmiao: One is the environment background (skybox), and the other is a solid color background. My expectation is that when the background is checked, the skybox background will be displayed, and when unchecked, a default solid color background will appear. Adding both backgrounds seems to work correctly (when displayed together, the skybox will override the solid color background).

For reference: The reason for this approach is that I encountered an issue while implementing Physically-Based Transparency. When generating the transmission light sampling texture, the background in the framebuffer should not be transparent. I will continue to explore solutions in #974.

Korijn: Locally this works for me as described :+1: cool!

pygfx/materials/_mesh.py
```diff
@@ -1081,12 +1089,27 @@ class MeshPhysicalMaterial(MeshStandardMaterial):
         How much the clearcoat normal map affects the material. This 2-tuple
         is multiplied with the clearcoat_normal_map's xy components (z is unaffected).
         Typical ranges are 0-1. Default is (1,1).
+    iridescence : float
+        The intensity of the iridescence layer, simulating RGB color shift based on the angle
+        between the surface and the viewer,from 0.0 to 1.0. Default is 0.0.
```
Korijn, line 1094:
```suggestion
        between the surface and the viewer, from 0.0 to 1.0. Default is 0.0.
```

pyproject.toml
```diff
@@ -49,6 +50,7 @@ docs = [
     "trimesh <4.6",
     "gltflib",
     "imgui-bundle>=1.6.0",
+    "httpx",
```
Korijn, line 53: do we have examples that are built by docs that require httpx? I'm not sure if that's desirable.

Korijn:
This PR looks good to me! 1 question about httpx dependency, otherwise good to go.

pyproject.toml
panxinmiao: The "iridescence.py" example utilizes httpx (via the glTF loader) to load sample models from remote sources. Future examples may also load resources remotely, eliminating the need to include these assets in the pygfx repository. Of course, we can disable automatic testing and documentation generation for this type of example, but i think it's worthwhile to include certain examples and demos that showcase distinctive features in the documentation. :)

PR #982, opened 2025-02-14
Clamp colormap by default when texture is given.

### LINKED ISSUE #979: #957 causes issues with colormap clamping on lavapipe on github actions
After #957 colormaps for textures no longer clamp on the min value, but this occurs only on lavapipe. 

For example, here the colorbar on the right is an `Image` with a `Texture` that contains values between -51 (bottom of the bar) to 303 (top of the bar). The image material clim is (0.0, 255.0). The values below zero should be black. 

Image

Correct image:

Image

Weird thing is this issue is only on lavapipe on github actions, the mac runners do not produce this issue and I can't produce this locally on any GPU (integrated, nvidia, AMD).

I tried using lavapipe locally but getting segfaults, I'm on debian 11 will try on a newer OS tomorrow.

almarklein:
Closes #979 (I think).

This makes the colormaps a bit more as expected for scientific users. However it also introduces inconsistency:
```py
mesh.map = some_texture  # clamp
mesh.map = gfx.TextureMap(some_texture)  # repeat
```

The thing is that from general 3D and games, "repeat" is the common default, whereas for scienctific applications "clamp" is more appropriate.

Maybe `fastplotlib` and co should just use colormaps using:
```py
mesh.map = gfx.TextureMap(some_texture, wrap="clamp")
```

### DIFF
--- a/pygfx/materials/_background.py
+++ b/pygfx/materials/_background.py
@@ -139,7 +139,7 @@ class BackgroundImageMaterial(BackgroundMaterial):
     def map(self, map):
         assert_type("map", map, None, Texture, TextureMap)
         if isinstance(map, Texture):
-            map = TextureMap(map)
+            map = TextureMap(map, filter="linear", wrap="clamp")
         self._store.map = map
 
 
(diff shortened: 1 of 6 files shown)


almarklein:
Oh, maybe a better way would be to turn the default colormaps in `pygfx.cm` into `TextureMap` objects with `wrap='clamp'`!

Issue #866, opened 2024-10-16
Grid bugs 2.0


hmaarrfk:
So I know we worked on a few grid shader bugs in the past, but I reinstalled pygfx on python 3.12 on my M1

and just with panning, I can create an effect where the grid just "disappears" along 1 axis.

![image](https://github.com/user-attachments/assets/02c8295c-b124-4db5-b8c6-8f8f50d3e2cd)

Pretty quickly, after panning, I can create flicker effects:
On pygfx 0.3.0
https://vimeo.com/1020267512

Then again on the main branch (v0.5.0-5-ga3f8723) mov file, sorry....
https://github.com/user-attachments/assets/29a1132d-033b-4c1b-a04b-22b2dcee087b

then eventually i find the magical position where things fail

hmaarrfk:
For fear of derailing your thoughts on this one, as I've been writing my own shaders, I think that we need to take into account the "w" position of the camera.

For example in 
https://github.com/pygfx/pygfx/pull/776/files#diff-27f45e757d50e7970d72faa3ee914220e5a854ef152c8026d94c66becf67356eR238
i feel like we aren't taking into account the "w" component in the xyz which will make things strange when you are "way too zoomed in" or "way too zoomed out".


I've seen this crop up when I tried to build my own fragment shader to keep text "sticky" to the screen in
https://github.com/hmaarrfk/pygfx/pull/8/files#diff-27f45e757d50e7970d72faa3ee914220e5a854ef152c8026d94c66becf67356eR106


Just an idea...

hmaarrfk:
To recreate the issue one can use 

ˋ`ˋ
camera.set_state(dict(position=(2000, 6, 0)))
ˋ`ˋ

In the line plot code

hmaarrfk:
I’m now somewhat convinced that that it is the ruler object.

hmaarrfk:
Adding a print statement https://github.com/pygfx/pygfx/blob/main/examples/feature_demo/line_plot.py#L92 shows that the major steps, computed from the ruler object, can actually be 0 in certain causes causing the reported visual artifacts.

PR #984, opened 2025-02-17
No need to divide by the length of a unit vector in grid

hmaarrfk:
Micro optimization, but I think this is helpful. Maybe.... I hope compilers are one day able to do this kind of trivial stuff.


hmaarrfk:
This was found when I was trying to debug:
https://github.com/pygfx/pygfx/issues/866

Korijn (approved):

PR #987, opened 2025-02-18
Colormap clamp

### LINKED ISSUE #979: #957 causes issues with colormap clamping on lavapipe on github actions

almarklein:
Closes #979

* [x] Clamps the values after the gamma correction step. In the image/volume shader the values to sample the colormap are always between zero and one, because clim etc.
* [x] Turns the default colormaps into `TextureMap` objects with `wrap="clamp"`.


Korijn (dismissed):

almarklein:
I accidentally added my intended changes to a previous branch. I cleaned up and force-pushed.

Korijn (approved):

Issue #985, opened 2025-02-18
> In pygfx you can independently pick a blending method, and toggle scene sorting. If you enable scene sorting, and select the `ordered2` blender, you'll get the expected result.


panxinmiao:
> In pygfx you can independently pick a blending method, and toggle scene sorting. If you enable scene sorting, and select the `ordered2` blender, you'll get the expected result. 

In the current main branch, the renderer does not distinguish between transparent and opaque objects. Even with sorting enabled, the transparent and opaque objects are sorted together from near to far based on their z-values, resulting in incorrect rendering.

Additionally, I've always had reservations about Blender's approach. Ordered2 Blender renders the scene twice using two passes, discarding transparent and opaque fragments separately, and then combines the results. This method incurs significant performance overhead. (Firstly, it rendered twice, and secondly, the "discard" operation disrupted the Early-Z capability of the GPU.).
The only potential advantage I can see is that the renderer doesn't need to distinguish between transparent and opaque objects or manage their rendering order. However, as mentioned earlier, without proper sorting, transparent objects often result in incorrect alpha blending outcomes.

_Originally posted by @panxinmiao in https://github.com/pygfx/pygfx/pull/974#discussion_r1959240522_

panxinmiao:
The following are the results of testing using the “validate_blend_ordered2.py” example.  Set “blend_mode” to “ordered2” and “ordered1” respectively, and disable and enable sorting.

**Test on the main branch

**With Ordered2 Blender:**

- **When sorting is disabled:**  

![Image](https://github.com/user-attachments/assets/5ed12787-e628-415e-b9d1-3cb8a48fc0a4)

- **When sorting is enabled:**  

![Image](https://github.com/user-attachments/assets/4e91b4f4-7fb0-4829-a5bb-232a0de19f67)

**With Ordered1 Blender:**  

- **When sorting is disabled:**  

![Image](https://github.com/user-attachments/assets/55a985ff-b876-48be-a6c0-9c76c208497d)

- **When sorting is enabled:**  

![Image](https://github.com/user-attachments/assets/b4b16834-48ca-4ba6-9504-9d677a356858)


**All the above results are incorrect. For comparison, the following tests were conducted using the transmission branch.**

Note: We sort transparent objects from far to near by z, Since materials do not currently have a "transparent" flag and are instead determined by opacity, we set opacity to 0.99999 to indicate to the renderer that they are transparent.

```python
[...]

PR #989, opened 2025-02-20
Alpha mode and Transparent

panxinmiao:
This pull request refactors the logic related to rendering transparent objects, which is essentially part of #974.

### Key Changes:

- Material systems now feature a settable `transparent` property. To ensure correct rendering of transparent objects, developers must explicitly set this property to True, which triggers the appropriate transparent rendering workflow in the renderer.

- Added support for alpha_test - an efficient technique primarily used for achieving mask-like effects (e.g., vegetation rendering for grass and foliage).

With these changes, we can better support logic related to transparent objects and these enhancements improve pygfx's compliance with glTF specifications for transparency handling (alpha mode).

The following test uses “gltf_viewer.py" and "ordered1" Blender mode (no need for two renderings).

Test case 1: [**Alpha Blend Mode Test**](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/AlphaBlendModeTest/README.md)

image

Test case 2: [**Compare Alpha Coverage**](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/CompareAlphaCoverage/README.md)

image

Ps: Most of the work for this pull request was completed before #974. However, I considered this constitutes a significant behavioral change in pygfx. To provide a more comprehensive rationale for approval, I opened draft PR #974 in advance to facilitate thorough discussion. This is also the work that the implementation of physics based transparency in #974 relies on.


panxinmiao:
Another issue to pay attention to is whether the renderer defaults to enabling z-sort.
I strongly recommend **enabling it by default**, at least for transparent objects.

At present, I have maintained the behavior of not enabling it by default, so that the behavior of certain transparent examples remains consistent with before. But in fact, the rendering result may be incorrect.

Korijn:
Looks good to me. Curious to hear Almar's thoughts.

Korijn:
Is transparency still properly supported in points, lines and text?

almarklein:
The `material.transparent` looks a lot like `WorldObject.rendermask`, let's use that instead, or unify these api's somehow?

https://github.com/pygfx/pygfx/blob/e08965a0f456770c8d3d9d6eb9640b0d1007f0a7/pygfx/objects/_base.py#L263-L285

panxinmiao:
> The `material.transparent` looks a lot like `WorldObject.rendermask`, let's use that instead, or unify these api's somehow?
> 

I'm not quite sure because theoretically, setting transparent only tells the renderer to process transparent objects according to their flow, and it doesn't necessarily have to be a truly "transparent" object. If it's not set to True, it doesn't necessarily mean it's not a transparent object (such as typical tranimissive objects based on physics). It is an identifier for reference by the renderer and has no direct relationship with the rendering process (shader).

panxinmiao:
> Is transparency still properly supported in points, lines and text?

I think there should be no problem, and the test doesn't seem to have any abnormalities.
The only issue is Stats, which is essentially not a part of the scene and should not be involved in sorting objects in the scene. At present, we have dealt with it separately, but I think that in the future, this type of component should be placed in a separate UI layer and not together with the world objects in the scene.

pygfx/renderers/wgpu/wgsl/mesh.wgsl
```diff
@@ -355,7 +355,11 @@ fn fs_main(varyings: Varyings, @builtin(front_facing) is_front: bool) -> Fragmen
     $$ else
         let physical_albeido = albeido;
     $$ endif
-    let opacity = color_value.a * u_material.opacity;
+    var opacity = color_value.a * u_material.opacity;
+
+    $$ if USE_ALPHA_TEST is defined
```
panxinmiao, line 360: For @almarklein reference:

In this code, theoretically, we could avoid using the USE_ALPHA_TEST macro and instead check if alpha_test is greater than 0 within the shader. If it is, we would perform the alpha test and discard the fragment.

However, we define the USE_ALPHA_TEST macro to enable or disable the alpha test specifically to prevent the discard operation from interfering with Early-Z. When alpha_test is not enabled ( USE_ALPHA_TEST macro not defined), the shader does not have any discard operations. The discard operation is only exist when marco USE_ALPHA_TEST is active, ensuring that Early-Z is disabled solely for this rendering pipeline.

This approach helps maintain the integrity of Early-Z optimizations by ensuring that discard operations, which can disrupt Early-Z, are only present when necessary.

Korijn: @almarklein docs: https://www.khronos.org/opengl/wiki/Early_Fragment_Test

panxinmiao:
> Another issue to pay attention to is whether the renderer defaults to enabling z-sort. I strongly recommend **enabling it by default**, at least for transparent objects.

A major consideration in deciding whether to enable z-sorting by default is the additional performance overhead. Because for each object, z-sorting requires an additional position conversion (from world space to camera space).

I have thought about this issue again today. 
Currently, we use the la.vec_transform() method for calculation, but it seems that directly using numpy's matrix multiplication speed will be much faster. Moreover, most importantly, because the same camera matrix is used, they can be calculated in batches.

I did a test:

```python
import numpy as np
import pylinalg as la
import timeit

v = np.array([1, 2, 3], dtype=np.float32)
v4 = la.vec_homogeneous(v)
m = la.mat_compose(np.array([1, 2, 3]), la.quat_from_euler((1, 2, 3)), np.array([1, 2, 1]))

print(la.vec_transform(v, m))


def apply_matrix(v, m):
    vv = m @ v.T
    return vv.T

print(apply_matrix(v4, m))

v4_batch = np.stack([v4] * 100000)

vo = apply_matrix(v4_batch, m)
print(vo[:5])


print(timeit.timeit(lambda: la.vec_transform(v, m), number=100000))
print(timeit.timeit(lambda: apply_matrix(v4, m), number=100000))
print(timeit.timeit(lambda: apply_matrix(v4_batch, m), number=1))
```

Outpus:
```
[-3.02585975  2.94074763  0.01546533]
[-3.02585975  2.94074763  0.01546533  1.        ]
[[-3.02585975  2.94074763  0.01546533  1.        ]
 [-3.02585975  2.94074763  0.01546533  1.        ]
 [-3.02585975  2.94074763  0.01546533  1.        ]
 [-3.02585975  2.94074763  0.01546533  1.        ]
 [-3.02585975  2.94074763  0.01546533  1.        ]]
0.45319540007039905
0.13281040010042489
0.0017979999538511038
```
If I haven't made any mistakes, there seems to be a lot of room for optimization here, so the performance cost of z-sorting can be almost negligible.

Korijn:
I believe vec_transform also accepts batches.

But I see the major difference is that you are not performing the division operation by w. Is that right?

panxinmiao:
I'm not quite sure, but it seems like there's more to it. The performance gap is still quite noticeable.

Korijn:
> I'm not quite sure, but it seems like there's more to it. The performance gap is still quite noticeable.

I looked into this, see https://github.com/pygfx/pylinalg/pull/102

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -48,14 +49,26 @@
...
-def _get_sort_function(camera: Camera):
+def _get_sort_function(camera: Camera, z=None):
     """Given a scene object, get a function to sort wobject-tuples"""
 
-    def sort_func(wobject: WorldObject):
+    sort_func = attrgetter("render_order")
+
+    # todo: we should use the bounding-sphere center to get a more accurate z.
+    def sort_func_z(wobject: WorldObject):
         z = la.vec_transform(wobject.world.position, camera.camera_matrix)[2]
```
Korijn, line 59: would it be sufficient to use the `view_matrix` instead of the `camera_matrix`? that way you can skip the perspective projection with `projection=False` in https://github.com/pygfx/pylinalg/pull/102

if not, then it should still be a lot faster with the new implementation, even if `projection=True` is needed

panxinmiao: I think it's possible. The z-value in the view space obtained solely through the view matrix is sufficient to accurately reflect the front to back relationship of the object. The transformation of the projection matrix will maintain the relative order of the z-values in the view space .

panxinmiao: I tested it and if I didn't make any mistakes, it seems that using numpy's matrix multiplication directly (the apply_matrix method above) is still much faster...

pygfx/renderers/wgpu/wgsl/mesh.wgsl
almarklein: I was aware that setting the depth in the shader, which we do in some shaders kills early z-testing. Good to know the same applies to using `discard` (even if conditional).

almarklein:
Awesome how discussions like this result in performance boosts 😄 🚀.  Making sorting the default feels better now.

pygfx/renderers/wgpu/engine/renderer.py
Korijn: So the mistake you are making in your comparison is that you have pulled the call to `la.vec_homogeneous` out of your timing test for `apply_matrix`, while `la.vec_transform` includes. It's a more fair comparison to define `apply_matrix` like this:

```py
def apply_matrix(v, m):
    v = la.vec_homogeneous(v)
    vv = m @ v.T
    return vv.T[..., :-1]
```

Anyway, I have also done some timings, and `la.vec_transform` performs in the same order of magnitude as `apply_matrix`. However `apply_matrix` is faster for single vectors, while `la.vec_transform` is faster for batches.

Timings:

```
vec_transform (projection=True) 0.5506380000006175
vec_transform (projection=False) 0.37464429999818094
apply_matrix 0.32120049999502953
vec_transform (batch, projection=True) 0.004962200007867068
vec_transform (batch, projection=False) 0.002444199999445118
apply_matrix (batch) 0.002591600001323968
```

Benchmark script:

```py
import numpy as np
import pylinalg as la
import timeit


def vec_transform(
    vectors, matrix, /, *, w=1, projection=True, out=None, dtype=None
) -> np.ndarray:
    matrix = np.asarray(matrix)

    if projection:
        vectors = la.vec_homogeneous(vectors, w=w, dtype=float)
        vectors @= matrix.T
        vectors[..., :-1] /= vectors[..., -1, None]
        vectors = vectors[..., :-1]
    else:
        vectors = np.asarray(vectors, dtype=float, copy=True)
        vectors @= matrix[:-1, :-1].T
        vectors += matrix[:-1, -1]

    if out is None:
        out = vectors
        if dtype is not None:
            out = out.astype(dtype, copy=False)
    else:
        out[:] = vectors

    return out


def apply_matrix(v, m):
    v = la.vec_homogeneous(v)
    vv = m @ v.T
    return vv.T[..., :-1]


v = np.array([1, 2, 3], dtype=np.float32)
m = la.mat_compose(np.array([1, 2, 3]), la.quat_from_euler((1, 2, 3)), np.array([1, 2, 1]))

print(la.vec_transform(v, m))
print(vec_transform(v, m, projection=False))
print(apply_matrix(v, m))

v_batch = np.stack([v] * 100000)

print(vec_transform(v_batch, m, projection=False)[:5])
print(apply_matrix(v_batch, m)[:5])


print("vec_transform (projection=True)", timeit.timeit(lambda: vec_transform(v, m), number=100000))
print("vec_transform (projection=False)", timeit.timeit(lambda: vec_transform(v, m, projection=False), number=100000))
print("apply_matrix", timeit.timeit(lambda: apply_matrix(v, m), number=100000))
print("vec_transform (batch, projection=True)", timeit.timeit(lambda: vec_transform(v_batch, m), number=1))
print("vec_transform (batch, projection=False)", timeit.timeit(lambda: vec_transform(v_batch, m, projection=False), number=1))
print("apply_matrix (batch)", timeit.timeit(lambda: apply_matrix(v_batch, m), number=1))
```

Korijn: Edit: I didn't notice the `dtype=np.float32` parts you sprinkled in, that also makes a big performance difference. We could tweak `vec_transform` further to also work with `float32` instead of `float64`, it makes it even faster.

Forcing everything to operate with `float32` timings:

```
vec_transform (projection=True) 0.3528847000125097
vec_transform (projection=False) 0.42849849999765866
apply_matrix 0.3166431999998167
vec_transform (batch, projection=True) 0.002431900007650256
vec_transform (batch, projection=False) 0.0015787000011187047
apply_matrix (batch) 0.0023511000035796314
```

almarklein: Switching to float32 will result into precision problems earlier than float64, but I reckon that this is less of a problem if we don't include the projection matrix. Something to keep in mind though.

Korijn: As part of this PR, can you do the following:

* Bump pylinalg minimum version to 0.6.6
* In the sorting code:
  * Pass `projection=False` to `la.vec_transform`
  * Use `camera.view_matrix`

We can look into boosting the performance by batching operations separately.

Korijn:
I want to propose:

* We try to merge this PR as is
* We enable sorting by default and improve its performance in a separate PR

@almarklein how do you feel about that?

almarklein:
This PR highlights multiple issues with the current blending, and possibilities for improvements. This is very valuable indeed! But I also have a few objections to (the current state) of this pr. Let's discuss these.

The **alpha_test** makes sense, no objections.

The  **depth_write** is also a good addition, but I think it perhaps needs a default "auto" value, because transparent objects should by default not write their depth?

I believe the stats object does not need special handing, because it's always rendered in a separate pass anyway.

You make good points about the currently incorrect sorting, and I can see it being enabled by default now.

Also good point to try and make the `discard` conditional to help early-z.

The greater objection that I have: the new `transparent` prop, and categorizing objects based on it sort of adds a new system next to the one based around `render_mask`. The `render_mask` system is already pretty powerful, because the shader can in many detect whether an object is opaque or transparent. And users can also explicitly override it via `object.render_mask`. It feels to me like you're implementing a better version of the `ordered2` blender, but hard-coded in the renderer. I respond in more detail in #985

almarklein:
The approach of this PR assumes `ordered1` and ignores other blend modes. Which makes sense, since you say you want to get rid of the blender. But the blender has a purpose. Let me explain.

I strive to make things Just Work for our users. In the ideal case, a user populates a scene with different types of objects (meshes, lines, volumes, etc.) and everything looks as one would expect. That way scientists can explore data without having to be viz experts. We got a lot of things right in Pygfx, but making things Just Work for transparency is particularly tricky. It always has been.

The blender is my attempt to resolve this, by providing alternative blend modes, like weighted blending. In order to implement these advanced blend modes, they need control over the blending ops and whether or not depth is written. Maybe this helps understand the reason why the blender defines them and not the material.

I admit that the blender is not really successful at making things plug and play in regard to transparency. Users have to be aware of it, and chose (and understand) between the different modes. And even then, the results vary. E.g. weighted blending works well in some cases, but not so much in others.

I can also see how this inhibits the more game-engine-like feature, like specifying additive blending on a specific object.

But I have an idea ...

The blend mode which I find most interesting is the one based on dither (stochastic transparency), since its the only one that always produces correct results, regardless of ordering. And it's a single pass. And it can deal with objects that are partially transparent. Ok, it looks noisy, but we may be able to improve that.

Another big advantage (I realize now), is that it's compatible with classic blending; you can mix objects that use dither and alpha blending (you simply consider the dithered-object as opaque when you sort the objects).

So, a proposal ...

* We get rid of the weighted blending modes and the blender.
* The material gets a `blend_mode` property, by which the user can chose between classic alpha blending, additive, or dithered. And maybe allow specifying the full blend function.
* The shader still examines the color sources and determine whether the object is opaque, transparent, or possibly both (the `render_mask`), and will expose this attribute better.
* The shader can disable the dither code (if in use) when the object is known to be opaque, enabling early-z.
* The detected opaque/transparent-ness is used for sorting the objects by the renderer.
* Users can explicitly state the opaqueness too (as in `object.render_mask` or `material.transparent`).

With this:
* There's always a single "pass" (no double-rendering as ordered2 does for some objects).
* Simpler code.
* Enables gltf compatibility / classic blending.
* Transparency that just works with dithering.

panxinmiao:
> But the blender has a purpose. Let me explain...
> 
> I strive to make things Just Work for our users ...
> 

Thank you for your detailed explanation. Taking this opportunity, I would like to share some of my thoughts here, and I hope it doesn't come across as too presumptuous. 😅 

As a rendering engine, I think Pygfx's core value lies in its high-level encapsulation of wgpu functionalities and rendering pipelines. It abstracts and simplifies wgpu capabilities and rendering processes through Pygfx-specific concepts and data structures.

In the rendering pipeline, we've introduced several key abstractions. The traditional programmable rendering pipeline typically consists of a geometry processing stage (vertex shader), rasterization stage, and pixel processing stage (fragment shader). For the geometry processing stage, we've designed the Geometry class as an abstraction, while for the rasterization and pixel processing stages, we use the Material class. Therefore, our renderable objects (world objects) consist of geometry and material components.

The core module of the engine should automatically handle the mapping between these abstract concepts and their underlying implementations, freeing users from dealing with the low-level details of WGPU API. This includes managing the lifecycle of various GPU objects, assembling rendering pipelines, data transfer, byte alignment, and other complex technical details.

Ideally, when users define vertex attributes in Geometry, these attributes should be automatically available in the vertex shader without additional effort. Similarly, when users define uniform properties or textures in Material, these resources should also be automatically accessible in shaders. Corresponding WGSL structure definitions, GPU object generation, and binding should all be handled automatically by the engine.

In essence, Pygfx users can conveniently define, configure, and implement their rendering pipelines using intuitive concepts like Geometry and Material. And it can cover all or most of the capabilities provided by wgpu. With these features, developing various applications based on Pygfx becomes straightforward, without requiring in-depth knowledge of wgpu's underlying implementation.

This level of functionality primarily targets developers who use Pygfx as a rendering engine.

Additionally, our various built-in Material classes (with accompanying shaders) and objects can be seen as predefined "classic rendering pipelines" based on the core functionalities mentioned above. These components are ready-to-use, and by adjusting parameters, they can meet the needs of most general scenarios and typical tasks. This level of functionality is aimed at more "front-end" users or developers who may not need to deeply understand the details of rendering pipelines.

As for higher-level functionalities targeting specific scenarios (such as scientific computing and data visualization), I believe these are better suited for implementation by libraries built on Pygfx (like fastplotlib). Even if some of these capabilities are to be included in Pygfx, they should be positioned as “core capability demonstrations and applications” rather than framework essentials.

almarklein:
Thanks for sharing your view on Pygfx. It's very helpful. And I mostly agree 😄 

> As for higher-level functionalities targeting specific scenarios (such as scientific computing and data visualization), I believe these are better suited for implementation by libraries built on Pygfx (like fastplotlib). Even if some of these capabilities are to be included in Pygfx, they should be positioned as “core capability demonstrations and applications” rather than framework essentials.

I do think that Pygfx's core framework can include features that make it more suitable for scientific purposes. We've always positioned Pygfx as a scientific-able render engine. 

Which parts of my proposal are you referring to exactly?

As for dithering, making it possible in the core means that a library/users that wants to use it, does not have to subclass the shader for every object they want to support.

As for the shader determining whether the object is opaque or transparent, this is a task that the shader can do relatively easily, but is much harder to implement by higher-level code (because it's shader-specific). It's a relatively small effort from the engine's end that makes the user-experience a lot more friendly.

almarklein:
I'm not sure if it was clear from what I wrote, but I'm basically proposing what you are in this PR and comments with two additions: 1) in addition to the common options of `material.blend_mode`, also enable stochastic transparency; 2) keep the auto-detection of transparent/opaque to the user (usually) does not have to specify it.

panxinmiao:
> Which parts of my proposal are you referring to exactly?

The comments above are not directed towards a specific proposal you have put forward, just taking this opportunity to express my thoughts, 😅 

My idea is that we should expose and map more comprehensive wgpu capabilities through basic abstract concepts such as Material, rather than hiding them (such as depth and blend related configurations), in order to provide maximum flexibility. Ideally, for all configurable and programmable places in wgpu's programmable rendering pipeline, we should provide corresponding APIs through this level of conceptual abstraction (Geometry and Material, etc.), rather than embedding some fixed pattern.

almarklein:
This is superseded by #1002. The discussions in this PR initiated that work 😉  @panxinmiao I think the main thing to salvage here is the addition in `load_gltf.py`. Could you submit that in a new PR?

Issue #930, opened 2025-01-10
Issues with Color Mode


panxinmiao:
I’ve recently been refactoring the code and logic related to `colormap`, `colorspace`, `vertex color`, and other related components.

Currently, our Materials object has a `color_mode` property, which can take values such as `uniform`, `vertex`, `face`, `vertex_map`, `face_map`, `auto`, and `debug`. These are enumerations (parallel relationship) that determine the method of obtaining an object’s base color.

Here, there are a few issues:

In Mesh rendering, according to the glTF specification (and generally in 3D rendering practice), "uniform color", "vertex color", and "colormap" are not separate, independent rendering modes; instead, they collectively influence the final diffuse color. In fact, "uniform color" and "vertex color" are factors for the colormap.

The diffuse color equation would be:
`diffuse_color = base_color_factor * vertex_color_factor? * sample(colormap)?`

Where `base_color_factor` is the `uniform_color (material.color)`, which defaults to (1,1,1).
When vertex data includes color, `vertex_color_factor` is present and takes effect, multiplying with "base_color". (Some engines might also have an additional boolean variable, like `use_vertex_colors`, to control whether vertex color is used when it exists.)
When a colormap exists, it is sampled and multiplied by "color_factor".
All color factors should be in linear space.

So I think they should not exist side by side as a "color mode".

Additionally, we've also allowed for `face` and `face_map` modes, which feels a bit odd. Typically, geometry stores vertex attributes (per-vertex data), which are the inputs to the vertex shader. In our abstraction model and rendering pipeline, there shouldn’t be per-face data. Per-face data in the shader should come from the interpolation of vertex data. In practice, there seem to be very few use cases where per-face data is provided externally. If the goal is to give each face of the mesh a unique color for a "flat" shading effect, WebGPU actually provides a flat interpolation mode for vertex data (options are `perspective`, `linear`, `flat`, with the default being `perspective`).

almarklein:
That's an interesting point that you are raising. I've always assumed that the color came from one source. I think the simplest solution is to introduce another value to the enum that uses uniform + vertex + colormap, and implement that (at least) for the mesh shader.

As for `face` and `face_map`: engines that are based on vertex buffers have no access to the face index, so supporting per-face colors would be difficult. A side-effect of our approach using storage buffers is that we do have the face index, and supporting per-face colors is therefore easy. I agree that face-colors are less common, but there are certainly (scientific?) use-cases where a user has per-face data. I remember that I did at some point in my research.

[...]

PR #991, opened 2025-02-22
Use correct diffuse color

panxinmiao:
Fixed the logic of diffuse color, retained color_mode, compatible with previous behavior. By default, when color_mode is auto,   diffuse color will use all  "uniform color”, “vertex color”, and “color map".

See: https://github.com/pygfx/pygfx/issues/930 for detail.

Test Case 1:  [Vertex Color Test]( https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/VertexColorTest/README.md)
image

Test Case 2:  [Compare Base Color](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/CompareBaseColor/README.md)
image

_Note:  In this test case, Due to the fact that extension “KHR_texture_transform” has not been implemented yet, there are some differences( The "glTF" logo appears slightly smaller) between the screenshot and the example_


almarklein:
Thanks for this! In the pr's current state, it changes the behavior of colormode quite a bit. I think the existing colormode options should select a single color source, as the do in `main`. And a new value should be added to the enum that will use all color values available and multiply them (IIUC that's the option you want to expose here).

panxinmiao:
> In the pr's current state, it changes the behavior of colormode quite a bit. I think the existing colormode options should select a single color source, as the do in main

Uh, I don't quite understand, but the current color mode does have only one color source, which is consistent with the original behavior. The only difference from before is that when the color mode is not explicitly set (i.e. when using the default auto), Mesh rendering will integrate all colors.
If not, it may be a bug in the logic of the code. 🤔

panxinmiao:
> My initial preference was to maintain the behavior of 'auto' (which selects one color source) and add a new option 'all' that combines them, but I'm also open to change the behavior of auto, but then we need to update its docs (in the above enum).

From my usage perspective, the explicit "color_mode" configuration for mesh rendering appears counterintuitive. Based on typical workflow patterns:

1. Most users will naturally expect all color sources to be active by default when working with materials. Requiring explicit "color_mode" configuration to enable standard color behavior seems unnecessarily complex.

2. Texture Slot Paradigm: The "diffuse_map" operates consistently with other texture slots in the material system. Users instinctively:
    - Add textures to slots when needing specific effects
    - Remove textures from material when not required
  
   This aligns with standard material editing practices across 3D applications.

3. The base color property is universally recognized as a core material attribute, and user usually don't think of disabling it, but rather modifying its RGB value, for example, changing it to white (1, 1, 1).

I feel that color_mode may become a hidden feature, and most users may not pay attention to this attribute unless specific users are aware of color_mode and may use it in certain limited situations.

pygfx/utils/enums.py
```diff
@@ -57,7 +57,7 @@ class EdgeMode(Enum):
 class ColorMode(Enum):
     """The ColorMode enum specifies how an object's color is established."""
 
-    auto = None  #: Use either ``uniform`` and ``vertex_map``, depending on whether ``map`` is set.
+    auto = None  #: Use all the color sources in (material.color, geometry.colors, material.map) if available.
```
almarklein, line 60:
```suggestion
    auto = None  #: Use (multiply) all the color sources in (material.color, geometry.colors, material.map) if available.
```

almarklein:
I understand that some groups of users expect the behavior that you describe. But I also believe there are also users, especially in interactive settings (data explorations) who may have vertex colors and want to switch between a uniform color, vertex colors etc. at runtime.

Anyway, with the current implementation we have the default that you want, but also the option to support the special cases I care about 😄 

Are the CI errors related to the recent changes in pylinalg?

almarklein (approved):

Korijn:
> > Are the CI errors related to the recent changes in pylinalg?
> 
> Yes.. looks like we need to support 3D and higher ndim arrays as well. I will add it to the test suite and release a patch.

Done. :)

Korijn (approved):

PR #992, opened 2025-02-22
Emulate rgb with monochrome textures

hmaarrfk:
I know I said I wouldn't have time to do this, but I'm experimenting in whether or not this can help us improve performance for our shader

Some learnings:

- Forcing this optimization might be annoying for users of colormaps. The colormap sampler will have to be adjusted to "sample" the packed RGB structure. Alternatively, colormaps could just always be specified as RGBA.

- This is somewhat annoying for anybody using textureDimensions

```
    $$ if emulate_rgb
    let size = textureDimensions(texture);
    let sizef = vec2<f32>(f32(size.x / 3), f32(size.y));
    $$ else
    let sizef = vec2<f32>(textureDimensions(texture));
    $$ endif
```

- Relying on the sampler interpolation is no longer valid for RGB.

### Questions I (Mark) needs to answer before requesting external input:

1. [ ] Does this change the "largest" texture that we can send to the GPU? 
2. [ ] How many files need their shaders adjusted for this.

### API questions

- [ ] Do we need to keep the "old" mode alive?


hmaarrfk:
I basically spent most of my day debugging the one shader I enabled this optimization for. 

It is game changer, but it does really introduce peculiarities in how textures are addressed. 

The way this is done, forces every shader that accesses triplets to need to do this. Seems like an over reach.

almarklein:
Thanks for trying tho!

PR #993, opened 2025-02-22
Micro optimize video decoding

hmaarrfk:
I've been trying to understand why our video player is so slow.

The video decodes at about 60-90 fps (it really fluctuates) But the playback is much closer to 10 fps on my laptop.

I used this video_yuv as my benchmark of the "simplest" PyGFX video player I have.

It sent from benchmarking a 3kx2x video from 20fps to 3kx2k video at 70 fps on my machine.

I'm really not sure why, I took the optimization parameters I had from my own source code. I didn't really leave any better comments sadly

### DIFF
--- a/examples/feature_demo/video_yuv.py
+++ b/examples/feature_demo/video_yuv.py
@@ -157,11 +157,15 @@ def video_width_height():
 def benchmark_video_read():
     start_time = time.perf_counter()
     with av.open(FILENAME) as container:
+        stream = container.streams.video[0]
+        stream.thread_type = "AUTO"
         frames_read = 0
-        for frame in container.decode(video=0):
-            frame = frame.reformat(format=FORMAT)
-            frame.to_ndarray()
-            frames_read += 1
+        for packet in container.demux(stream):
+            for frame in stream.decode(packet):
+                if FORMAT != frame.format.name:
+                    frame = frame.reformat(format=FORMAT)
+                frame.to_ndarray()
+                frames_read += 1
 
             if frames_read > 100:
                 break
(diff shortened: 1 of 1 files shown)


examples/feature_demo/video_yuv.py
```diff
@@ -157,11 +157,15 @@ def video_width_height():
 def benchmark_video_read():
     start_time = time.perf_counter()
     with av.open(FILENAME) as container:
+        stream = container.streams.video[0]
+        stream.thread_type = 'AUTO'
```
hmaarrfk, line 161: I'm 90% sure it is due to this line 

```
        stream.thread_type = 'AUTO'
```
since when i comment it, the frame rate goes back to 20.

almarklein (approved):
Thanks!

hmaarrfk:
Illl try to upload my 4K video that I was using as a benchmark some time in the coming weeks

Korijn (approved):

kushalkolar:
If anyone's looking at this issue and curious, yup it is indeed due to the "AUTO" parameter, more info here: https://pyav.org/docs/stable/api/codec.html#av.codec.context.ThreadType

copy pasted relevant part here in case the link ever breaks:

<dt class="sig sig-object py" id="av.codec.context.ThreadType">
<em class="property"><span class="pre">class</span><span class="w"> </span></em><span class="sig-prename descclassname"><span class="pre">av.codec.context.</span></span><span class="sig-name descname"><span class="pre">ThreadType</span></span><a class="headerlink" href="https://pyav.org/docs/stable/api/codec.html#av.codec.context.ThreadType" title="Permalink to this definition"></a></dt>
  <dd><p>Bases: <a class="reference internal" href="https://pyav.org/docs/stable/api/enum.html#av.enum.EnumFlag" title="av.enum.EnumFlag"><code class="xref py py-class docutils literal notranslate"><span class="pre">av.enum.EnumFlag</span></code></a></p>
  <p>Which multithreading methods to use.
  Use of FF_THREAD_FRAME will increase decoding delay by one frame per thread,
  so clients which cannot provide future frames should not use it.</p>
  
  ThreadType Name | Flag Value | Meaning in FFmpeg
  -- | -- | --
  NONE | 0x0 | -
  FRAME | 0x1 | Decode more than one frame at once
  SLICE | 0x2 | Decode more than one part of a single frame at once
  AUTO | 0x3 | Either method.
  
  </dd>

Auto determines the best method (either FRAME or SLICE) depending on the codec.

PR #995, opened 2025-02-23
Generate mipmaps using RenderPipline and linear interpolation downsampling

panxinmiao:
While debugging transmission light sampling textures, I discovered that we cannot correctly generate mipmaps for -srgb format textures. This issue arises and found because our transmission light maps are copied from Blender's color_tex, which is in the `rgba8unorm-srgb` format.

Currently, we generate mipmap textures using compute shaders with advanced (Gaussian) filtering algorithms. However: 

- Compute shaders do not support “-srgb” format textures. 
- Moreover, for standard mipmap usage—such as smooth sampling at varying distances and scaling levels—linear interpolation downsampling is more accurate. (For textures with sizes that are powers of two, this method is equivalent to a box filter, where each pixel in the mipmap uses the average of the four adjacent pixels from the previous mipmap level.) 
- Additionally, performing interpolation filtering and computing in the srgb space for srgb textures is not entirely correct (should in linear space).

Our current mipmap generation is primarily used for IBL (Image-Based Lighting) ambient light calculations. The existing IBL employs an approximation algorithm that uses standard mipmap textures combined with this approximation, rather than PMREM (Pre-Filtered Environment Map). Therefore, this issue is not critical, and Gaussian filtering might even provide visually better results for IBL. However, as we expand support for more advanced rendering techniques, correctly generating standard mipmap textures becomes increasingly important.

Regarding IBL, there's no need to worry. After completing work related to texture formats and HDR support, I will implement a standard PMREM generator and the corresponding IBL algorithms.


almarklein (approved):

pygfx/renderers/wgpu/engine/mipmapsutil.py
```diff
@@ -144,17 +149,15 @@ def get_bind_group(
...
             )
-        dst_view = texture._wgpu_object.create_view(
-            base_mip_level=mip_level,
-            mip_level_count=1,
-            dimension="2d",
-            base_array_layer=base_array_layer,
-        )
+        global mipmap_sampler
+        if mipmap_sampler is None:
+            mipmap_sampler = device.create_sampler(min_filter="linear")
```
Korijn, line 154: are you sure it's okay to have a global cache variable? what if this function is invoked for a different `device`?

panxinmiao: Uh, it will result in an error. However, we have seemingly assumed that only one device instance can exist during program execution (not sure if it's possible to allocate multiple different device instances). Also, it appears that our pipeline cache (and other GPU objects) is global too.

Korijn: @almarklein can you comment on this? I'm not sure

almarklein: Yes, in pygfx we have one single device.

Korijn (approved):

PR #999, opened 2025-02-25
Support texture transform

panxinmiao:
This PR attempts to introduce the texture transform feature.

Related glTF extension: [KHR_texture_transform](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_transform/README.md)

For 3D rendering engines, this is a common feature. Essentially, it is used to optimize the resource utilization of 3D scenes, with the most important being the ability to minimize the number of textures that the GPU must load. Many models package low resolution textures of many objects into a single large texture image set. Then, the region of the result atlas corresponding to each object is defined by vertical and horizontal offsets, as well as the width and height of the region. 
Additionally, there are also some texture animations or special effects that can be achieved through it.

Test Case 1: [Texture Transform Test](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/TextureTransformTest/README.md)
image

Test Case 2: [Texture Transform Multi Test](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/TextureTransformMultiTest/README.md)
image


panxinmiao:
The core workflow has been successfully implemented and is operational, but there are some issues that need to be addressed.

The main issue is that currently, the "map" of MeshMaterial is very special. Our shader contains significant colormap-specific logic that accommodates various texture formats and scenarios through extensive compatibility code (I must admit this has consistently been one of the most challenging aspects when extending texture-related functionality for me). However, for general Mesh rendering, it is just a regular 2D texture map (representing the most basic diffuse color), which is not much different from other textures. For Mesh (surface) rendering, all workflows expect it to be 2D, filter sampling, and used by a 2D UV coordinate. And the essence of texture transform is also to transform the 2D UV coordinates. 

I'm not sure if we have any other considerations regarding the current definition of the colormap concept for Mesh, but I do think that supporting 1D and 3D doesn't make much sense for Mesh. 1D textures can essentially be seen as 2D textures  with single-pixel width/height, while 3D textures should be used for volumetric rendering domains, and irrelevant to surface-based mesh rendering.
﻿
Additionally, we need to address some secondary issues including dependency cycles and propagation update mechanisms.

almarklein:
> I'm not sure if we have any other considerations regarding the current definition of the colormap concept for Mesh, but I do think that supporting 1D and 3D doesn't make much sense for Mesh.

You should think about it as being able to attach some sort of value to each vertex in the, mesh and then visualizing that value trough a colormap. E.g. in my PhD I had a mesh that modeled a stent graft, with an estimate of the force present in the wire:

image

I really like to maintain this feature for normal colormaps. For all the other special maps (environment, etc.) I don't think this is needed. I also don't think we need to support this in combination with this texture transform feature. I.e. When using texture transform, just assume a 2D texture. I hope this opens up some shortcuts for you.

panxinmiao:
> You should think about it as being able to attach some sort of value to each vertex in the, mesh and then visualizing that value trough a colormap.

> When using texture transform, just assume a 2D texture. I hope this opens up some shortcuts for you.

I see. 
I believe we should define a dedicated material for scientific data visualization (perhaps named MeshColormapMaterial?) and separate it from materials used in standard 3D scene rendering. This way, for regular 3D scenes, our map only needs to handle 2D textures, while in scientific data visualization scenarios, we can have more flexibility in customizing material properties and rendering effects. For example, I think this material should not be affected by lighting to ensure that the visualization remains accurate and consistent.

almarklein:
> I believe we should define a dedicated material for scientific data visualization (perhaps named MeshColormapMaterial?)

I can see how that would make things easier, but I think such use-cases should still support Phong lighting. 

Would it make sense to make the `MeshBasicMaterial` simpler and support nd colormaps there? And move advanced stuff like light maps / environment / etc, to the `MeshStandardMaterial`?

panxinmiao:
> I can see how that would make things easier, but I think such use-cases should still support Phong lighting.

Adding Phong lighting support to MeshColorMaterial is straightforward since the lighting code is already well-modularized. I noticed that in our previous examples related to colormap usage, most cases utilized MeshPhongMaterial, though some used MeshBasicMaterial as well. Therefore, we might need to design two materials specifically for scientific visualization—one with lighting and one without.

> Would it make sense to make the `MeshBasicMaterial` simpler and support nd colormaps there? And move advanced stuff like light maps / environment / etc, to the `MeshStandardMaterial`?

The workflow of MeshBasicMaterial is also based on 2D texture mapping. Essentially, it corresponds to a unlit  model for rendering 3D meshes. However, it should not be the parent class of MeshPhongMaterial and MeshStandardMaterial as it is now. Instead, it should be on the same level as them because it follows a distinct shading model with its own unique logic. I plan to reorganize these materials in the future.

almarklein:
One thing I found a bit odd is that `MeshBasicMaterial` is an unlit model, but has support for various lighting features like reflection and environment maps. Do you mean to refactor that too?

About the nd colormaps, I think if this functionality is supported via separate material(s), it complicated the material design. Would it not be feasible to support nd colormap support for the basic `.map`. and not for all the others?

panxinmiao:
> One thing I found a bit odd is that `MeshBasicMaterial` is an unlit model, but has support for various lighting features like reflection and environment maps. Do you mean to refactor that too?

Here, "unlit" means unaffected by real-time lighting, and no lighting calculation,  both light_map and env_map are just simply sampled as regular textures.

> About the nd colormaps, I think if this functionality is supported via separate material(s), it complicated the material design. Would it not be feasible to support nd colormap support for the basic `.map`. and not for all the others?

However, colormap does differ logically from standard 2D texture processing. If a separate material is not used, both colormap and regular diffuse color texture share the material’s map property and represent the base color of the Mesh—albeit with different processing logic—which make things much more complexity.

almarklein:
> Here, "unlit" means unaffected by real-time lighting, and no lighting calculation, both light_map and env_map are just simply sampled as regular textures.

Ok, that makes sense, although it still feels rather an advanced type of mesh for a class called `Basic` 😉 

> However, colormap does differ logically from standard 2D texture processing

This is an important point to get clear. I responded in #1008

panxinmiao:
Ready from my end :)

examples/feature_demo/gltf_viewer.py
```diff
@@ -27,7 +27,11 @@
```
hmaarrfk, line 27: You are doing so much work on gltf!

is it worth adding some gltf validation examples?

pygfx/renderers/wgpu/wgsl/mesh.wgsl
```diff
@@ -340,22 +340,22 @@ fn fs_main(varyings: Varyings, @builtin(front_facing) is_front: bool) -> Fragmen
     $$ endif
 
 
+    {$ include 'pygfx.uv_fragment.wgsl' $}
```
hmaarrfk, line 343: I'm curious on the design choice here. Why use directly inlined code instead of letting the compiler handle this for you?

These should be trivial optimizations for it to unfold no?

panxinmiao: The main purpose is to make the code more modular.

examples/feature_demo/gltf_viewer.py
panxinmiao: I have an open mind on this. Currently, pygfx's support for advanced rendering features in the glTF specification is validated by comparing it against the model asset library and test cases provided by the Khronos Group. Additionally, each major advanced rendering feature has its own dedicated example. The screen validation examples in pygfx primarily serve to prevent unintended modifications to rendering effects during the development of other features. However, in the current development process, most changes to rendering effects are intentional.

pygfx/renderers/wgpu/wgsl/mesh.wgsl
```diff
@@ -340,22 +340,22 @@ fn fs_main(varyings: Varyings, @builtin(front_facing) is_front: bool) -> Fragmen
...
+
     $$ if color_mode == 'normal'
         var diffuse_color = vec4<f32>((normalize(surface_normal) * 0.5 + 0.5), 1.0);
     $$ else
         // to support color modes
         var diffuse_color = u_material.color;
-        $$ if colorspace == 'srgb'
-            diffuse_color = vec4f(srgb2physical(diffuse_color.rgb), diffuse_color.a);
-        $$ endif
+        diffuse_color = vec4f(srgb2physical(diffuse_color.rgb), diffuse_color.a);
```
almarklein, line 350: What's the purpose of this change?

pygfx/renderers/wgpu/shaders/meshshader.py
```diff
@@ -105,6 +105,7 @@ def _define_texture_map(self, geometry, map, name, view_dim="2d", check=True):
         bindings = [
             Binding(f"s_{name}", "sampler/filtering", sampler, "FRAGMENT"),
             Binding(f"t_{name}", "texture/auto", view, "FRAGMENT"),
+            Binding(f"u_{name}", "buffer/uniform", map.uniform_buffer, "FRAGMENT"),
```
almarklein, line 108: I wonder if this should only be done when the transform is not the identity. Otherwise cases that don't use texture transforms have a lot more buffers and lookup in these buffers. What do you think?

pygfx/resources/_texturemap.py
```diff
@@ -117,3 +133,90 @@ def wrap_t(self):
     @wrap_t.setter
     def wrap_t(self, value):
         self._store.wrap_t = value
+
+    @property
+    def offset(self):
+        """The offset of the texture map.
```
almarklein, line 139: Can we add a note in the docstring of TextureMap or in the docstring of these new props, that they currently only apply to Mesh objects?

pygfx/renderers/wgpu/wgsl/mesh.wgsl
panxinmiao: Uniform color is `gfx.Color`, not related to the colorspace of TexturemMap, and `gfx.Color` is all in the srgb space now.

pygfx/renderers/wgpu/shaders/meshshader.py
panxinmiao: I've considered this approach, but I believe it would be more cumbersome.

Firstly, we would need to perform identity matrix checks on texture transforms for every frame. 
More critically, when dealing with materials that are otherwise identical, the mere presence or absence of texture transforms could create 2^N distinct shader variants (where N represents the number of texture properties). This combinatorial explosion would lead to multiple rendering pipeline variations, potentially increasing pipeline state switches during scene rendering.

Imagine a scene with several objects using the same material, which has N texture properties. this could potentially require up to 2^N different rendering pipelines. Frequent switching between these pipeline states during rendering operations would significantly increase draw call overhead and degrade rendering performance.

Therefore, I think it's best to handle them in a unified and straightforward manner.

pygfx/resources/_texturemap.py
```diff
@@ -117,3 +133,90 @@ def wrap_t(self):
...
+            self._scale = scale
+
+        if rotation is not None:
+            self._rotation = rotation
+
+        self.update_matrix()
+
+    # Currently, we let users to manage the matrix update timing themselves.
+    # todo: auto update matrix when offset, scale, rotation are changed before the next draw call.
+    def update_matrix(self):
```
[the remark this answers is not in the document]
panxinmiao: I think both approaches are viable. The main concern is the potential addition of texture animations in the future, which could involve modifying multiple transform properties of a texture multiple times within the same frame, leading to unnecessary computations.

Currently, when texture animations are not in use, a texture's transform is typically assigned manually during loading or initialization. Allowing users to manage the matrix themselves wouldn’t introduce much burden and would actually provide more flexibility.

In fact, this consideration extends beyond just the texture transform buffer. I'm also exploring better ways to establish a unified per-frame update mechanism for similar types of buffers.

pygfx/resources/_texturemap.py
panxinmiao: Theoretically, this should be a property of TextureMap and should apply to all 2D textures. However, at the moment, it only affects Mesh. We can document this limitation accordingly.

almarklein:
I have two concerns. One is that with this change, every created `TextureMap` creates an additional buffer, which adds overhead, because e.g. it also gets synced. The other is that the API added to `TextureMap` assumes that the texture is 2D.

How do you feel about adding a subclass `class TextureMapWithTransform(TextureMap)` (or a better name).

pygfx/resources/_texturemap.py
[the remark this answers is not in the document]
almarklein: Makes sense.

panxinmiao:
> I have two concerns. One is that with this change, every created `TextureMap` creates an additional buffer, which adds overhead, because e.g. it also gets synced. 

The primary reason for adopting this approach is that creating a separate uniform buffer for each 2D texture is relatively easy to implement from an engineering perspective, 😅.
I previously considered placing it within the material's uniform buffer. However, since most textures are optional properties of a material, we currently don’t have a straightforward way to dynamically modify the structure of the material uniform buffer. As a result, we would have to preallocate transforms for all possible textures in advance. 
Moreover, the material's uniform buffer is currently updated by setting material properties (rather than bcollected uniformly before rendering). This makes it difficult to modify the material's uniform buffer directly from the TextureMap.


> How do you feel about adding a subclass `class TextureMapWithTransform(TextureMap)` (or a better name).

I don’t think it’s necessary to create a separate subclass for this, , as it seems a bit verbose in use.. Instead, we can simply clarify in the documentation that the texture transform applies only to 2D textures (and textures in UV mapping mode). 
In fact, for other types of textures (such as 3D textures or cube Textures), users are unlikely to be concerned with these transformation-related APIs. Also, Users who notice these APIs generally realized that they are used for 2D UV transformations.

If we want to design subclasses for TextureMap (or Texture), I think a better approach is to divide them into UVTexture (2d texture), CubeTexture, Data3DTexture, and so on, because they have significant differences in both purpose and usage. But that's another topic.

panxinmiao:
Additionally, adding a separate uniform buffer for textures could help extend other texture-related functionalities.

I have considered passing a boolean value via the texture uniform buffer to indicate whether the texture requires a color space conversion in the shader (i.e., whether its colorspace is "srgb"). Another approach is to use template macro variables, generating multiple rendering pipeline variants based on the texture’s colorspace properties..
I am currently hesitating  between the two options.  😅

almarklein:
> I am currently hesitating between the two options  

In general, stuff that may change at runtime can/should be in a uniform. Otherwise via a template var. Whether a texture requires an srgb transform feels like the latter to me 😄


>> How do you feel about adding a subclass class TextureMapWithTransform(TextureMap) (or a better name).
>
> I don’t think it’s necessary to create a separate subclass for this, , as it seems a bit verbose in use.. Instead, we can simply clarify in the documentation that the texture transform applies only to 2D textures (and textures in UV mapping mode).
In fact, for other types of textures (such as 3D textures or cube Textures), users are unlikely to be concerned with these transformation-related APIs. Also, Users who notice these APIs generally realized that they are used for 2D UV transformations.

What do you mean with "verbose in use"? 

My two concerns still stand: This creates quite a lot of extra buffer objects, which are only used in a rather specific use-case. An the added API is rather larger although it only applies to a subset of usages of the `TextureMap`.

I think creating a separate class would solve this. Another possible route is to make this work for 1D (and eventually 3D) textures as well, and to not create the buffer unless a transform is actually used. I actually quite like this idea.

* Make sure that the transform props also work for 1D textures. I think we can later probably make it work for 3D as well, but let's start with 1D.
* The matrix calculation should triage between 1D and 2D. I.e. fit a 2x2 or 3x3 matrix in the 3x4 buffer.
* Start out with the buffer being None, and create it in `update_matrix` if its None.
* Add a prop `has_buffer` that stores its value on `self._store`. Use this in the `meshshader.py` to determine whether to add the buffer binding. That way when the transform gets set later, the shader will automatically update.

panxinmiao:
> This creates quite a lot of extra buffer objects, which are only used in a rather specific use-case. 

My concern is whether texture transform is used to textures, which is a very small difference for shaders. Is it necessary to generate different shader programs (and different rendering pipelines) just for this small difference.
For the case where texture transform is not used, in reality, only the identity matrix needs to be passed. Although a GPU buffer will be generated, it should be static,  and if not updated, there should be no additional performance overhead here. The only extra overhead is the addition of a 3x3 identity matrix multiplication operation in the shader, which is very, very lightweight, and may be much smaller than the overhead caused by switching between different shader program pipelines during rendering.

For such minor differences, in practice, in most cases, a unified shader (consistent shader logic, consistent GPU resource binding group) should be chosen, so that the GPU does not have to frequently switch rendering pipelines and can also perform optimizations such as merging Drall Calls.

The concern about different texture space formats is also the same. If different objects in a scene use different color spaces for their material textures, just this small difference can cause these two objects to use different rendering pipelines, resulting in unnecessary pipeline state switching costs during rendering, which is not worth it.

panxinmiao:
> What do you mean with "verbose in use"?

What I mean is that it's a bit unnecessary and strange to design a separate class based on whether texture transform is supported. All 2D UV textures should support texture transform. 
If we want to classify Texture, a better approach is to define it according to its purpose and usage, such as UVTexture (which definitely supports texture transform APIs),  CubeTexture, Data3DTexture, DataArrayTexture and so on.

panxinmiao:
I think maybe we can add a class UVTextureMap, which represents the standard 2D texture mapping used for Mesh. We can only  support texture transform for this UVTextureMap, and we can also specify that the color space of this class can only be `tex-sgb` or `physical`.

almarklein:
> so that the GPU does not have to frequently switch rendering pipelines and can also perform optimizations such as merging Draw Calls.

Thanks for explaining this. I was not aware that multiple object with the same pipeline can be optimized, makes sense though.

That said, I imagine that if you load multiple mesh objects, you do so using the same code, e.g. the gltf loader, and they all will either have or not have a transform on their textures.

The performance penalty I'm more worried about is the initialization. If the transform does not change, indeed no sync is needed, but it does need an initial sync.

> I think maybe we can add a class UVTextureMap, which represents the standard 2D texture UV mapping mode used for Mesh. We can only support texture transform for this UVTextureMap, and we can also specify that the color space of this class can only be tex-sgb or physical.

That also sounds good. I like how that also supports the srgb/physical feature witout affecting the simpler `TextureMap`  objects.

almarklein:
@panxinmiao I completely agree. This crossed my mind earlier and I think this approach is a really good compromise for (partially) supporting 3D textures, while avoiding complexity.

panxinmiao:
@almarklein 
I'm planning to implement this PR based on the consensus we reached earlier, enabling UV transforms for 1D, 2D, and 3D textures.

I revisited the handling of colormap. Currently, our treatment of colormap is essentially the same as for regular textures, with one exception: in the `sample_colormap` method, we allow i32 and u32 texture formats.

https://github.com/pygfx/pygfx/blob/c6436d8f0a0f112d8a9cfe58ecb6041fd40b9c00/pygfx/renderers/wgpu/wgsl/colormap.wgsl#L10-L34

However, this support doesn't seem meaningful for colormaps.  🤔 
In our code, these textures are sampled as integers and then converted to f32. But on the GPU, color values are typically represented as floating-point numbers in the range [0, 1]. Integer textures can only take on discrete values like 0 or 1, which makes them ineffective for representing continuous color information.

In my opinion, integer textures are mainly intended for storing and transmitting custom data, and are not suitable for use as color maps.

almarklein:
Well, the vast majority of colormaps / texturemaps are simply uint8. The 0-255 value is simply converted to 0..1 upon sampling the value. I think integer colormaps can still be useful, e.g. for 16 bit integer data (common for e.g. CT data).

The values that you then get are indeed way bigger then. The purpose of `clim` is to bring that back to 0..1.

Korijn:
@almarklein what do you think about this now?

pygfx/renderers/wgpu/wgsl/uv_fragment.wgsl
```diff
@@ -0,0 +1,77 @@
+// todo: now, to prevent byte align issue, we use mat3x4f as map_transform, but it should be mat3x3f.
+// We should avoid do conversion in shader, to improve performance and simplify code.
```
almarklein, line 2: I'm looking into this now. I expect a pr tomorrow.

pygfx/renderers/wgpu/shaders/meshshader.py
```diff
@@ -108,6 +108,12 @@ def _define_texture_map(self, geometry, map, name, view_dim="2d", check=True):
             Binding(f"t_{name}", "texture/auto", view, "FRAGMENT"),
         ]
 
+        # todo: Only "UVMapping mode" have "transform" and "uv_channel"
+        # todo: introduce "MappingMode" in TextureMap
```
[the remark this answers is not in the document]
panxinmiao: I plan to introduce a `MappingMode` property in `TextureMap` to specify the mapping method of the texture. This property is an enumeration, with possible values including `UVMapping` (the default and most commonly used), `CubeReflectionMapping`, `CubeRefractionMapping`, and `EquirectangularMapping`, among others. The `uv_transform` and `uv_channel` settings are only meaningful when the texture uses the `UVMapping` mode.

pygfx/resources/_texturemap.py
```diff
@@ -117,3 +133,90 @@ def wrap_t(self):
...
+        """
+        return self._rotation
+
+    @rotation.setter
+    def rotation(self, value):
+        self._rotation = value
+
+    def update_matrix_components(self, offset=None, scale=None, rotation=None):
+        """Update the texture map's matrix with the given offset, scale and rotation.
+
```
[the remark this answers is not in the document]
panxinmiao: This method is designed to automatically update the matrix by setting all components at once.

pygfx/renderers/wgpu/shaders/meshshader.py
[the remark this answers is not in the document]
almarklein: Do you intend to do that in this pr or do you want to get this merged first and do that in a next pr?

pygfx/resources/_texturemap.py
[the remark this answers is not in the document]
almarklein: Sorry! I thought I checked wheter it called `update_matrix()` but it very obviously does 😅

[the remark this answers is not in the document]
almarklein: Would it solve the whole issue with manually calling `update_matrix()` if we remove the properties, and only have this function?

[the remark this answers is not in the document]
panxinmiao: This method is only useful when setting all components at once. 
However, in some scenarios, different properties may be updated separately multiple times within a single frame—for example, potential future support for texture animations. Therefore, I think keeping the individual property setters along with an explicit update_matrix method provides greater flexibility.

pygfx/renderers/wgpu/shaders/meshshader.py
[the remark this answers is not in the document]
panxinmiao: I plan to implement this in a separate PR when we add support for `EquirectangularMapping`.

Korijn (approved):

Issue #905, opened 2024-12-16
Inconsistency in text rendering on the CI


hmaarrfk:
I had to think of this again when the image comparison failed in #902. I had an idea what could explain the different results. Writing this down so we can look into it another time.

I think what's going on is that when the sdf is sampled, it is not fully contained within that sdf's patch. Either pixels from the neighbouring glyph affect the sampling, or a different result may happen depending on whether the glyph is at the edge of the atlas or not.

This could explain why certain glyphs look slightly different depending on whether other tests were run first; the atlas will be pre-populated differently, so the glyph ends up in a different slot.

Would be good to implememt a proper test for this, and/or limit texcoords when a glyph is sampled (similar as for the u and v plane for yuv420p in #873).

_Originally posted by @almarklein in https://github.com/pygfx/pygfx/issues/824#issuecomment-2545375636_

hmaarrfk:
if the atlas sampling is the problem i'm wondering if it is "as simple as" padding things with an extra pixel of "-1" or the appropriate value.

almarklein:
<img width="650" alt="Image" src="https://github.com/user-attachments/assets/695e1468-f56a-46a1-9dcc-fe8a9fa89c39" />

I think what happens (but have not actually checked) is that we sample the atlas glyph from the edge of the first/last pixel. But we the center of that pixel should be the last point, otherwise you also interpolate from the pixel in the neighboring atlas slot.

Korijn:
So pad with half a pixel. :) I've seen this half pixel bug in so many different places, haha. It's the difference between index space and pixel space. 0,0,0 is the corner of the first pixel, not it's center.

almarklein:
Mind that what I said above is based on an assumption, I have not taken the time to properly look into this yet.

hmaarrfk:
#1001 adds two pixel padding. One on the left, the other on the right side of the atlas. I don’t really like my implementation though. 

I think the same text results can be obtained with adding 1 pixel padding (only on the left), though this complicates how one deals with the very edge of the atlas. 

🤔Maybe Korins idea of half a pixel padding is the same of having 1 pixel padding on « left and bottom ».

almarklein:
Perhaps we can also solve this in the shader by clamping the texcoords to `[0.5, slot_size - 1.5]`.

hmaarrfk:
> Perhaps we can also solve this in the shader by clamping the texcoords to [0.5, slot_size - 1.5].

You know I love to pixel peep. 

[...]

PR #1001, opened 2025-02-25
Glyph atlas, pad with 1 pixel

hmaarrfk:
I feel like I'm not padding exactly right.
It still seems too close to the "top and left edge"

I've been testing by watching the example

```
python text_waterfall.py
```

an attempt to close https://github.com/pygfx/pygfx/issues/905


hmaarrfk:
I believe that the fact that the example tests are failing just now is corroborative evidence that we are sampling and interpolating with the background of the atlas.

I think one would need to rework the logic, but I think this should be correct.... I wonder how we can validate that this is in fact the cause of the error.....

pygfx/utils/text/_atlas.py
```diff
@@ -231,18 +255,26 @@ def _set_new_glyphs_array(self, size):
...
             w1, h1 = info["size"]
-            if not w1 or not h1:
-                continue  # freed region
+            # An index can have 0 size, for example the
+            # character 32, a non-printable character.
+            # has a glyph that is empty, but it should still
+            # take space and have an index in our atlas.
+            if index in self._free_indices:
+                # freed region
+                continue
```
hmaarrfk, line 264: On the main branch, If you add a print statement when width = 0 you can see a few characters are allocated in the waterfall demo.....

Those seem to be various whitespace, one of them 32 (0x20)

pygfx/utils/text/_atlas.py
```diff
@@ -231,18 +255,26 @@ def _set_new_glyphs_array(self, size):
...
 
-        assert allocated_area == self._allocated_area
+        assert allocated_area == self._allocated_area, (
+            f"{allocated_area} != {self._allocated_area}"
+        )
 
         self._free_area = 0
         self._allocated_area = allocated_area
-        self._free_indices.clear()
+        # self._free_indices.clear()
```
hmaarrfk, line 277: I actually think this used to be a bug, where _free_indices was never cleared. They were always just "there". and available for the taking since the `self._info` array isn't repacked.

pygfx/utils/text/_atlas.py
```diff
@@ -211,15 +226,24 @@ def _set_new_glyphs_array(self, size):
         """
         assert size > 0
 
+        # Set the fill value to 255 to debug the text. it should create
+        # slight artifacts around the text, but it should be very hard to see.
+        # However our tests should be able to see it
+        fill_value = 255
```
hmaarrfk, line 232: actually, if you watch the waterfall demo with the with this, you can see a faint "grey" or light blue color appear around some of the characters. it is most prominent around characters in the graveyard.
Screenshot from 2025-02-26 18-06-50

image
image
image

pygfx/utils/text/_atlas.py
hmaarrfk:
```suggestion
```
if you agree, you can press commit.

tests/utils/test_text_atlas.py
```diff
@@ -58,35 +60,36 @@ def test_atlas_resize():
     prev_array = atlas._array
 
     # Allocate 4 regions
-    for _i in range(4):
-        atlas.allocate_region(8, 8)
+    for i in range(4):
+        index = atlas.allocate_region(7, 7)
+        atlas.set_region(index, i + 1)
```
hmaarrfk, line 65: The reason you want to wrap the sampling is that in this case:
```
ipdb> self._array
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 2, 2, 2, 2, 2],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4],
       [0, 3, 3, 3, 3, 3, 3, 3, 0, 4, 4, 4, 4, 4, 4, 4]], dtype=uint8)
```

you want the samples at the very edge of the "4" to sample the "0" on the other side.

pygfx/renderers/wgpu/shaders/textshader.py
```diff
@@ -42,7 +42,11 @@ def get_bindings(self, wobject, shared):
         ]
 
         tex = shared.glyph_atlas_texture
-        sampler = GfxSampler("linear", "clamp")
+        # We use a repeat since we have a half pixel padding on each glyph
+        # however, we only have half on the '0' edge of the x and y dimensions
+        # therefore on the high value edge, we need to 'wrap' back to the beginning
+        # to get the '0' from the 0 edge.
+        sampler = GfxSampler("linear", "repeat")
```
hmaarrfk, line 49: I don’t get it. This should have worked…..

do we provide sampling options in the two dimensions????

hmaarrfk:
@almarklein i'm wondering if:
image

is showing that there is an "off by 1" error in the sampler, causing there to be 2 pixels of boundary. but maybe this is due to SSA?

hmaarrfk:
I wish that the error was from a boundary pixel....

https://github.com/user-attachments/assets/5d2d1cdd-2bf8-4b0f-99bc-ff5d6ba3eb95

hmaarrfk:
I definitely got bored of using gimp for pygfx pixel peeping

https://github.com/user-attachments/assets/f93008c0-be3a-4e46-b5d8-d85a7d0e4859

pygfx/renderers/wgpu/shaders/textshader.py
hmaarrfk: This seems to be working now.

examples/tests/test_examples.py
```diff
@@ -148,7 +148,7 @@ def test_examples_compare(filename, pytestconfig, force_offscreen, mock_time):
     stored_img = iio.imread(screenshot_path)
 
     # assert similarity
-    atol = 2
+    atol = 1
```
hmaarrfk, line 151: While I thought that we could decrease this tolerance, it seems clear to me that some of the errors are mostly introduced due to the rounding of the atlas sampler.

```suggestion
    atol = 2
```

If you agree, you can commit and get this pushed through.

hmaarrfk:
Sadly after this whole exploration, maybe the most useful thing that came out of it is the comparison tool.

I think that part of the challenges is that the sampler will interpolate, and this interpolation will not always be "down to 2 values accurate".

examples/feature_demo/compare_images.py
```diff
@@ -0,0 +1,281 @@
...
+    reference = imageio.imread(args.images[0])
+    image = imageio.imread(args.images[1])
+elif len(args.images) == 0:
+    gfx_examples_screenshot_dir = (
+        Path(__file__).parent.parent / "screenshots"
+    ).resolve()
+    reference = imageio.imread(
+        gfx_examples_screenshot_dir / "validate_blend_weightedplus.png"
+    )
+    image = imageio.imread(gfx_examples_screenshot_dir / "validate_blend_dither.png")
```
hmaarrfk, line 52: I picked these as the two most comparable examples that already exist the validation screenshots.

hmaarrfk:
Todo:

look into using a full character like this

█████ █████

to force a test that looks at the boundary of glyphs

almarklein:
> look into using a full character like this

I'm working on a test that directly modifies the atlas texture and atlas info array to try and reproduce these edge effects.

hmaarrfk:
I’ll let you rebase and stuff. 

I am also half curious if there is a real usecaze for this optimization. 

So the “manual writing” of atlas seems a little contrived to me, but it might be robust.

hmaarrfk:
I was trying to get the atlas to appear on the left, and some text on the right, eventually I gave up and just got the text.

`hello e` is enough to create the bug.... however, even this patch with padding of 1 or 2 or 10... doesn't fix it.

https://github.com/user-attachments/assets/a0b5070a-bcd6-4dfb-bf0b-cf613b753cf2

But this example is so contrived it might not even matter...

examples/tests/test_examples.py
almarklein: Well, we have not updated the reference screenshots yet.

pygfx/utils/text/_atlas.py
almarklein: Nice catch!

Korijn:
Do I understand correctly that the padding in the atlas works just like margin in css? In the sense that if you have multiple blocks with each a 1 pixel margin, the margins overlap, such that there is 1 pixel around the collection of blocks, and 1 pixel in between the blocks (and not 2).

hmaarrfk:
> We can do that in the shader too: https://github.com/pygfx/pygfx/pull/1014

yes. But I don’t know if it ok to resize a glyph from 10X24 pixels to 9.5x21.5 pixels. 

This kind of scaling is not symmetric in both dimensions. 

> Do I understand correctly that the padding in the atlas works just like vertical margin collapse in css? Except of course we apply it also horizontally. In the sense that if you have multiple blocks with each a 1 pixel margin, the margins overlap, such that there is 1 pixel around the collection of blocks, and 1 pixel in between the blocks (and not 2).

yes. I honestly thought this was your concept with the idea of padding by “half a pixel”

examples/tests/test_examples.py
hmaarrfk: I have a few times. Always reverted since it seemed like it wasn’t helping. 

You can use the pixel peeper application to see that the errors are coming from the middle of the letter k and this is not something that is thought to be affected by this PR. At least in theory.

almarklein:
> yes. But I don’t know if it ok to resize a glyph from 10X24 pixels to [...] 9x21 pixels.

There's no resizing, just clamping the pixels when they become too close to the edge.

hmaarrfk:
Ah.... is that is right. But will clamping now introduce new kinds of "stretching"? 

I think I need to read through the free type documentation to learn about this

almarklein:
No, no stretching. It's basically what value we assume to be beyond the bitmap that freetype produced. I assume that freetype produces a bitmap that is large enough so that the value at the edge is zero. Wait ... I can test this 🤔

hmaarrfk:
The clamping strategy would be "cool" if it were "correct".

However, I updated my glyph pixel peeper to actually look at the boundaries of the glyph.
Unfortunately, the last pixel seems "meaningful" and non 0.


https://github.com/user-attachments/assets/c92dd751-e4b8-4152-9244-0eb8e8e6048c

almarklein:
> Unfortunately, the last pixel seems "meaningful" and non 0.

Yeah, this code confirms that:
```py
atlas_indices = text_object.geometry.glyph_data.data["atlas_index"]
from pygfx.utils.text import glyph_atlas as global_atlas

atlas_array = global_atlas._array
for atlas_index in atlas_indices:
    if atlas_index == 0:
        continue  # might be the first char, or an empty slot
    info = global_atlas._infos[atlas_index]
    x, y = info["origin"]
    w, h = info["size"]
    patch = atlas_array[y:y+h,x:x+w]
    edge_patch = patch.copy()
    print(patch.min(), patch.max(), edge_patch[0,:].max(), edge_patch[-1,:].max(), edge_patch[:,0].max(), edge_patch[:,-1].max())
```

A lot of glyphs have nonzero values at the edges. Funny enough, never on the right side.
```
...
0 169 2 0 0 0
0 165 0 0 0 0
0 168 0 0 5 0
0 166 0 1 7 0
0 161 0 0 0 0
0 161 0 0 1 0
0 167 4 0 0 0
0 165 0 0 0 0
0 166 0 9 7 0
0 165 0 0 0 0
0 173 5 3 1 0
...
```

But nothing we can't fix in the shader :)   Actually, if relying only on padding, one row of padding is not enough; on the pixel's edge, you get the mean of that pixels value and the zero from the padding, which is still non-zero.

almarklein:
But thanks for bringing it up! The fix gets rid of artifacts like these, which I previously overlooked:
image

almarklein:
> Are you saying this PR address these artifacts?

No, but #1014 does.

> I think it is still “more correct” to pass in padded glyphs and use WRAP as a trick to “share the padding”,

```                    
          │                 
   pad    │  1st     2nd    
  ┌───────┤───────┌───────┐ 
  │       │       │       │ 
  │   0   │   8   │  20   │ 
  │       │       │       │ 
  └───────┘───────└───────┘ 
  a   b   c   d   e         
          │                 
          │                           
```

* If you use padding, and the leftmost point is c, than interpolating from c-d goes from 4 .. 8, which is still nonzero.
* If you use padding, and the leftmost points is a or b, than you indeed start from zero. But very likely the slope b-d does not match the slope of d-e.
* In #1014,  the leftmost points is a c, but the part c-d is tweaked so it goes from 0 to 8. This slope also does not match the slope of d-e, but is arguably closer.

hmaarrfk:
Sorry, i definitely missed this addition
image

Coding definitely needs more than 4 inch screen.

PR #1002, opened 2025-02-26
Simplify blending

almarklein:
Alt to #989. See #1065

## Intro

This drops our system of per-renderer blend modes, instead using per-object blending. We managed to keep the "weighted" and "dithered" blending modes. The multi-pass "ordered2" and "weighted_plus" are gone. We include examples that demonstrate how these can be mimicked.

The new system:
* allows mixing objects with different blending in the same scene.
* allows more control over blending, by controlling e.g. `depth_write` and other props.
* is more performant.

But also:
* the user may have to be more aware of opaque vs transparent fragments to avoid artifacts.

## Context

You may ask why we have different blend modes in the first place. The short answer is that blending semi-transparent objects is really hard and there is not a single solution that *does the trick*. And then there are some more special types of blending, like additive.

In the previous system we kind of tried to offer a few blend modes that did their best to "fix" blending for the user. The new system accepts that transparency is non-trivial, relies on the user to think about the problem, and it provides the user with more tools to approach the problem.

### Ordered2 blending

The previous default blending was "ordered2", which had the great feature that distinguishing between opaque and transparent fragments happens per-fragment instead of per object. It did this by rendering most objects twice,
discarding transparent fragments in the first pass, and opaque fragments in the second. This solves a range of problems, (but not everything). 

The costs are quite high:
* The code is more complex.
* Objects are rendered twice, which is not good for performance.
* The shader had a conditional `discard`, which prevents the GPU from doing early-z optimizations.

See #1003 for more details.

## API changes

- Removed `WorldObject.render_mask`.
- Removed `Renderer.blend_mode`, use `Material.blending` instead.
- Added `Material.blending`.
- Added `Material.transparent`, default is None (auto). 
- Added `Material.depth_write`, default is None (auto).  
- Added `Material.depth_compare, default is "<".
- Added `Material.alpha_test`.
- Added `Material.alpha_compare`, default is "<".
- Removed `material.color_is_transparent` for `Line`, `Mesh`, `Points`, `Text`. Use `material.color.a < 1` instead.
- Removed `material.edge_color_is_transparent` for `Points`. Use `material.edge_color.a < 1` instead.
- Removed `material.outline_color_is_transparent` for `Text`. Use `material.outline_color.a < 1` instead.

## Other changes

- The `renderer.sort_objects` is True by default.
- The renderer's sorting depends on  opaque vs transparent objects, and the performance is improved.
- Add support for 'multiply' and 'subtractive' blending.
- More control over 'weighted' blending so it's use extends to other use-cases.
- Background objects have a default `render_order` of -1e6, so they are usually drawn first.


almarklein:
I was able to fit in the weighted blending (user sets `material.blending = "weighted"). Yes you can also mix e.g. dithered objects with weighted ones 🚀

examples/feature_demo/image_stitching.py
```diff
@@ -0,0 +1,85 @@
...
+camera = gfx.OrthographicCamera()
+camera.local.scale_y = -1
+camera.show_object(scene1, match_aspect=True, scale=1.05)
+
+controller = gfx.PanZoomController(camera, register_events=renderer)
+
+
+def animate():
+    renderer.render(scene1, camera, flush=False)
+    renderer.render(scene2, camera)
```
hmaarrfk, line 80: This example and really cool since I think it hints at a way to continue to use the weighted blending modes.

I had one comment in discussing with Almar live:

One of the features we do is that the "order of the passes" is something that we actually "designed".

Our preferred order would be:

1. Objects marked as "opaque"
2. Objects marked as "weighted" <-- these are, for all intents and purposes, "opaque" to the ones behind them, but "blended among each other".
3. Then the transparency pass.

The reason to keep things in a single scene is that we make use of the PyGFX graph quite heavily when we create the object "hierchy" to ensure objects are positioned appropriately. Maintaining "two scenes" like this is dooable when you have "3 objects" such as in this example, but it becomes very easy to have them in and out of sync.

In talking with @kingbedjed he mentionned that perhaps we can have a flag that could "order" the objects as well. or at least "group" them together, so that the "groups" can have a "priority"???

examples/feature_demo/transparency1.py
```diff
@@ -34,43 +43,40 @@
...
-            "ordered1",  # 3
-            "ordered2",  # 4
+            None,
+            "no",  # 1
+            "normal",  # 2
+            "additive",  # 3
+            "dither",  # 4
             "weighted",  # 5
-            "weighted_depth",  # 6
-            "weighted_plus",  # 7
```
hmaarrfk, line 67: I think it would be interesting to "rebuild" this with the `image_stitching` ideas.

On clear win already from this new architecture is that the text is always beautiful.

examples/feature_demo/image_stitching.py
Korijn: > In talking with @kingbedjed he mentionned that perhaps we can have a flag that could "order" the objects as well. or at least "group" them together, so that the "groups" can have a "priority"???

* All objects have a `render_order` property
* You can use properties such as `depth_test` on Material to always draw on top

almarklein: The thing is that in the renderer we make categories, one for opaque objects, one for transparent, and one for weighted-blended objects. I imagine Pan will add a category for translucent objects after this. While it is natural that the opaque objects should be rendered first, the order in which the other categories should be rendered is not so obvious. 

So it makes sense to allow users to have control over that. Maybe simply by using `render_order` 🤔 ... the API will need some thought.

almarklein: I changed the object sorting logic, so that `ob.render_order` is leading, so it can always be used to force a certain order. The categories I mentioned earlier are now just a way to help get good ordering by default. I will write up how it works in more detail later.

panxinmiao:
I noticed that [`dual source blending`](https://www.w3.org/TR/webgpu/#dom-gpufeaturename-dual-source-blending) was added to the WebGPU specification as an extension recently. 
IIUC, With this capability, it's possible to implement weighted blending for OIT in a single render pass. As a result, it seems that the `Blender` class is no longer necessary, which helps simplify the renderer.

almarklein:
Thanks for the heads up!

I tried to search, but could not find any reference to implement WBOIT using dual source blending. 

So I figured I ask chatGTP, and this got weird. It firtst told me that one of the main use-cases of dual source blending is WBOIT. It sayd that dual source blending is mentioned in the original paper, event showing a quote (which does not exist). Upon asking further it says it indeed was not referenced in the original paper, but is demonstrated in later posts and examples, it gives links to these, but these all still use a reslove pass. So I point that out and well, eventually it says "Dual-source blending does not eliminate the resolve pass in Weighted Blended Order-Independent Transparency (WBOIT) as originally proposed — and it’s rarely useful for WBOIT specifically". 

This funny and super-annoying at the same time.

> As a result, it seems that the Blender class is no longer necessary, which helps simplify the renderer.

The good news that in this PR there is still a blender, but there is only one. It's just a sub-component of the renderer that handles the render targets and blending.

panxinmiao:
Ah, after reconsidering, I realized that while `dual-source blending` allows the use of a single color attachment—avoiding multi-target rendering—it still seems to require a post-processing pass for color normalization, 😅

examples/feature_demo/blending_ordered2.py
```diff
@@ -0,0 +1,82 @@
...
+            ob.material.alpha_test = 0.999
+            ob.material.depth_write = True
+    renderer.render(scene, camera, flush=False)
+
+    # Draw transparent fragments (alpha < 0.999)
+    for ob in scene.iter():
+        if ob.material:
+            ob.material.alpha_test = -0.999
+            ob.material.depth_write = False
+    renderer.render(scene, camera, flush=False)
```
Korijn, line 74: Instead of updating material properties, it is more common to predefine the materials and swap them out on the objects. That should also have better performance (caching) right?

almarklein: Well spotted! Yeah, I already changed the example to insert duplicate objects in the scene that have different alpha_test and depth_write props.

almarklein:
Ok, this is about ready from my end! Reviews welcome.

Let's do a release before we merge this.

hmaarrfk:
are we now 'suppoed' to be able to pick transparent objects with this refactor? 

if so, should we add a test for this?

almarklein:
> are we now 'suppoed' to be able to pick transparent objects with this refactor?

Yes!

> if so, should we add a test for this?

Yeah, will look into that.

hmaarrfk:
>> if so, should we add a test for this?
>
> Yeah, will look into that.

Thanks that is awesome!

Korijn:
Have you done any benchmarking or profiling to see if the sorting is not negatively impacting the performance?

almarklein:
> Have you done any benchmarking or profiling to see if the sorting is not negatively impacting the performance?

I did a bit, just tried some more:
* using the Sponza scene. I could not observe a difference. I was hoping/expecting it to be more performant because it could prevent overdraw, but I guess the scene is not big enough to see that effect. All the meshes are textured, but I guess the texture is RGB, so in `main` (before this pr)  the shader detects that it does not have transparent fragments, so it did not hit the ordered2-penalty.
* Same for the skinning animation example. Maybe 1-2 fps more, but not sure.
* The `collection_line.py` example performs significantly better (65 vs 50 fps). I think this is because of the dual-pass for transparent fragments.

pygfx/renderers/wgpu/shader/resolve.py
```diff
@@ -351,18 +352,31 @@ def update_lines(self, lines):
...
-                is_return_line = line.lstrip().startswith("return")
-                indent = indent_from_line(line)
-                extra_line = f"{indent}apply_virtual_fields_of_fragment_output({', '.join(args)});"
-                new_lines = [extra_line, line] if is_return_line else [line, extra_line]
-                lines[linenr] = "\n".join(new_lines)
+                line = f"{last_indent}apply_virtual_fields_of_fragment_output({', '.join(args)});"
+                extra_last_lines.append(line)
+
+            if self.uses_alpha_test:
+                line = f"{last_indent}if (select(out.color.a > -u_material.alpha_test, out.color.a < u_material.alpha_test, u_material.alpha_test > 0.0)) {{ discard; }}"
```
panxinmiao, line 370: In what situations would alpha test and color.a be less than 0 🤔 ?

Also, I assume the reason alpha testing is handled in the shader fragment within resolve, rather than using template_vars to inject it directly into WGSL, is to free users from having to deal with alpha testing when writing their own WGSL shaders—is that correct?

However, I think it would be better to handle alpha testing explicitly within the WGSL shader. The logic is quite simple, and it helps make the overall structure clearer. First, whether alpha testing is enabled or used is part of the shading logic itself and should be controlled within the shader. Second, in the current approach, alpha testing is done after all shading calculations and just before outputting the final color. But in fact, it could be performed earlier to avoid unnecessary computations—discard operations should ideally be done as early as possible.

almarklein: > In what situations would alpha test and color.a be less than 0 🤔 ?

A negative value indicates the test is reversed; you can use it to discard fragments with alpha larger than the reference value. 

I'll look into your other comments.

I wonder if the compiler is capable of hoisting the check to an earlier point 🤔

almarklein: I refactored the alpha testing:
* Added `.alpha_compare` to allow fancy comparisons.
* Shaders can/need to explicitly call `do_alpha_test(my_alpha)`, which is a function defined in `std.wgsl`.

almarklein:
I addressed the feedback so far.

almarklein:
Somewhat related, I recently found that ThreeJS also supports dithered blending! It does so via [`Material.alphaHash`](https://threejs.org/docs/?q=materia#api/en/materials/Material.alphaHash).

almarklein:
@Korijn I would like to move this forward.

I plan to do a few quick follow-ups on this PR because objects with mixed opaque/transparent fragments are prone to artifacts (they already were, but ordered2 avoided a subset of problems). In short, my plan is to add support for fxaa and turn off aa on lines/points by default. I have this working as a POC now and it looks great! 

So what I would like to do, step by step:

* Get this PR approved.
* I'll draft a follow-up PR for post-effect passes, so it can receive review asap.
* Once these two are merged I'll add fxaa. I think best is to make it optional but very easy to enable, so we enable it in all but the validation examples.
* Then change the default values for aa on line/points.

Once that is done we have the same quality (in terms of jaggyness), with much better support for different kinds of blending, and less artifacts. 🚀

Korijn (approved):

Korijn:
Sorry for making you wait, I wasn't sure if you had looked into all the feedback yet. Nice work here, it's really a major improvement.

Do you think #989 and #974 are unblocked after this PR? Or not yet?

almarklein:
This implements most of #989. I think only the gltf loading should be added from it.

#974 should be relatively easy to fit in now. The way that the blender now sorts and categorizes objects was inspired by that pr.

PR #1004, opened 2025-02-28
Add "KHR_materials_unlit" support

panxinmiao:
Support [KHR_materials_unlit](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_unlit/README.md) 

This extension defines an unlit shading model for use in glTF 2.0 materials, as an alternative to the Physically Based Rendering (PBR) shading models provided by the core specification.
 It is very useful in situations where resources are limited for mobile devices or stylized materials that do not require lighting (such as "anime" or "hand drawn" appearances).
 
 For Pygfx, it basically corresponds to our MeshBasicMaterial.
 
This example demonstrates the differences in rendering the same model using unlit and standard PBR, with unlit material on the left unaffected by lighting and PBR material on the right affected by lighting.
 
image


panxinmiao:
Currently, in the example, canceling IBL lighting will throw an exception, See #1005.

almarklein:
> Currently, in the example, canceling IBL lighting will throw an exception

Is this a new error, or simply this use-case exposing it? In other words, do you want to merge this, or get #1005 fixed first?

panxinmiao:
Yes, this is just a use case that exposes the issue in #1005.   
This PR is unrelated to #1005. I think it can be merged first.

panxinmiao:
I was planning to fix the bug in #1005 while I was at it,  but I found that it seems to be a bit tricky. 😓

Korijn (approved):

PR #1006, opened 2025-02-28
Text performance tweaks

almarklein:
## Benchmarks

I've written a few benchmarks in https://github.com/pygfx/pygfx-benchmarks/pull/7.

Before new text:
```
               large_static_text (10x) - cpu:  2.58 ms
   changing_one_word_in_big_text (10x) - cpu: 17.85 ms
 changing_one_word_in_small_text (1000x) - cpu:  0.08 ms
           multiple_text_objects (10x) - cpu: 20.05 ms
              multitext_geometry (10x) - cpu:  -
```

After new text (#834):
```
              large_static_text (10x) - cpu:  2.52 ms
  changing_one_word_in_big_text (10x) - cpu:  0.66 ms
changing_one_word_in_small_text (1000x) - cpu:  0.12 ms
          multiple_text_objects (10x) - cpu: 22.16 ms
             multitext_geometry (10x) - cpu:  2.86 ms
```

We can see that rendering a big static text does not make a significant difference.
The extra lookup in the positions buffer does not seem to affect the rendering performance.

Changing a single word in a big text is now *much* more efficient, because of how text blocks and text items are re-used, and re-layout mostly confined to the changed paragraph. And this is with improved layout (e.g. `line_width`).

Updating a short piece of text is measurable slower. This is one of the things I try to improve in this PR.

Using `MultiTextGeometry` is a lot faster than rendering separate text objects, as expected. (Tested with 1000, but the improvement is likely observable with much smaller numbers.)

After this PR:
```
changing_one_word_in_small_text (1000x) - cpu:  0.07 ms
```
A tiny number but (with this number of iters) quite stable, so a 40% speedup.

## What I did

I ran one of the benchmarks in isolation to test the performance as I tweaked the code, and used cprofile to identify bottlenecks: 
```
python -m cProfile -o program.prof my_little_benchmark_script.py && python -m snakeviz program.prof
```

## Tasks

In this PR I focus on optimizing the setting of text. 

* [x] Added an example to render a large piece of text.
* [x] Tweaks for performance related to setting text:
  * [x] tokenizer
  * [x] shaping
  * [x] font selection    
  * [x] some general tweaks
  * [x] text item syncing
  * [x] layout 

Directions for future optimizations:
* Avoid attribute access of geometry, which is expensive due to the custom `__getattr__`.
* Pack per-glyph info into a single buffer.

### DIFF
--- a/examples/feature_demo/test_big.py
+++ b/examples/feature_demo/test_big.py
@@ -1,0 +1,70 @@
+"""
+Big static text
+===============
+
+Render 10 paragraphs of Lorem ipsum.
+"""
+
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
+from rendercanvas.auto import RenderCanvas, loop
+
+import pygfx as gfx
+
+
+canvas = RenderCanvas(
+    size=(800, 600), title="Big static text $fps", update_mode="fastest"
+)
+renderer = gfx.renderers.WgpuRenderer(canvas)
+
+scene = gfx.Scene()
+
+
+scene.add(gfx.Background.from_color("#fff"))
+
+text = """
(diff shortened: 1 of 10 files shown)


Korijn:
Just out of curiosity: are you using a profiler?

almarklein:
Ready from my end.

Korijn:
The docs build is failing, so we can't merge before that's addressed.

almarklein:
I missed that. Fixed.

Korijn:
🏇

PR #1008, opened 2025-03-02
Unified mesh texture map logic and special treatment for colormaps

panxinmiao:
~~Add MeshColormapMaterial, it is usually used to present scientific data (which stores in geometry's texcoords) in a visual way. 
It can be 1D, 2D or 3D, but should match the number of columns in the geometry's texcoords.~~ 

~~Because  some previous colormap examples in pygfx using MeshPhongMaterial, the current MeshColormapMaterial adopts the Phong lighting shading model. However, I feel that it would be better to design it as an unlit model, perhaps with a built-in lighting model like the earliest PhongMaterial.~~

See: https://github.com/pygfx/pygfx/pull/999#issuecomment-2692573984

Unify the logic related to Mesh texture maps and handle colormaps separately.


almarklein:
> However, I feel that it would be better to design it as an unlit model, perhaps with a built-in lighting model like the earliest PhongMaterial.

I definitely think Phong lighting should be included, as it helps a lot to navigate 3D objects, visually estimate curvatures, etc.

Korijn (dismissed):
These changes look good to me (I love the compromise you found here, it's an elegant application of the Material abstraction) but I would really like Almar to review the changes to the renderer. So he can hit merge when he's comfortable with it.

almarklein:
From what I understand, the problem that this pr tries to solve is the difference in handling of a colormap vs a regular texture map. Let's try to establish what that difference is, exactly.

If I look at the `define_colormap()`  and `_define_texture_map()` methods, that represent the two different paths, the handling is quite similar. If the colormap is 2D the result seems to be the same.

https://github.com/pygfx/pygfx/blob/d2eb8e9d47916216a22a8b0c0420e1c2313b7208/pygfx/renderers/wgpu/shader/base.py#L208-L239

https://github.com/pygfx/pygfx/blob/d2eb8e9d47916216a22a8b0c0420e1c2313b7208/pygfx/renderers/wgpu/shaders/meshshader.py#L93-L123

In WGS, the colormap path will  call `sample_colormap()`  instead of wgsl's builtin `textureSample()`. When the texture is 2D and unorm or float, `sample_colormap()` boils down to *almost* the same as `textureSample`. 

The only difference that I can see is this bit:

https://github.com/pygfx/pygfx/blob/d2eb8e9d47916216a22a8b0c0420e1c2313b7208/pygfx/renderers/wgpu/wgsl/colormap.wgsl#L14-L19

Let's call it tight-clamping: it maps the values 0 and 1 to the center-pixel of the end-points of the colormap. See `validate_image_colormap.png`: 
image

Am I missing other differences in the colormap vs texturemap handling?

----

I acknowledge the problem, but TBH I'm not sure if adding material classes is the best solution. My main argument against it is that it adds a mental burden on users, in selecting the right mesh material.

I've thought of other approaches, and the best that I can come up with right now:
* Add a prop to `TextureMap` that hints it to be used as tight-clamping.
* Add a `class ColorMap(TextureMap)`  that hints the meshshader.

In both cases, the colormaps in our `cm.py` are easily adjusted, so things Just Work for most users.

panxinmiao:
> Am I missing other differences in the colormap vs texturemap handling?

The differences in sampling behavior are just one aspect. More importantly, functionalities—including texture transformations—are designed with the expectation that textures are 2D (and often even assume the sampling type is float). For standard 3D mesh rendering, ensuring broad compatibility with various colormap formats remains a burden when developing mesh rendering features. In high-quality scene rendering, the diffuse map is a crucial resource, and shaders expect it to be a conventional texture—one that can generate mipmaps and leverage hardware-accelerated smooth filtering, including filtering between mip levels. Fundamentally, it behaves like any other texture, and WebGPU provides efficient implementations for conventional texture usage (e.g., textureSample, textureSampleLevel, etc.).

As for the cognitive load on users when choosing materials, this is unavoidable. That’s why I have always advocated for a well-structured conceptual layering in pygfx. Scientific visualization is a specific application domain that should be abstracted with dedicated concepts and classes rather than trying to fit within general-purpose concepts. The best approach would be to build a separate library based on pygfx, specifically tailored for scientific visualization, with its own encapsulated concepts. Alternatively, even if it were to be included in pygfx, it should exist as a dedicated submodule.

Additionally, I believe there is some excessive concern here. Users who directly use pygfx for scientific visualization often already exhibit some unconventional usage patterns—for example, there is no well-defined abstraction for visualization data, and instead, it is simply stored in geometry.texcoords, which feels somewhat odd. In my view, asking users to explicitly select a specialized scientific visualization material is far more intuitive than requiring them to manually configure geometry.texcoords (at least from my perspective). Geometry is a fundamental abstraction representing the geometric stage of the rendering pipeline. A better approach would be to design a dedicated class for this specific scenario, with well-defined properties to store and represent the required visualization data.

almarklein:
> The best approach would be to build a separate library based on pygfx, specifically tailored for scientific visualization, with its own encapsulated concepts. Alternatively, even if it were to be included in pygfx, it should exist as a dedicated submodule.

 It's impossible to build good scientific visualizations on a tool that it not tailored to it.

As I've said before, Pygfx has always targeted scientific use-case. These use-cases are equally valid / important as "standard mesh rendering". From my point of view it started more as a scientific visualization tool to which we (mostly you) added standard mesh functionality. This is great, and it also creates some friction, as in this use-case here. I want to try and find an API that works well for both groups of users, without making one more important than the other.

Back to the point of this PR. We want a way for the user to express that the `meshmaterial.map` is either a (classic) texture-map or a (dataviz) color-map. You've now done that by introducing new material classes. I'm trying to see if we can find a solution that avoids that.

panxinmiao:
> As I've said before, Pygfx has always targeted scientific use-case. These use-cases are equally valid / important as "standard mesh rendering". From my point of view it started more as a scientific visualization tool to which we (mostly you) added standard mesh functionality. This is great, and it also creates some friction, as in this use-case here. I want to try and find an API that works well for both groups of users, without making one more important than the other.
> 

I understand that the primary goal of pygfx was initially scientific data visualization. However, I believe it also has the potential to become an outstanding 3D scene rendering engine—perhaps even the best general-purpose 3D rendering engine within the Python ecosystem. In fact, to some extent, I think it already is.

That being said, throughout the development process, I have noticed certain differences in focus and design philosophy between visualization-oriented rendering and general-purpose 3D scene rendering. As more advanced features are introduced, I am somewhat concerned that these differences could lead to inconsistencies and inefficiencies. Therefore, my approach is to properly separate the logic and concepts of both use cases while ensuring modularity, encapsulation, and reusability.

panxinmiao:
> Add a class ColorMap(TextureMap)

I think this approach is feasible. Whether it's the MeshColormapMaterial or the ColorMap class, their primary purpose is to enable the engine to distinguish between scientific data visualization and general-purpose 3D rendering. This distinction allows the engine to apply different processing workflows accordingly.

Korijn:
> I believe it also has the potential to become an outstanding 3D scene rendering engine—perhaps even the best general-purpose 3D rendering engine within the Python ecosystem. In fact, to some extent, I think it already is.

❤️

almarklein:
> I have noticed certain differences in focus and design philosophy between visualization-oriented rendering and general-purpose 3D scene rendering.

I recognize that 😄  It's not only a difference in technical features, but also a somewhat different way to think about things. I'm glad we're talking about this.

> As more advanced features are introduced, I am somewhat concerned that these differences could lead to inconsistencies and inefficiencies. Therefore, my approach is to properly separate the logic and concepts of both use cases while ensuring modularity, encapsulation, and reusability.

I totally agree. Though I want to guard against making the scientific case subordinate to the general-purpose case.

----

Ok, unsepected update: I thought some more about this, and had another look at how Matplotlib does colormaps. Turns out their colormaps are usually 256 values, using nearest-neigbour. They have a util to create a colormap that interpolates between a few colors, but that simply produces a 256-element table.

The feature that we now have in Pygfx, where one can create a colormap with just a few values, was nice, but we now see that it conflicts with how texture maps are usually applied (the whole topic of this pr). I propose we follow mpl's approach instead, thereby reducing this whole issue to the ability to create have multidimensional colormaps. I imagine that this is much easier to handle (e.g. require the map to be 2d when using features like texture transforms.

almarklein:
I think that with #1025, we could do:

```py
map = material.map
map_fmt, map_dim = to_texture_format(map.texture.format), map.texture.dim
is_standard_map = map_dim == 2 and ("norm" in map_fmt or "float" in map_fmt)

if not is_standard_map and have_features_that_need_regular_map:
       raise RuntimeError(...)
```

Does that make sense, Pan?

panxinmiao:
Uh, it seems that the sampling behavior of single channel textures (such as r32float) is inconsistent, but there are no other issues.

panxinmiao:
Usually, diffuse maps are not single channel (or dual channel) textures, so I think it should be fine to process them according to the colormap logic in this case, but I think we need to remember this difference.

Korijn (approved):
I hope we are all satisfied with the outcome? 🙏🏻

almarklein (approved):
I'm happy with the variant!

For a future tweak-PR (I don't want this to hold up this PR),
I think it may be good to change the naming a bit, because in mesh rendering, the `material.map` is often referred to as the color-map (e.g. in ThreeJS). So to make things more explicit and avoid confusion, we can rename:
* `define_colormap()` -> `define_generic_colormap()`
* `sample_colormap()` -> `sample_generic_colormap()`

Where "generic" refers to it being 1D/2D/3D, float/int, and the fact that the logic is shared between mesh/image/line/points.

PR #1010, opened 2025-03-03
Use single buffer to store per-glyph data

almarklein:
This PR changes the text geometry to use a single buffer for per-glyph data, instead of 4 separate buffers. This makes the code cleaner, and easier to add more data if we want later. It's also faster on the CPU side, because only one call to `.update_indices()` instead of 4. And possibly faster on the GPU (could not measure that effect though).

### DIFF
--- a/pygfx/geometries/_text.py
+++ b/pygfx/geometries/_text.py
@@ -53,9 +53,8 @@ WHITESPACE_EXTENTS = {}
 
 GLYPH_DTYPE = np.dtype(
     [
-        ("x", "f4"),
-        ("y", "f4"),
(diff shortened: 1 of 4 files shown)


hmaarrfk:
We have definitely been feeling the effects of multiple calls to send data to the GPU!!!! Thanks for caring about this kind of optimization while everything is fresh in the mind!

Korijn (approved):

hmaarrfk:
I'm pretty sure this broke text_waterfall.py


image

almarklein:
Lol, we found out at the same time. I'm looking into #1001

PR #1013, opened 2025-03-04
Add test to reproduce glyph-bleeding

almarklein:
Submitting as a separate pr, to make sure that it indeed fails on CI. It fails locally. And I confirmed locally that #1001 fixes it.

### DIFF
--- a/tests/utils/test_text_atlas.py
+++ b/tests/utils/test_text_atlas.py
@@ -340,7 +340,9 @@ def test_against_glyph_bleeding():
     # diff_image = diff.astype("u1") * 255
     if diff_count:
         print("Number of pixels that differ:", diff_count)
-    assert diff_count == 0
+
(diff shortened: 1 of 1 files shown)


almarklein:
BTW, I've only found the error to show when using an outline (though it could be that glyphs that are close to an edge are effected too).

almarklein:
CI indeed fails. Closing this pr. I'll just add to to #1001.

hmaarrfk:
Btw. Not sure if you saw that the waterfall demo showed some bleeding between. Glyphs on main

https://github.com/pygfx/pygfx/pull/1001#discussion_r1972562087

you have to be a ninja to catch them!

hmaarrfk:
Oh. Sorry. I had deliberately padded with the worst case of 255 to force them to happen

PR #1014, opened 2025-03-04
Prevent glyphs from bleeding into neighbours in atlas

almarklein:
### ... and a few other improvements to the text shader

Alternative to #1001

* [x] Update the text shader to constrain the texcoord to be within the glyphs own patch. 
* [x] Also deal with the sdf's edge-pixels being nonzero, preventing (more) artifacts in outlines.
* [x] This update also reduces the chance on rarer differences due to lower precision in varyings (from experience I know that the precision of varyings is less than 32 bits).
* [x] Tweak smoothness factor for aa a bit.
* [x] Don't diminish alpha for text that is "unreadably small". 
* [x] Instead render such text as little blobs, and at smaller text sizes.


almarklein:
I saw a call to `clamp()` in the shader that looked wrong. Upon further investigation, it turns out that you can interchange `clamp(x, min_val, max_val)` with `clamp(min_val, x, max_val)` 🤯  I will still update the call though ...

almarklein:
There's a few odd effects I'm looking into while i'm here:

Weird ghosting of the glyph bg when using outline:
image

Weird outline edges. Their shape looks like they might be sampling from neighbouring glyph, but that does not seem to be it:
image

Text becomes near-transparent from far off. This effect was added deliberately, but happens a bit too early. It starts at a scale when the text is still readable:
image

almarklein:
Tiny text better visible:
image

No background, no odd shapes on edges:

image

Korijn (approved):

examples/tests/test_examples.py
```diff
@@ -148,7 +148,7 @@ def test_examples_compare(filename, pytestconfig, force_offscreen, mock_time):
     stored_img = iio.imread(screenshot_path)
 
     # assert similarity
-    atol = 2
+    atol = 1
```
hmaarrfk, line 151: wow. i somewhat don't believe this.....

i was seeing variations in the middle of the `k` .... i guess your changes made it more stable for "small font" which I think we were always operating in that limit.

examples/screenshots/validate_blend_dither.png
hmaarrfk, line 1: The file `validate_blend_dither.png` lost the word "dither" at the very top.

almarklein: Sorry about that. The example changed, and I thought to just remove that text because the aa is dithered too. Thanks for bringing it back!

PR #1015, opened 2025-03-05
Reset shader state when update bindings

### LINKED ISSUE #1005: Bug when dynamically applying or removing env_map
While testing the example in #1004, I found that dynamically applying or removing the env_map was not working correctly. Initially, I thought it was a small issue, assuming we weren't tracking the env_map in MeshStandardMaterial. However, after modifying the code to enable tracking, the program threw the following exception:

```

  File "F:\pygfx\pygfx\renderers\wgpu\engine\pipeline.py", line 135, in get_cached_render_pipeline
    result = device.create_render_pipeline(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\wgpu-py\wgpu\backends\wgpu_native\_api.py", line 1717, in create_render_pipeline
    id = libf.wgpuDeviceCreateRenderPipeline(self._internal, descriptor)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\wgpu-py\wgpu\backends\wgpu_native\_helpers.py", line 457, in proxy_func
    raise wgpu_error
  File "F:\wgpu-py\wgpu\backends\wgpu_native\_api.py", line 1717, in create_render_pipeline
    id = libf.wgpuDeviceCreateRenderPipeline(self._internal, descriptor)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Caused by:
      Shader global ResourceBinding { group: 0, binding: 9 } is not available in the pipeline layout
        Binding is missing from the pipeline layout
```

Upon further investigation, I realized the issue was more complicated than expected. Here are my findings:

- When bindings change, the `update_shader_data` method of `PipelineContainer` recollects the bindings, but the `_binding_codes` in `BindingDefinitions` were not cleared (which caused the above exception). 
- After I added a clear operation for `_binding_codes` in the shader's `get_bindings()` method, I found that the shader's `kwargs` were not updated too, which led to some incorrect "macro switches" in the shader. This seems a bit tricky, as we cannot clear all `kwargs` as we did with `_binding_codes`.

panxinmiao:
Fix #1005 

This seems to be the simplest way to fix it.

It seems that there is no situation where `kwargs` are passed in the initialization of the shader in our actual code. In order to keep the `kwargs` state consistent, I deleted it in the initialization.

### DIFF
--- a/pygfx/renderers/wgpu/engine/pipeline.py
+++ b/pygfx/renderers/wgpu/engine/pipeline.py
@@ -358,6 +358,7 @@ class PipelineContainer:
         if "bindings" in changed:
             self.shader.unlock_hash()
             with tracker.track_usage("!bindings"):
+                self.shader.__init__(wobject)
                 self.bindings_dicts = self.shader.get_bindings(wobject, self.shared)
             self.shader.lock_hash()
             self._check_bindings()
--- a/pygfx/renderers/wgpu/shader/base.py
+++ b/pygfx/renderers/wgpu/shader/base.py
@@ -95,11 +95,11 @@ class BaseShader(ShaderInterface):
     type = "render"  # must be "compute" or "render"
     needs_bake_function = False
 
-    def __init__(self, wobject, **kwargs):
+    def __init__(self, wobject):
         super().__init__()
 
         # The shader kwargs
-        self.kwargs = kwargs
+        self.kwargs = {}
 
         # The stored hash. If set, the hash is locked, and kwargs cannot
         # be set. This is to prevent the shader implementations setting
--- a/tests/renderers/test_shader_composer.py
+++ b/tests/renderers/test_shader_composer.py
@@ -28,11 +28,12 @@ def test_templating():
             """
 
     # Missing variables
-    shader = MyShader(foo=True)
+    shader = MyShader()
     with raises(ValueError):
         shader.generate_wgsl()
 
     # Fill in value
+    shader["foo"] = True
     shader["bar"] = 42
     assert shader["bar"] == 42
     assert shader.generate_wgsl().strip() == "x = 42"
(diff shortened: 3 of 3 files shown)


panxinmiao:
`gltf_unlit.py`  is a use case  that reveals this issue.  you can switch the IBL checkbox to try. Moreover, I believe that dynamically remove any textures (and other bindings) will result in this exception.

image


When the bindings change, both the `_binding_codes` and `kwargs` are not fully updated.
For this example, the binding code related to env_tex in `_binding_codes` has not been cleared, and the key `use_ibl` in `kwargs` has not been updated.

almarklein:
It works for regular colormaps. I think we even have unit tests for that, so that they can be updated, and when the texture format matches, it does not invoke a shader recompile or pipeline construction. I can also confirm that this still works.

I will have a look. I  need to refresh my memory about the how the trackable-system works to see what's the best solution.

almarklein:
Oh right, I remember. You want to make sure that all the kwargs are set in the `Shader.__init__`. I.e. you should not call `self['something'] = x` in e.g. `get_bindings()`.

almarklein:
> Switching material bindings is unlikely to be a high-frequency operation, and it almost certainly involves shader recompilation and re-binding of bind groups. 

I think frequent enough to optimize for this case. User's selecting a different colormap, objects that switch to larger buffers as needed (like the text), etc.

No offense, but I'm going to close this, as it is not the correct approach. Will comment in #1005

panxinmiao:
It's okay, I also don't think this solution is ultimatel the right one, the more important purpose of this PR is to expose and verify the cause of the #1005 issue.

PR #1018, opened 2025-03-05
Ruler use multitext

almarklein:
Apart from being more performant, this also allows users to tweak the text (change font size, color, etc).


almarklein:
It appears this also fixes an artifact near the endpoint (see the updated screenshot).

examples/screenshots/validate_ruler.png
Korijn, line 1: Is this expected?
image

pygfx/geometries/_text.py
```diff
@@ -638,6 +638,8 @@ def _ensure_text_block_count(self, n):
                 block = self._text_blocks.pop(-1)
                 block._clear(self)
 
+        self.positions.draw_range = 0, n
```
Korijn, line 641: is this needed if `n == len(self._text_blocks)`?

examples/screenshots/validate_ruler.png
almarklein: Ah, I misread that this artifact was *removed* with the new screenshot, but it's actually added. Thanks for catching that! Will have a look, because it is indeed *not* expected.

pygfx/geometries/_text.py
almarklein: Yes, because `_text_blocks`  is exactly the size of the number of blocks in used, while `positions` is sized with factors of two.

For the text rendering itself, the `draw_range`  does not have to be set. But it helps when the `positions` buffer is also used to render points/lines, as we do in the ruler. Will add a comment to clarify this.

examples/screenshots/validate_ruler.png
almarklein: It is a partial join towards the point (0, 0) 😉  - fixed.

Korijn (approved):
Looks better. I can't even tell what changed in the screenshot.

almarklein:
> I can't even tell what changed in the screenshot.

🤔 I wonder whether only some meta-data (date?) has changed

PR #1019, opened 2025-03-05
Add image comparator example

hmaarrfk:
Is there a better place for this kind of tool? I feel like it uses text replace on the WGSL code and shouldn't really be encouraged by users that don't want their code to break every other version.

Let me know if you think this tool would be a useful addition, I've added it to my personal arsenal.

Taken from #1001

### DIFF
--- a/examples/feature_demo/compare_images.py
+++ b/examples/feature_demo/compare_images.py
@@ -1,0 +1,282 @@
+"""
+A Tool to Compare Two Images of Identical Shapes
+================================================
+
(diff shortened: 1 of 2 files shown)


almarklein:
> Is there a better place for this kind of tool? 

Maybe in `examples/other` since it's rather advanced, and does not demo one particular feature.

> I feel like it uses text replace on the WGSL code and shouldn't really be encouraged by users that don't want their code to break every other version.

Just add a comment to better not do that 😉

almarklein (approved):

Korijn (approved):

PR #1020, opened 2025-03-05
Fix TransformGizmo.set_object when object is `None`

schlegelp:
This is a tiny fix to allow instantiating `gfx.TransformGizmo` without immediately associating a world object.

The issue was a `isinstance(object, (WorldObject, None))` call that failed because `None` is not a type.

### DIFF
--- a/pygfx/helpers/_gizmo.py
+++ b/pygfx/helpers/_gizmo.py
@@ -88,7 +88,7 @@ class TransformGizmo(WorldObject):
 
         """
 
-        if not isinstance(object, (WorldObject, None)):
+        if not isinstance(object, (WorldObject, type(None))):
             raise ValueError("The object must be None or a WorldObject instance.")
 
         self._object_to_control = object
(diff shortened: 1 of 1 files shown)


almarklein (dismissed):

Korijn (dismissed):
I have never seen this solution before (why not `object is None`?) but I also hesitate to block :)

almarklein:
If `NoneType` was a builtin than `isinstance(x, (Some, NoneType))`would be the most elegant :)

schlegelp:
> I have never seen this solution before (why not object is None?) but I also hesitate to block :)

I thought I'd stick with the existing `isinstance` call but yes, it could be something like this:
```python
if object is None or isinstance(object, WorldObject):
   self._object_to_control = object
else:
   raise ValueError("The object must be None or a WorldObject instance.")
```



> If NoneType was a builtin than isinstance(x, (Some, NoneType))would be the most elegant :)

Yeah, you're right. Do you want me to define type `NoneType = type(None)` at the beginning of `_gizmo.py` and change the `isinstance` call?

schlegelp:
Changed to `isinstance(x, (WorldObject, NoneType))`.

almarklein (approved):
I think either way is fine 😄  We were just discussing the various options.

Korijn (approved):

PR #1021, opened 2025-03-05
fix two bugs in atlas allocator

hmaarrfk:
1. The free list shouldn't be cleared. It refers to a pointer array on the GPU which can in fact be readily reused since each "chunk" is independent.
2. There are 0 sized glyphs in the world....


Korijn (dismissed):
This makes sense to me but I'll leave it to Almar to hit the merge button as I'm not 100% sure about the internals of the text atlas code

almarklein:
I added some tests. Looked into what it actually means when an empty region is selected, because it looks like an oversight, but apparently this Just Works.

almarklein (approved):

Korijn (approved):

PR #1026, opened 2025-03-07
Improve tracking of bindings

### LINKED ISSUE #1005: Bug when dynamically applying or removing env_map

almarklein:
Fixes #1005

* [x] Rename "kwargs"  to "template-vars".
* [x] Split into two groups, one for init, and one for bindings, so that the latter can be cleared, so that when values are *not* set, this is detected, changing the hash, and thus triggering a recompile.
* [x] Check clearing of `_binding_codes` in `BindingDefinitions` -> ~Not necessary, since on a recompile all definitions are recreated.~ done, use `clear()`.
* [x] Have a look at the performance of the hashing. -> with the mesh in `gltf_unlit.py` re-hasing takes 12us 🤷 
* [x] Documentation.


Korijn:
Remark about hashing using json: isn't it a lot faster to convert to a frozendict and call hash on that?

almarklein:
> Remark about hashing using json: isn't it a lot faster to convert to a frozendict and call hash on that? Or (named)tuples?

I tried. Yes it's faster, going from 12us to 3us for the mesh in `gltf_unlit.py`. However, the mesh shader has one template var which is a dict (`used_uv`), so the code will have to check for these. 

Implemented a variant where dicts and lists are tuple-ized, and with 9 us it's still faster. Will open a new pr, because this is rather orthogonal.

Korijn (approved):

almarklein:
> Implemented a variant where dicts and lists are tuple-ized, and with 9 us it's still faster. 

To do this well, checking for instances quickly adds up, bringing it to about the same performance as the json implementation. Also we can calculate 100 hashes in 1 ms, so I'm not looking into this further right now.

Issue #996, opened 2025-02-24
Point rotation does not affect marker shape


MrLixm:
Hello,
I have an issue when using `Points` object and a `PointsMarkerMaterial`:
Any rotation applied to the `Points` object does not affect the orientation of the marker shape.

Here is an example where we use a "cross" marker shape and rotate the Points object:
```python
import math

import numpy
import pylinalg
from wgpu.gui.auto import WgpuCanvas, run
import pygfx as gfx


canvas = WgpuCanvas()
renderer = gfx.WgpuRenderer(canvas)

scene = gfx.Scene()

positions_x = numpy.linspace(0, 50, 10, dtype=numpy.float32)

for index, pos in enumerate(positions_x):
    geometry = gfx.Geometry(positions=[(0, 0, 0)])
    material = gfx.PointsMarkerMaterial(
        marker="cross", size=2.5, size_space="world", edge_width=0
    )
    points = gfx.Points(geometry, material)
    points.local.x = pos
    points.local.rotation = pylinalg.quat_from_axis_angle(
        (0, 0, 1),
        math.radians(15 * index),
    )
    scene.add(points)

camera = gfx.OrthographicCamera()

grid = gfx.GridHelper(size=100, divisions=10)
grid.local.rotation = pylinalg.quat_from_axis_angle(
    (1, 0, 0),
    math.radians(90),
)
scene.add(grid)

camera.show_object(scene)

controller = gfx.PanZoomController(camera=camera)
controller.register_events(renderer)

if __name__ == "__main__":
    canvas.request_draw(lambda: renderer.render(scene, camera))
    run()
```

Which executed renders the following:

![qt wgpu canvas screenshot](https://github.com/user-attachments/assets/f2328f6b-65f3-4974-b8e6-2ed746344e04)

I assume that this is kind of expected as Points doesn't seem to have an orientation in 3D space but I would personally like an option where the point marker behaves like a plane shape and can be rotated in the space.

So I'm not sure if this is a bug, a feature request, or simply a misuse of the points and I should use another workflow.

I am using a Points + PointsMarkerMaterial because it is handy for me to draw a custom shape using Sdf and don't have to worry about the geometry.

Is it possible to have any guidance on this problem?
Thanks,
Liam.

Korijn:
Are you expecting 2d rotation only in the plane of the screen, or full 3d rotation, as though it were a texture on a mesh?

MrLixm:
[...]

PR #1027, opened 2025-03-09
Support marker rotation

hmaarrfk:
I want to create a few arrows and whatnot, eventually I think this kind of rotation will be useful for us as a whole.

The SDF is also the only way we have to "draw a circle perfectly". However, I realize in all my enginuity that circles don't rotate!

- [x] What should "auto" mean?
   - Just like size mode, I removed auto and just kept it simple with "uniform"
- [x] I'm not sure what "vertex_map" is or if it is meaningful for rotations?
   - Confirmed with Korin that this is not applicable
- [x] Which direction should positive rotate in?
  - According to Korin this should be counter clockwise (need to triple check).

It isn't quite what @MrLixm wanted, but should be able to do what they wanted.

Closes https://github.com/pygfx/pygfx/issues/996

https://github.com/user-attachments/assets/667c7536-51f5-4b7f-9eee-95b00ffad024


Korijn:
I can already see the feature request for per-point configurable center of rotation :)

Awesome work!

pygfx/materials/_points.py
```diff
@@ -216,6 +221,29 @@ def map(self, map):
...
+    @property
+    def rotation_mode(self):
+        """The way that rotation is applied to the markers."""
+        return self._store.rotation_mode
+
+    @rotation_mode.setter
+    def rotation_mode(self, value):
+        value = value or "auto"
+        if value not in ["auto", "uniform", "vertex"]:
+            raise ValueError(f"PointsMaterial.rotation_mode does not support {value!r}")
```
Korijn, line 243: I never saw `!r` before, it's a useful trick for error messages! Will definitely remember that one.

pygfx/renderers/wgpu/shaders/pointsshader.py
```diff
@@ -76,8 +76,10 @@ def __init__(self, wobject):
             self["edge_color_mode"] = "uniform"
             self["edge_color_buffer_channels"] = 0
 
-        self["edge_color_mode"] = edge_color_mode
```
Korijn, line 79: did you intentionally remove this line?

Korijn:
I guess `auto` would mean you detect if the `rotations` buffer is available or not.

pygfx/materials/_points.py
hmaarrfk: this exact line was ChatGPT. sorry i didn't even note...

pygfx/renderers/wgpu/shaders/pointsshader.py
hmaarrfk: notice that it is defined in both "if/else" paths

hmaarrfk:
> I guess `auto` would mean you detect if the `rotations` buffer is available or not.

Should i consider "edge_mode" being auto -> "uniform" a bug? or just keep "auto" as an alias for "uniform"?

pygfx/renderers/wgpu/shaders/pointsshader.py
Korijn: gotcha!

pygfx/materials/_points.py
Korijn: I realize now we are already using it in the codebase in multiple places. It's a nice way to format the error. Nothing to apologize for

Korijn:
> * Which direction should positive rotate in?

Normally a positive angle is counter clockwise I think.

> * I'm not sure what "vertex_map" is or if it is meaningful for rotations?

https://github.com/pygfx/pygfx/blob/d39e5fca4c851099b443ff8988b15885c503e44c/pygfx/utils/enums.py#L68

It's unrelated/not meaningful for rotations.

> Should i consider "edge_mode" being auto -> "uniform" a bug? or just keep "auto" as an alias for "uniform"?

Hm. I think I jumped the gun with my previous suggestion. Looking at `pygfx/utils/enums.py`, `auto` has different meanings for different options. It can just mean that the default behavior is automatically suggested. That's fine. So no, not a bug.

Maybe it would be better to use an Enum in your work here as well.

examples/screenshots/validate_points_markers.png
hmaarrfk, line 1: @Korijn I've made the angle of the pin `pi/4` please confirm that this is the effect that matches your intuition.

> Normally a positive angle is counter clockwise I think.

Korijn: That's perfect 👍🏻

hmaarrfk:
Ok my personal checklist is over. Let me know if you have anything else that comes to mind.

almarklein (approved):

Korijn (approved):

hmaarrfk:
I'm curious how we can get rid of the computation of `cos` and `sin`.

I tried to doing it and for the "uniform", and it was easy enough.

However, it would be great to keep the same "API" in the geometry of using a single "N length vector". We would have to expand it to a precomputed `cos_rotation` and `sin_rotation` vector (or maybe a Nx2 vector) to sample directly in the shader.

almarklein:
>  We would have to expand it to a precomputed cos_rotation and sin_rotation vector (or maybe a Nx2 vector) to sample directly in the shader.

I don't think the added complexity of that is worth only to avoid two calls to `cos`/`sin`; I don't expect them to be expensive.

hmaarrfk:
IQ has like 3 blog posts about avoiding them and one is recent!

https://iquilezles.org/articles/noacos/


https://iquilezles.org/articles/noatan/
> This now runs 2.8 times faster than the original atan()

Korijn:
> IQ has like 3 blog posts about avoiding them and one is recent!
> 
> https://iquilezles.org/articles/noacos/
> 
> https://iquilezles.org/articles/noatan/
> 
> > This now runs 2.8 times faster than the original atan()

Well yes, but not because they moved tan to the CPU. It got faster because they found that the problem can be solved without trigonometry.

hmaarrfk:
yeah i guess by doing the 1D -> 2D tranformation we are only "precomputing" the results of cos/sin.

Issue #1011, opened 2025-03-03
Turn geometry into a simple store (and move TextGeometry api to Text object)


almarklein:
This has been discussed to some degree in several pull requests.

The geometry is a bit of an odd object, where we add buffers and textures without having properly documented how they should be named. To support this dynamic behavior, and to support properly tracking the buffers stored on it, it implements `__getattribute__`:

https://github.com/pygfx/pygfx/blob/735544f6b01bb171766eb7ce650f831ab81252d0/pygfx/geometries/_base.py#L95-L105

This makes attribute access on the geometry relatively slow. For most geometry there is not *that* much attribute access. But there is for `TextGeometry`; in benchmarks in #1006 I saw this take a significant portion of the CPU cycles in various stages of the processing of text.

 My proposal:

* Make `Geometry` subclass `Store` so it simply tracks all its attributes. (I have to double-check whether this works well.)
* Document clearly that `Geometry` is not to be subclassed. Just use it as a container to store buffers, textures, and maybe some other info like morph stuff.
* Document common buffer/texture names that are stored on a geometry.
* Each Material class documents what geometry attributes are supported.
* Move the text API to the `Text` world object, and a new `MultiText` world object.

Korijn:
FYI in Three.js a Geometry is a really dumb simple object with a dictionary of buffers in it. And you can add buffers as you wish/need to. Materials require different buffers to be available - so that's where they are documented.

almarklein:
FWIW we started with `Geometry` being a simple object. I think we added the `__getattribute__` when we added the trackable system, so that you can still add buffers with any name, and it would be stored on the `geometry._store` automatically. Then we used private variables (probably when we first implemented `TextGeometry` and added exception for private attributes to `__getattribute__`. So it kind of grew.

PR #1034, opened 2025-03-10
Move Text API and implementation to Text world object

almarklein:
* [x] Move API from `TextGeometry` to `Text` world object.
* [x] Move API from `MultiTextGeometry` to `MultiText` world object.
* [ ] Update examples.
* [ ] Update tests.


Korijn (approved):

almarklein:
🚀

hmaarrfk:
Would it be possible to get a quick release for this?

I ask because its quite disruptive in the API usage and I can't really think of a good "refactor" strategy. it would be nice to lean on the new API asap since there haven't been any other major changes since 0.8.0.

kushalkolar:
> Would it be possible to get a quick release for this?
> 
> I ask because its quite disruptive in the API usage and I can't really think of a good "refactor" strategy. it would be nice to lean on the new API asap since there haven't been any other major changes since 0.8.0.

Same we also want to do a release.

hmaarrfk:
If it helps, i did find my "easy compatibility layer"

```
from pygfx import Text as _Text
if not has_text_geometry:
    Text = _Text
else:
    class Text(_Text):
        def __init__(
            self,
            *args,
            text=None,
            markdown=None,
            font_size=12,
            family=None,
            screen_space=False,
            anchor="middle-center",
            text_align="start",
            **kwargs,
        ):
            geometry = gfx.TextGeometry(
                text=text,
                markdown=markdown,
                font_size=font_size,
                family=family,
                screen_space=screen_space,
                anchor=anchor,
                text_align=text_align,
            )
            super().__init__(*args, geometry=geometry, **kwargs)

        def set_text(self, text):
            self.geometry.set_text(text)

        @property
        def anchor(self):
            return self.geometry.anchor

        @anchor.setter
        def anchor(self, value):
            self.geometry.anchor = value

```

something like this can help. helped me.

I guess this release isn't as critical as I thought for me.

almarklein:
Yes, my plan is to get #1011 done quick (this was the biggest hurdle) and then to a release. (and then #1002)

PR #1040, opened 2025-03-16
Add back the text "dither" in the validation example

hmaarrfk:


Korijn (approved):

almarklein:
Note that the aa-edge is also dithered, so it would be better to set `aa=False` in the material. But even better is that after #1002 you can mix objects that use dither and objects that blend in the classic way.

PR #1041, opened 2025-03-17
No longer set the Bone object as invisible by default

panxinmiao:
While testing glTF rendering, I discovered that a very small number of models failed to render correctly, with certain Mesh components missing during rendering, as shown below: 

(The wheels of the car are missing)

image

After a thorough investigation, I identified the root cause of the issue: a bug in how we handle invisible objects during scene traversal.

Consider a parent-child relationship in the scene graph:
`... -> a -> b`
Here, `a` is an invisible scene control node, while `b` is a Mesh object that should be rendered. During scene traversal, since `a` is invisible, the current logic skips `a` along with its entire subtree. However, this is incorrect. Only `a` itself should be skipped, while its renderable child nodes should still be processed.

The correct result:
image

### DIFF
--- a/pygfx/objects/_skins.py
+++ b/pygfx/objects/_skins.py
@@ -18,7 +18,6 @@ class Bone(WorldObject):
 
     def __init__(self, name=""):
         super().__init__(name=name)
-        self.visible = False
 
 
 class Skeleton:


Korijn (dismissed):
Nice find!

almarklein:
mmm, I have always assumed that this was intended behavior: marking an object as invisible would hide that object *and* all its children.

almarklein:
We have at least one example that uses this feature (`labelled_image_grid.py`), and users may rely on it too. So let's first establish whether this is indeed a bug.

As for the original issue. If "a is an invisible scene control node", it does not need to set its `.visible` prop. Objects without a material are not draw anyway.

panxinmiao:
> mmm, I have always assumed that this was intended behavior: marking an object as invisible would hide that object _and_ all its children.

In this example model, the parent node of the car's wheels is an invisible Bone object (I guess it is convenient to add animations related to controlling the wheels), and under our current logic, they will all be skipped.

I believe that the visibility of an object should only affect itself and not its child nodes.

almarklein:
It's the same behavior of ThreeJS, which we probably modelled this after. So it's intended:

https://stackoverflow.com/a/25571352/2271927

We can probably support this use-case (hiding an object but not its children) in another way. The answer in the above link mentions that there also is `Material.visible`, seems like a good solution?

panxinmiao:
> mmm, I have always assumed that this was intended behavior: marking an object as invisible would hide that object _and_ all its children.

The parent-child relationship in the scene diagram should only be used for transform transformations, and visibility should not be affected by parent-child relationships.
In fact, for the most front-end users, the concept of scene graph and its parent-child relationships may be completely transparent to them, and they may not care at all.
Consider a very common scenario where users attempt to hide a certain Mesh in the scene. Obviously, if the child nodes of this Mesh are hidden together, they will definitely feel confused. And this clearly does not meet the expected usage.

panxinmiao:
> It's the same behavior of ThreeJS, which we probably modelled this after. So it's intended:
> 
> https://stackoverflow.com/a/25571352/2271927
> 
> We can probably support this use-case (hiding an object but not its children) in another way. The answer in the above link mentions that there also is `Material.visible`, seems like a good solution?

I conducted a simple test on this same scenario using Three.js, and the current version of Three.js can handle it correctly, which means that the current version of Three.js should not skip visible child nodes of invisible objects. I will confirm again.

panxinmiao:
> It's the same behavior of ThreeJS, which we probably modelled this after. So it's intended:
> 
> https://stackoverflow.com/a/25571352/2271927
> 
> We can probably support this use-case (hiding an object but not its children) in another way. The answer in the above link mentions that there also is `Material.visible`, seems like a good solution?

The usage scenarios of `material.visible` are very different from those of `object.visible`, as materials can be shared by multiple objects.

almarklein:
> The parent-child relationship in the scene diagram should only be used for transform transformations

I think that's a choice. Who says what is correct and what is not?

panxinmiao:
> > The parent-child relationship in the scene diagram should only be used for transform transformations
> 
> I think that's a choice. Who says what is correct and what is not?

Yes, that's true.
I wrote a test case using three.js, and indeed three.js skips the entire subtree of invisible objects, 😓

almarklein:
We don't have this feature yet, but layers may be a way to solve the original issue? https://waelyasmina.net/articles/manage-object-visibility-in-three-js-with-layers/

Or a `material.visible` -> perhaps all these control nodes can share a material?

Or using a `Group`, and adding the bone as a separate child.

panxinmiao:
This is indeed a matter of API selection.

- three.js will skip all invisible nodes and their subtrees.
https://jsfiddle.net/pq3g8jy9/
- By default, babylon.js only skips the invisible nodes themselves.
https://jsfiddle.net/qw72gz6m/

And, the Mesh object of babylon.js has an [`inheritVisibility`](https://doc.babylonjs.com/typedoc/classes/BABYLON.AbstractMesh#inheritvisibility) property (default is `false`), which when set to `true`, its visibility is affected by its ancestor nodes.

Korijn (approved):

PR #1042, opened 2025-03-17
Remove get_bounding_box from geometry

almarklein:
Ref #1011

The current implementation / API for object bounds could use some love. I looked into this, but I realized that this results in much more work than I wanted to do on this right now 😅, e.g. supporting CPU-based picking, and all the tiny API decisions that come with it. So I minimized my work, making it do exactly what is used to do, but the new `Bounds` class sets things up for improvements later.

* [x] Adds a module `utils.bounds` that provides a simple `Bounds`  class. Plus tests.
* [x] Remove `geometry.get_bounding_box()`, replaced API with `geomety.get_geometry_bounding_box()`.
* [x] Remove `geometry.get_bounding_sphere()` -> it was not even used. And broken `*`
* [x] Added `object._get_bounds_from_geometry()` that produces a `Bounds` object from the geometry.
* [x] Update other `bound_xx` methods accordingly.

`*` The now-deleted `geometry.get_bounding_sphere()` actually had two errors `**`:
* the center of the sphere was calculated using mean(), making it different from the center
  of the bounding box. 
* the distances from the center were calculated using  np.linalg.norm(pos_finite - center, axis=0)
    but that should have been axis=1, basically producing bullocks radii! 

`**`: my mistake 😢 

## API changes

* Remove `geometry.get_bounding_box()`, use the new `object.get_geometry_bounding_box()` instead.
* Remove `geometry.get_bounding_sphere()`


pygfx/utils/bounds.py
```diff
@@ -0,0 +1,95 @@
...
+    the bounds are the intersection of the box and the spere. This
+    allows applying transforms without growing the bounds because of
+    the box's corners.
+
+    When the radius is None, or near-equal to half the diagonal of the aabb,
+    the bounds respresents a box. The radius is at least the
+    max(width,height,depth) of the aabb, i.e. only cutting off its corners.
+    """
+
+    def __init__(self, aabb, radius=None):
```
Korijn, line 17: This seems like a good candidate for `__slots__` as well?

almarklein: Makes sense!

Korijn (approved):

PR #1043, opened 2025-03-17
Allow wgpu 0.19 - 0.21

almarklein:
All are compatible.


pygfx/renderers/wgpu/wgsl/mesh.wgsl
```diff
@@ -441,8 +441,8 @@ fn fs_main(varyings: Varyings, @builtin(front_facing) is_front: bool) -> Fragmen
                 var tbn = getTangentFrame(view, normal, varyings.texcoord{{normal_map_uv or ''}} );
             $$ endif
 
-            tbn.x = tbn.x * face_direction;
-            tbn.y = tbn.y * face_direction;
+            tbn[0] = tbn[0] * face_direction;
+            tbn[1] = tbn[1] * face_direction;
```
almarklein, line 445: @panxinmiao I think `matrix.x` stopped working in wgsl. I think this is the correct fix, can you please confirm?

Korijn (approved):

PR #1044, opened 2025-03-17
Make geometry a simple Store

### LINKED ISSUE #1011: Turn geometry into a simple store (and move TextGeometry api to Text object)

almarklein:
Closes #1011 

* [x] Refactor trackable logic a bit, so that a store can be used directly.
* [x] Reclass and strip Geometry.
* [x] Update Geometry docs .
* [x] Document usage of geometry in docstrings.

### DIFF
--- a/pygfx/geometries/_base.py
+++ b/pygfx/geometries/_base.py
@@ -26,8 +26,8 @@ class Geometry(Store):
     * ``normals``: an Nx3 buffer representing the surface normals of a ``Mesh``.
     * ``texcoords``: an Nx1, Nx2, or Nx3 set of per-vertex texture coordinates. The dimensionality
       should match that of the dimension of the colormap's texture (``material.map``).
-    * ``textcoords1``, ``textcoords2`` etc.: for additional texture coordinates. Usually Nx2.
-      E.g. a ``TextureMap`` with ``uv_channel`` set to 4 will use "textcoords4".
+    * ``texcoords1``, ``texcoords2`` etc.: for additional texture coordinates. Usually Nx2.
+      E.g. a ``TextureMap`` with ``uv_channel`` set to 4 will use "texcoords4".
     * ``colors``: per vertex or per-face color data for e.g. ``Mesh``, ``Line``, ``Points``.
       Can be Nx1 (grayscale), Nx2 (gray plus alpha), Nx3 (RGB), or Nx4 (RGBA).
     * ``sizes``: per vertex sizes for e.g. ``Points``.
--- a/pygfx/geometries/_base.py
+++ b/pygfx/geometries/_base.py
@@ -7,8 +7,13 @@ from ..utils.trackable import Store
 from ..resources import Resource, Buffer, Texture
 
 if TYPE_CHECKING:
-    import collections.abc
     from numpy.typing import ArrayLike
+
+    ArrayOrBuffer = ArrayLike | Buffer
+    ArrayOrTexture = ArrayLike | Texture
+    ArrayOrTextureOrBuffer = ArrayLike | Texture | Buffer
+
+Optional = None
 
 
 class Geometry(Store):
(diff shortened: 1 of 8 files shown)


pygfx/geometries/_base.py
```diff
@@ -3,44 +3,68 @@
...
+    Common texture attributes:
+
+    * ``grid``: a 2D or 3D texture for the ``Image`` and ``Volume`` objects, respectively.
+
+    For mesh morphing the following attributes are used:
+
+    * ``morph_targets_relative``: a bool indicating whether the morph positions are relative.
+    * ``morph_positions``: a list of arrays of per-vertex morph positions.
+    * ``morph_normals``: a list of arrays of per-vertex morph normals.
+    * ``morph_colors``: a list of arrays of per-vertex morph colors.
```
almarklein, line 46: @panxinmiao is this more or less correct? 😄

pygfx/geometries/_base.py
```diff
@@ -3,44 +3,68 @@
...
+    are optional. However, there are several common names.
+
+    Common buffer attributes:
+
+    * ``positions``: an Nx3 buffer representing vertex positions. Used by e.g. ``Mesh``, ``Line``, ``Points``, ``Text``.
+    * ``indices``: an Nx3 buffer representing the triangular faces of a ``Mesh``.
+    * ``normals``: an Nx3 buffer representing the surface normals of a ``Mesh``.
+    * ``texcoords``: an Nx1, Nx2, or Nx3 set of per-vertex texture coordinates. The dimensionality
+      should match that of the dimension of the colormap's texture (``material.map``).
+    * ``textcoords1``, ``textcoords2`` etc.: for additional texture coordinates. Usually Nx2.
```
Korijn, line 29: Tex not text right?

Korijn:
I wonder if somehow it would be possible to provide completion?

pygfx/geometries/_base.py
almarklein: 🤦 woops

pygfx/geometries/_base.py
panxinmiao: Yes, that's correct.

panxinmiao: Just out of curiosity, we didn't previously consider "morphy_xx" as a trackable attribute. Now, do they become trackable attributes like others.

In its current implementation, "morph_xx" behaves uniquely: once a Geometry object has been rendered for the first time, any modifications to "morph_xx" will not take effect. (Under normal usage scenarios, "morph_xx" is not expected to be modified. If changes are required, the Geometry object should be recreated.)

almarklein: Thanks for bringing this up. I suspect that it does not matter. Now that it is on the store, it will detect it when `geometry.morph_xx` is set, but since this is unlikely to happen, it does not matter.

If you think it does, we can also access these lists in the `meshshader.py` using `geometry["morph_xx"]`, which allows access to these attributes without triggering the tracking.

almarklein: Do you recall why we made the exception for these morph lists?

panxinmiao: > I suspect that it does not matter. Now that it is on the store, it will detect it when `geometry.morph_xx` is set, but since this is unlikely to happen

Yes, I think so.

> Do you recall why we made the exception for these morph lists?

Since morph animation data can be quite large—especially for high-precision animated models, where the data size often exceeds that of regular vertex data in geometry—to improve access speed, we encode the morph data into a texture and bind it to the geometry before the first render. In general, morph animation data is determined before the initial rendering and should not be dynamically modified at runtime.

almarklein:
> I wonder if somehow it would be possible to provide completion for buffer attributes?

 We could add the named kwargs to the `__init__`  ...

pygfx/geometries/_base.py
almarklein: That's the `_gfx_morph_texture`, created in the meshshader, IIUC. I made a change so that at least this one is accessed using `geometry["_gfx_morph_texture"]` so it does not participate in the tracking.

almarklein:
> We could add the named kwargs to the __init__ ...

Gave this a shot. Let me know if this is in line with what you were thinking of.

Korijn:
Looks good to me. Hope it helps our users find their way a bit more quickly!

almarklein:
Then this is ready from my end.

Issue #683, opened 2024-03-20
Camera.show_object() often has large margins


almarklein:
In some examples it feels as if the scene is "zoomed out".

Korijn:
It's because of the way bounding volumes are merged. Bounding spheres are already too big usually. Merging two spheres gives an even wider sphere. And so on.

almarklein:
I think a thing that also plays a role is that it picks a scale such that when the window is resized, it still fits.

hmaarrfk:
I spent more time that I would like troubleshooting this so hopefully this will help others. 

My usecase is limited to an orthographic camera (Perspective with FOV 0) so it may not really apply to other workflows.

What I arrived to was that there was some interaction between the aspect ratio of the viewport and the width and height of the camera.

```python
def _camera_show_bbox(camera, bbox, canvas_size):
    width = abs(bbox[0, 0] - bbox[1, 0])
    height = abs(bbox[0, 1] - bbox[1, 1])

    center_x = (bbox[0, 0] + bbox[1, 0]) / 2
    center_y = (bbox[0, 1] + bbox[1, 1]) / 2

    object_aspect = height / width
    canvas_aspect = canvas_size[1] / canvas_size[0]

    # This formula was obtained by consider the 6 cases below
    # and simplifying it
    if canvas_aspect < 1:
        scale = max(min(object_aspect, 1), canvas_aspect)
    else:
        scale = max(min(1, 1 / object_aspect), 1 / canvas_aspect)

    # if canvas_aspect < 1:
    #     if object_aspect > 1:
    #         scale = 1
    #     elif canvas_aspect < object_aspect:
    #         scale = object_aspect
    #     else:
    #         scale = canvas_aspect
    # else:  # canvas_aspect >= 1:
    #     if object_aspect < 1:
    #         scale = 1
    #     elif object_aspect > canvas_aspect:
    #         scale = 1 / canvas_aspect
    #     else:
    #         scale = 1 / object_aspect

    square = max(width, height) * scale

    state = camera.get_state()
    state['position'] = (center_x, center_y, square)
    state['width'] = square
    state['height'] = square

    camera.set_state(state)
```

I hope this helps you all get to the bottom of things, but for us, this gives us what appears to be visually "pixel perfect" results no matter the aspect of the canvas and the bounding box.

hmaarrfk:
PS. i don't think i needed to play with the z position, i was just trying to mimic the zoom controller.

hmaarrfk:
[...]

Issue #81, opened 2020-11-02
Geometry based raypicking


Korijn:
Picking by rendering is now supported (see #77) but we don't have geometry based raycasting yet which has other use cases.

PR #1049, opened 2025-03-18
Refactor bounds

almarklein:
When working on #1042 I actually was on the way for a bigger refactoring. I trimmed it down to avoid stalling. This PR shows what I was working towards.

This is very experimental and requires much more thoughts and discussions. I don't consider it a priority, so I may pick this back up (much) later.

### Also see

* #81
* #683

### Thoughts

* The radius trick helps get tighter bounds, and prevents them from becoming larges are aabb's get transformed.
* But it's also more CPU intensive. Its it worth it? Maybe some benchmarks.
* Maybe make it optional?
* What do we want a cpu-based picking API to look like?

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -521,82 +521,133 @@ class WorldObject(EventTarget, Trackable):
             self._bounds_geometry = None
         return self._bounds_geometry
 
-    def get_geometry_bounding_box(self) -> np.ndarray | None:
-        bounds = self._get_bounds_from_geometry()
-        if bounds is not None:
-            return bounds.aabb
+    def get_bounds_local(self):
+        """Get the Bounds based on this objects's geometry only, or None."""
+        return self._get_bounds_from_geometry()
 
-    def get_bounding_box(self) -> np.ndarray | None:
-        """Axis-aligned bounding box in parent space.
-
-        Returns
-        -------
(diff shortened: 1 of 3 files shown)


pygfx/objects/_base.py
```diff
@@ -521,82 +521,133 @@ def _get_bounds_from_geometry(self):
...
         return self._bounds_geometry
 
-    def get_geometry_bounding_box(self) -> np.ndarray | None:
+    def get_bounds_local(self):
+        """Get the Bounds based on this objects's geometry only, or None."""
+        return self._get_bounds_from_geometry()
+
+    def get_bounds_local_recursive(self):
+        """Get the Bounds, local to this objects space, of itself and all its children."""
+        # TODO: this replicates get_bounding_box(), but should it not include this local.matrix??
```
Korijn, line 530: All bounding geometry should be local - in the space of the geometry buffers.  It's much cheaper to transform a ray into this space than to transform bounding geometry into ray space (may or may not be world).

Korijn:
I could maybe take this thing out of your hands, and free up some of your time? I have experience implementing this stuff. Maybe we can call so you can share your vision and ideas.

almarklein:
> I could maybe take this thing out of your hands

I was hoping this'd nerd-snipe you 😉  But no rush from my end.

PR #1051, opened 2025-03-19
Add `match_aspect` arg to camera.show_object()

almarklein:
This makes it possible to fit a 2D scene tightly to a window. Ref #683.

When `show_object(.. match_aspect=True)` is used. It will do a refinement step at the end. It takes the objects bounding box, transforms it with the (inverse) camera transform, and then matches the camera's `width` and `height`.

## Examples!

Here's an example showing an image using `camera.show_object()`:

image

Now if we align the bounding box to the screen:

image

... better, the image will touch the edge of the screen when the window is narrow. But when the window is wide there's a margin.

If we also match the `camera.aspect` (the solution in his pr), the image fits snugly, even if the window is resized:

image

image

And this also works when `maintain_aspect=False`. E.g. the line plot example with
```py
camera.show_object(scene, match_aspect=True, scale=1.1)
```
image

This means many calls to `show_rect()` can be replaced. I only did it in a few examples for now.


almarklein:
cc @kushalkolar

almarklein:
I deliberately used the new feature in a validation screenshot, so that the new feature is tested.

Korijn (approved):

PR #1053, opened 2025-03-24
Refactor engine's resolving of fragment shader output 

almarklein:
**a.ka. get rid of `get_fragment_output`**

This is part of the work in #1002, but it seemed orthogonal enough to roll into a separate PR, which makes review easier.

### In short

Earlier, (custom) shaders ended like this:
```wgsl
@fragment
fn fs_main(varyings: Varyings) -> FragmentOutput {
   ...
    var out = get_fragment_output(varyings.positions, the_final_color);
    out.picking  = ..  # maybe
    out.depth = ... # some shaders
    return out;
}
```

The `get_fragment_output()` provided a hook, allowing the blenders to provide an implementation to "do their thing".

In the new situation of this pr, the hook is implemented differently. A bit more complex, but leading to simpler shaders for users:
```wgsl
@fragment
fn fs_main(varyings: Varyings) -> FragmentOutput {
   ...
    var out = FragmentOutput;
    out.color = the_final_color;
    out.picking  = ..  # maybe
    out.depth = ... # some shaders
    return out;
}
```

### Motivation

* The first arg in `get_fragment_output()` feels arbitrary. It's only needed by the dither blending and weighted with depth. It also does not 'scale'.
* For classic blending, the code is now as one would expect, and the shader resolving indeed will not change it (after #1002).
* For special blending (mainly dither and weighted), the code also looks natural.
* It will allow users to more easily tweak the special blending, e.g. changing the seed for the dither, or the weight for weighted blending, using simply `out.weight = ...`.
* It will make it easier / more natural to support custom/additional render targets in the future.

### Compatibility

To help users who write custom shaders transition, shaders can still be written using the old approach. The wgsl is converted to the 'new approach' as part of the shader resolving. We can remove that in a later version.

### Performance

I also tweaked the performance of the shader resolving step. I made changes that made it a lot faster, but also some more checks (e.g. the resolving now correctly knows whether it's in a vertex/fragment entrypoint). In total the resolving is still more than twice as fast. Don't expect to 'feel' the difference, I don't think this was a bottleneck, but all bits help.

### Changes in a list

* [x] Refactor `resolve.py` for performance.
* [x] Refactor output resolving to do more work, to allow simpler shader.
* [x] Added some trickery to support the old usage. 
* [x] Added many tests. 
* [x] Updated blender.py accordingly ... which will be completely overhauled in #1002. 
* [x] Docs.
* [x] Update our shaders and examples that use custom shaders.


almarklein:
This was harder than I thought when I started it. And all it does is change the API for writing custom shaders a bit. I still think its worth it though, especially if we formalize the API for custom-shaders more in the furture, since its cleaner/simpler, and provides an easy way to turn the knobs of e.g. weighted blending for specific use-cases.

almarklein:
Another way to put it: that `get_fragment_output` was a nice trick that provides a very easy shortcut, but now let's do it properly.

Korijn (approved):

Korijn (approved):

docs/advanced_shaders.rst
```diff
@@ -223,11 +223,39 @@ All fragment functions in Pygfx are somewhat like this:
...
+
+... or set the weight for weighted blending:
+
+.. code-block::
+
+    ...
+    var out: FragmentOutput;
+    out.color = vec4<f32>(...);
+    $$ if blending == 'weighted'
+    out.weight = f32(...);
```
hmaarrfk, line 254: This isn't exactly consistent with the implementaiton.

In the implementation you use

```
depth_for_weight
```

Your also at the limit of basically creating your own shading language!!!

almarklein: Yeah, that will be fixed in #1002. Not so much create a shading language, but providing entry points for custom shaders, I think.

PR #1056, opened 2025-03-27
Add support for loops to line object

### LINKED ISSUE #1047: Beatufully closed lines
One of our favorite tools is the line shader.

However when transparency gets thrown into the mix, it isn't so "beautiful" in one aspect. Drawing closed shapes, such as squares, or circles, causes the starting point to be "drawn twice", giving it a different color than the other points on the line.

Image

In my mind, the API i'm looking for is something like:

1. Provide a long list of positions.
2. Assume these all have the same number of segments.
3. Provide that "number of segments" to the shader code.
4. It can then us that information (in the bake function????) to close off the ends of the lines adequately.

Often times, we have a an array, `Nx4` for rectangular boxes that specify the 4 points that make up the corners. I think it would be great if we could "throw that" by reshaping it to `1x4N` the shader and let it draw the square for us, instead of extending it to to `1x6N` to include the start/end point twice, and a nan to separate out the lines.

Code to generate the the image above (well I shared an older version, but the same principle applies)
[validate_closed_loops.py.txt](https://github.com/user-attachments/files/19318483/validate_closed_loops.py.txt)

almarklein:
Closes #1047 

Every line piece between two nan's is closed. The bonus node of the point that is nan is used to close the line.

* [x] Working POC for n=4.
* [x] Fixed a bug in the join when the corner is exactly 90 degrees and aligned with the coordinate frame.
* [x] Generalize to any piece between two nans, using the bake function.
* [x] Also close final point of a line, or only where there's a nan?


examples/validation/validate_line_loop.py
```diff
@@ -0,0 +1,119 @@
...
+# sphinx_gallery_pygfx_test = 'compare'
+
+import numpy as np
+from rendercanvas.auto import RenderCanvas, loop
+import pygfx as gfx
+
+
+canvas = RenderCanvas(size=(1000, 800))
+renderer = gfx.WgpuRenderer(canvas)
+scene = gfx.Scene()
```
hmaarrfk, line 20:
```suggestion
scene = gfx.Scene()

# This example ensures there are no visual artifacts with transparent lines
# for closed loops
# We add a background to demonstrate that it behaves well with
# transparent lines
scene.add(gfx.Background.from_color("#fff", "#000", "#f00", "#0f0"))
```

examples/validation/validate_line_loop.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+positions = np.vstack(
+    [
+        # Special case
+        all_corners,
+        nanpoint,
+        # Increasingly more corners
+        circle(0, -7, 0),
+        nanpoint,
```
hmaarrfk, line 62:
```suggestion
        nanpoint,
        nanpoint,
```

ensure successive nanpoints don't crash the shader

examples/validation/validate_line_loop.py
```diff
@@ -0,0 +1,119 @@
...
+        circle(2, -1, 0),
+        nanpoint,
+        circle(3, +2, 0),
+        nanpoint,
+        circle(4, +5, 0),
+        nanpoint,
+        # Even more corners
+        circle(5, -7, -3),
+        nanpoint,
```
hmaarrfk, line 73:
```suggestion
        nanpoint,
        nanpoint,  # ensure many successive nanpoints don't crash the shader
        nanpoint,
        nanpoint,
        nanpoint,
        nanpoint,
```

examples/validation/validate_line_loop.py
```diff
@@ -0,0 +1,119 @@
...
+all_corners += (-8, 3, 0)
+
+
+positions = np.vstack(
+    [
+        # Special case
+        all_corners,
+        nanpoint,
+        # Increasingly more corners
+        circle(0, -7, 0),
```
almarklein, line 61:
```suggestion
        circle(0, -7, 0),  # actuall zero points, so this results in two successive nanpoints
```

examples/validation/validate_line_loop.py
almarklein: Sure why not 😅

Korijn (approved):

PR #1057, opened 2025-03-27
Fix pick info inheritance

limdaepl:
When picking a `Mesh`, I get a pick info like this:

```
{
  'rgba': Color(0.8039, 0.7529, 0.698, 1.0),
  'world_object': <pygfx.Mesh  at 0x2544b7ec510>,
  'face_index': 2181,
  'face_coord': (0.015873015873015872, 0.6666666666666666, 0.2857142857142857)
}
```

When picking an `InstancedMesh`, I get a pick info like this:

```
{
  'rgba': Color(0.9765, 0.9098, 0.8471, 1.0),
  'world_object': <pygfx.InstancedMesh  at 0x24841da8050>,
  'instance_index': 5
}
```
Note that `'face_index'` and `'face_coord'` are missing. This is not only inconvenient but also a LSP violation as `InstancedMesh` derives from `Mesh`.

The expected pick info for an `InstancedMesh` would look like this:

```
{
  'rgba': Color(0.8039, 0.7529, 0.698, 1.0),
  'world_object': <pygfx.InstancedMesh  at 0x2544b7ec510>,
  'face_index': 2181,
  'face_coord': (0.015873015873015872, 0.6666666666666666, 0.2857142857142857),
  'instance_index': 5
}
```

The issue is that many implementations of `_wgpu_get_pick_info()` create a new pick info dict, when they should call the base class version and add their own keys to it.

This pull request updates all implementations of `_wgpu_get_pick_info()` to call and extend the base implementation.

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -598,7 +598,7 @@ class WorldObject(EventTarget, Trackable):
         aabb = self.get_world_bounding_box()
         return None if aabb is None else la.aabb_to_sphere(aabb)
 
-    def _wgpu_get_pick_info(self, pick_value):
+    def _wgpu_get_pick_info(self, pick_value) -> dict:
         # In most cases the material handles this.
         return self.material._wgpu_get_pick_info(pick_value)
 
--- a/pygfx/objects/_instanced.py
+++ b/pygfx/objects/_instanced.py
@@ -69,8 +69,8 @@ class InstancedMesh(Mesh):
         """get the matrix for the instance at the given index."""
         return self._store["instance_buffer"].data["matrix"][index].T
 
-    def _wgpu_get_pick_info(self, pick_value):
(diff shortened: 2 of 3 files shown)


almarklein:
Thanks! I think calling super() makes sense for the WorldObject classes, but not so much for the materials. How the picking info is encoded differs per material and is not necessarily compatible with the superclass. For example the normal line material does not have picking data.

almarklein:
Just revert the supers in the material subclasses.

The class hierarchy for the materials is something we want to address at some point; it does not always follow OO principles very tightly. Some classes (e.g. `MeshNormalLinesMaterial` and `LineThinMaterial`) don't really extend their base classes, but are reduced / completely different.

limdaepl:
I reverted the changes to the materials. Let me know if you need anything else changed!

PR #1058, opened 2025-03-27
Instanced lines

limdaepl:
Introduces instanced line rendering:
* Generalize `InstancedMesh(Mesh)` to `InstancedObject(WorldObject)` (no real changes) and refactor `InstancedMesh` to inherit from `Mesh` and `InstancedObject`
* Introduce `InstancedLine` class inheriting from `Line` and `InstancedObject`
* Add instancing support to `LineShader` (copied the relevant parts from `MeshShader`)
* Add instancing support to `line.wgsl` (copied the relevant parts from `mesh.wgsl`)

The duplications in the shader classes and shader code can probably be generalized to some degree as well, but I am not familiar enough with the templating engine to know how to do that.

Requires https://github.com/pygfx/pygfx/pull/1057 because the multiple inheritance breaks the pick info even more.

### DIFF
--- a/pygfx/objects/_instanced.py
+++ b/pygfx/objects/_instanced.py
@@ -1,26 +1,26 @@
 import numpy as np
 
 from ._base import id_provider
-from . import Mesh
+from . import WorldObject, Mesh
 from ..resources import Buffer
 
 
-class InstancedMesh(Mesh):
-    """Display a mesh multiple times using instances.
+class InstancedObject(WorldObject):
+    """Display an object multiple times using instances.
 
-    An instanced mesh with a matrix for each instance.
+    An instanced object with a matrix for each instance.
 
     Parameters
     ----------
     geometry : Geometry
-        The mesh's geometry data.
+        The object's geometry data.
     material : Material
-        The material with which to render the mesh.
+        The material with which to render the object.
     count : int
         The number of instances to create.
     kwargs : Any
         Additional kwargs get forwarded to the :class:`base class
-        <pygfx.objects.Mesh>`.
+        <pygfx.objects.WorldObject>`.
 
     """
 
@@ -75,3 +75,7 @@ class InstancedMesh(Mesh):
         id = pick_value & 1048575  # 2**20-1
         info["instance_index"] = self._idmap.get(id)
         return info
+
+
+class InstancedMesh(Mesh, InstancedObject):
+    pass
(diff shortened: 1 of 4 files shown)


pygfx/renderers/wgpu/wgsl/line.wgsl
```diff
@@ -97,14 +97,37 @@ fn rotate_vec2(v:vec2<f32>, angle:f32) -> vec2<f32> {
...
+    $$ if instanced
+    @builtin(instance_index) instance_index : u32,
+    $$ endif
+};
+
+$$ if instanced
+struct InstanceInfo {
+    transform: mat4x4<f32>,
+    id: u32,
 };
```
almarklein, line 109: As you said, I think we could e.g. move this to `std.wgsl`, but there's still the `instance_index` and `s_instance_infos` to account for. Plus it's not *that* much duplicate code, so let's leave it like this 👍

pygfx/objects/_instanced.py
```diff
@@ -75,3 +75,11 @@ def _wgpu_get_pick_info(self, pick_value):
...
         info["instance_index"] = self._idmap.get(id)
         return info
+
+
+class InstancedMesh(Mesh, InstancedObject):
+    pass
+
+
+class InstancedLine(Line, InstancedObject):
+    pass
```
almarklein, line 85: Docstrings woulld be nice. Other than that, this pr looks GTM!

limdaepl: Sure, what would you prefer?

1) Copy the docstring from `InstancedObject`, substituting mentions of `object`/`WorldObject` with `mesh`/`Mesh` and `line`/`Line`, so essentially the same description three times.
2) As above, but in code: `__doc__ = InstancedObject.__doc__.replace('object', 'mesh').replace('WorldObject', 'Mesh')`.
3) Something brief that refers to the base classes for further information.

almarklein: I think we can keep this mixin class an implementation detail. So (1) or (2), your pick :)

limdaepl: Went with option 2. Let me know if you need anything else!

almarklein (approved):
Nice!

almarklein:
The liter complaints. We use ruff for linting (`pip install ruff` or similar):
* `ruff .` in the repo root auto-formats everything.
* `ruff check` checks for linting errors.

Korijn (approved):

PR #1062, opened 2025-04-01
Don't throw AttributeError when calling dict attributes

berendkleinhaneveld:
This makes Geometry behave more like a proper dict.

### DIFF
--- a/pygfx/utils/trackable.py
+++ b/pygfx/utils/trackable.py
@@ -54,9 +54,8 @@ there is no persistent knowledge of hierarchy.
 
 """
 
-import weakref
 import threading
-
+import weakref
 
 global_id_counter = 0
 global_lock = threading.RLock()
@@ -150,6 +149,8 @@ class Store(dict):
             value = self[key]
             return value
         except KeyError:
+            if hasattr(dict, key):
+                return dict.__getattribute__(self, key)
             raise AttributeError(key) from None
         finally:
             if global_context:
--- a/pygfx/utils/trackable.py
+++ b/pygfx/utils/trackable.py
@@ -149,9 +149,7 @@ class Store(dict):
             value = self[key]
             return value
         except KeyError:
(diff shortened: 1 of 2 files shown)


berendkleinhaneveld:
I can imagine that you would not want to support calling any other methods on Store objects, so please let me know what you think.
I encountered this `AttributeError` in collagraph/observ when a geometry is part of a reactive state object. It calls the `values()` method on any dict within the reactive state. I could also imagine just adding a try/except there instead of this PR.

pygfx/utils/trackable.py
```diff
@@ -150,6 +149,8 @@ def __getattribute__(self, key):
             value = self[key]
             return value
         except KeyError:
+            if hasattr(dict, key):
+                return dict.__getattribute__(self, key)
```
Korijn, line 153: I feel like a getattr with a try/except clause would be more performant, to avoid looking up the key twice?

berendkleinhaneveld: I tried the following:
```python
try:
    value = self[key]
    return value
except KeyError:
    return dict.__getattribute__(self, key)
```
and the happy path seems ever so slightly faster. When the attribute is not there, it is a bit slower, but that code path should not be hit too often. Certainly not within regular pygfx usage.

Korijn: No, I mean, the code path raises `AttributeError(key)` anyway, so why not just:

```py
    def __getattribute__(self, key):
        if key.startswith("__"):
            return dict.__getattribute__(self, key)
        value = undefined
        try:
            value = self[key]
            return value
        except KeyError:
            return dict.__getattribute__(self, key)
        finally:
            if global_context:
                global_context.tracker._track_get(
                    self, key, value, global_context.label
                )
```

almarklein (approved):
Makes sense to restore this functionality, especially since its a `dict` subclass.

PR #1063, opened 2025-04-02
Example to compute histogram using a compute shader

### LINKED ISSUE #1060: Request: A compute shader example
I'm hoping this request goes well into the standard "pygfx examples".

I'm hoping to learn how to build up the compute shader pipelines.

Here i've created a simple example where we get an image as a texture, and we wish to compute its histogram

https://github.com/hmaarrfk/pygfx/pull/12

I've tried to organize it in a way that:

1. We have a single buffer for the image
2. We output the data as a single texture
3. The shader rescales as  necessary

I'm hoping to get an example where the code
```
            # Trigger recomputation of the histogram
            hist_r, hist_g, hist_b, hist_l, computation_time = compute_histogram(img)
            update_histogram(hist_r, hist_g, hist_b, hist_l)
```

can be replaced by a compute shader.

Efficiency isn't the key, but mostly the mechanics of pygfx as they relate to the compute shader.

We can focus on the `main` branch. no need to backport the example to a "stable" version. 

Thanks for the help!

(ps. i know my log representation of the histogram doesn't look the best)

https://github.com/user-attachments/assets/0710b457-7d06-4349-a9d1-76c30eb3941d

almarklein:
Closes #1060

* [x] Working POC.
* [x] Parallellize.
* [x] Use a separate stage to 'draw' the result.
* [x] Maybe just write positions to a a buffer used as `positions` in a line object?
* [x] Show hist for r, g, b.


examples/other/image_histogram.py
```diff
@@ -0,0 +1,263 @@
...
+    compute_pipeline = device.create_compute_pipeline(
+        layout="auto",
+        compute=compute,
+    )
+
+    # Create bindings
+    bindings = [
+        {
+            "binding": 0,
+            "resource": image_wgpu_texture.create_view(
```
kushalkolar, line 179: Wait so when this is done, is it sharing the same buffer for rendering the Texture as an image shown on the screen as well as sharing this buffer for the compute shader? It's that's easy? I have so many ideas!

Korijn: Turns out compute shaders are very useful! 😁

almarklein: But also the other way around. Let ppl generate a texture using GPU compute, and the just show it in Pygfx. Same for buffers that can be used for positons, colors, etc.

almarklein:
I think this is nearly done.

I abstracted away internal (pygfx) stuff into a nice little `ComputeShader` class. My first thought was to put this in `pygfx.utils`, but I think we should think a little bigger. If we do https://github.com/pygfx/wgpu-py/issues/704, this class can be more generic, and put in `wgpu.utils.compute`, or even a new library. What are your thoughts @Korijn, @hmaarrfk, @kushalkolar ?

examples/other/image_histogram.py
```diff
@@ -0,0 +1,582 @@
...
+# Create GUI renderer
+gui_renderer = ImguiRenderer(renderer.device, canvas)
+gui_renderer.set_gui(draw_imgui)
+
+
+def animate():
+    if hist_calc_shader.changed:
+        size = image_object.geometry.grid.size
+        hist_calc_shader.dispatch(
+            int(size[0] / 16 + 0.499999), int(size[1] / 16 + 0.499999)
```
hmaarrfk, line 571: What is the +0.499999 for?

examples/other/image_histogram.py
```diff
@@ -0,0 +1,582 @@
...
+        )
+        hist_write_shader.dispatch(1)
```
hmaarrfk, line 573: How guaranteed is it that this will complete before the next scene draw?

examples/other/image_histogram.py
```diff
@@ -0,0 +1,582 @@
...
+                usage=wgpu.TextureUsage.STORAGE_BINDING
+                | wgpu.TextureUsage.TEXTURE_BINDING,
+            )
+            # Update the image
+            image_object.geometry.grid = image_texture
+            # Update histogram
+            hist_calc_shader.set_resource(0, image_object.geometry.grid)
+            histogram_object.local.scale_x = image_texture.size[0] / (nbins - 1)
+
+        # imgui.text(f"Histogram computation time: {computation_time * 1000:.1f} ms")
```
hmaarrfk, line 554: Semi related to the other questions. Is there any way to measure the computation time of the histogram function?

examples/other/image_histogram.py
```diff
@@ -0,0 +1,582 @@
...
+            v = color[channel];
+        }
+        let binIndex = min(u32(v * f32(binCount)), lastBinIndex) * 4u + channel;
+        atomicAdd(&bins[binIndex], 1u);
+    }
+
+    workgroupBarrier();
+
+    // Copy the bin values from our workgroup array to the storage buffer.
+    let binValue = atomicLoad(&bins[binIndex]);
```
hmaarrfk, line 456: Why do you need an atomic load here? Doesn't this thread own the variable?

examples/other/image_histogram.py
almarklein: poor-man's `int(round(..))`. But actually should be `ceil`. Will fix

examples/other/image_histogram.py
almarklein: I'm pretty sure that Wgpu manages this.

examples/other/image_histogram.py
almarklein: It's measured! (this line can be deleted).

examples/other/image_histogram.py
almarklein: The example at webgpufundamentals does it too. I'll have another look. I think I also changed back and forth a bit as I was trying to get things to work.

panxinmiao: I think this is because `bins` is a workgroup variable shared among all threads within the workgroup. Therefore, in a multithreaded context, its read and write operations need to be atomic to ensure data consistency.

examples/other/image_histogram.py
panxinmiao: Yes, It should be `ceil`,  to ensure that every pixel is properly handled by a corresponding workgroup.

Additionally, the value 16 here corresponds to the compute shader's workgroup_size,  and it would be better to pass it in as an override constant through the pipeline.

examples/other/image_histogram.py
kushalkolar: > Let ppl generate a texture using GPU compute, and the just show it in Pygfx

Yes definitely something I'm interested in!

examples/other/image_histogram.py
almarklein: In this (and one other) case there is no risk on race-conditions, because each thread only touches its own index. I can indeed use normal array opeations, and that works. However, the WGSL spec states:

> the only valid operations on atomic objects are the [atomic builtin functions](https://www.w3.org/TR/WGSL/#atomic-builtin-functions).

So, I expect that at some point that will be forbidden. Note that atomicity is address specific, to the fact that we have `atomicLoad` here does not mean that the different threads wait for each-other, since they never touch the same address.

examples/other/image_histogram.py
```diff
@@ -0,0 +1,582 @@
...
+    nbytes=nbins * 4 * 4,
+    nitems=nbins,
+    format="4xu4",
+    usage=wgpu.BufferUsage.STORAGE | wgpu.BufferUsage.COPY_DST,
+)
+
+# Create a buffer that holds the line positions. This is written to by a compute shader.
+# It consists of 4 pieces (red, green, blue, luminance), with nans in between.
+histogram_line_buffer = gfx.Buffer(
+    nbytes=(4 * nbins + 3) * 3 * 4,
```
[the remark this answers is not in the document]
almarklein: Are you referring to the RGB -> RGBA conversion for `gfx.Texture`? That is only for textures. Buffers for positions and colors can be Nx3.

examples/other/image_histogram.py
almarklein: Also note the `device._poll_wait()`. That will actually halt the CPU until the GPU finished its work. We only do that when we want to measure the GPU's spent time.

almarklein:
I moved the new `ComputeShader` in `pygfx.utils.compute` for now. That allows ppl to start playing with it. We can refactor it into a more generic version, and move it to another place, when we have done https://github.com/pygfx/wgpu-py/issues/704.

examples/other/image_histogram.py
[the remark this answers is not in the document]
hmaarrfk: Interesting. Yes indeed I was referring to that.

pygfx/utils/compute.py
```diff
@@ -0,0 +1,254 @@
...
+        compute_pass.set_bind_group(0, self._bind_group)
+        compute_pass.dispatch_workgroups(nx, ny, nz)
+        compute_pass.end()
+
+        # Submit!
+        device.queue.submit([command_encoder.finish()])
+
+        # Timeit
+        if self._report_time:
+            device._poll_wait()  # wait for the GPU to finish
```
Korijn, line 251: Should poll wait be outside the if statement?

almarklein: No, the poll is only needed so that the timing measures the actual time the GPU is working. Without the poll, the GPU is doing its thing while the CPU can also proceed.

Korijn (approved):

PR #1066, opened 2025-04-03
Make the Store track hashable objects by value

almarklein:
* [x] ~Can only store hashable objects on the `Store`.~
* [x] Objects that are not primitive types, but are hashable, are now tracked by their hash. 
* [x] Implement simple `ReadOnlyDict` that is hashable. Not used yet, but I plan to use it in #1002

### DIFF
--- a/pygfx/utils/__init__.py
+++ b/pygfx/utils/__init__.py
@@ -219,3 +219,43 @@ def assert_type(name, value, *classes):
 
         # Raise message with alt traceback
         raise TypeError(msg).with_traceback(tb) from None
+
+
+class ReadOnlyDict(dict):
+    """A read-only dict, for storing structured data that can be hashed."""
+
+    ___slots__ = ["_hash"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Calculate hash in a way that requires any value to also be hashable
+        parts = []
+        for k in sorted(self.keys()):
+            v = self[k]
+            parts.append(str(hash(k)))
+            parts.append(str(hash(v)))
+        self._hash = hash(" ".join(parts))
+
+    def __setitem__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


Korijn:
Why not frozendict? It's built in?

pygfx/utils/__init__.py
```diff
@@ -219,3 +219,43 @@ def assert_type(name, value, *classes):
 
         # Raise message with alt traceback
         raise TypeError(msg).with_traceback(tb) from None
+
+
+class ReadOnlyDict(dict):
+    """A read-only dict, for storing structured data that can be hashed."""
+
+    ___slots__ = ["_hash"]
```
Korijn, line 227: You have one underscore too many :)

pygfx/utils/__init__.py
```diff
@@ -219,3 +219,43 @@ def assert_type(name, value, *classes):
 
         # Raise message with alt traceback
         raise TypeError(msg).with_traceback(tb) from None
+
+
+class ReadOnlyDict(dict):
```
Korijn, line 224: I wonder if it would be more performant to wrap a dict instead of subclassing?

almarklein: I want it to be a dict, e.g. with `isinstance()`.

almarklein:
> Why not frozendict? It's built in?

There is `frozenset`, but the PEP for `frozendict` was rejected, apparently.

pygfx/utils/__init__.py
almarklein: nice catch! Also added a test for it.

almarklein:
Ah, mesh morphing stores a list on the store 🤔

almarklein:
I restored being able to strore non-hashable types. So this simply adds better support for hashable types.

Korijn (approved):

PR #1067, opened 2025-04-03
create background subtraction example

kingbedjed:
Here is an example of how to use mipmaps to perform a background subtraction. The benefit of this over other approaches is that the background creation is done  through pygfx's mipmap creation so that it is only done at texture loading, not every render cycle.

### DIFF
--- a/examples/other/background_subtraction.py
+++ b/examples/other/background_subtraction.py
@@ -1,3 +1,16 @@
+"""
+A Method to Subtract Background from an Image
+================================================
+
+This example shows how to subtract a background from an image.
+
+The background is created via the mipmap levels of the image so that the background does not need to be
+computed per pixel during each render cycle. The mipmap level is sampled from to create the background.
+
+One can adjust the background level via the dropdown menu. Each level subtracts a subsampled background
+from the image with the subsampling increasing by a power of 2 with each level. Level 0 is the original
+image with no background subtraction.
+"""
 import imageio.v3 as iio
 import numpy as np
 from wgpu.gui.auto import WgpuCanvas, run
@@ -6,7 +19,6 @@ from pygfx.renderers.wgpu import register_wgpu_render_function
 from pygfx.renderers.wgpu.shaders.imageshader import ImageShader
 from wgpu.utils.imgui import ImguiRenderer
 from imgui_bundle import imgui
-
 
 
 # Load image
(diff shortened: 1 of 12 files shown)


examples/other/background_subtraction.py
```diff
@@ -0,0 +1,189 @@
...
+    def background_level(self):
+        return self.uniform_buffer.data["background_level"]
+
+    @background_level.setter
+    def background_level(self, value):
+        self.uniform_buffer.data["background_level"] = int(value)
+        self.uniform_buffer.update_range()
+
+class BackGroundRemovedImage(gfx.Image):
+    pass
```
almarklein, line 57: I don't think you need a new `Image` subclass, just the new material.

examples/other/background_subtraction.py
```diff
@@ -0,0 +1,189 @@
...
+current_image_index = 0
+
+def draw_imgui():
+    global current_background_index
+    global current_image_index
+    global im, image_texture
+
+    imgui.new_frame()
+
+    imgui.set_next_window_size((300, 0), imgui.Cond_.always)
```
almarklein, line 153: The label is cut off (for me), this fixes it. 
```suggestion
    imgui.set_next_window_size((400, 0), imgui.Cond_.always)
```

examples/other/background_subtraction.py
```diff
@@ -0,0 +1,189 @@
...
+    var b1: vec4<f32> = mix(
+        textureLoad(t_img, 
+            vec2<i32>(texcoords_u + vec2<f32>(-0.5, 0.5)), 
+            i32(u_material.background_level)),
+        textureLoad(t_img, 
+            vec2<i32>(texcoords_u + vec2<f32>( 0.5, 0.5)), 
+            i32(u_material.background_level)),
+        f.x,
+    );
+    let background = mix(b0, b1, f.y);
```
almarklein, line 98: Instead of using `textureLoad` to read 4 pixels and combine them, you can also use `textureSampleLevel` to let the hardware do the interpolation. Unless I'm overlooking a reason for using `textureLoad` here.

almarklein (commented):
Cool example! Thanks for contributing this trick here ❤️ 

Left a few comments. Could you also add a module docsting explaining the idea (and mentioning that level 0 can be selected to see the original (unfiltered) image)?

Korijn (approved):

almarklein:
LGTM, except for the two linting errors:

    Error: examples/other/background_subtraction.py:25:17: F541 f-string without any placeholders
    Error: examples/other/background_subtraction.py:73:9: F841 Local variable `material` is assigned to but never used

And all examples need directives for the docs/tests (below the docstring):
```py
# sphinx_gallery_pygfx_test = 'run'
```

Korijn (approved):

Korijn (approved):

PR #1068, opened 2025-04-08
Ray tracing example

panxinmiao:
This PR implements a simple ray tracer and tests it on the well-known Cornell Box scene. This is an experimental demonstration of ray tracing with compute shaders in pygfx.

Ray tracing is a rendering technique distinct from rasterization. By simulating the physical behavior of light in a scene, it produces visuals that are closer to real-world lighting, with natural support for soft shadows and global illumination. This makes it particularly well-suited for offline rendering scenarios that demand high visual fidelity, such as film-quality rendering.

Below are the rendering results of the Cornell Box scene at different sample-per-pixel (spp) levels:

1000 spp:
image

10,000 spp:

100,000 spp:
image

Replacing the small cube in the scene with a transparent glass and applying a mirror-like surface to the front of the big rectangular box, we can get the following rendering result (around 30,000 spp):
image

I think this example serves as an exploratory starting point for implementing ray tracing with Pygfx.  Perhaps in the future, we can have a powerful offline renderer using ray tracing technology in Pygfx :)


Vipitis:
Cool stuff, worth noting that we will eventually get access to hardware raytracing (on supported hardware)  in wgpu: https://github.com/gfx-rs/wgpu/issues/6762

panxinmiao:
> Cool stuff, worth noting that we will eventually get access to hardware raytracing (on supported hardware) in wgpu: [gfx-rs/wgpu#6762](https://github.com/gfx-rs/wgpu/issues/6762)

Thank you for the information. I'm also keeping an eye on this :)

PR #1072, opened 2025-04-16
Add LineInfSegmentMaterial for infinite lines

almarklein:
The purpose is to support the rendering of infinite lines.

## Approach

I first considered allowing `inf` and `-inf` values for the vertices, but we could probably make work, except this does not go well when transforms are involved.

So instead, the idea is to provide two points that define the vector along which the infinite line is rendered. This looks similar to the `LineSegmentMaterial`; you create multiple infinite line segments. Thus the name `LineInfSegmentMaterial`.

What happens in the shader is that the vertex is displaced along the vector, in NDC coordinates so it sits exactly at the edge of the screen.

## Tasks

* [x] Added `LineInfSegmentMaterial`
* [x] Rename to `LineInfiniteSegmentMaterial`?
* [x] Added a shader for it.
* [x] Make dashing work.
* [x] Document what does and what does not work, and how. For e.g. per-vertex coloring and thickness. 

----

image


kushalkolar:
Thanks this looks great!! 🥳

almarklein:
> I wonder if this could also help simplify and improve performance of the Ruler since endpoints no longer have to be calculated, just use a basis vector and an offset.

Endless rulers! Will still have to calculate the ticks and labels though ...

hmaarrfk:
Do you consider «  one sided » infinite lines?

almarklein:
> Do you consider «  one sided » infinite lines?

I was wondering when someone was going to suggest that :P  We could add a prop on the new material.

mmm, maybe even just add `LineSegmentMaterial.infinite_end` which can be 'start', 'end' and 'both' 🤔

kushalkolar:
> Document what does and what does not work, and how. For e.g. per-vertex coloring and thickness

Does it make sense to just use a uniform for the color? But in the example I see color mode is "face" and you're using a cmap O_O. What do faces of a line mean? :laughing: Unless you mean the triangles that represent the line segments that join the vertices?

almarklein:
> Does it make sense to just use a uniform for the color?

The line shader is one big shader, so the inf lines inherit all the things, like dashing, per-vertex coloring etc. But since we move the vertices, some of these features now work a bit different.

> What do faces of a line mean?

It's the whole area between two points/nodes of the line. Maybe segment would be a better name, but we took the definition from the mesh object.

Korijn:
Maybe it helps for your understanding to clarify that although you specify points, the shader is actually generating triangles (which have faces) to render the lines.

Korijn (approved):

PR #1074, opened 2025-04-16
Increase upper bound to wgpu version

almarklein:


hmaarrfk:
is there any reason why I shouldn't backport this change to pygfx 0.9.0

almarklein:
Well, except that the change of `GPUAdapterInfo` breaks cases where a dict is expected. Looking into it.

almarklein:
See https://github.com/pygfx/wgpu-py/pull/709

hmaarrfk:
> See https://github.com/pygfx/wgpu-py/pull/709

does this mean that 0.22.0 should be considered "broken" or will it work with pygfx 0.10.0?

almarklein:
> does this mean that 0.22.0 should be considered "broken" 

Only for the diagnostic reports, as far as I can tell.

hmaarrfk:
Ah. I see thanks. I might experiment. 

We do print the diagnostics in one of our windows lol

Korijn (approved):

Korijn (approved):

PR #1076, opened 2025-04-17
`Camera.get_state()` returns new copy of the array instead of a view

### LINKED ISSUE #1073: broken example
controllers no longer have a `{save,load}_state` method, do they?  if not, this example is out-of-date:

https://docs.pygfx.org/stable/_gallery/feature_demo/panzoom_camera.html

relatedly, what is the current way to reset the pan/zoom of a PanZoomController?

kushalkolar:
fixes #1073 

For `Camera.get_state()` to return a dict that can be used to restore states it must return a copy of the arrays instead of a reference that shares the same buffer. I cannot think where this might be detrimental. I would think that the typical use of `get_state()` is to _get and keep_ a snapshot of the camera state so it makes sense to return a copy of the arrays.

Also updates the example, enabling & disabling the controller is necessary to cancel any in-progress actions, see vid for what I mean.

https://github.com/user-attachments/assets/4d6d50c6-285b-4b87-a643-f81528a56a38

Bonus feature: if you guys think it makes sense we can have a `Controller.pause` decorator which can be useful in many situations.

```python
def pause(self, func):
    def wrapper(*args, **kwargs):
        state = self.enabled
        self.enabled = False
        func(*args, **kwargs)
        self.enabled = state

    return wrapper
```

It can then be used like this:

```python
@controller.pause
def do_stuff_while_controller_paused()
    ...
```

### DIFF
--- a/examples/feature_demo/panzoom_camera.py
+++ b/examples/feature_demo/panzoom_camera.py
@@ -57,13 +57,10 @@ def on_key_down(event):
 
     if event.key == "s":
         camera_state = camera.get_state()
-        print("save")
     elif event.key == "l":
         camera.set_state(camera_state)
-        print("load")
     elif event.key == "r":
         camera.show_object(scene)
-        print("reset")
 
     controller.enabled = True
 
--- a/examples/feature_demo/panzoom_camera.py
+++ b/examples/feature_demo/panzoom_camera.py
@@ -41,13 +41,33 @@ camera.show_object(scene)
 controller = gfx.PanZoomController(camera, register_events=renderer)
 
 
+# initial camera state
+camera_state = camera.get_state()
+
+
 def on_key_down(event):
+    if event.key not in ["s", "l", "r"]:
+        return
+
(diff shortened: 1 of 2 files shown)


almarklein:
Oh, nice catch! The idea is indeed that you can snapshot the state, so yes, they should be copies.

I like the idea of that decorator. Though now that I think of it, a context manager might be more appropriate?

Korijn:
Good call, this is the effect of our changes to the Transform API. It used to return copies by default but this was very detrimental for performance (since these properties are read so often). Copies now have to be made explicitly.

Korijn (approved):

kushalkolar:
> I like the idea of that decorator. Though now that I think of it, a context manager might be more appropriate?
> ```py
> with controller.pause():
>   ...
> ```

I agree this is better, will do in separate PR

almarklein (approved):

almarklein:
Thanks!

PR #1077, opened 2025-04-22
BF: register_wgsl_loader accepts package path instead of callable function.

maharshi-gor:
- This eliminates the func requirement and introduces the package path
- The function resulted in the error saying function does not have attribute `load` as it was expecting a `PackageLoader`.
- The PrefixLoader expects dict of `[str, PackageLoader]` which this function will serve.

Hi,
I saw this function while adding include in the my wgsl files in another wgsl file like pygfx.std.wgsl But when I try the above function it gives me the error explained above. I realized that we accept a func but never use it in the code. 

I am creating this PR so it becomes easy to add downstream wgsl package in the loader.

Let me know if you any other concerns.

### DIFF
--- a/pygfx/renderers/wgpu/shader/templating.py
+++ b/pygfx/renderers/wgpu/shader/templating.py
@@ -14,7 +14,7 @@ jinja_env = jinja2.Environment(
 )
 
 
-def register_wgsl_loader(context, func):
+def register_wgsl_loader(context, package):
     """Register a source for shader snippets.
 
     When code is encountered that looks like::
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks for your contrinution!

Can you please provide a minimal example of how you're trying to use `register_wgsl_loader`, so I can see how it fails? Using a function instead of a package-name was deliberate to allow more flexibility for downstream libs.

maharshi-gor:
Hello @almarklein 

Thank you for the feedback!

I tried following signatures:
```
def register(name):
    return "Name of the package."
register_wgsl_loader("fury", register)
```

```
def register(name):
    return jinja2.PackageLoader("Name of the package.")
register_wgsl_loader("fury", register)
```

Both of these resulted in the error function does not have attribute `load`. As I was trying to understand the error I saw the loader.mapping is directly set the func, instead of calling the function.

The value of the loader mapping i.e. `loader.mapping[context]` should be a type of `jinja2.PackageLoader`. But we are setting a function. Thus getting the error function does not have attribute load as it is the attribute of the package loader class.

We can change the definition back to making it a callback function but we need to call the function and extract the package loaders that the user is providing. 

Also, we might need to explain that the return type of the function should be a PackageLoader or list of PackageLoader. 

Currently, while doing the PR i took the inspiration from the line #4 where we define the `pygfx` context.

almarklein:
Sorry for the late reaction; I was away on holiday.

I looked into this issue closer, and indeed it looks like `register_wgsl_loader` is completely broken! It's simply never used. I created #1083 to match the implementation with the intended behavior.

Thanks for reporting this issue!

PR #1080, opened 2025-04-30
add `Controller.pause` context manager

kushalkolar:


almarklein (approved):

pygfx/controllers/_base.py
```diff
@@ -153,6 +154,29 @@ def enabled(self, value):
...
+            # do things while controller.enabled is False
+
+        # outside context manager, controller.enabled is set back to the original value
+
+        """
+        is_enabled = self.enabled
+
+        self.enabled = False
+
+        yield
```
Korijn, line 176: I would suggest to add try/finally to ensure line 178 always runs even if an exception occurs.

almarklein: wow, I assumed that the `@contextmanager` handled that ... but it does not:
```py
from contextlib import contextmanager

@contextmanager
def foo(x):
    print(f"enter {x}")
    yield
    print(f"exit {x}")


with foo(1):
    print("test 1")

with foo(2):
    print("test 2")
    1/0
```

prints
```
enter 1
test 1
exit 1
enter 2
test 2
  File "<tmp 1>", line 46, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero
```

Korijn: Yeah... try/finally is a pattern I apply by default when writing context managers!

Vipitis: there is some more weirdness with finally and return - but that will finally change in 3.14: https://peps.python.org/pep-0765/

Korijn: That is really surprising, I never knew about that... I don't think it works that way in other languages... :/

pygfx/controllers/_base.py
```diff
@@ -153,6 +154,29 @@ def enabled(self, value):
...
+
+        self.enabled = is_enabled
```
almarklein, line 178:
```suggestion
        try:
            yield
        finally:
            self.enabled = is_enabled
```

pygfx/controllers/_base.py
almarklein: I added your suggestion.

pygfx/controllers/_base.py
kushalkolar: Thanks! I forgot about this PR

Korijn (approved):

PR #1083, opened 2025-05-06
Properly implement `register_wgsl_loader`

almarklein:
Ref #1077

We implemented `register_wgsl_loader()` with the idea to let downstream code register their own loader. It seems like the implementation was a stub that was never properly finished: there is no test, we don't use it ourselves, and it's simply broken.

This PR fixes that, allowing downstream code to provide either a `jinja2.BaseLoader`, a function, or a dict.


maharshi-gor:
Thank you @almarklein for the work this will be really helpful.

Korijn (approved):

PR #1091, opened 2025-05-20
Uniform buffers can now handle vec3 and matnx3

almarklein:
Ref #999

This adds support for `vec3` and `mat2x3`, `mat3x3`, and `mat4x3`, which need more care to handle alignment properly (which is why they were previously not supported). For the matrices, padding is needed, because the GPU wants to read the matrix in one block. For the `vec3`, a scalar field can be put next to it to it. Doing this results in more efficient padding, which can result in a slightly smaller buffer.

Also took `f2` (16bit float) fields into account to some degree, so that we can add them relatively easy later.

### DIFF
--- a/pygfx/utils/__init__.py
+++ b/pygfx/utils/__init__.py
@@ -88,6 +88,16 @@ def array_from_shadertype(shadertype, count=None):
     }
 
     class Field:
+        __slots__ = [
+            "align",
+            "array_size",
+            "intrinsic_padding",
+            "name",
+            "primitive",
+            "shape",
+            "size",
+        ]
+
         def __init__(self, name, format):
             self.name = name
             if format[-2:] not in primitives:
--- a/pygfx/utils/__init__.py
+++ b/pygfx/utils/__init__.py
@@ -80,43 +80,138 @@ def array_from_shadertype(shadertype, count=None):
     assert isinstance(shadertype, dict)
     assert count is None or count > 0
 
+    # Note: we have yet to add f2 (16 bit float)
     primitives = {
         "i4": "int32",
         "u4": "uint32",
         "f4": "float32",
     }
 
+    class Field:
+        def __init__(self, name, format):
+            self.name = name
+            if format[-2:] not in primitives:
+                raise RuntimeError(
+                    f"Values in a uniform must have a 32bit primitive type, not {format}"
+                )
+            self.primitive = primitives[format[-2:]]
+            primitive_size = 4  # no support for f2 yet
+            # Get shape, excluding array part
+            arraystr, _, shapestr = format[:-2].rpartition("*")
+            shape = [int(i) for i in shapestr.split("x") if i]
+            shape = shape or [1]
+            # Include array size
+            self.array_size = None
+            if arraystr:
+                array_names.append(name)
+                self.array_size = int(arraystr)
+            # Calculate size and alignment, in bytes
+            self.shape = shape = tuple(int(i) for i in shape)
+            self.size = int(np.prod(shape)) * primitive_size
+            self.align = shape[-1] * primitive_size
+            # Snap align to factors of 2.
+            # 3xf2 -> 6 -> 8
+            # 3x3xf4 -> 12 -> 16
+            for ref_align in [2, 4, 8, 16]:
+                if self.align <= ref_align:
+                    self.align = ref_align
+                    break
+            # Matrices of shape nx3 need padding. As in the GPU wants to read more bytes than needed,
+            # so we can't fill that space up with other fields, but need empty bytes.
+            self.intrinsic_padding = 0
+            if len(shape) > 1 and shape[-1] == 3:
+                full_size = int(np.prod(shape[:1] + (4,))) * primitive_size
+                self.intrinsic_padding = full_size - self.size
+                self.size = full_size
+
+        def use(self):
+            nonlocal pad_index
+            shape = self.shape
+            if self.array_size is not None:
+                shape = (*shape, self.array_size)
+            shape = tuple(reversed(shape))
+            result = []
+            result.append((self.name, self.primitive, shape))
+            if self.intrinsic_padding:
+                pad_index += 1
+                result.append(
+                    (f"__padding{pad_index}", "uint8", (self.intrinsic_padding,))
+                )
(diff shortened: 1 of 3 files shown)


pygfx/utils/__init__.py
```diff
@@ -80,43 +80,138 @@ def array_from_shadertype(shadertype, count=None):
...
 
+    # Note: we have yet to add f2 (16 bit float)
     primitives = {
         "i4": "int32",
         "u4": "uint32",
         "f4": "float32",
     }
 
+    class Field:
+        def __init__(self, name, format):
```
Korijn, line 91: recommend to add __slots__ to this thing, will make it a lot faster.

Korijn (approved):

almarklein:
Note that a 3x3 still takes the size of a 3x4 (the gpu requires it), but at least you don't have to worry about the packing/unpacking.

hmaarrfk:
ah good point! Yeah it might be worth while just for the simplicity.....

thanks for the note!

panxinmiao:
Uh, I think the handling of nx3 matrix layout here is incorrect.
The current implementation adds  padding at the end of the entire matrix. However, according to WGSL and standard shader layout rules, each column should be aligned and padded individually.

For example: In the current approach, a 3x3xf4 matrix is treated as a contiguous 36-byte block (3×3×4), with 12 bytes of padding added at the end:
```
[3x3 matrix values (36 bytes)] [padding (12 bytes)]
```

But the correct layout should align each column as a vec4, meaning each column of 3 f32 values should be individually padded with 4 bytes, resulting in:
```
[3 floats (12 bytes)] [implicit padding (4 bytes)]
[3 floats (12 bytes)] [implicit padding (4 bytes)]
[3 floats (12 bytes)] [implicit padding (4 bytes)]
```

panxinmiao:
I've thought about this issue before, but realized that solving it properly isn’t as straightforward—so I temporarily put it aside. 😅 

To address the problem thoroughly, I believe we need to introduce a dedicated class to represent WGSL uniform objects (e.g., UniformBuffer). This would also offer much better flexibility.

Currently, in objects like Material, we simply use a class-level dictionary (uniform_type) to define the uniform data, and map it directly to an ndarray with a corresponding dtype. I think this is the root of the problem—we're using NumPy’s ndarray both to represent the WGSL struct's logical data layout and its underlying memory layout. We then construct WGSL structs based on the fields and shape defined in the dtype, but this conflates memory layout with data type structure. In WGSL, memory layout and data structure are not always in one-to-one correspondence. For example, mat3x3 and mat3x4 share the same memory layout (both are 48 bytes, arranged as a (4,3) shape), but they represent different data types.

In my proposal, the UniformBuffer class could still be backed by a NumPy ndarray, but should provide higher-level APIs and properties that directly comply with WGSL semantics—rather than exposing raw ndarray interfaces.

Users should be able to dynamically add or remove fields from a UniformBuffer, and access or update uniform values directly by field name. The engine would be responsible for memory layout and alignment, GPU resource creation and binding, and auto-generating the corresponding WGSL code from the UniformBuffer.

Additionally, Material should have an instance-level UniformBuffer, rather than a static class-level one. It should be constructed dynamically, since in practice, different instances of the same material type may require different uniforms. For advanced materials, many properties are optional, and uniforms should only be added when those properties are present—i.e., when certain shader subroutines are actually being used.

almarklein:
> The current implementation adds padding at the end of the entire matrix. However, according to WGSL and standard shader layout rules, each column should be aligned and padded individually.

Doh! :facepalm: this crossed my mind, but not hard enough to look closer.  Thanks for figuring this out!

>  In WGSL, memory layout and data structure are not always in one-to-one correspondence

From what I can tell, this only applies to nx3 matrices, right?  A solution could be to require users to still submit an nx4 matrix in these cases. In most cases it would be the a pygfx class that will have to deal with this.

> Users should be able to dynamically add or remove fields from a UniformBuffer,

With this I think you mean that the *material* can dynamically add/remove fields? E.g. when `alpha_test` becomes nonzero add the value for `alpha_test` as a field to the buffer. (alpha test is not supported yet, but it was the first thing that came to mind)

> [...] access or update uniform values directly by field name.

Do you have anything specific in mind with this? Like exposing the uniform buffer to the user?

panxinmiao:
> From what I can tell, this only applies to nx3 matrices, right? A solution could be to require users to still submit an nx4 matrix in these cases. In most cases it would be the a pygfx class that will have to deal with this.

Yes, but the issue is that in this case, the underlying data layout of an nx3 matrix and an nx4 matrix is completely identical (according to the WGSL specification, their layout is indeed the same). Since we're directly using an ndarray to represent the uniform, they end up having the same shape. As a result, when the engine generates WGSL code, it can't determine whether the matrix is a mat3x3 or a mat3x4 based solely on the layout. This is the crux of the problem: we've lost the original type information, and WGSL ends up using mat3x4, because the engine has no way to distinguish it from a mat3x3 just by looking at the layout.

If we want to preserve and carry over the original type information, we need a custom Uniform class (which uses an ndarray internally for data storage) instead of relying on a plain NumPy ndarray.

> With this I think you mean that the _material_ can dynamically add/remove fields? E.g. when `alpha_test` becomes nonzero add the value for `alpha_test` as a field to the buffer. (alpha test is not supported yet, but it was the first thing that came to mind)

Yes, for example, many properties in MeshPhysicalMaterial—such as clearcoat, iridescence, sheen, and others—are optional. The uniform values related to these features are only needed when the corresponding features are enabled. However, currently all instances of MeshPhysicalMaterial share the same uniform structure, including all these fields, regardless of whether they are actually used or not.

> Do you have anything specific in mind with this? Like exposing the uniform buffer to the user?

My idea is to introduce a dedicated UniformBuffer class that encapsulates these low-level details. For example, a field in this class might represent a mat3x3 type in WGSL, which is internally stored as a NumPy ndarray with shape (4, 3), but is exposed to the user as a standard 3×3 matrix—both for reading and writing. For most users, constructing a (4, 3) shaped array from a 3×3 matrix is not intuitive, especially when they also have to consider row-major vs. column-major layouts and possible transposition. 
And, users also can freely manage the fields within the UniformBuffer according to their own use cases, rather than being constrained by pre-defined structures in classes like Material.

PR #1092, opened 2025-05-20
Add option to ScreenCoordsCamera to invert y

almarklein:
cc @kushalkolar 

This should do the trick. One issue I saw is that text drawn with this camera is upside down 🤔


almarklein:
>  what if the y_scale is inverted on the world transform?

That does the trick! I updated one example to use this.

Korijn (approved):

PR #1099, opened 2025-05-26
Render lines using SDF

almarklein:
## Problem I try to solve

When rendering a semi-transparent line, relatively sharp corners result in 'brighter' parts because there is overlap:

image

This happens, because at some angle, the vertices can no longer be placed so that the line is continuous. What angle this is, depends on the length of the two segments and the line width. When the angle is two small, it 'breaks' the line causing overlap.

## The idea of the approach of this PR

The idea is to render each line segment as a separate quad, using an SDF. But within each quad, the fragment shader calculates the sdf for the current as well as the *next*  line segment. It will then subtract the color/alpha of the next fragment from the current color, so that the final result shows no overlap. The cool thing is that we can also make it look 2 segments further and/or turn the overlap on 'smoothly'.

## State of this PR

For now this is a POC that demonstrates that the basic idea works:

image

But I have ignored per-vertex colors, texcoords, dashing, loops, and probably more.

## Moving forward

I think the first next step is making sure this can work with colors and texcoords. If it does, this approach is viable IMO. Next up would be fixing the caps and loops (but I see no issues there).

As for dashing, I kindof hope we can use SDF trickeries for that as well, but I'm not sure yet whether this is possible. If not, or if it's hard and we want to postpone that, we use fall back to the old approach when there is dashing. 

One big advantage of this approach, if it can be made to work, is that it's actually quite simple as it has much less 'geometry-juggling' than the current implementation.

cc @hmaarrfk @kingbedjed @claydugo

### DIFF
--- a/pygfx/renderers/wgpu/shaders/lineshader.py
+++ b/pygfx/renderers/wgpu/shaders/lineshader.py
@@ -99,6 +99,8 @@ class LineShader(BaseShader):
         # ):
         #     # self["line_type"] = "quickline"
 
+        self["quickline"] = False
+
         # Handle looping. The line_loop_buffer is one larger to enable looping the last point.
         if material.loop:
             self["loop"] = True
@@ -300,15 +302,17 @@ class LineShader(BaseShader):
         }
 
     def get_pipeline_info(self, wobject, shared):
+        # Note: making quads using triangle_strip and 6 verts, with degenerate triangles is apparently
+        # much faster than using triangle_list with 6 verts.
         return {
-            "primitive_topology": wgpu.PrimitiveTopology.triangle_list,
+            "primitive_topology": wgpu.PrimitiveTopology.triangle_strip,
             "cull_mode": wgpu.CullMode.none,
         }
 
     def _get_n(self, positions):
         offset, size = positions.draw_range
-        if self["loop"]:
-            size += 1
+        if not self["loop"]:
+            size -= 1
         return offset * 6, size * 6
 
     def get_render_info(self, wobject, shared):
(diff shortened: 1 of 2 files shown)


hmaarrfk:
On my run today I was thinking of using SDFs for this as well.... crazy the coincidence.

almarklein:
This weekend I was breaking my head thinking of different ways to approach the problem, when my 2yo took my pen and started scribbling on my drawings, and then suddenly I had to think of subtracting SDF's 💡 . The mind is weird.

almarklein:
> Isn't the union of two SDFs `min(sdf1, sdf2)`?

Yeah, but we're subtracting at the level of the fragment color. Basically, if you know that the next segment will draw the fragment too, you can omit it to prevent double-drawing it. 

> I also wanted to ask: I guess all of this only works if the lines drawn are entirely 2D?

Good point, but I think it can still work if we only subtract/discard dubplicate fragments when they're the one furthest from the camera.

almarklein:
Update: My enthusiasm about this approach is going a bit back and forth 😄   On one hand, it does not completely remove overlap, it just prevents overlap of successive line segments. If the line is very "noisy", you'll still get overlap. On the other hand, in a line shape like this:

image

the new method "never" has overlap, while in the current method it depends on the zoom level (more specifically, the thickness in relation to the length of the segments) whether there is overlap.

To do it right, you have to look at the next *and* the previous line piece. This means that we need to sample 4 elements from the positions buffer, instead of 3 as the current method does. I just did some quick benchmarks and it looks like the new method is actually faster in most use-cases! This is likely because the fragment shader is simpler. 

All in all, I think it's worth to move this forward.

hmaarrfk:
i was working on a little UI to better understand the existing algorithm. i'll try to share tonight!

PR #1104, opened 2025-06-01
Drop Python 3.9 because its already broken, and bump CIs

hmaarrfk:
The syntax added in https://github.com/pygfx/pygfx/pull/960 makes it incompatible with python 3.9

lets move on?

### DIFF
--- a/examples/feature_demo/image_click_events.py
+++ b/examples/feature_demo/image_click_events.py
@@ -38,7 +38,9 @@ yy = [145, 131, 112, 59, 29, 14, 48, 91, 136, 137, 172]
 colors = np.vstack([[0.0, 0.0, 1.0, 1.0]] * len(xx)).astype(np.float32)
 
 points = gfx.Points(
-    gfx.Geometry(positions=[(x, y, 0) for x, y in zip(xx, yy, strict=True)], colors=colors),
+    gfx.Geometry(
+        positions=[(x, y, 0) for x, y in zip(xx, yy, strict=True)], colors=colors
+    ),
     gfx.PointsMaterial(color_mode="vertex", size=10),
 )
 points.local.z = 1  # move points in front of the image
--- a/pygfx/objects/_more.py
+++ b/pygfx/objects/_more.py
@@ -384,7 +384,9 @@ class Volume(WorldObject):
         values = unpack_bitfield(pick_value, wobject_id=20, x=14, y=14, z=14)
         texcoords_encoded = values["x"], values["y"], values["z"]
         size = tex.size
-        x, y, z = [(v / 16383) * s - 0.5 for v, s in zip(texcoords_encoded, size, strict=True)]
+        x, y, z = [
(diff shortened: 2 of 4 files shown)


examples/feature_demo/scene_subplots_video.py
```diff
@@ -85,15 +85,15 @@ def animate():
         img.geometry.grid.update_full()
 
     # render the viewports
-    for viewport, s, c in zip(viewports, scenes, cameras):
+    for viewport, s, c in zip(viewports, scenes, cameras, strict=True):
```
hmaarrfk, line 88: i feel like "strict=True" is the correct way to zip is that right?

hmaarrfk:
Did my deletion of the jobs make this "impossible to report"?
image

Korijn:
> Did my deletion of the jobs make this "impossible to report"?

No there are branch protection rules that require certain jobs to pass, and it's a literal whitelist of job names. I don't like it but that's how GitHub works.

hmaarrfk:
> No there are branch protection rules that require certain jobs to pass, and it's a literal whitelist of job names. I don't like it but that's how GitHub works.

Ok i guess the rest is on you. just wanted to make sure i did everything I could here.

examples/feature_demo/scene_subplots_video.py
almarklein: I agree that it's the more correct way. I wish it was the default. Since it's also more verbose I personally don't care in cases were you know both iterables are the same length ... but I'm fine with these changes 😉

almarklein (approved):
For reference: Python 3.9 is scheduled to be EOL somewhere this year. The [scientific pythpn spec 0](https://scientific-python.org/specs/spec-0000/) is on 3.11+

examples/feature_demo/scene_subplots_video.py
hmaarrfk: Just fyi. Ruff complained without this

jacopoabramo:
> For reference: Python 3.9 is scheduled to be EOL somewhere this year. The [scientific pythpn spec 0](https://scientific-python.org/specs/spec-0000/) is on 3.11+

Sorry for the intrusion; are you planning to adhere to SPEC 0?

almarklein:
> Sorry for the intrusion; are you planning to adhere to SPEC 0?

No, I generally like to support Python versions longer than that spec. And e.g. wgpu-py should IMO follow the Python lifetimes.

hmaarrfk:
Just FYI.

I would recommend your merge this, release 0.10.1, and yank 0.10.0 on pypi since it technically declares itself as compatible.

Alternatively, you could restore Python 3.9 compatibility by using `Union` (I find `Union` to be nicer syntax than `|` but maybe i'm just old....)

if you leave 0.10.0 as the latest "python 3.9" on pypi, everybody on Python 3.9 will have a "miserable experience".

Now, I think people shouldn't be using Python 3.9 in 2025, but who am I to stop them....

almarklein:
@Korijn can you approve?

Korijn (approved):

PR #1109, opened 2025-06-06
Scene and Group accept childern on init

almarklein:
Just a s small convenience so you can create objects, and then wrap them in a scene in one call. 

I did not add tests, but I'm writing tests in #1002 that uses it ...

### DIFF
--- a/pygfx/objects/_more.py
+++ b/pygfx/objects/_more.py
@@ -22,8 +22,10 @@ class Group(WorldObject):
 
     """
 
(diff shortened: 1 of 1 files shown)


Korijn:
Feels weird to only do it on these objects. Sets expectations that other objects support it too.

I think it's better that we teach users to use add.

almarklein:
Fair enough. One can always do `gfx.Scene().add(ob1. ob2)`

PR #1110, opened 2025-06-08
Increasing the GUI thick lines examples to allow it to be an interactive

hmaarrfk:
GUI

@almarklein I think we should be able to "extend" the validity of the standard line shader.

This is my visualization GUI that I was speaking about efore.


almarklein:
>  "extend" the validity 

What validity do you mean?

hmaarrfk:
More like

- not changing to SDF
- using more mathematical if statements to keep things working the old way. 

But this interactive example has discouraged me a bit. 

We’ll see I have to explore a bit more. 

Truthfully, the basic line material might be all I need for my most immediate demo.

Issue #181, opened 2021-10-15
Shadow: ambient occlusion / SSAO


berendkleinhaneveld:
Would be great to have shadows working. It can improve the perception of 3D objects.

almarklein:
We have shadows since #324. Let's keep the issue to track ambient occlusion, I'll update the title accordingly.

I think we can still have SSAO, if our shaders store depth and normals for opaque fragments. We'll probably have to pretend that non-opaque surfaces led through all ambient light. I think that's ok in practice.

Issue #75, opened 2020-10-15
FXAA


almarklein:
FXAA is an antialiasing technique that smoothes an image along the direction of the edge. It is very cheap (performance-wise) and gives moderate results. Can be nice to combine with SSAA.

Years back I looked at FXAA and felt that it can be improved by using scale-space (Gaussian) kernels. Maybe dive into that as well ...

Issue #1065, opened 2025-04-02
Blending refactor and improvements


almarklein:
A meta-issue to track the refactoring and further improvements to the blending.

### Tasks

* [ ] Initial refactoring: #1002 
* [ ] Clean up
  * Move what's left in `blender.py` to `renderer.py`?
  * Simplify `pipeline.py` and `environment.py`.
* [ ] Transmissive transparency #974
* [ ] Alpha test
* [ ] Dither can feel less noisy if the noise is static, which we can get with an object-id that is an index in the scenegraph, plus some element index. See ST paper.
* [ ] A form of post-processing AA (like FXAA, but better)
* [ ] Set ``material.aa`` off by default, avoiding a range of artifacts.

### Related 

* #989
* #1003
* #75 
* #901

PR #1114, opened 2025-06-13
Add support for post-processing effects

### LINKED ISSUE #72: API for adding fog and custom full-screen post-processing effects
In #71 we added the ability to add post-processing effects to the rendered image. 
```py
renderer.post_processing_steps.append(NoisyPostProcessingStep(renderer.device))
```

We should give this API some more thought. Maybe alternatives work better.

One idea could be to just use our existing tools, and create a new renderer that takes the result of the first renderer and renders it using geometry that maps it to the full viewport. But that will result in a lot of boilerplate?

Another idea: these post processing classes, like `NoisyPostProcessingStep` look a lot like materials, in that they define the shaders. Maybe they can even be specific kind of materials?

In all of this we should be aware that these post-processing steps are specific to the wgpu renderer backend.

almarklein:
* Adds `EffectPass` base class.
* Adds `CopyPass`, `NoisePass`, `FogPass`, `DepthPass` builtin passes.
* The flusher also inherits from `Effectpass`  and is now called `OutputPass`.
* Closes #72
* Paves the way for #181
* Paves the way for post-processing aa #75
* Enables tons of new effects for uses to contribute!

In ThreeJS post-processing is an addon. The API proposed here is more integrated with the renderer, making it simpler to add effects to an existing visualization. 

Screen Recording 2025-06-13 at 13 00 04

### DIFF
--- a/pygfx/renderers/wgpu/engine/blender.py
+++ b/pygfx/renderers/wgpu/engine/blender.py
@@ -591,7 +591,7 @@ class Blender:
             @fragment
             fn fs_main(varyings: Varyings) -> @location(0) vec4<f32> {
                 let epsilon = 1e-6;
-                let texIndex = vec2i(round(varyings.position.xy));
+                let texIndex = vec2i(varyings.position.xy);
 
                 // Sample
                 let accum = textureLoad(r_accum, texIndex, 0).rgba;
--- a/pygfx/renderers/wgpu/engine/effectpasses.py
+++ b/pygfx/renderers/wgpu/engine/effectpasses.py
@@ -338,13 +338,16 @@ class EffectPass(FullQuadPass):
         fn fs_main(varyings: Varyings) -> @location(0) vec4<f32> {
 
             // Available variables:
-            // colorTex - the texture containing the rendered image, or the previous effect pass
-            // depthTex - the texture containing the renderd depth values
-            // texSampler - a sampler to use for the above
-            // varyings.position - the position in physical pixels (a vec3f).
+            // colorTex - the texture containing the rendered image, or the previous effect pass.
+            // depthTex - the texture containing the renderd depth values.
+            // texSampler - a sampler to use for the above.
+            // varyings.position - the position in physical pixels (a vec4f).
             // varyings.texCoord - the coordinate in the textures (a vec2f).
             // u_effect.time - the current time in seconds, changes each frame.
             // u_effect.xx - whatever uniforms you added.
+
+            // Calculate the pixel index, e.g. if you want to use textureLoad().
+            let texIndex = vec2i(varyings.position.xy);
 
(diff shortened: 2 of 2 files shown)


almarklein:
Ready for review. I think I made a small change in the output pass (formerly flusher) causing the screenshot comparison tests to fail.

almarklein:
And CI is green!

Korijn (approved):
Awesome to finally have access to the depth buffer here :) looks great!

hmaarrfk:
Really cool. One of the effects I’ve wanted has been a « grayscale » effect. 

It helps us generate assets that are 1/3 in size 

@claydugo and I have been playing with workarounds (mostly doing the conversion on the CPU) for the grayscale filter.

almarklein:
Also some ideas for more effects:

https://github.com/pygfx/pygfx/blob/fa71a9369733323d9b62e39b9ef6b5408bb7a6f5/pygfx/renderers/wgpu/engine/effectpasses.py#L521-L527

Vipitis:
Feel like some potential for interop with existing shadertoys. Especially those that read a texture (or two for depth?) and then draw to screen.

almarklein:
> Feel like some potential for interop with existing shadertoys. Especially those that read a texture (or two for depth?) and then draw to screen.

Yeah, I guess the post-processing effects is like Shadertoy Light (TM) 😉.

They can actually have more textures as input (and also output); the base class makes no assumptions about the number of render and target textures. In terms of API, the `EffectPass` subclass should likely have properties that allow users to set a texture.

Minigoer:
Hello, 
Does renderer`.snapshot()` capture `EffectPass` ? 
In the viewport I can see the fog pass, 
but when I try to save several `.snapshot` as numpy arrays to make a video with imageio, the effect is not (yet?) rendered.
I know snapshot not exactly save the viewport, as it "samples from the internal texture, so the result may be different then what's shown on screen."

Anyway, thanks for your excellent library.

almarklein:
@Minigoer Good question! This is something that has crossed my mind, and which was considered in the design, but I have not addressed snapshot yet. Thanks for bringing it up. I'll add an entry in #1065.

The way that post-processing is implemented, is that it pingpongs between the texture that it renders to, and an alt texture. So snapshot cannot capture the original rendered image, but it can/should capture the image including post-processing. For what it's worth, I also considered the option where snapshot captured the render result without effects, but that made the code more complex and requires an additional render texture, so I decided against it.

Anyway, to fix this issue, the renderer should keep track of what the "final" texture is to take the screenshot from. 

In the mean time, as a workaround, you can add a `CopyPass` if needed, so the final effect is in the original texture again 😃

Minigoer:
@almarklein , it does the trick very well ! Thanks a lot !

PR #1116, opened 2025-06-13
Cloth Simulation with Compute Shaders

panxinmiao:
This is an example implements a simple cloth simulation with a verlet syatem using compute shaders.

https://github.com/user-attachments/assets/2a158414-3a6d-49cd-87e9-6fb79dd946ce

### DIFF
--- a/examples/feature_demo/compute_cloth.py
+++ b/examples/feature_demo/compute_cloth.py
@@ -1,0 +1,703 @@
+"""
+Cloth Simulation with Compute Shaders
+=====================================
+
+This example demonstrates how to implement a Verlet integration cloth physics simulation using compute shaders,
+including a spring system, gravity, wind forces, and sphere collisions.
+"""
+
+import numpy as np
+import pygfx as gfx
+import wgpu
+import imageio.v3 as iio
+from wgpu.gui.auto import WgpuCanvas, run
+from pygfx.utils.compute import ComputeShader
+from imgui_bundle import imgui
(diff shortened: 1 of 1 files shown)


panxinmiao:
When writing compute shaders, `ComputeShader` util  is very convenient to use. 👍,  and it would be even better if the storage buffer and data layout could be more convenient.

In PR #1091, we added support for vec3 and mat3x3 in uniform buffers. However, in storage buffers, vec3 is still not supported, and we also lose data layout information. This leads to some inconvenience in usage. and, I thick there's likely a performance cost due to the need to re-pack data manually to accommodate the missing layout support.

examples/feature_demo/compute_cloth.py
```diff
@@ -0,0 +1,703 @@
...
+    sphere_y: f32,
+    sphere_z: f32,
+    sphere_active: f32,
+    wind: f32,
+    stiffness: f32,
+    time: f32,
+}
+
+@group(0) @binding(0) var<uniform> params: Params;
+@group(0) @binding(1) var<storage, read> positions: array<f32>;
```
panxinmiao, line 208: Here,  the positions buffer should actually be of type array<vec3<f32>>, However, due to data alignment reasons, we do not support it, so we can only flatten it. 
As you can see in the code below, it is not very convenient for both data reading and storage.

almarklein:
Cool example BTW, and nice to see another use-case for compute!

Korijn (approved):

PR #1117, opened 2025-06-13
Add support for FXAA and DDAA

### LINKED ISSUE #75: FXAA

almarklein:
Closes #75 

* Add `FXAAPass`  and `DDAAPass` effects.
* Add `renderer.ppaa`  property that can be set to 'default', 'none', 'fxaa', or 'ddaa'.
* This adds/removes the corresponding effect  to the `renderer.effect_passes`.
* Add `PYGFX_PPAA` to control what 'default' resolves to. The primary use-case is to turn it off for image tests.
 
----

Line without aa, on a regular (no-hidpi) screen, with pixel ratio 1:

afbeelding

With pixel ratio 2 (the trick we use by default):

afbeelding

With ddaa and pixel ratio 1:
afbeelding

More comparisons: https://almarklein.github.io/ppaa-experiments/viewer.html#image=lines.png&algs=ori,ssaax8,fxaa3,ddaa2

----

DDAA is bases on an idea I had in the early days of Visvis. I revived the idea and worked on it outside of "Pygfx hours". A small improvement in the perceived smoothness (without getting blurry) can be a big win, because it means we can turn the aa of objects off, which removes a whole range of artifacts, and maybe even reduce the ssaa (pixel_ratio).

Differences of DDAA w.r.t. FXAA:
* Better smoothing for diagonal-ish edges.
* Uses similar edge-search technique as FXAA, does not skip pixels, but (therefore) searches a bit less far with default settings.
* Better preservation of small structures such as thin lines and parts of glyphs.
* More performant than FXAA (with default settings).
 

For more info see https://github.com/almarklein/ppaa-experiments


almarklein:
I got stuck on #1118 when trying to create a validation example. Will proceed with this after #1121.

pygfx/renderers/wgpu/engine/effectpasses.py
```diff
@@ -493,6 +494,41 @@ def filter(self, filter):
...
+class DDAAPass(EffectPass):
+    """An effect pass implementing Directional Diffusion anti-aliasing.
+
+    DDAA produces better results than FXAA for near-diagonal lines, at the same performance.
+    It estimates the direction of the edge, and then diffuses (i.e. smoothes) in that direction.
+    For near-horizontal and near-vertical a technique similar to FXAA is used.
+    """
+
+    wgsl = """
+        {$ include 'pygfx.ddaa.wgsl' $}
```
Korijn, line 523: it seems like you forgot to include the file?

almarklein:
@Korijn  this is also ready, btw

PR #1120, opened 2025-06-16
Support alpha mode

panxinmiao:
Complete #989 

### Changed:
When `transparent` is set to `False`, the alpha value of the shader's output color should always be 1.0.

Currently, This scenario mainly occurs when alpha testing is used—where the alpha value does not represent transparency, but instead determines whether a fragment passes the alpha test and should be retained.


panxinmiao:
Moreover, I firmly believe that a material's transparency should not be automatically inferred from its alpha value. It is essential to explicitly specify whether a material is transparent. Alpha does not always represent transparency; it may be used as a mask or for alpha testing.

When transparent is `False`(most material case), blending in the render pipeline should be completely disabled to improve performance (We are not currently doing this). In this case, the alpha output is ignored (i.e., not used for blending) and should always be set to 1.0. Otherwise, in special scenarios—such as when alpha-to-coverage is enabled—an incorrect alpha value may lead to rendering artifacts. Similarly, if post-processing or a deferred rendering pipeline is used, an incorrect alpha value may cause inconsistencies in the G-buffer or transparency channels.

Transparency should be always controlled through explicit flags rather than implicit logic. Explicitly declaring `material.transparent = True/False` is a more robust and reliable approach. This is especially important in complex rendering pipelines.

almarklein:
> Transparency should be always controlled through explicit flags rather than implicit logic. Explicitly declaring material.transparent = True/False is a more robust and reliable approach. 

I disagree with the first statement, and agree with the latter statement. 

Starting with the latter, users can set `.transparency` and it should work as expected. I think loaders like the gltf one should always set `.transparency` to either True or False.

As for the first statement, I think it's counterintuitive to declare a material with an opacity < 1 and then *also* having to state that it's transparent.

Therefore, if the transparency is not set (i.e. None), I tried to make things behave as a user may expect. Partly to be intuitive, partly for backwards compatibility. In this case, the implicit `transparency` value is based on `.opacity`, which I think makes perfect sense. And for materials that also have a `.color`, the color's alpha is also used, but only when the `color_mode` is such that the material's color is used. I figured this made sense too, but I agree it's more arguable, and I am open to reconsider this. When alpha values are used as a  mask, I imagine it's done via the alpha values in the texture, right?

In any case, if you prefer the explicit approach, just always set `.transparency`, and there's no problem?

Korijn:
> tried to make things behave as a user may expect

I think this is the challenge here; different user groups have different expectations. I certainly wouldn't expect a transparency flag to be set automatically, because I am used to graphics library conventions.

But I can imagine a newcomer will not be aware of the intricacies of transparency and will expect this to be handled automatically.

panxinmiao:
> I think it's counterintuitive to declare a material with an opacity < 1 and then _also_ having to state that it's transparent.
> 

That makes sense. However, there's another important consideration: for opaque materials, the rendering pipeline does not need to enable blending at all. The transparent property of a material essentially serves as a flag to determine whether blending should be enabled in the render pipeline. Explicitly setting this property helps clarify that intent and ensures the correct rendering behavior.

> When alpha values are used as a mask, I imagine it's done via the alpha values in the texture, right?
> 

Yes, in this case, we need to set the alpha value of the output color to 1.0. In other words, under our current design, it's essential to explicitly set transparent to False—not None—even if opacity is 1. Otherwise, it may lead to unexpected rendering issues. I believe this could actually cause more confusion for users.

The issue is illustrated in Test Case: [**Alpha Blend Mode Test**](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/AlphaBlendModeTest/README.md)

almarklein:
> I think this is the challenge here

Exactly. I think it's going to be a bit awkward one way or another. I do appreciate that making a material opaque unless `.transparent` is set to True, is *really* easy to explain. It does minimize confusion. But it also feels "stupid" (or really low level?) that you have to do that when it's clear that the material is going to be transparent.

@panxinmiao So if I understand it correctly, your main concern then, is that opaque is not the default, right?

So right now the `.transparent` flag resolves into one of 3 possible values: False, True and None. With None being a bit weird state. I think we can make it work so that it at least resolves into either False or True (i.e. opaque and transparent). Then, the auto-behaviour makes the object opaque, unless `.opacity<0` or `.color.a<0`.

You're also saying that the `.blending` is only applied if the object is marked as transparent. Is that how ThreeJS works as well? I think we need some sort of decision tree of what props affect what (in ThreeJS and in Pygfx) ...

panxinmiao:
> But it also feels "stupid" (or really low level?) that you have to do that when it's clear that the material is going to be transparent.
> 

But I think that’s exactly where the problem lies — whether a material is transparent or not is often not as “obvious” to the engine as it might seem to us as developers.

This is indeed a question of design philosophy. One way to look at it is: the opacity property only conveys transparency if and only if the material is already marked as transparent. It shouldn’t be the property that determines whether the material is transparent.

In some cases, opacity and color.a are used purely for alpha testing, not for blending — for example, consider the case where opacity = 0.5 but transparent = False. This is valid if you're relying on alpha testing or masking, and blending is not desired.

While we can currently support this by explicitly setting transparent = False, I actually think that allowing transparency to be inferred from opacity or color.a might lead to more confusion, not less.

> So if I understand it correctly, your main concern then, is that opaque is not the default, right?
> 

Yes, my main concern is that materials should not remain in an "undefined" (None) transparent state by the time they reach the renderer. Regardless of whether we allow transparent to be inferred from other props earlier, the renderer needs to know definitively whether a material is opaque or transparent in order to choose the correct rendering behavor.

> You're also saying that the `.blending` is only applied if the object is marked as transparent. Is that how ThreeJS works as well?
 
Yes, in Three.js,  for non-transparent objects, blending will not be enabled unless the user explicitly specifies the blending method - this is actually the case for most materials, which is a sensible default settings for performance in most cases.

WebGL: https://github.com/mrdoob/three.js/blob/1f603c32b8a21d43102e9b31dc8bf271679784bc/src/renderers/webgl/WebGLState.js#L774-L776

WebGPU: https://github.com/mrdoob/three.js/blob/1f603c32b8a21d43102e9b31dc8bf271679784bc/src/renderers/webgpu/utils/WebGPUPipelineUtils.js#L120-L124

almarklein:
I've read up a bit about this a bit, and I feel the `.transparent` property, and how it affects things like blending, is actually confusing. I find the `alpha_mode` attribute of gltf and babylon to be much more intuitive. This may feel a bit of a side-track, but bear with me :)

What about:
* Refactor the `blending` property into `alpha_mode`:
  * "opaque"
  * "blend" (the over operator)
  * "add", "subtract", "multiply"
  * "dither"
  * "weighted"
  * "mask" (use alpha_test, which will now be 0.5 by default).
* Remove the `.transparent` property, it's now *blended* (pun intended) into `alpha_mode`.
 
So instead of setting `material.transparent = True` one sets `material.alpha_mode = "blend"`.

Since this prop defines how to handle the alpha, this *also* fixes the ambiguity when using alpha for masking 🥳 

As for automatically setting `alpha_mode`, I'd like to propose, as a compromise, to only make it depend on `.opacity`. I think we can train (sciency) users to use the opacity instead of a color with `a<1`. It's also easy to explain and document: *by default the alpha_mode is 'opaque', unless opacity < 1, in which case it's 'blend'.*

panxinmiao:
> I find the alpha_mode attribute of gltf and babylon to be much more intuitive

I think the transparency handling in Babylon, glTF, and Three.js are conceptually quite similar—they just express the separation between *transparency intent* and *blending method* differently.

In Babylon, the `transparencyMode` property maps as follows:

* `OPAQUE`: `transparent = false`, `alphaTest = 0`
* `ALPHATEST`: `transparent = false`, `alphaTest > 0`
* `ALPHABLEND`: `transparent = true`,  `alphaTest = 0`
* `ALPHATESTANDBLEND`: `transparent = true`, `alphaTest > 0`

Meanwhile, Babylon’s `alphaMode` refers specifically to the *blending function* (`ADD`, `MULTIPLY`, etc.), which is comparable to Three.js’s `blendMethod`.

> Refactor the `blending` property into `alpha_mode`...

In my opinion, this proposal conflates two orthogonal concerns:

* Whether a material is transparent or not (Babylon's `transparencyMode`, Three.js's `transparent` flag)
* How the blending should be performed (Babylon's `alphaMode`, Three.js's `blendMethod`)

While it may simplify the API surface, it introduces semantic ambiguity and could make it harder for users (and the renderer) to reason about the rendering path.

From the renderer’s perspective, it **still needs to clearly distinguish** opaque from transparent materials. Babylon checks `transparencyMode`, Three.js checks `transparent`. This distinction is essential for render ordering, depth writing, batching, and pipeline setup.

> As for automatically setting `alpha_mode` based on `.opacity`...

I think any form of implicit inference will still lead to confusion. Even in Babylon, `transparencyMode` must be explicitly set. Three.js requires an explicit `transparent = true`. These engines *could* have inferred the intent—but they intentionally don't, for good reason.

That said, I understand that `pygfx` can have its own philosophy. But if auto-inference is allowed, it's inevitable that edge cases will occur where what the user expects and what the engine infers do not align—that’s where confusion comes in.

What's most important, I believe, is that **by the time the renderer and shaders receive the material**, its transparency state must be clearly defined—no ambiguity, no `None`. Regardless of whether the API layer or material layer allows automatic inference, the underlying renderer must obtain a clear opaque/transparent judgment.

almarklein:
> In my opinion, this proposal conflates two orthogonal concerns:
> 
>    * Whether a material is transparent or not (Babylon's transparencyMode, Three.js's transparent flag)
>    * How the blending should be performed (Babylon's alphaMode, Three.js's blendMethod)

They are not orthogonal, because the `alphaMode` / `blendMethod` is only used depending on the state of the first point; the second point are extra details about how to use alpha values.

But I think I see where the confusion comes from. Perhaps a better way to explain my proposal is to say that `.alpa_mode` can have these values: *opaque, blend, mask, dither, weighted*.

Then, depending on the value, extra information can be provided (like blend equations for "blend", and special args for "weighted"). In the current implementation, the `material.blending` is actually a dict, with a field `mode` being one of the aforementioned values, and the other keys additional information. And `blending` can be set with several presets (e.g. 'normal', 'additive', etc.)  Maybe we can refactor this into:
```py
material.alpa_mode : str
material.blend_details : dict  # only used when alpha_mode = 'blend'
material.weighted_details : dict  # only used when alpha_mode = 'weighted'

# Using presets:
material.alpha_mode = "add"  # -> sets alpha_mode to "blend" and fills in appropriate blend_details
```

> From the renderer’s perspective, it still needs to clearly distinguish opaque from transparent materials.

And it's very clear indeed: `material.alpha_mode == "opaque"`.

>  I think any form of implicit inference will still lead to confusion. 

I think so too. And without it there is confusion too. Depends on the user and use-case. It won't be perfect, but I think *some* inference will help a certain group of users a great deal. I understand that this can cause some confusion elsewhere, but I think these will be rare and easily resolved because we can document it well. It's a tradeoff.

> What's most important, I believe, is that by the time the renderer and shaders receive the material, its transparency state must be clearly defined.

I agree.

panxinmiao:
Thanks for the clarification.

> They are not orthogonal, because the `alphaMode` / `blendMethod` is only used depending on the state of the first point; the second point are extra details about how to use alpha values.

I’d like to add a nuance here — while blending is typically associated with transparent materials, it's not always strictly coupled to transparency.

There are legitimate use cases where blending is applied to opaque objects, depending on the user’s intent or effect logic — for example, additive blending for glow effects, modulation in custom passes, or accumulation-based rendering techniques.

So while `transparent = true` often implies "enable blending", the reverse is not always true (mostly but not always): enabling blending doesn't necessarily mean the material is semantically "transparent". This distinction becomes especially important when users are doing custom or advanced rendering effects.

In this sense, I think it's helpful to decouple "transparency intent" from "blending config" as separate concepts. Otherwise, we risk oversimplifying scenarios where blending is used creatively beyond just visual transparency.

> And it's very clear indeed: material.alpha_mode == "opaque".

If `.alpha_mode` is going to take over the role of `.transparent`, I think that’s fine — as long as it’s explicit, and always resolvable into either "opaque" or "transparent" from the renderer's perspective. I think it’s vital that this model maintains a clear, top-level declaration of transparency intent, for both technical correctness and user clarity.

Korijn:
I will add that we probably need to add some tables to our docs, to show all the different combinations of options and how they are ultimately resolved. 

Additionally we can include a table that shows a mapping from all the different combinations of options in three.js to the equivalent configuration in pygfx.

In the end, I think people will figure it out if we document it well. Both graphics and science people are great with docs. We can always iterate again if it turns out we can do better.

almarklein:
> There are legitimate use cases where blending is applied to opaque objects, depending on the user’s intent or effect logic — for example, additive blending for glow effects, modulation in custom passes, or accumulation-based rendering techniques.

I'm having trouble understanding what this means. What does "semantically opaque" mean and imply?

*edit: and how does it add to the things we already know of the object, like whether it blends, and whether it writes depth?*

panxinmiao:
> I'm having trouble understanding what this means. What does "semantically opaque" mean and imply?
> 

What I mean by “semantically opaque” is: an object that is considered opaque in terms of rendering logic (e.g., included in the opaque render pass, writes depth, etc.), even if blending is enabled for some reason — like additive effects, screen compositing, or custom rendering tricks.

This brings up a core issue with the proposed model of removing `.transparent` and relying solely on `alpha_mode`. If we use alpha_mode == "blend"  to imply transparency, and only then enable blending in the render pipeline, we’re conflating two orthogonal concepts:

- Whether the material is transparent (i.e., goes into the transparent pass, sorted back-to-front, etc.)

- Whether blending is used in the pipeline (i.e., enable blending config, set blend equations, etc.)

More specifically:
```python 
 material.alpha_mode: str
 material.blend_details: dict  # only used when alpha_mode == "blend"
```
This creates confusion because blending can be useful even when alpha_mode != "blend", and blend_details may still be relevant — e.g., a user might want additive blending on an otherwise opaque object. Restricting blend_details to "blend" limits flexibility and introduces inconsistency.

Also:
```python
# Using presets:
material.alpha_mode = "add"  # -> sets alpha_mode to "blend" and fills in appropriate blend_details
```
This blurs the line further — the field alpha_mode is now being overloaded with both intent (blend, mask, opaque) and implementation detail (add, multiply, etc.).

So to reiterate:

- Whether an object is transparent or opaque should remain a top-level, explicit decision.

- How blending is done should be a separate, composable configuration.

almarklein:
Thanks to the explanation. I'm still a bit confused though. From what I understand now, semantically opaque only really means that it's rendered in the opaque pass, before the "transparent" objects. Whether or not it writes depth seems irrelevant, because that can be set independently.

But that raises another question. If you have an object that applies some kind of blending, and you render it together with other opaque objects, that blending will only work correctly if the other objects are rendered first. So you actually want to render it later??

Maybe the renderer can do this:
* Render objects with `alpha_mode = 'opaque'` (i.e. objects that don't blend)
* Render objects that blend and write depth (the special use-case)
* Render objects that blend  and don't write depth (i.e. "transparent" objects)

> ```py
>  material.alpha_mode: str
>  material.blend_details: dict  # only used when alpha_mode == "blend"
> ```
>
> This creates confusion because blending can be useful even when alpha_mode != "blend", and blend_details may still be relevant — e.g., a user might want additive blending on an otherwise opaque object

My counter question here is, what is "otherwise opaque"? I have a feeling that it means that it writes depth, and you can set set independently.

panxinmiao:
When I say "semantically opaque," I'm referring to how the renderer should treat the object in terms of sorting, batching, and rendering order, not necessarily whether it writes depth or enables blending.

Here's how I think about it:

- Semantically opaque objects are those you intend to render in the opaque pass. They:

  -  Typically don’t rely on blending for their visual result.
  -  Can optionally enable blending (e.g., additive effects), but are still rendered early.
  -  Usually write to depth — but not always.

- Semantically transparent objects are those that:

  -  Visually depend on blending (e.g., alpha compositing).
  - Require back-to-front sorting to appear correct.
  - Usually don’t write to depth (but there are exceptions, like order-independent transparency techniques).
  
  
> Maybe the renderer can do this:
>
> Render objects with alpha_mode = 'opaque' (i.e. objects that don't blend)
> Render objects that blend and write depth (the special use-case)
> Render objects that blend and don't write depth (i.e. "transparent" objects)

That ordering makes sense and is actually quite close to what many engines do:

- Opaque pass — objects that don't blend and typically write depth.
- Post-opaque blending pass — objects that blend but don’t need strict sorting, e.g., additive glowing particles that still write depth.
- Transparent pass — objects that blend and don’t write depth, sorted back-to-front.

> If you have an object that applies some kind of blending, and you render it together with other opaque objects, that blending will only work correctly if the other objects are rendered first. So you actually want to render it later??

This is not necessarily the case. opaque  objects do not require strict sorting to ensure accurate rendering results (due to depth writing), and the blending here does not require strict sorting (such as ADD, MULTIPLY). Therefore, theoretically, category 1 and 2 above do not actually must need to be strictly separated.

panxinmiao:
I’d also like to point out that including "weighted blending" as an `alpha_mode` alongside options like "blend" or "mask" might not be semantically appropriate.

The core reason is that alpha_mode is intended to describe the rendering behavior of a single object, independent of how other objects in the scene are rendered.

But weighted blended OIT (order-independent transparency) doesn’t work this way — it requires all transparent objects in the scene to participate in a specific global rendering strategy. You can’t mix it freely with other blending modes like alpha, add, on a per-material basis. It has to be applied consistently across the scene (or at least across all transparent passes).

In that sense, "weighted" is not a material-level "alpha_mode", but rather a renderer-level or scene-level rendering mode — a global technique, not a local behavior.

While individual materials might still use alpha_mode = "blend" to opt into the transparent pass, the actual blending technique (weighted、other OIT tech or not) should be owned by the `renderer` or the `scene`, At this point, the behavior of the entire renderer will change.

almarklein:
> This is not necessarily the case. opaque objects do not require strict sorting to ensure accurate rendering results (due to depth writing), and the blending here does not require strict sorting (such as ADD, MULTIPLY). Therefore, theoretically, category 1 and 2 above do not actually must need to be strictly separated.

This is simply not true. Consider two "opaque" objects, where A does not blend and object B is in front of A and does blend. Then if A is rendered later, then B is not blended with A; A is simply hidden.

@panxinmiao I know you use an LLM for translating which is fine, but some of your comments make me feel like I'm talking to ChatGTP, the above one being a good example. I know you're a smart person and understand graphics, so an argument like this feels really out of place. Please don't let an LLM argue on your behalf. That said, I might be wrong, and if I am I apologize. 

----

> But weighted blended OIT (order-independent transparency) doesn’t work this way — it requires all transparent objects in the scene to participate in a specific global rendering strategy.

This actually works pretty well. The way it works is that objects are sorted also based on whether they are "weighted". And during rendering, a group of successive objects will be rendered as one OIT pass. There is indeed a catch, if the weighted objects are not successive after sorting, they have their own OIT pass, and are blended with the OVER operator. But this can only happen when users set `render_order`. 

The benefit of being able to render objects with all these different blend/alpha modes is huge: you can combine dithered objects with weighted, add objects with additive blending etc.

So I think that combining it makes perfect semantic sense: it defines how an object's fragments are combined with those of other objects.

panxinmiao:
> This is simply not true. Consider two "opaque" objects, where A does not blend and object B is in front of A and does blend. Then if A is rendered later, then B is not blended with A; A is simply hidden.
> 

Not really, because opaque will write depth, and A's fragment will be discarded due to depth testing

panxinmiao:
I just read [Babylon's instructions on handling transparency](https://doc.babylonjs.com/features/featuresDeepDive/materials/advanced/transparent_rendering/), and Babylon's logic is more complex, while Babylon actually allows for inferring transparency. But as mentioned earlier, when the material reaches the renderer, its specific classification must be clearly defined. When there are classification errors in automatic inference, there are advanced customization options that allow users to manually specify the settings.

almarklein:
> Not really, because opaque will write depth, and A's fragment will be discarded due to depth testing

Yeah, that's my point. But if B is supposed to blend, e.g. an additive glow, it needs A to be rendered underneath.

panxinmiao:
> Please don't let an LLM argue on your behalf. That said, I might be wrong, and if I am I apologize.

Sorry to make you feel that way. Perhaps the logic and expression are a bit confusing and divergent, but all of the above (including examples) are just to illustrate that: 
The Blend (transparent) in `alpha_mode` is not the same concept as the blending state configuration of RenderPipline. To be precise, we implement AlphaBlend (transparent) rendering through a specific configuration of the blending state of RenderPipline. However, for other (non AlphaBlend (transparent) category objects), the blending state of RenderPipline can also be configured, depending on the specific logic (whether it is post-opaque effects, particles or sprites, or any specific logic), and cannot Simply put, only objects in the AlphaBlend (transparent) category need to be configured with a blending state.

panxinmiao:
> Yeah, that's my point. But if B is supposed to blend, e.g. an additive glow, it needs A to be rendered underneath.

I think I now understand your point more clearly, as well as the core of the problem and differences.

There’s no universal rendering classification or sorting model that fits all scenarios — especially when dealing with specific logic or visual effects.

If our classification model includes a “post-opaque” category (for example, for particles or sprites), then of course these objects should be rendered after opaque objects. This gives more flexibility for complex compositing.

However, in a simplified two-category model like the one used in Three.js (just "opaque" and "transparent"),  I think such "post-opaque" objects are typically grouped under opaque, because their rendering and blending behavior do not depend on Z-sorting. Their behavior is more like opaque objects. For example,, rendering flame effects only in the scene (or sub scene) without considering their interaction with other objects.

In these cases, for more complex scenarios, it becomes necessary for the user to intentionally control render order, depth settings, and blending behavior, depending on the specific scene and logic. The important thing is that the engine exposes the right APIs for users to control this.

Even in more sophisticated engines like Babylon.js, which offers a more detailed classification model, things are not always handled perfectly. For example, say I want to render a flame (particles) in front of a piece of glass. According to the default logic, the flame (blended particles) would be rendered first, followed by the transparent glass — but the result could be visually incorrect (if flame's depth_write is disabled). What matters is that the user has the ability to override this behavior via the API to achieve the correct visual outcome.

almarklein:
I started writing docs for how Pygfx could work in #1124. It includes a list of use-cases, with the (envisioned) API for Pygfx and for ThreeJS. Feel free to suggest more use-cases.

PR #1124, opened 2025-06-23
Add docs for blending

almarklein:
Ref discussion in #1120.

I refactored the current guide, moving the content to multiple files, to make room to go into some topics deeper, like we do here for transparency.

The text does not describe how Pygfx currently work, but how I think it should work. Open for suggestions. The idea is to get (more or less) agreement on this, then implement it accordingly.

### DIFF
--- a/docs/transparency.rst
+++ b/docs/transparency.rst
@@ -130,7 +130,7 @@ The renderer sorts these objects front-to-back to avoid overdraw (for performanc
 
 In contrast, the "composite" and "weighted" methods result in semi-transparent fragments,
 and by default have ``depth_write=False``. The renderer sorts these object back-to-front to
-improve the chance of correct blending. Note that with the 'auto' mode, de default ``depth_write`` depends
+improve the chance of correct blending. Note that with the 'auto' mode, the default ``depth_write`` depends
 on the ``opacity``.
 
 


docs/blending.rst
```diff
@@ -0,0 +1,261 @@
...
+------------
+
+For objects that have their fragments blended, the order of rendering these
+objects is important to get the correct result. The renderer sorts objects based
+on three factors:
+
+* Objects are first sorted by their ``object.render_order``. Users should generally not need this, but it allows full control over the render order.
+* Then bases on their ``material.alpha mode`` and ``material.depth_write``, in order: "opaque" and "dither", "blend" with depth write, "blend", "weighted".
+* The object's distance to the camera. Note that this won't help all cases, since objects
+  can still intersect other objects (and themselves).
```
Korijn, line 73:
```suggestion
  can still intersect other objects (and themselves). This sorting step is intended primarily to reduce overdraw.
```

Korijn:
Focusing on the API design: it looks fine to me. I'm curious to hear from @panxinmiao. Are there any overlooked use cases?

docs/blending.rst
```diff
@@ -0,0 +1,261 @@
...
+
+Render order
+------------
+
+For objects that have their fragments blended, the order of rendering these
+objects is important to get the correct result. The renderer sorts objects based
+on three factors:
+
+* Objects are first sorted by their ``object.render_order``. Users should generally not need this, but it allows full control over the render order.
+* Then bases on their ``material.alpha mode`` and ``material.depth_write``, in order: "opaque" and "dither", "blend" with depth write, "blend", "weighted".
```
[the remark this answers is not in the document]
Korijn: Then I would propose this formatting change to make it easier to parse the order

panxinmiao:
I still believe this design conflates two distinct concepts: the definition of transparency (a high-level API concept at the engine level) and the blending configuration (a low-level API concept at the GPU level).

What we commonly refer to as “alpha blending” is just one specific configuration of the blending settings in the rendering pipeline — a particular blending mode.

It corresponds to the classic alpha blending formula:
`FinalColor = SrcColor * SrcAlpha + DstColor * (1 - SrcAlpha),`
also known as “pre-multiplied alpha blending”. This approach is order-dependent and is one of the standard methods to achieve transparency.

As I understand it, `alpha_mode = "blend"` — given that it's grouped alongside options like "opaque" — is intended to represent an "alpha blended object", which is, to some extent, a high-level API concept for transparent objects. It does not represent the low-level blending configuration of the rendering pipeline.

Moreover, blending in the rendering pipeline is not exclusive to transparent objects. It's quite common for opaque objects to use blending as well (e.g., for glowing or light-accumulation effects). The engine should not restrict this capability.

Additionally, if we follow this API design, how should we categorize objects during rendering — i.e., assign them to different render passes? Furthermore, the extra parameters under `alpha_mode = "blend"` such as "over", "add", "subtract", and "multiply" do not fully cover the range of blending options available in the rendering pipeline.

panxinmiao:
I'm not entirely sure if I expressed my point clearly. If my wording caused any confusion, I sincerely apologize.

I would like to explain my concerns again:

### 1、`alpha_mode` conflates object classification semantics and pipeline configuration semantics, leading to conceptual confusion.

Values like `alpha_mode="blend"`, alongside opaque, weighted, and dither, essentially serve as a classification system indicating how an object should be rendered. This is a high-level, engine-side logic.
On the other hand, values like "add" or "multiply" under blend_mode (or sometimes directly under alpha_mode) represent specific GPU-level blending configurations. These define how colors are computed in the rendering pipeline and belong to the low-level graphics API domain.

Mixing these two kinds of semantics into a single enum or parameter tree leads to confusion. And, blending settings in the rendering pipeline are not exclusive to transparent objects, using “whether blending is enabled” as the criterion for determining if an object is transparent is problematic and misleading.

For example:
```
m.alpha_mode = "add"
```
This raises the question:
- Does this mean the object is a transparent additive blend, requiring back-to-front sorting?
- Or is it an opaque object using additive blending, rendered in a forward (unsorted) opaque pass?

<del>
### 2、The current presets under `alpha_mode="blend"`—such as "over", "add", "subtract", and "multiply"—are too limited.
They do not adequately cover the full range of blending configurations supported by modern rendering pipelines. 
For instance:

- Min/Max blending
- Custom src/dst blend factors
- Blend constants
</del>

Edit: 
I just realized that in this API design, there is a setting for `blend_mode` when `alpha_mode=blend`, which allows for custom blending methods, so this statement does not hold.

This brings us back to the core issue: using `alpha_mode=blend` as a classification mechanism for objects is conceptually different from configuring blending operations in the GPU rendering pipeline, and conflating the two can be misleading. 
In other words, `blend_mode` is a low-level parameter that controls how the GPU performs blending, and it should not be exclusive to `alpha_mode=blend`. Even opaque objects may require specific `blend_mode` configurations in certain scenarios. Therefore, `alpha_mode` (which classifies objects for rendering strategy) and `blend_mode` (which configures pipeline-level blending behavior) should be treated as separate concepts.

docs/blending.rst
```diff
@@ -0,0 +1,250 @@
...
+that the result will depend on the order in which the objects are rendered.
+
+
+Render order
+------------
+
+The renderer sorts objects based on three factors:
+
+* Objects are first sorted by their ``object.render_order``. Users should generally not need this, but it allows full control over the render order.
+* Then based on ``material.alpha mode`` and ``material.depth_write``, in order:
```
panxinmiao, line 68: I think 'render_order' property can only control the rendering order within a specific rendering classification. In other words, an object's classification (e.g., opaque, alpha-blended, weighted) is the primary factor in determining render order, and `render_order` only applies within that classification group.

This is because, fundamentally, different classifications correspond to different render passes, which are executed in separate phases. Therefore, cross-pass sorting is not feasible and generally makes no sense. `render_order` is only meaningful for controlling the order of objects within the same render pass.

almarklein: This is a great point. I was actually thinking about this yesterday. I agree that often you want to offset the object render order *within a group* of objects (opaque/blended). However, I think there are also cases where the user wants to re-order beyond groups, one example being objects in the weighted and blended class; the renderer renders weighted last, but not all use-cases may want this.

Maybe we can distinguish two super-groups, within which shifts makes sense. One for objects producing opaque fragments (opaque and dither alpha modes), and one blended for the rest.

almarklein:
> Moreover, blending in the rendering pipeline is not exclusive to transparent objects. It's quite common for opaque objects to use blending as well

In the proposed API, what you refer to is an object with `alpha_mode='blend'` and `depth_write=True`. Is that transparent or opaque? I don't know and it also irrelevant because `.transparent` is not a property anymore. 

> Additionally, if we follow this API design, how should we categorize objects during rendering — i.e., assign them to different render passes?

I believe objects can be unambiguously categorized (from the pov of the renderer) based on just their `alpha_mode` and `depth_write` props.

> Values like `alpha_mode="blend"`, alongside opaque, weighted, and dither, essentially serve as a classification system indicating how an object should be rendered. This is a high-level, engine-side logic. On the other hand, values like "add" or "multiply" under blend_mode (or sometimes directly under alpha_mode) represent specific GPU-level blending configurations. [...] Mixing these two kinds of semantics into a single enum or parameter tree leads to confusion.

I don't agree with this distinction of high-level vs GPU-level to be honest. There is certainly a categorization, where we have four main categories, where each category has a number of options, but all it does it define how alpha values are used in rendering the object. 

Or are you objecting to the fact that the presets for e.g. `blend_mode` ('over', 'add', 'subtract'), can also be passed to the `alpha_mode` prop? (which is a reasonable objection IMO)

docs/blending.rst
panxinmiao: The core issue here is that sorting across different render passes is difficult to implement or can introduce significant overhead, while offering very little benefit in most practical scenarios.

Regarding the example you mentioned, I have also added comments in the renderer (see #974) to explain the behavior when handling weighted blending objects.

To summarize: it is generally not advisable to mix alpha blending and weighted blending objects within the same scene, because they represent fundamentally different transparency strategies. A single scene should adopt a consistent approach to transparency—either order-dependent alpha blending or OIT via weighted blending. Moreover, weighted blending relies on all participating objects contributing to the same compositing process.
Even if, for some reason, different transparency strategies are used in the same scene, correct rendering is not guaranteed. This is because weighted blending objects do not participate in depth sorting—which is, in fact, one of the main reasons to use weighted blending (OIT) in the first place.

I understand you might argue that users can still rely on `render_order` to carefully control the draw order when interleaving weighted and alpha-blended objects.(But if so, what’s the point of using OIT in the first place?). Weighted and alpha blending require different render passes. Interleaving them would mean switching between render passes multiple times during a single frame, which is extremely costly. Modern GPUs are explicitly designed to minimize render pass switches for performance reasons.

In general, from a renderer’s perspective, the number and order of render passes should be determined by the engine’s rendering strategy—not dynamically changed based on object arrangement in the scene.

The reason we categorize objects into different groups is precisely because they use different render passes and must be rendered in different phases of the pipeline. All objects in a scene should be clearly grouped and rendered during well-defined stages—not arbitrarily interleaved based on per-object sorting.

If users have special needs where an object must be rendered out-of-order (e.g., a transparent object rendered between opaque layers), they should render it in a separate scene, with explicit control over rendering order and timing.

panxinmiao:
> In the proposed API, what you refer to is an object with `alpha_mode='blend'` and `depth_write=True`. Is that transparent or opaque? I don't know and it also irrelevant because `.transparent` is not a property anymore.

No, it's important. This is exactly where the problem lies, we need clear and unambiguous categorization. Without it, it's unclear which render pass should handle the object, nd which stage of the pipeline it belongs to.

Regarding object classification, we indeed didn't pay much attention to it before (prior to the simplification and refactoring of Blender), because at that time, the renderer did not yet distinguish between different rendering passes or stages — so the classification of an object didn’t matter much.
However, as we started introducing more advanced rendering features, this classification and render pass scheduling has become crucial. That’s also why I opened #974 in advance—it brings significant changes to the renderer’s behavior and needs to be discussed and reviewed.

> There is certainly a categorization, where we have four main categories, where each category has a number of options, but all it does it define how alpha values are used in rendering the object.

Not just that, it also determines which render pass is used and at what stage in the pipeline the object is rendered. This is not a minor technicality, it's a fundamental architectural concern in the subsequent renderer design and render pass scheduling.

> I believe objects can be unambiguously categorized (from the pov of the renderer) based on just their `alpha_mode` and `depth_write` props.

I agree that `alpha_mode` can be used as the primary way to classify objects—but we must clarification that `alpha_mode='blend'` represents a category of objects that should be rendered using a specific render pass at a specific pipeline stage. It does not mean that GPU blending is being used per se. In other words, we must decouple `alpha_mode` (for classification) from `blend_mode` (which configures the GPU pipeline).

As for using `depth_write` to infer categorization—I don't think that's appropriate. Properties like `depth_write` and `depth_test` serve many flexible purposes and are frequently controlled directly by the user. They are not strongly tied to render pass decisions and should not drive classification.

> Or are you objecting to the fact that the presets for e.g. `blend_mode` ('over', 'add', 'subtract'), can also be passed to the `alpha_mode` prop? (which is a reasonable objection IMO)

Yes, that's exactly my point. `blend_mode` should be treated as an independent concept.

docs/blending.rst
almarklein: The argument about switching pipelines makes sense if all objects were of the same type (and to some degree the same material), but the use cases we're dealing with is e.g. stiching multiple images using 'weighted' and then rendering semi-transparent points on top. The pipelines are not the same anyway, so switching order will not affect performance.

> To summarize: it is generally not advisable to mix alpha blending and weighted blending objects within the same scene, because they represent fundamentally different transparency strategies.

I see no objection why this should not be possible 🤷  (except for this sorting issue).

In any case, I shall address the original issue.

panxinmiao: > The argument about switching pipelines makes sense if all objects were of the same type (and to some degree the same material), but the use cases we're dealing with is e.g. stiching multiple images using 'weighted' and then rendering semi-transparent points on top. The pipelines are not the same anyway, so switching order will not affect performance.

**the number and order of render passes should be determined by the engine’s rendering strategy—not dynamically changed based on object arrangement in the scene.**

In our current grouping design (as implemented in #974), we have three fixed rendering passes: the opaque pass, the transparent pass (which may use two passes render twice), and the weighted blending pass. **The number and order of these render passes are fixed**.

In an extreme case, if there are 100 alpha-blended objects and 100 weighted-blending objects interleaved with each other, does that mean we would need to use 200 render passes if we strictly follow their `render_order`?

almarklein:
> No, it's important. This is exactly where the problem lies, we need clear and unambiguous categorization.

Please help me understand, by providing an example where an object has `transparent=False`, which cannot be derived from `depth_write`. Or maybe another way to phrase it, you seem to suggest that an object can be `transparent=False, alpha_mode='blend', depth_write=False`. How do you expect the renderer to handle that?

panxinmiao:
> > No, it's important. This is exactly where the problem lies, we need clear and unambiguous categorization.
> 
> Please help me understand, by providing an example where an object has `transparent=False`, which cannot be derived from `depth_write`. Or maybe another way to phrase it, you seem to suggest that an object can be `transparent=False, alpha_mode='blend', depth_write=False`. How do you expect the renderer to handle that?

For example, scene backgrounds (or objects meant to serve as scene backgrounds), decals, or helper objects and markers for debugging or visualization purposes. I once had a use case where an object had a hidden proxy geometry—only the proxy geometry wrote to the depth buffer, while the visible object itself only wrote to the color buffer but not to depth

almarklein:
Yeah but how do you think the renderer should use the value of `.transparent`? From what I understand, objects with `transparent=True` are rendered first, but for the case of backgrounds this is not enough, because you want to render before all the other opaque objects. I'm not sure how the decals etc. fit in. Can you elaborate?

panxinmiao:
For example, some skybox or skydome models are modeled as a box or a sphere. See this model on Sketchfab: https://skfb.ly/oIpX9 and https://skfb.ly/oIHQL

When used, these typically have the following settings: 
```python
transparent=False, 
render_order=-99,  # to ensure they render first
depth_test=False, 
depth_write=False  # so they don’t participate in any depth-related operations.
```
And they are unit-sized and have their center position follow the camera every frame.


As for decals, they refer to "localized texture layers attached on top of existing geometry" to enhance visual detail—such as cracks, stains, bullet holes, or signs—without modifying the object's base geometry or main texture. In essence, a decal might be a small piece of geometry (like a plane) that is projected onto the target surface using geometric projection techniques.

While these effects are usually achieved using textures with transparency (i.e., alpha blending), making them visually semi-transparent, they should not be classified as transparent objects from the rendering pipeline's perspective. Decals do not participate in the transparent object depth sorting done in the main geometry pass or any other complex rendering behavior for transparent objects (such as double-sided/double passes rendering, as well as possible future more complex processes such as depth peeling, dual depth peeling, etc.). they behave more like a screen-space overlay. Decals are typically a separate pass or a separate render group in opaque pass following opaque objects (to control the rendering order), and are directly overlaid on the result of the opaque pass.


A typical decal setup maybe like this:
```python
transparent = False
depth_test = True
depth_write = False
blend_mode = "normal"
```

Of course, the above parameter settings can be adjusted based on the actual needs of the scene. However, the key point is that low-level GPU pipeline features such as depth_write and blending method are atomic capabilities and should not be used as the basis for object classification. These features are flexible and can be freely combined by users to meet various creative needs, and we should not impose restrictions on the engine's ability to support such flexibility.

We can say that a certain category of objects (e.g., "opaque" or "transparent") is typically implemented using a specific set of parameters, but we should not assume that a particular parameter configuration necessarily implies the object belongs to a certain category.
I think the design of an engine's API should be guided by semantics rather than being constrained by low-level implementation details. In other words, APIs should focus on expressing the intrinsic properties or intended behaviors of objects, rather than exposing low-level pipeline settings such as "depth_write" or blend method. These parameters should serve as implementation mechanisms—not as the basis for object categorization or interface design.

almarklein:
Thanks for the explanation; I understand your point better now. It has everything to do with controlling the order that objects are rendered in. I'll give this some more thought ...

almarklein:
I'm moving rather slow since I am on holiday and try to get some Pygfx done in the late evenings / early mornings.

I've updated this pr and the #1144. I've not included a `transparency` property, because from the use-cases put forward it seems mostly used as a trick to force things to be rendered earlier or later than usual. Instead I added `object.render_group` which allows controlling multiple "layers", as if using multiple calls to `render()` except defined in the scene graph, which covers use-cases such as backgrounds and overlays. 

I went a bit back and forth with naming things but I settled on:
* `alpa_mode`: a string preset for the majority of cases (and to which we can add if we want).
* `alpha_config`: a dict that fully describes the alpha handling in one place (instead of separate dicts for composite, dither, etc. which felt awkward).
* `alpha_method`: represents the 'group': either "opaque", "composite", "stochastic"  or "weighted".
* Note that the names of alpha-method are more technical, while the names in alpha-mode are a bit more user-friendly, e.g. "solid" and "blend" (for the over-operator).

Further:
* The renderer employs an opaque and transparency pass, like most render engines do. Which pass to assign an object to is based on the alpha-method, but users can override this in the `alpha_config`. These cases should be really rare.
* Objects with alpha-method 'weighted' are grouped together because of the way the z-sorting happens. Though they can still be "grouped" using `ob.render_order`.
* I added alpha-mode 'auto' as a somewhat plug-n-play mode, especially for 2D-ish scenes. I'm not entirely happy about it, but I think it more or less represents the earlier behavior.

almarklein:
mmm, things still don't feel right to me. 

I've been reading a bit and I quite like Unity's [`renderQueue`](https://docs.unity3d.com/6000.3/Documentation/Manual/built-in-rendering-order.html). Basically instead of ThreeJS's approach of having 2 queues (opaque and transparent), it has 5 builtin queues (including background and overlay), but users can also assign objects a `renderQueue` int value in between two builtin queues.

I feel like this is a quite elegant and flexible solution. Better than ThreeJS (setting `.transparent` and `.renderOrder` to try to control an objects render order) and also better than the `ob.render_group` I implemented that tries to do a similar thing.

We can derive a suitable value from `alpha_method` and allow the user to set it to override it. The `ob.render_order` still has value for controlling the order *within* a queue, and be per-object instead of per-material.

I'll have a stab at this soon, but figured I'd post the idea here to see what others think.

almarklein:
I updated this pr and #1144. IMO the ``render_queue`` idea solves the whole render-order issue in an elegant way that can also be super-flexible for special use-cases. I hope we can start to wrap this up ...

Korijn:
> I updated this pr and #1144. IMO the `render_queue` idea solves the whole render-order issue in an elegant way that can also be super-flexible for special use-cases. I hope we can start to wrap this up ...

To me this looks fantastic but I also think you should just move on and accept the outcome. In the interest of other priorities. :)

panxinmiao:
I believe there is one more point that needs to be clarified and documented.

Specifically, which rendering passes/stages our renderer internally uses (currently, we have three passes/stages in #974: Opaque, Transparent, and Weighted),

and according to our API design, what rules govern the assignment of objects to these specific rendering passes/stages.

Additionally, is `render_queue` intended to be user-configurable? Since the different rendering stages are inherent internal passes of the renderer and cannot be reordered arbitrarily, could this conflict with `render_queue`?

In other words, if an object is assigned to the Transparent stage based on the above rules, but the user sets its `render_queue` to 2000, what would happen?

panxinmiao:
I submitted a draft PR #1163  that adds a glTF Inspector along with several examples, which I believe could help us verify the practicality and soundness of our transparency-related API design.

Korijn:
@almarklein I'm going to answer the questions, but they are based on my understanding of the diff. Correct me if I get anything wrong!

> Specifically, which rendering passes/stages our renderer internally uses (currently, we have three passes/stages in #974: Opaque, Transparent, and Weighted),

I believe they are [listed here](https://github.com/pygfx/pygfx/pull/1124/files#diff-f23d1daa42e4084c5313c03856d50c36dd3830a5bddfa12295aee3b00242e62eR178-R193).

```
Render queue
------------

The ``material.render_queue`` is an integer that represents the group that the renderer uses to sort objects.
The property is intended for advanced use; it is determined automatically
based on ``alpha_method``, ``depth_write`` and ``alpha_test``. It's values lays between 1 and 4999,
with the following default values:

* 1000: background.
* 2000: opaque non-blending objects.
* 2400: opaque objects with a discard based on alpha (i.e. using ``alpha_test`` or "stochasric" alpha-mode).
* 2600: transparent objects that write depth.
* 3000: transparent objects that don't write depth.
* 4000: overlay.

Objects with ``render_queue`` between 1501 and 2500 are sorted front-to-back. Otherwise objects are sorted back-to-front.
```

> and according to our API design, what rules govern the assignment of objects to these specific rendering passes/stages.

I think this is also covered in the linked section of the documentation.

> Additionally, is `render_queue` intended to be user-configurable?

Yes. It allows you to override the assignment done by the renderer. You can assign to one of the predefined queues, or to a custom value. If you use a custom value, you define a new custom rendering stage.

> Since the different rendering stages are inherent internal passes of the renderer and cannot be reordered arbitrarily, could this conflict with `render_queue`?

As [discussed here](https://github.com/pygfx/pygfx/pull/1124/files#diff-f23d1daa42e4084c5313c03856d50c36dd3830a5bddfa12295aee3b00242e62eR205-R215) the value you assign determines when the custom stages get executed.

```
How the renderer sorts objects
------------------------------

The renderer sorts objects based on the following factors:

* The ``material.render_queue``.
* The ``object.render_order``.
* The object's distance to the camera, either front-to-back or back-to-front, depending on the ``render_queue``. Weighted objects are not sorted.

Even with this sorting, objects can still intersect other objects (and themselves).
To prevent drawing the (parts of) objects that are occluded by other objects, a depth buffer is used.
```

So let's say you set an object's render_queue to 2002, then you get this ordered list of render queues:

```
* 1000: background.
* 2000: opaque non-blending objects.
* 2002: <your custom render queue>
* 2400: opaque objects with a discard based on alpha (i.e. using ``alpha_test`` or "stochasric" alpha-mode).
* 2600: transparent objects that write depth.
* 3000: transparent objects that don't write depth.
* 4000: overlay.
```

I am not sure if it is possible to create conflicts. I don't think so. An object is always part of a render queue, and that can be predefined or custom.

Also note the sorting rules for objects within stages depend on the render queue value chosen, as [discussed here](https://github.com/pygfx/pygfx/pull/1124/files#diff-f23d1daa42e4084c5313c03856d50c36dd3830a5bddfa12295aee3b00242e62eR193): `Objects with ``render_queue`` between 1501 and 2500 are sorted front-to-back. Otherwise objects are sorted back-to-front.`

> In other words, if an object is assigned to the Transparent stage based on the above rules, but the user sets its `render_queue` to 2000, what would happen?

Then you assign the object to the render queue `2000: opaque non-blending objects`.

---

As a side note, this system is inspired by [Unity's system](https://docs.unity3d.com/6000.3/Documentation/Manual/built-in-rendering-order.html). Almar linked it in [this comment](https://github.com/pygfx/pygfx/pull/1124#issuecomment-3143729522).

---

> I submitted a draft PR https://github.com/pygfx/pygfx/pull/1163 that adds a glTF Inspector along with several examples, which I believe could help us verify the practicality and soundness of our transparency-related API design.

Awesome! That's very helpful. It would be great if we can test many use cases. :)

panxinmiao:
Thank you for the detailed explanation. Now I have a better understanding of the API.

Korijn:
I would love to merge this!

almarklein:
> I'm going to answer the questions, but they are based on my understanding of the diff. Correct me if I get anything wrong!

I agree with all the answers 👍  The `render_queue` is an alternative to the binary (opaque vs transparent) render-pass concept that ThreeJS uses. It solves cases like background and overlays, allows precise insertions, custom passes, and is explicit.

almarklein:
> I would love to merge this!

Should ideally happen together with / right after #1144

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+
+Transparency is expressed using an ``alpha`` value. This can be the fourth
+component in an RGBA color, e.g. in a color property, colormap or image texture. But it
+can also be an explicit value, e.g. ``material.opacity``.
+
+Although alpha values are commonly used to represent transparency, this is not always
+the case; they can e.g. be used as the reference value in alpha testing.
+
+In any case, the alpha value represents a weight for how the object is combined with
+other objects, and it's applicance is fully defined by ``material.alpha_config``.
```
Korijn, line 24:
```suggestion
other objects, and its application is fully defined by ``material.alpha_config``.
```

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+are opaque (alpha=1) are rendered correctly, and objects that are
+(partially) transparent too, as long as the objects
+don't intersect (i.e. can be sorted based on their distance from the
+camera). Objects with ``material.opacity<1`` behave the same as with
+``alpha_mode`` "blend".
+
+2. Set ``material.alpha_mode`` to a preset string:
+
+These provide configurations for common cases. Examples are "solid",
+"blend", "dither", and several more.
```
Korijn, line 50:
```suggestion
"blend", "dither", and several more. See below for a full list of options and descriptions.
```

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+
+* "solid": alpha is ignored.
+* "solid_premul": the alpha is multipled with the color (making it darker).
+
+Method "composite" (per-fragment blending of the object's color and the color in the output texture):
+
+* "auto": classic alpha blending, with ``depth_write`` defaulting to True if ``.opacity==1``.
+* "blend": classic alpha blending using the over-operator.
+* "add": additive blending that adds the fragment color, multiplied by alpha.
+* "subtract": subtractuve blending that removes the fragment color.
```
Korijn, line 110:
```suggestion
* "subtract": subtractive blending that removes the fragment color.
```

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+``material.alpha_config`` directly to take full control over all available
+options. In this case the 'mode' field and ``material.alpha_mode`` become "custom".
+
+
+Render queue
+------------
+
+The ``material.render_queue`` is an integer that represents the group that the renderer uses to sort objects.
+The property is intended for advanced use; it is determined automatically
+based on ``alpha_method``, ``depth_write`` and ``alpha_test``. It's values lays between 1 and 4999,
```
Korijn, line 183:
```suggestion
based on ``alpha_method``, ``depth_write`` and ``alpha_test``. Its values lays between 1 and 4999,
```

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+
+
+How the renderer sorts objects
+------------------------------
+
+The renderer sorts objects based on the following factors:
+
+* The ``material.render_queue``.
+* The ``object.render_order``.
+* The object's distance to the camera, either front-to-back or back-to-front, depending on the ``render_queue``. Weighted objects are not sorted.
```
Korijn, line 212: What are weighted objects?

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+
+    .. code-block:: py
+
+        # Pygfx
+        m.alpha_mode = "solid"
+
+    .. code-block:: js
+
+        // ThreeJS
+        m.transparent.false;  // default
```
Korijn, line 250:
```suggestion
        m.transparent = false;  // default```

File: docs/transparency.rst [hunk 10]
```diff
@@ -0,0 +1,421 @@
...
+        m.transparent = true;
+        m.blending = THREE.AdditiveBlending;
+        m.depthWrite = False;
+
+* Additive blending (glowy opaque objects)
+
+    .. code-block:: py
+
+        # Pygfx
+        # (the object gets automatically rendered at the very start of the transparency-pass)
```
Korijn, line 284: because of render queues right? or... why?

Korijn:
I figured it would be worthwhile to give this a linguistic review and to check for clarity :) if you can address my comments we can merge this.

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+        // ThreeJS
+        m.transparent = True;
+        m.alphaTest = 0.5;
+
+* A background
+
+    .. code-block:: py
+
+        # Pygfx
+        ob.material.render_queue = 1000
```
Korijn, line 376:
```suggestion
        ob.material.render_queue = 1000  # the render queue for backgrounds
```

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+Most users just set ``material.alpha_mode`` which implicitly sets
+``material.alpha_config``. In advanced/special cases, users can set the
+``material.alpha_config`` directly to take full control over all available
+options. In this case the 'mode' field and ``material.alpha_mode`` become "custom".
+
+
+Render queue
+------------
+
+The ``material.render_queue`` is an integer that represents the group that the renderer uses to sort objects.
```
Korijn, line 181: Maybe also mention here that it is an integer and not an Enum because that would prevent custom assignment? I can already picture the PRs/issues about this.

docs/transparency.rst
almarklein: yeah, I updated this comment.

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+
+    .. code-block:: py
+
+        # Pygfx
+        ob.material.render_queue = 1000
+
+    .. code-block:: js
+
+        // ThreeJS
+        // (put at the beginning of the opaque-pass, so no blending possible.)
```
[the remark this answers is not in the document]
almarklein: I'll remove the comment about no blending possible, because it's confusing. But its a limitation of ThreeJS: blending is only enabled if you use `transparent=false`.

docs/transparency.rst
```diff
@@ -0,0 +1,421 @@
...
+Alpha methods
+-------------
+
+Most users don't have to worry much about what the methods mean. Though it's good to understand
+that the "opaque" and "stochastic" methods produce opaque fragments, and by default have ``depth_write=True``.
+The renderer sorts these objects front-to back to avoid overdraw (for performance).
+
+In contrast, the "composite" and "weighted" methods result in semi-transparent fragments,
+and by default have ``depth_write=False``. The renderer sorts these object back-to-front to
+improve the chance of correct blending. Note that the 'auto' mode is an exception to this rule.
```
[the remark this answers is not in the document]
Korijn: @almarklein you missed this comment I think!

Korijn (dismissed):

docs/transparency.rst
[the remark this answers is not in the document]
almarklein: Woops! Yes I did. Addressed it by making explicit what I meant.

docs/transparency.rst
```diff
@@ -124,13 +124,14 @@ Method "weighted" (order independent blending):
...
 
-Most users don't have to worry much about what the methods mean. Though it's good to understand
+Most users don't have to worry much about what the alpha-methods mean. Though it's good to understand
 that the "opaque" and "stochastic" methods produce opaque fragments, and by default have ``depth_write=True``.
 The renderer sorts these objects front-to-back to avoid overdraw (for performance).
 
 In contrast, the "composite" and "weighted" methods result in semi-transparent fragments,
 and by default have ``depth_write=False``. The renderer sorts these object back-to-front to
-improve the chance of correct blending. Note that the 'auto' mode is an exception to this rule.
+improve the chance of correct blending. Note that with the 'auto' mode, de default ``depth_write`` depends
```
Korijn, line 133:
```suggestion
improve the chance of correct blending. Note that with the 'auto' mode, the default ``depth_write`` depends
```

Korijn:
I guess it was a little frustrating process, but I am really impressed by the outcome. ❤️

Korijn (approved):

hmaarrfk:
uge

almarklein:
Next is up is the actual implementation, which is ready in #1144

panxinmiao:
Sorry to bring up this topic again, 😓

After several days of hands-on practice and investigation, I can confirm that the renderQueue here is not the object classification mechanism we need.

Essentially, it is a higher-level sorting mechanism. On the CPU side, it buckets and sorts all objects based on their renderQueue to determine the order and grouping strategy of DrawCalls.  
Although it ultimately determines which rendering phase/Pass an object belongs to in the renderer, it functions more like an internal identifier (even though it can be directly specified by users).  
It still doesn’t solve the fundamental question: “How are objects classified (i.e., assigned to a specific phase) by default?”  
Even in Unity, users are required to set the Queue or RenderType tag in the material shader to determine the final RenderQueue.

- "Opaque" → 2000  
- "TransparentCutout" → 2450  
- "Transparent" → 3000  

The only difference is that Unity allows users to directly set the RenderQueue.

However, Queue or RenderType itself is explicit, unambiguous, and requires user input—the engine does not infer it (and in my opinion, cannot infer it, as it is essentially a meta-attribute that reflects user intent).  
If not set, the engine simply assigns a default value (e.g., Queue=2000) without any additional inference.

The Queue or RenderType values—Opaque, TransparentCutout, and Transparent—are exactly the internal phases and object classifications we need in the renderer.  
Just like the `transparent` property in three.js or the `TransparencyMode` property in Babylon.js, these are meta-attributes that determine user intent and directly assign objects to the appropriate rendering phase. I believe the engine should not attempt to infer them (and in fact, cannot reliably do so).

_Additionally, while reviewing Babylon.js documentation, I noticed that `TransparencyMode` can be left empty, and Babylon.js uses a very complex (though still explicit) logic to automatically infer it—unlike other engines that simply assign a default value. I was initially curious why Babylon.js adopted this approach, but after some research, I believe it’s merely a historical burden.
As mentioned in https://github.com/BabylonJS/Babylon.js/pull/16144, this might be removed in the future._

---

Our current issue is that although `alpha_method` seems usable as a classification basis, it is an internal value that cannot be set directly or independently, and it’s opaque to users.  
Even if it can be configured via the `alpha_config` dictionary, `alpha_method` is tied to the GPU pipeline’s blending settings and cannot serve as an independent intent-based identifier for grouping or phase assignment. This brings us back to square one. I’ve expressed this viewpoint multiple times that internal renderer phase division and GPU blending configuration are orthogonal concepts and should not be forcibly coupled.

I’d like to reiterate: object classification and renderer phase division are based on the engine’s own scene rendering logic and design philosophy—they exist to enable specific engine capabilities, e.g., supporting Weighted Blend OIT objects by introducing a dedicated Weighted phase (the rationale for treating it as a separate phase can be discussed elsewhere).  
Most engines classify objects into at least opaque and transparent categories (with corresponding internal render phases) because, in most scenarios, opaque and transparent objects require different processing logic. Moreover, between rendering opaque and transparent objects, the renderer may need to perform some certain tasks (e.g., generating SSR-based transmissive light sampling maps for physically based transparency).

GPU blending, on the other hand, is a rendering pipeline capability with various applications. It is not synonymous with transparency or alpha blending, nor is it exclusive to a specific object category or render phase.

A simple example:  
Even for opaque objects (from the renderer’s perspective, meaning they are rendered in the opaque phase—not necessarily visually opaque), users can leverage blending. A common use case is using additive blending for color accumulation.

Furthermore, if desired, users can achieve correct alpha-blended transparency effects even in this phase.  
(Assuming source colors are pre-multiplied by α for simplicity)  

When blending back-to-front iteratively (where `dst` is the accumulated result and `src` is the current fragment), the standard blend equation is:  
  `dst ← src + (1 - src_α) * dst`, 
  configured with:  
  `SrcFactor = ONE`, `DstFactor = ONE_MINUS_SRC_ALPHA`.  

When blending front-to-back (typical for the opaque phase), use:  
  `dst ← dst + (1 - dst_α) * src`,
  configured with:  
  `SrcFactor = ONE_MINUS_DST_ALPHA`, `DstFactor = ONE`.  

The two are mathematically equivalent.  

However, if users do this, the renderer still treats it as an opaque object and processes it in the opaque phase. This means all engine logic after the opaque phase (e.g., generating SSR buffers for physically based transparency, special handling for transparent-phase objects like dual-side dual-pass rendering, etc.) occurs after this object is rendered.

---

At last, I feel the current design of concepts like `alpha_mode`, `alpha_config`, and `alpha_method` is overly complex—even for someone relatively familiar with the API like myself, it’s easy to get confused.

Additionally, while well-intentioned to minimize user cognitive load (by reducing the number of values users need to know and set), having `auto` behaviors for many enums may actually cause more confusion. The engine needs to know the user’s explicit intent, and users working directly with the engine should understand its characteristics. Higher-level applications built on top of the engine should be responsible for providing abstractions based on specific use cases.

This is similar to how developers using pygfx must understand what Geometry represents, what Material represents, and what vertex positions in Geometry mean—while end-users of applications built on pygfx don’t need to know these details.

For example, the current code defaults `alpha_mode='auto'` but uses `"blend"` as the fallback behavior. I understand the reasoning: even for opaque objects, using blending won’t cause issues (since alpha=1), ensuring broad compatibility. However, this causes all default objects to be classified as transparent (renderQueue 3000) and assigned to the transparent phase (using the Transparent Pass), leading to unexpected results. It took me about an hour to figure out the reason why my rendering result was incorrect.

I think, trying to “do everything for the user” in a rendering engine often leads to bigger problems.

almarklein:
I have a feeling the confusion comes from a need to want to classify an object as opaque or transparent.

The short answer is this:
```py
matrerial.render_queue = 2000  # == material.transparent = False
material.render_queue = 3000  # == material.transparent = True
```

Yes, the value of `render_queue` is automatically derived from `alpha_method`, `depth_write`, and whether alpha test is used. But as far as "intent" goes, users should probably just set `render_queue` directly.

> Even if [`alpha_method`] can be configured via the `alpha_config` dictionary, `alpha_method` is tied to the GPU pipeline’s blending settings and cannot serve as an independent intent-based identifier for grouping or phase assignment. This brings us back to square one.

You are correct that alpha-method is a poor identifier for intent by itself. You should probably set `material.render_queue` instead.


> A simple example:
Even for opaque objects (from the renderer’s perspective, meaning they are rendered in the opaque phase—not necessarily visually opaque), users can leverage blending.
> [...]
> However, if users do this, the renderer still treats it as an opaque object and processes it in the opaque phase. 

Same answer, set `render_queue`.

> At last, I feel the current design of concepts like `alpha_mode`, `alpha_config`, and `alpha_method` is overly complex—even for someone relatively familiar with the API like myself, it’s easy to get confused.

I agree it's a rather complex system. I have struggled with this a lot, and it still frustrates me. I would very much present users with a system that Just Works. I actually think we could get close if we adopt techniques such as adaptive transparency, mlab etc. which insert output fragments into a multi-layer render target. But such techniques suffer a significant performance and memory hit. So I've somewhat accepted that we cannot really solve this problem for the user. Instead we offer more solutions that most engines do.

So the complexity comes partly from the fact that we have included support for stochastic and weighted methods. This offers more ways to deal with transparency, at the cost of a  more complex API 🤷 

> having auto behaviors for many enums may actually cause more confusion.

Also true. I can then recommend to explicitly set most such enums. I think it makes sense to recommend users to use explicit values in the docs.

BTW: the 'auto' alpha-mode also comes from the fact that we have multiple objects (lines, points, text) that have semi-transparent fragments for aa even if the object is opaque. Such objects can still be rendered fine, especially in 2D scenario's.

panxinmiao:
I believe that render_queue is not suitable as a user-facing API for expressing intent. It should be treated as an internal mechanism, since its meaning is neither clear nor intuitive to most users. In fact, the value of render_queue is usually derived from other properties, and users rarely need to set it directly.

As I understand it, automatic alpha blending was introduced mainly to support anti-aliasing (AA) for points, lines, text, and similar objects. However, this actually highlights that object classification/phasing and GPU blending capabilities are orthogonal concepts, and should not be coupled together.

For example, when the opacity=1, these objects should be classified as opaque, with no reason to place them in the transparent rendering stage. This applies even more strongly to Meshes: by default, they should not be treated as transparent objects (when opacity=1.0), since doing so violates user expectations, introduces potential issues, and conflicts with common usage patterns.

That said, it is reasonable for these 2D objects to require GPU blending when AA is enabled. In my view, such objects should still belong to the opaque category, but enabling AA should automatically configure the GPU blending state of their rendering pipeline (or adjust the corresponding material's GPU blending parameters). Additionally, because these 2D objects essentially reside on a flat z=0 plane, they neither need depth testing nor complex sorting. A simple sequential rendering order—based on layer or order—is sufficient.

almarklein:
There is no "opaque rendering stage" or "transparent rendering stage". And there is no way to classify an object as either opaque or transparent.

There's two main things:
* How is the alpha value used, e.g. blended, dithered, ignored. This is defined by `alpha_config`. Most users will set it via the `alpha_mode` presets.
* In what order the objects are rendered. This is defined by `render_queue`. It's default value is derived by guessing intent from alpha mode and depth_write, covering common cases. Setting the value manually, gives users *a lot* of flexibility in controlling object order.

There is also `ob.render_order` to control an object's order within a render queue. In ThreeJS this is used in certain cases to push objects to the end/beginning of the current pass. With the concept of the `render_queue` user will need it less often.

> For example, when the opacity=1, these objects should be classified as opaque, with no reason to place them in the transparent rendering stage.

To address this point specifically as an example: objects that are blended but also write depth, by default end up in render_queue 2600, which sits somewhere in between the default queues for opaque (2000) and transparent (3000) objects, and objects are sorted back-to-front (so that any aa edges blend correctly).

almarklein:
> There's two main things:

The `alpha_config` and `render_queue`  are indeed orthogonal.

panxinmiao:
> There is no "opaque rendering stage" or "transparent rendering stage". And there is no way to classify an object as either opaque or transparent.

Dividing the rendering process into stages is a necessary prerequisite for implementing some modern real-time rendering techniques. I am certain that almost all modern rendering engines at least separate the opaque and transparent stages, since many techniques require the renderer to perform specific operations between these two stages—for example, generating the SSR buffer. This was also the original motivation behind introducing object classification and stage division in #974.

I will explain in detail in #974 when I have time.

panxinmiao:
> BTW: the 'auto' alpha-mode also comes from the fact that we have multiple objects (lines, points, text) that have semi-transparent fragments for aa even if the object is opaque. Such objects can still be rendered fine, especially in 2D scenario's.

I’d also like to add a note here: as I understand it, the current "aa" for certain opaque objects (such as points, lines, and text) relies on transparent fragments, which makes the API design and implementation somewhat awkward.
In my view, the reason for adopting this fragment-based alpha blending approach to “aa” is largely due to the lack of a proper implementation of PMA (premultipliedAlpha). If I’m not mistaken, this “aa” is mainly intended to mitigate edge artifacts and black borders, but the same can be achieved correctly with PMA. This would make the notion of “opaque” more consistent across API semantics, visual results, and internal implementation. Combined with MSAA and other techniques, it can further improve the overall anti-aliasing quality.
My plan is to systematically investigate and refine the logic of premultipliedAlpha after resolving the ColorSpace-related issues, in order to establish a complete and correct workflow for textures and color management. Until then, including in changes like #974, I will aim to maintain compatibility and preserve the current “aa” behavior.

almarklein:
> In my view, the reason for adopting this fragment-based alpha blending approach to “aa” is largely due to the lack of a proper implementation of PMA (premultipliedAlpha). If I’m not mistaken, this “aa” is mainly intended to mitigate edge artifacts and black borders, but the same can be achieved correctly with PMA. This would make the notion of “opaque” more consistent across API semantics,

The purpose of aa is to blend the edges of objects with the things rendered behind it, to avoid jaggies. It's a very efficient way to prevent aliasing because it in the shader you can compute the exact pixel coverage. The downside is that it needs blending, so it relies on sorting to prevent artifacts. 

I think premultiplied alpha is a different thing.

almarklein:
*copied from #974*

render stages vs render_queue

I feel like they're two different ways to approach the same problem, where `render_queue` is actually a more powerful model. But now they're used both, which defeats the purpose of the `render_queue`. But I'm pretty sure this can be made to work without dividing objects into 4 render stages (opaque, transmissive, transparent, weighted)...

First the problems that I see with the currently proposed approach (in #974):

The three render stages representing the generally-transparent objects, are forced in the order: transmissive, transparent, weighted. Ignoring the `render_queue` set by the user.

You added a comment for the weighted blending, saying that it should absolutely not be mixed with other transparent objects. I think there are cases where it could, as long as both groups of objects are spatially separate. And the same argument can be made for transmissive vs transparent objects!

In the renderer logic in `main`, the `iter_render_pipelines_per_pass_type` basically dealt with this problem. I believe we can keep using that and include support for transmissive objects, by adding a `pass_type` "transmissive". 

In that way, the renderer would first sorts all objects in the normal way. Then `iter_render_pipelines_per_pass_type` iterates over them, grouping by `pass_type`. If the scene is "clean", you'll get at most one group of weighted and/or transmissive objects. But the user can also use `render_queue` to force multiple groups, and can control the order of transmissive/transparent/weighted objects. Or put all objects in the same render_queue, and let depth_sorting create the grouping, resulting in better quality of scenes with multiple "clusters" of transmissive  objects.

Some things that would be possible with this approach:
* transparent/translucent/weighted objects can be used in the background.
* In case of multiple translucent objects, e.g. multiple wine-glasses, the translucent objects could be put in a different render_queue than the normally-transparent objects, so they are rendered neatly in order (i.e. only one copy of the transmission texture)
* Or they could be the render_queue, resulting in one copy of the transmission texture for each wine glass.
* Maybe we can allow users to force a transmission-copy for each translucent object to get more accurate results (at cost of performance), by putting each object in a differtent render_queue.

panxinmiao:
> But I'm pretty sure this can be made to work without dividing objects into 4 render stages (opaque, transmissive, transparent, weighted)...
> 

The rendering process is currently divided into three stages: opaque, transparent, and weighted blend. Transmissive and transparent objects are handled within the same rendering stage/render pass. However, I deliberately sort transmissive objects before regular transparent ones using renderQueue. This ordering brings two benefits, as transmissive objects typically do not rely on GPU blending and usually `depth_write=True` (from the perspective of GPU rendering pipeline, it behaves like an "opaque" object), :

- When `depth_write = True`, it allows better utilization of early-Z testing.

- When users intentionally set its `depth_write = False`, it can, to some extent (though not perfectly), simulate the visibility of regular transparent objects behind them.

> The three render stages representing the generally-transparent objects, are forced in the order: transmissive, transparent, weighted. Ignoring the `render_queue` set by the user.

This is exactly the issue I mentioned earlier: renderQueue is essentially a higher-level sorting mechanism, while the division of rendering stages is an inherent part of the renderer’s logic. No matter how renderQueue is adjusted, the sorting only takes effect within a given rendering stage, and cannot cross stage boundaries.
That said, in the current implementation, the distinction between opaque and transparent stages is determined by whether renderQueue < 2500. As a result, apart from weighted blend objects, the rendering order can be regarded as fully following the user-defined renderQueue; weighted blend objects are the only exception.

---

Before considering how to handle Weighted Blended objects, I attempted to research the implementations of current mainstream rendering engines. However, I found that the vast majority of engines do not have built-in OIT (Order-Independent Transparency) pipelines. Even when they do, OIT is only treated as a global transparent rendering solution, separate from traditional sorting-based transparent methods. Therefore, there is a lack of readily available reference cases or best practices for using Weighted Blended objects alongside ordinary alpha blend objects.

We aim to support Weighted Blended OIT and alpha blend objects simultaneously to some extent. However, the unique aspect of Weighted Blended OIT is its inherently multi-pass nature (at least requiring Pass A: rendering OIT objects to an accumulation/revealage buffer; Pass B: compositing to the main framebuffer via a full-screen pass). This differs from the standard forward rendering process. When mixing Weighted Blended OIT objects with ordinary alpha blend objects, frequent GPU rendering state switches are difficult to avoid:

- Rendering OIT objects requires switching to an OIT-specific render pass (binding the accumulation/revealage buffers).
- Rendering ordinary alpha objects requires switching back to the main framebuffer's render pass.

If these two types of objects appear alternately, it will trigger expensive render pass switches.

A reasonable idea is: since Weighted Blended OIT objects do not participate in depth sorting themselves, there is no need to handle interleaving scenarios between them and ordinary alpha blend objects. Therefore, all OIT objects can be processed uniformly at a certain stage of the rendering pipeline, thereby avoiding the complexity introduced by their multi-pass nature—that is, treating OIT as a completely independent rendering stage, decoupled from the rendering process of other objects.

Based on this, categorizing Weighted Blended OIT objects as a separate type and assigning them an independent rendering stage is a natural implementation. Although some users might desire the flexibility to manually specify the rendering order to achieve interleaved rendering of OIT and ordinary transparent objects, the value of this flexibility is likely outweighed by the performance cost and implementation complexity it introduces. Frequent render pass switches are not only prone to unintentional misuse by users, leading to performance degradation, but mixing Weighted Blended objects with ordinary alpha blend objects might also cause unexpected rendering issues. Furthermore, this would impose constraints on the engine's feature expansion and maintenance. For example, adding support for two-pass rendering of double-sided transparent objects would necessitate additional handling for compatibility with Weighted Blended OIT objects, introducing unnecessary complexity.

**I believe this design choice is essentially an engineering trade-off. The core of real-time rendering technology lies in finding a balance between visual realism, performance cost, and implementation complexity. This is also the fundamental distinction between modern real-time rendering and offline rendering.** While the basic laws of physics are inherently simple—light propagation, energy conservation, laws of reflection and refraction, all expressible precisely in mathematics—offline rendering (e.g., path tracing, Monte Carlo integration) tends to follow these physical laws directly, approximating the real world through extensive sampling and computation. In contrast, real-time rendering heavily employs approximations and simplifications (e.g., BRDF approximations, precomputed IBL) to achieve acceptable visual quality within the constraints of limited performance budgets and engineering complexity. Handling Weighted Blended OIT as a separate stage adheres to this philosophy.

A similar trade-off is evident in the handling of transmissive light in #974. For transmissive objects, screen-space refraction (SSR) is needed to gather information about the scene behind them.
I initially considered generating a separate scene framebuffer for each transmissive object, but this would be too expensive. Each transmissive object would require re-rendering the entire scene, which is impractical for real-time rendering.
Subsequently, I thought about utilizing transparency sorting and layer-by-layer accumulation (hence the most significant change initially made to the Pygfx renderer in this PR was the requirement for transparency sorting). The result after drawing each object could serve as the background for the next object, thus avoiding the costly overhead of rendering the scene multiple times. However, since the GPU does not allow simultaneous reading and writing to the same texture, double buffering (ping-pong buffers) or an intermediate copy mechanism would be required. This would still cause render pass interruptions. When there are multiple transmissive objects in the scene, although it wouldn't require extra re-rendering of the scene, frequent render pass interruptions and switches would still lead to performance degradation (This is the same as the problem with weighted blended objects.). Moreover, this approach still cannot handle cases where transmissive objects intersect or self-occlude.

Therefore, I tried looking at implementations in mature engines, and the final answer was straightforward. The methods adopted by the reference implementation of the glTF 2.0 Transmission extension, as well as engines like Three.js, Babylon.js, and Unity HDRP, are all very "simple and direct": after all opaque objects are rendered, a copy of the background color buffer is generated, and all transmissive objects share this buffer for sampling. Although this scheme cannot handle the superposition of multiple layers of transmissive objects, it is simple, efficient, and practical.

I believe this is a simple trade-off, and one that makes sense. Firstly, in the vast majority of scenarios, there won't be overlapping and superposition of multiple transmissive objects. Furthermore, adopting this fixed strategy also inversely requires scene designers and technical artists to optimize scene design accordingly, rather than abusing the feature. Content creators inherently need to optimize asset structure and scene layout within the engine's capabilities to achieve the best visual results.


All these engines divide the rendering process into multiple stages—at least an opaque stage and a transparent stage—and perform specific tasks between these stages based on logical requirements. Generating the SSR framebuffer is one such task. Some global illumination, AO, volumetric fog, and even certain post-processing effects also need to be inserted between the opaque and transparent stages. Unity even exposes injection points between these stages to let users plug in custom logic or passes.

panxinmiao:
Sorry for post such a “long essay” without properly dividing it into paragraphs and section titles, 😅 . I’ll add some examples and supporting facts to certain arguments and viewpoints in the text when I have more time.

almarklein:
>  No matter how renderQueue is adjusted, the sorting only takes effect within a given rendering stage, and cannot cross stage boundaries.

I think you're confusing `material.renderQueue` with `ob.renderOrder` here. The renderOrder sorts objects *within* a pass/queue, and the renderQueue *is* the pass/stage.

From what I understand how Unity works. It sorts all objects based on the renderQueue, then on the depth. Then per renderQueue it renders all objects in that queue, which can be a mix of objects requiring different render passes. The SSR texture is copied right when the first (transparent) object that needs it is to be rendered.

> Therefore, I tried looking at implementations in mature engines, and the final answer was straightforward. The methods adopted by the reference implementation of the glTF 2.0 Transmission extension, as well as engines like Three.js, Babylon.js, and Unity HDRP, are all very "simple and direct": after all opaque objects are rendered, a copy of the background color buffer is generated, and all transmissive objects share this buffer for sampling. 

Awesome! This can be done perfectly fine with the way we iterate over objects in current `main`. What we'd have to do, is copy the SSR texture as soon as it encounters an object that needs it. 

Imagine the case where a user does not set `render_queue`, so all transparent (normal and translucent) are in 3000. So right before  the transparent objects are rendered, we copy the SSR texture. This is quite literally the behavior that you call the "simple and direct" approach in your comment.

> Although this scheme cannot handle the superposition of multiple layers of transmissive objects [...]

But with `render_queue` we can! If we copy the SSR texture for every renderQueue that contains objects that need it, the user can create scenes as complex as he wants. Also, the user can easily control whether e.g. weighted objects are drawn before or after, by setting renderQueue.

As for mixing the different kinds of transparent objects in a single render_queue. I don't really see a problem. The normal and translucent objects can be mixed, as long as we just copy the SSR texture beforehand. The weighted objects are also fine, because they get grouped separately from the other objects automatically because of the depth-sorting (their key is zero).

So there is no need to fear these "render pass interruptions"; by default there would be just one SRR copy, and one group of weighted objects (if any). Only when the user uses `render_queue` there can be multiple such passes, enabling better rendering of specific scenes.

panxinmiao:
In fact, the engine is deliberately designed to avoid creating multiple SSR textures for different transmissive objects.

In the rendering pipeline, a full-resolution screen-sized copy is already a very heavy operation, and generating mipmaps for the SSR color buffer adds even more overhead. When multiple transmissive objects are present, this quickly becomes a performance bottleneck.
Moreover, in the transparent/transmissive stage, the engine sorts all objects (back-to-front) and renders them in batches. If a “copy framebuffer” barrier were inserted every time a transmissive object is encountered, it would break the rendering pipeline, disrupt batching, and cause a GPU pipeline flush — which is extremely costly.

I can confirm that Unity HDRP only maintains a single SSR buffer, generated at a fixed stage in the renderer. Unity also provides a BeforePreRefraction injection point, which is specifically placed before SSR generation.

This is the documentation for Unity HDRP:

https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@17.3/manual/how-hdrp-calculates-color-for-reflection-and-refraction.html
> The color buffer HDRP uses is the first color pyramid that contains only opaque objects, so refractive objects won't be visible through other refractive objects. 

https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@17.3/manual/Custom-Pass-buffers-pyramids.html#depth-pyramid-and-color-pyramid-generation-in-hdrp
> HDRP generates color pyramids at these points in the rendering pipeline:
> 1. After the BeforePreRefraction injection point (in Color Pyramid PreRefraction). This contains opaque objects and objects rendered in BeforePreRefraction and is used for Screen space reflections next frame and the refraction effect in the current frame.

Blender seems to be the same:
https://community.khronos.org/t/objects-with-transmission-hiding-other-objects-with-transmission/107669

I believe the same principle applies to weighted blended objects: they should be handled in a single batch. Otherwise, the overhead of frequently interrupting the rendering pipeline and switching rendering states would far outweigh any potential benefits.


> But with render_queue we can! If we copy the SSR texture for every renderQueue that contains objects that need it, the user can create scenes as complex as he wants...

As I mentioned earlier, I believe these measures are, in fact, intentional restrictions on users creating arbitrarily complex scenes. Such complexity is usually unnecessary, and may be abused. and the engine in fact may not be able to handle them well, which could lead to potential issues.

panxinmiao:
> I think you're confusing material.renderQueue with ob.renderOrder here. The renderOrder sorts objects within a pass/queue, and the renderQueue is the pass/stage.

https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition%4017.0/manual/rendering-execution-order.html

This documentation clearly outlines Unity’s rendering pipeline stages, including the fundamental opaque and transparent stages, and further explains the exact timing of tasks executed within or between these stages.  In the transparent stage, Unity also processes refraction first and then transparents, which is consistent with our logic.

As I mentioned earlier: RenderQueue is essentially a higher-level sorting mechanism, while the division of rendering stages is an inherent part of the renderer’s logic. 


I haven’t found detailed documentation on how Unity assigns renderable objects to different rendering stages, but I believe the division is based on whether the object’s renderQueue is greater than 2500 (opaque stage or transparent stage).  Therefore, it can also be said that RenderQueue determines which rendering stage an object is assigned to.

https://docs.unity3d.com/ScriptReference/Rendering.RenderQueue.html
The documentation emphasizes that users should “**Use the render queue section that matches the object type**”, and it also specifies that **2500 is the last render queue treated as opaque**.

almarklein:
> I believe these measures are, in fact, intentional restrictions on users creating arbitrarily complex scenes. Such complexity is usually unnecessary

Sure, any added flexibility is also an opportunity for users to do something stupid 😉. I think the design would naturally allow users to create advanced use-cases. But I'm fine with restricting the SRR to being created only the first time it's needed. 

> Therefore, it can also be said that RenderQueue determines which rendering stage an object is assigned to.

Yes, you could say that. I'd even say there is not really an explicit "rendering stage". Objects are sorted by `render_queue`, and some things (like whether objects sort front-to-back) are different based on certain *ranges* of the render-queue. In the common case where objects only use render-stages 2000 and 3000 (and maybe 1000 for background), you have exactly the same mechanics as a "two-part render stage".

panxinmiao:
> Yes, you could say that. I'd even say there is not really an explicit "rendering stage". 

No, for a renderer, rendering stages are explicitly defined—this is a fundamental aspect of pipeline design. Unity follows the same principle. Tasks such as SSR generation are also executed at a specific time points. This has already been explained in the Unity document that I mentioned earlier: 
https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition%4017.0/manual/rendering-execution-order.html

https://docs.unity3d.com/ScriptReference/Rendering.RenderQueue.html

According to Unity’s documentation, whether an object belongs to the transparent stage is determined by whether its RenderQueue value is greater than 2500. In other words, both 2600 and 3000 are rendered in the transparent stage, with lower values rendered earlier. RenderQueue itself is not the render stage; rather, it is a higher-level sorting mechanism within a stage, intended to provide flexibility for batching optimizations. For most users, Unity recommends following the predefined RenderQueue conventions to avoid potential rendering issues.

PR #1126, opened 2025-06-26
Fix that snapshot may not include the last effect

almarklein:
See https://github.com/pygfx/pygfx/pull/1114#issuecomment-2991507972)

### DIFF
--- a/pygfx/renderers/wgpu/engine/renderer.py
+++ b/pygfx/renderers/wgpu/engine/renderer.py
@@ -287,6 +287,9 @@ class WgpuRenderer(RootEventHandler, Renderer):
 
         self._blender = Blender()
         self._effect_passes = ()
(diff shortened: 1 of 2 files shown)


pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -888,9 +899,16 @@ def snapshot(self):
...
+            self._name_of_texture_with_effects or "color"
+        )
         size = texture.size
-        bytes_per_pixel = 4
+        if texture.format == "rgba16float":
+            bytes_per_pixel = 8
+            dtype = np.float16
+        else:
+            bytes_per_pixel = 4
+            dtype = np.uint8
```
almarklein, line 911: Just an artifact because I took this out of #1125.

PR #1127, opened 2025-06-30
Use rendercanvas in all our examples

almarklein:
High time we switch ... so https://github.com/pygfx/wgpu-py/issues/627 can move forward.

* Replace `from wgpu.gui import WgpuCanvas, run` with `from rendercanvas import RenderCanvas, loop`
* Replace `run()` with `loop.run()`
* Examples that use Qt required a bit more care.
* Use `update_mode='fastest'` where `fps=-1` was used. See https://github.com/pygfx/rendercanvas/pull/99
* I tested quite a lot (but not all) examples. In the majority of examples the change is trivial.

### DIFF
--- a/examples/feature_demo/animation_mixer.py
+++ b/examples/feature_demo/animation_mixer.py
@@ -33,14 +33,16 @@ except NameError:
 
 import pygfx as gfx
 from pygfx.utils.text import FontProps
-from wgpu.gui.auto import WgpuCanvas, run
+from rendercanvas.auto import RenderCanvas, loop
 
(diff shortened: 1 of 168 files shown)


Korijn:
Require new rendercanvas version maybe?

Korijn (dismissed):

almarklein:
Oh dear, the `validate_text_md` example is glitching. It did that in an early stage of #1121, but later the glitch disappeared. I propose that we add a little exception, allowing this example to have more leeway.

Korijn (approved):

PR #1139, opened 2025-07-10
Improve dithered alpha_mode

almarklein:
The dithered alpha_mode (currently called blending) has some really nice properties, being able to handle transparency for complex scenes in an order-independent way. The only downside is that it produces noisy images.

This PR improves that, by at least making the noise stable across frames *and* make it move along with the object. 

* [x] Use blue noise.
* [x] Also support for Bayer pattern. 
* [x] Use an element id as seed to prevent overlapping parts from getting the same pattern. 
* [x] Prevent different objects from getting the same pattern, by letting the renderer assign id's.
* [x] Tweaks to shader resolved so that varyings used by the blending are detected. 
* [x] example videos!

---

Screenshot of  dithered line, before and after this PR. 

image

This one is of a point cloud. Correctly rendering a point cloud like this is a typical problem that no other method can do well. (the noisy nature of these images result in mpeg artifacts ... I guess that can be considered a downside of this method).

https://github.com/user-attachments/assets/48169a39-d131-4107-8b73-b5d27724709c

Classic example of a simple scene that is impossible to get right with other blend modes

image

Here's a torus knot, with alpha 0.7. No way to render this with `alpha_mode='blend`, except if you cut the object in multiple pieces.

image

Maybe not a very practical example, but if we up the pixel-ratio x4, and use alpha=0.9. This illustrates that if performance is not the priority, you can use dithering with extreme upscaling to get super-sweet results.
image

Using the Bayer pattern for dithering looks really nice. The downside is that it cannot really blend multiple layers (only if later layers have higher alpha). But that can also be used as feature: e.g. render a transparent line without overlap 😉 

image

And the stanfort dagon:

https://github.com/user-attachments/assets/3fdbd964-85a4-4e95-a459-3f45b009720c


almarklein:
This idea came out of #1124 and https://github.com/fastplotlib/fastplotlib/pull/873. I'm still thinking a lot about what defaults should be and how and where we can automate common use-cases.

One idea is: if we require users to set `alpha_mode='blend'` for transparent objects, we might as well make 'dither' the default, because it can deal with objects that mix opaque and transparent objects, and transparent objects will actually be transparent (although a bit noisy).

almarklein:
@hmaarrfk thanks for the tip on blue noise: left old (white?) noise, right blue noise:

image

almarklein:
I found a better implementation for blue noise. It affects the results *a lot*:

image

hmaarrfk:
i find this hilarious... what is motivating this "dithered" mode???? it is cool, but it is useful in 2025?

Korijn:
> i find this hilarious... what is motivating this "dithered" mode???? it is cool, but it is useful in 2025?

For example Nintendo switch games apply it all the time, even the triple A titles. It's super efficient.

Transparency is computationally expensive, and so if you want to apply it to scenes with many elements, you can save some time by using dithering for scene elements that don't need to blend perfectly.

If applied tastefully, it is hardly noticeable.

almarklein:
> it is cool, but it is useful in 2025?

It's useful (and cool) because it offers *perfect* handling of transparency. In the sense that the order that objects are drawn does not matter, while it *does* result in correct contributions from fragments, based on their distance from the camera. So it's a very effective order-independent-transparency method. Further, it can deal with objects that are a mix of opaque and transparent fragments. There's basically no other method with these properties, unless you consider expensive methods like depth peeling.

To do this, it basically sacrifices spatial resolution, so that the "blending" does not occur per-fragment, but among a small region of fragments, using a stochastic process. Hence the noise. And apparently, we can improve the properties of the noise a lot. I also found that using SSAA (which we already do) is much more effective with the blue noise.

I have some examples running that show pretty amazing graphics, that are really hard to achieve with other blending approaches. Will post some video's soon.

Korijn:
There is a nice screenshot of dithering transparency in Mario Odyssey here: https://danielilett.com/2020-04-19-tut5-5-urp-dither-transparency/

And the video at the end shows an example of an animated transparent robot. Looks pretty great to me!

almarklein:
To add to that ... when using "classic blending", objects are either opaque or transparent, one has to deal with sorting, which helps but does not cover all cases, and artifacts can occur if things don't really work out. Users will need to be aware of the possibilities and limitations to deal with such issues. With dithering, it just works. 

Oh, also worth noting that we can mix dithered objects with blended objects. This is super-interesting, because when there's a conflict between say two objects that cannot be blended together well, you can likely use dither for one, and blend for the other (because the dithered object produces opaque fragments and writes to the depth buffer). In other words, using dither for some objects, can help render objects that use blending render better.

almarklein:
One more comment, and then I'll stop. In that Mario video you can see the dither has a distinct pattern (I think they call it Bayer) instead of just noise. This can also be interesting to add at some point, to give such objects a stylistic look.

almarklein:
I added screenshots and videos to the top post.

Korijn:
Awesome!! Did I nerdsnipe you with the bayer pattern? :)

almarklein:
> I nerdsnipe you with the bayer pattern? :)

Yeah, and now I found out we can blend it relatively well across objects. Look at this 😮 

image

almarklein:
I first nerd-sniped myself to let the pattern move along with the object. 

Then Mark nerd-sniped me with the blue noise.

You nerd-sniped me with the Bayer pattern.

BTW: I removed the part where the noise moved along with the object, because it resulted in flickery noise with multiple overlapping regions, and in blackish artifacts when triangles reach the size of a pixel. So now the noise pattern is stuck to the screen, but since the noise is much finer, it's much harder to notice.

Korijn:
> > I nerdsnipe you with the bayer pattern? :)
> 
> Yeah, and now I found out we can blend it relatively well across objects. Look at this 😮

I always wondered if it's possible to do dithering with subpixels (the literal individual LEDs of the pixels). 🤪

almarklein:
> subpixels

combines with ssaa, you're using subpixels quite literally

almarklein:
(wrong button)

Korijn:
> > subpixels
> 
> combines with ssaa, you're using subpixels quite literally

That's true! :) awesome. It looks fantastic.

Issue #1137, opened 2025-07-09
Using LRU strategy for GpuCache


panxinmiao:
Currently, our `GpuCache` uses a `WeakValueDictionary` to store objects, which means resources are immediately destroyed once they are no longer in use.
While this keeps memory usage minimal, it limits the effectiveness of the cache — especially in cases where a resource is briefly unused but likely to be reused shortly after.

For example, consider a rendering loop involving double-sided transparent materials,  we render the back and front separately:

```python
material. Side = "back"
# do render
material.side = "front"
# do render again
```
Each time the `side` property is switched, the corresponding `RenderPipeline` is destroyed and rebuilt, which incurs a significant performance cost.

Switching to an LRU strategy for managing the `GpuCache` could greatly improve this. By retaining a limited number of recently used resources, we can avoid frequent destruction and reconstruction, leading to better performance in dynamic rendering scenarios.

PR #1142, opened 2025-07-14
Implement GpuCache with LRU strategy (split from #974)

panxinmiao:
Implements #1137

Split from #974, The new cache improves performance during two-sided transmissive object rendering introduced in #974.

It is also beneficial in dynamic scenes where object properties change frequently—allowing GPU resources to be reused via cache switching rather than full recreation.

### DIFF
--- a/pygfx/renderers/wgpu/engine/pipeline.py
+++ b/pygfx/renderers/wgpu/engine/pipeline.py
@@ -21,10 +21,10 @@ from .binding import Binding
 
 PIPELINE_CONTAINER_GROUPS = WeakAssociativeContainer()
 
-# These caches use a WeakValueDictionary; they don't actually store object, but
-# enable sharing gpu resources for similar objects. It makes creating such
-# objects faster (i.e. faster startup). It also saves gpu resources. It does not
-# necessarily make the visualization faster.
+# These caches use an LRU strategy.
+# They store a limited number of recently used GPU resources to improve cache hit rates.
+# This enables reuse of GPU resources across similar objects, reducing the overhead of
+# repeatedly creating/destroying GPU resources, especially in dynamic rendering scenarios.
 LAYOUT_CACHE = GpuCache("layouts")
 BINDING_CACHE = GpuCache("bindings")
 SHADER_CACHE = GpuCache("shader_modules")
--- a/pygfx/renderers/wgpu/engine/utils.py
+++ b/pygfx/renderers/wgpu/engine/utils.py
@@ -3,7 +3,7 @@ Utils for the wgpu renderer.
 """
 
 import json
-import weakref
+from collections import OrderedDict
 
 import wgpu
 
@@ -284,21 +284,22 @@ gpu_caches = GpuCaches()
 
 
 class GpuCache:
-    """A cache for GPU objects."""
+    """A cache for GPU objects using LRU eviction strategy."""
 
-    def __init__(self, name):
+    def __init__(self, name, max_size=128):
         assert isinstance(name, str)
         assert not hasattr(gpu_caches, name)
         setattr(gpu_caches, name, self)
 
-        self._objects = weakref.WeakValueDictionary()
+        self._objects = OrderedDict()
+        self._max_size = max_size
         self._enabled = True
         self.hits = 0
         self.misses = 0
 
     def get_stats(self):
         """Get the number of (alive) objects in the cache."""
-        return len(list(self._objects.values())), self.hits, self.misses
+        return len(self._objects), self.hits, self.misses
 
     def enable(self):
         """Enable this cache."""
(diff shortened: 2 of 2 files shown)


Korijn:
Love it! Looks great.

almarklein:
I get the problem that this solves, and the implementation itself looks good to me, but I'm not sure if an LRU cache is the best solution here. The old implementation is removed, which allowed re-using a pipeline etc. from any objects that is currently alive. With the new implementation you can only resuse from the last 128 objects. Plus it may keep alive over 100 objects that are never going to be used again. I have no idea what would be the potential cost of that.

The solution does avoid weakrefs, which may help performance. 

Since this is a rather low-level bit, I'd like to understand a bit more about the implications.

If the performance benefit from avoiding weakrefs is not significant, we could stick with the old solution, and keep objects alive based on time, using something like `self._keep_alive.setdefault(int(time.perf_counter()), []).append(ob)`.

Korijn:
> I'm not sure if an LRU cache is the best solution here

I just want to note here that I am accustomed to having an LRU cache as the default in rendering libs. I think it's the common solution.

> which allowed re-using a pipeline etc. from any objects that is currently alive

I'm curious about the situation where there are more than 128 objects in use. Will the cache be essentially disabled in that case? I guess in that case the old implementation would be better then.

panxinmiao:
I think compared to `WeakValueDictionary`, LRU is definitely more suitable. 
Let me try to explain, I hope I can express my point as clearly as possible, 😅 

---

### Firstly, The purpose of a cache is performance, not minimal retention.

The core goal of this cache is to improve performance by avoiding redundant GPU resource creation, especially expensive objects like pipelines and shader modules.

The issue with WeakValueDictionary is that it offers zero retention: as soon as an object loses all strong references, it is immediately destroyed — even if it is about to be needed again in the same or next frame. This behavior defeats the core purpose of a cache. In dynamic rendering scenarios, this causes frequent GPU pipeline rebuilds, which has proven to be a real performance bottleneck.

___

> With the new implementation you can only resuse from the last 128 objects.

> I'm curious about the situation where there are more than 128 objects in use. Will the cache be essentially disabled in that case?

### Not last 128 objects — it keeps the 128 **most recently used** cache entries.  and  it targets exactly the use case where caching is needed.
This means:
- It naturally retains high-frequency objects, which are most likely to benefit from caching.
- It automatically evicts stale or one-off entries without additional bookkeeping.
- It gives us a **deterministic memory bound** and avoids potential memory bloat.

In practice, most GPU resources are limited and reused — e.g., the number of unique pipeline/shader combinations in a typical scene is small and stable. This means:

- The 128-entry cache is very likely to be sufficient.
- **In most scenes, eviction may never even occur.**

And if needed, we can always make the capacity configurable or auto-tuned based on usage patterns.

Most stable/static objects will retain their GPU resources directly — they don't rely on the cache. The cache becomes valuable only for dynamically created or reconfigured objects (e.g., procedural material tweaks or per-frame property switches).These are precisely the cases that benefit from short-term caching.

Once the cache is full, the items that are evicted are, by definition, the least frequently or recently used. **If you ever reach a state where even your 128 most recent items are being frequently evicted, your scene is absolutely pathologically inefficient (hundreds of unique pipelines per frame),  and I think this is not a problem that can be solved solely by caching, you definitely need other rendering strategies.**

___

> Plus it may keep alive over 100 objects that are never going to be used again. I have no idea what would be the potential cost of that.

Yes, the LRU cache may retain unused entries for a while. But I think it's not a big deal, as is mentioned above, the cache gives us a **deterministic memory bound** already. And this situation is actually a good thing in practice — because it means:
- The cache has spare capacity.
- We're avoiding unnecessary GC thrashing or recreation of GPU resources.

Also, this is a relatively simple (but highly effective,  😄 ) LRU strategy. If we later want to improve it further, Base on LRU, we could consider add Time-based eviction (e.g. evict if not accessed for N seconds) and Idle scanning during frame intervals. In this way, we can increase or even completely cancel the limit of cache size.

But in the current state, I think this version already gives us a major win with minimal complexity and good general-purpose performance.

Korijn:
I am a big fan of this PR. I've used LRU caches in practice many times to solve performance issues in interactive animated scenes with frequently toggling materials (can even be simple things like on hover highlighting or color changes).

I really think we should include this change in pygfx!

almarklein (approved):
Thanks for the explanation. Sorry for taking so long to respond. I was at EuroPython and now holidays, and I focused the little time I could find on the blending stuff.

I think you're right about it being highly unlikely to have more than 128 different pipeline objects in a scene.

I think I am mostly a bit reluctant of indefinitely storing stuff that may never be used again. Also because I'm not sure how large these objects can become and how it may affect the GPU. 

But I am going to trust both your judgement on this :)

Korijn:
Only one way to find out :)

But it's such common practice that I am not really concerned.

PR #1146, opened 2025-07-21
Add `output_colorspace` property to Renderer

panxinmiao:
This PR introduces a new output_colorspace property to the Renderer class, which specifies the desired output color space for the final rendered image. This addition provides better control over how linear color values are converted before being written to the screen or output textures.

Motivation:
Modern rendering workflows often involve rendering in a linear color space (as recommended for physically based rendering), but displaying or exporting images typically requires gamma encoding or color space transformation (Our default behavior). But for some cases and scenarios, we want to output the results to a linear space, such as scientific computing, post synthesis, and so on.

By adding this property, we make it explicit how the renderer should treat the final output. It will also affect the color space of the image saved by the screenshot function.


panxinmiao:
I’ve extended the `validate_srgb.py` example by adding a corresponding version that outputs to linear (physical) space.

Together, these two examples clearly demonstrate several key concepts: colorspace, interpolation space and the nonlinear nature of human brightness perception.

### Background:
The human visual system perceives brightness non-linearly — it is more sensitive to darker regions and less sensitive to brighter ones. However, physical light intensity is linear (based on photon count).

As a result, in order to match human perception, textures are often stored in sRGB space (a non-linear encoding), while shading and lighting calculations are usually performed in linear space.

### Example 1 (validate_output_physical.py): Output to Linear (Physical) Space

validate_output_physical

- Top image: The original data is linear and in the physical space, and it's output directly to the same space. No color space conversion is applied. This result is physically (as well as numerical ) correct, 

- Second image: The data is still linear but in the sRGB space, and the output is in physical space. Interpolation happens in sRGB space (the sRGB-to-linear conversion happens in the shader after interpolation).  It appears visually smooth, because interpolation takes place in a space aligned with human perception. It looks a litter darker (e.g., around midtones, sRGB 0.5 becomes ~0.214 in physical).

- Third image: The data is linear and in sRGB space. During sampling, hardware automatically converts it to physical space before interpolation. This approach is physically correct, but because the data is non-linear in physical space, interpolation can lead to visible banding.

### Example 2 (validate_output_srgb.py): Output to sRGB Space

validate_output_srgb

- Top image: The data is linear and stays in physical space during processing, then output to sRGB space (typically via GPU conversion). The image will become more brighter  (e.g., around midtones, physical 0.5 becomes ~0.73 in sRGB), it looks much brighter and non-linear. This often leads to an overexposed look, which is a common mistake in graphics rendering.

- Second image: The data is linear but in sRGB space, and also output in sRGB. It undergoes a round-trip conversion (sRGB → linear → sRGB), with interpolation happening in sRGB space. The final result appears visually similar to the top image from Example 1.  However, for scientific or precision-critical applications, this approach may introduce small inaccuracies. This is because the sRGB → linear conversion is performed in the shader (using an approximate formula), while the linear → sRGB conversion is handled by the GPU’s fixed-function sRGB OETF (opto-electrical transfer function). These two processes are not guaranteed to be mathematically symmetrical, and the mismatch can result in subtle but measurable errors in round-trip conversions.

- Third image: Again, the data is linear and in sRGB space. Hardware converts it to linear space for interpolation, then converts back to sRGB for output. banding may occur due to interpolating nonlinear (encoded) data in linear space, but this is physically correct,


### PBR Rendering vs Scientific Visualization

In Physically-Based Rendering (PBR), the entire rendering pipeline operates in linear space. Input textures (such as base color) are usually sRGB-encoded for artist convenience, but are converted to linear space when sampled. Lighting and material computations occur in physical space, and the final result is output in sRGB space, preserving dark detail and matching human visual perception. This is the standard in most art-driven pipelines.
This corresponds to the processing procedure of Third image in Example 2.

In scientific visualization or numerical computing, where data accuracy is critical, the workflow typically stays entirely in linear (physical) space — from input to output — without applying perceptual encoding and colorspace conversion. 
This usually use the processing procedure of Top image in Example 1.

panxinmiao:
Regarding color space, we've discussed this topic many times. I apologize for bringing it up again. 😓 

To meet the requirements of physically based rendering (PBR) and lighting correctness, you introduced the `"tex-srgb"` texture format in PR #906 .However, I haven't yet updated the related code to use it. This is not only because I think it makes shader logic more complex and verbose, but also because I genuinely believe that the current design and handling of texture and color space is problematic.

These concepts are quite tricky, and I’ve been trying to find better ways to express my thoughts. But I find it difficult to clearly and unambiguously articulate them (again, sorry for the confusion), so I’ve decided not to explain my opinion directly anymore. Instead, I will present a few specific examples to illustrate the facts.

### First, a declaration:

**Data is just data** — whether it lies in sRGB space or linear (physical) space depends on how we treat it. It's incorrect to assume that a `"unorm"` texture must be in linear space, or that a `"-srgb"` texture must be in sRGB space.

Each time I try to explain this, I tend to cause more confusion due to the limitations of language, so I’ll instead quote the following from the [wgpu wiki](https://github.com/gfx-rs/wgpu/wiki/Texture-Color-Formats-and-Srgb-conversions).
> First of all, it will be useful to clarify some terminology as the word “linear” and “sRGB” can be overloaded and lead to confusion:
>
> * sRGB space is a set of color primaries — this defines what red, green, and blue are.
> * In this sense (and perhaps counterintuitively), texture formats ending in `"Unorm"` and `"UnormSrgb"` are BOTH in the sRGB color space and encode their data in the same way, without any loss in precision (contrary to some explanations of “linear” vs “sRGB”). More on this later.
>
> Useful terminology to avoid confusion:
>
> * What we traditionally call "linear space" is more accurately called **scene-referred color**.
> * What we traditionally call "sRGB space" is more accurately called **monitor-referred color**.
> * The conversion from scene-referred to monitor-referred is handled by the **sRGB OETF** (opto-electrical transfer function \[light → bits]).
> * The conversion from monitor-referred to scene-referred is handled by the **sRGB EOTF** (electro-optical transfer function \[bits → light]).

---

### Let's examine a few specific cases

To make things easier, let’s assign IDs to the workflows described in the earlier examples:

**Workflow 1:** (First image in Example 1)
Input texture is in linear (physical) space, output is also in linear (physical) space — no any color space conversions.

**Workflow 2:** (Second image in Example 2)
Input texture is considered to be in sRGB space. Output is also in sRGB space, but interpolation happens in sRGB space. The shader manually converts sRGB to linear, and then output is converted back from linear to sRGB using the GPU’s hardware sRGB OETF.
This is our **default pipeline** — used in scientific computing, visualization, and all scenarios that are meant to work in **sRGB space (human perceptual linear space)**.

**Workflow 3:** (Third image in Example 2)
Input texture is in sRGB space (using the `-srgb` format), output is also in sRGB space. The GPU performs EOTF (sRGB to linear) before sampling. All interpolation and sampling happen in physical space, then the result is passed through the GPU’s OETF for output to sRGB space.
This is the **physically correct pipeline** required for **PBR rendering and lighting calculations**, which must operate entirely (interpolation sample and compute) in physical (linear) space. We introduced the `"tex-srgb"` format to support this workflow.

---

### A counterintuitive point:

For use cases that **work in sRGB space** (human perceptual linear space), we **should be using Workflow 1**, not Workflow 2.
In fact, Workflow 2 is problematic in all scenarios — and not just because the two transformations (`sRGB → linear` in shader and `linear → sRGB` via GPU OETF) may not be mathematically symmetrical.

Imagine a visualization use case:

* When data value `a = 0`, we use color(srgb) = 0 (black).
* When value `b = 100`, we use color(srgb) = 1.0 (white).
* When value `c = 50`, we want a perceptual mid-gray, color(srgb) = 0.5.

**In Workflow 1:**
We treat 0 and 1 as sRGB values (even though the texture is in physical space format), perform computation, and output 0.5 directly — which gives us the desired perceptual result.
Because the input colors are already in sRGB space, and **no space conversion** is done, we can consider the entire pipeline to operate in sRGB space, and the result is **numerically accurate**.

**In Workflow 2:**
Input 0.0 and 1.0 are treated as sRGB, then converted to linear in the shader. and 0,0 still 0.0, 1.0 still 1.0, After computation, the result is 0.5,  then passed through the GPU’s OETF — 0.5 is transformed to around 0.73. But this is **not** what we want — it looks too bright.
In fact, if we get result by interpolating directly from srgb texture(that's exactly what this workflow does), we got 0.5, then convert that to linear (e.g., 0.5 → \~0.218), and then pass it through the GPU OETF again (0.218 → \~0.5), the final value **happens** to be correct — but only **by coincidence**, because we are in a wrong workflow and doing the wrong operation, which happened to give a seemingly correct result.

I understand that you're using Workflow 2 and the `srgb2physical` shader function to support scientific visualization and other cases that rely on perceptual linearity (sRGB space). Since users work in sRGB space and interpret color data as sRGB-encoded, this seems natural. But as the examples show, **Workflow 2 is fundamentally flawed**.

For visualization (working in perceptual linear space), users don't need to care about color space at all — and should not use sRGB colorspace textures (neither input nor output).
Only for **PBR or lighting computation**, where **physical correctness** is required (working in physical space), should we use color space management and sRGB colorspace textures.

A general takeaway here is: if a shader operates in sRGB space — meaning that all computations and manipulations are intended to be interpreted as sRGB-encoded color — then you should simply write to a regular unorm texture and output the values as-is. There is no need to perform any additional color space conversion.

Conversely, if the shader works in physical (linear) space — where values represent scene-referred linear (physical) light — then you must use an -srgb texture format to ensure the correct conversion to monitor-referred sRGB space upon output.

This is also clearly illustrated in the `audio_visualizer.py` example. The shaders used there were originally created by artists on ShaderToy platform, where the creative process happens directly in perceptually linear (i.e., sRGB) space. There are no physical processes here, these shaders assume sRGB encoding by design and operate entirely within that space.

Before this PR, we mistakenly forced their output into -srgb textures — which applied a hardware linear → sRGB conversion. This caused a significant distortion (over-brightening) of the final image. To compensate, we had to insert an extra srgb2physical() step in the shader before final output to manually undo that conversion.

In reality, the correct solution is far simpler: just write directly to a unorm texture without any conversions — as this PR now does.

I think, for most modern GPU render workflows, you should never need to write srgb2physical or physical2srgb logic inside shaders — unless you have a very specific, non-standard requirement and fully understand the color semantics you're dealing with.

---

### One more problem with Workflow 2 in scientific use cases:

For scientific computing and visualization, even if users are okay with the potential inaccuracy of Workflow 2, there's another limitation:
By default, we use `unorm-srgb` textures, i.e., `uint8` data. But if users want **higher precision** (e.g., `float32`), they cannot achieve this without modifying shaders — because higher precision formats like `rgba32float` **do not support `-srgb`** formats. Therefore, users must modify the shader and manually apply the process of `physical2srgb`.
From this perspective, scientific computing and visualization **should not use sRGB textures at all** — not for input, and definitely not for output.

PR #1148, opened 2025-07-24
Enable weighted blending between calls to render

hmaarrfk:
This enables the blender to retain its state while it hasn't been completely cleared.

This is useful if users are trying to render a scene manually, object by object, streaming in the objects from disk or other source slowly.

One can recreate the same effect from the image_stitching.py example

Part of this comes from our desire to display truely huge images.


pygfx/renderers/wgpu/engine/blender.py
```diff
@@ -364,14 +364,14 @@ def get_color_attachments(self, pass_type):
                 "name": "accum",
                 "resolve_target": None,
                 "clear_value": accum_clear_value,
-                "load_op": wgpu.LoadOp.clear,
+                "load_op": wgpu.LoadOp.load,
```
hmaarrfk, line 367: Hey @almarklein i hope this 2 line change is OK. 

it allows one to keep these buffers alive between calls to `renderer.render` so that things can blend well!

You can see the streamed validation example for the effect.

hmaarrfk:
I can see an argument that some might need to clear the weights between calls to renderer especially when they are using differerent cameras for an unrelated "scene".

But i figured this would simulate a workflow that we often have where we "stream" in the data from a larger source and can't keep all the image data in GPU memory at the same time.

pygfx/renderers/wgpu/engine/blender.py
hmaarrfk: I just came by a usecase where being able to "clear" between calls to `render` is useful. not sure if can find a way to control this behavior

hmaarrfk:
I think with my latest commit, the user has enough control in a fashion similar to other buffers to create complex effects with overlays.

On the left is without the additions to `clear`, the right is `with` 
image
image

The parameter names can probably use some work, but it seems pretty cool now.

hmaarrfk:
The triangle draw using the new weighted pipeline to highlight the fact hat you can fully utilize the weighted blending properties for any "scene", not just the "first" one.

pygfx/renderers/wgpu/engine/renderer.py
```diff
@@ -536,22 +536,27 @@ def effect_passes(self, effect_passes):
...
             self._blender.clear()
-        elif color:
+            return
+
+        if color:
             self._blender.texture_info["color"]["clear"] = True
-        elif depth:
+        if depth:
             self._blender.texture_info["depth"]["clear"] = True
+        if blend:
```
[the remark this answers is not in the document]
almarklein: I made it "weights"

Korijn (approved):

hmaarrfk:
screenshot has been uploaded from CI.

Korijn (approved):
awesome example.

hmaarrfk:
> awesome example.

thx thx.

almarklein:
> screenshot has been uploaded from CI.

Probably not that important, but I'm a bit confused. The new validation example was not failing, but you uploaded a new image and it still works. Or is the new image only different in its metadata?

hmaarrfk:
Hmm. The last two commits were mine preceded by yours. 

Yours shows red. Likely because of aa changes. 

Then I just downloaded the screenshots and uploaded them.

almarklein:
Ah, now I see. My small tweaks were fine, but merging with main likely affected aa.

PR #1153, opened 2025-07-27
Adjust how version is reported when one checkouts a tag

hmaarrfk:
```
git checkout v0.12.0
python -c "import pygfx; print(pygfx.__version__)"
```

Old:
* `0.12.0+g15e3681e`
* `0.12.0`

But if it is "dirty"

* `0.12.0+g15e3681e.dirty`

### DIFF
--- a/pygfx/_version.py
+++ b/pygfx/_version.py
@@ -51,7 +51,9 @@ def get_extended_version():
     version = release
     if post and post != "0":
         version += f".post{post}"
-    if labels:
+        if labels:
+            version += "+" + ".".join(labels)
+    elif labels and labels[-1] == "dirty":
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Part of the problem is that we try to "condition" on if the current version is greater than 0.12.0 (for compatibility with the major refactoring of blending)

```
In [1]: from packaging.version import Version

In [2]: Version("0.12.0+g15e3681e") > Version("0.12.0")
True
```

This is somewhat annoying, we decided to change the comparison to `0.12.0.post1`

hmaarrfk:
but it has been great already since we were able to be proactive and build out compatibility before releases come out officially!

PR #1160, opened 2025-08-08
updated "expected 1.0" to 1 year from today

hmaarrfk:
I'm just making a guess. 

But it always feel like projects are "active" when they don't have these kinds of references to the past.


almarklein (approved):

almarklein:
I fixed the CI error in #1161

PR #1163, opened 2025-08-11
glTF model Inspector

panxinmiao:
This PR adds an Inspector to the glTF viewer, based on #974, and requires installing the latest main branch of `wgpu-py` (with imgui_bundle >= v1.92.0)

With the completion of #974, we can say that our support and implementation for glTF have become quite comprehensive. This Inspector helps users better inspect and debug glTF models.

**Motivation:**
glTF is increasingly becoming a universal 3D model format and standard. However, when converting from other formats to glTF, some issues often arise—mainly related to transparency. Different engines and modeling tools have their own unique transparency logic and API designs, whereas glTF standardizes this aspect primarily through the material's alpha_mode property (which can be OPAQUE, MASK, or BLEND). Accurate mapping during format conversion can therefore be challenging.

For example, when exporting models originally in FBX format from Sketchfab to glTF, the transparency mode often fails to correctly map to alpha_mode=MASK and is instead uniformly set to alpha_mode=BLEND, causing rendering issues in some models. Additionally, glTF does not specify the depth_write state for transparent materials. Our loader sets depth_write to False by default for transparent materials (alpha_mode=BLEND), but this is not always appropriate in certain cases. This has also been widely discussed in three.js:  (https://github.com/mrdoob/three.js/pull/18235) 

With this Model Inspector, users can conveniently and visually adjust these properties to achieve correct rendering results.

I opened this draft PR early because we are currently working on designing transparency-related APIs, and I believe this tool could be helpful for the design and refinement of these APIs, and verify the rationality.

Example 1:

In this model (https://skfb.ly/pyzTD), when exported to glTF, the materials for the clothes and hair were incorrectly marked as alpha_mode = BLEND, meaning transparent=True and depth_write=False.

image

After adjusting them to alpha_mode=MASK, the rendering appears correct.

image

Example 2:

In this model (https://skfb.ly/pwoKA), the situation is more complex. 

image

Although the material for the clothes correctly uses the transparency mode `alpha_mode=BLEND`, we must set `depth_write=True` in order to achieve correct rendering results.

image

Example 3:

For hair, `alpha_mode=MASK` is generally recommended, but in this model (https://skfb.ly/oXEAy), using `alpha_mode=BLEND `(with `depth_write=False`) produces the correct rendering result.

image

We also support a dither mode (where transparent=False and depth_write=True), which often yields better rendering results for hair.
image


Korijn:
Is this branch also using the work from #1144 and #1124 ?

panxinmiao:
No, it doesn’t. This branch is based on #974,  using the current main branch’s  transparency API design along with the OPAQUE-related revisions from #1120.

PR #1165, opened 2025-08-16
ColorSpace Management

panxinmiao:
**⚠️  This PR introduces significant changes to Pygfx and maybe has a broad impact.**

We have already discussed the color space issue many times. I apologize for submitting this PR so directly, but I believe that merely discussing it in words makes it difficult to fully convey the ideas, and can also cause misunderstandings.
By expressing the proposal directly in code, I can present it more clearly and unambiguously, while also showing the results.

### Current issues:
I have long believed that our current approach to color space management is overly simplistic, incorrect, and even somewhat chaotic. It is also not extensible, making it hard to support possible future HDR color spaces (e.g. Display-P3).

We currently lack the concept of a working color space.
We force Color values to be in sRGB space (as I understand, mainly for visualization use cases, since sRGB corresponds to perceptual linearity and is visually friendly).
However, for lighting calculations and PBR rendering, we require everything to work in linear space.

As a result, we are forced to assume shaders operate in linear space, but at the same time, in order to preserve visual friendliness for visualization use cases, we apply texture sampling interpolation in sRGB space. To be blunt, this is quite messy.

Currently, we only have one pipeline workflow, which is problematic in different scenarios and not entirely correct. I have listed some issues in past discussions, but to summarize briefly again here:

**Current workflow (example: input texture encoded in sRGB):**
```
Texture (sRGB space)
    --> GPU interpolation sampling (in sRGB encoded space)
    --> Shader srgb2physical conversion (but srgb2physical is often misused)
    --> Shading calculations, blending, etc.
    --> GPU OETF conversion to sRGB space for output
 ```
 
**Problems:**

- For 3D rendering use cases that require linear-space operation, interpolation results are incorrect (because they occur in the wrong space).On the CPU side, since colors are forced into sRGB space, any computation involving colors yields wrong results — such as light accumulation, intensity computation, or user-defined color processing.

- For visualization use cases (which are meant to work in perceptual linear space, i.e. sRGB), although interpolation is done in the right space, the subsequent conversion to linear space inside shaders again leads to incorrect results whenever computations (blending, etc.) are involved.

- In addition, the srgb2physical function is somewhat abused; we even see cases where conversions occur mid-calculation, which means the operations before and after are no longer in the same space.

- Finally, there is a lot of hard-coded color space logic in the code, making it very difficult to handle textures that may exist in different spaces (e.g. environment maps).

### I believe we should follow these principles:

1. All computations involving colors — including interpolation, sampling, CPU-side operations, GPU (shader) computations, blending, etc. — should happen in the same color space, i.e. the working color space.

2. Color space management should be simple, clear, and ideally transparent to the user.

3. Avoid unnecessary redundant conversions. For example, many models store color data in linear space (glTF specifies that all Color values are stored in linear space, and as far as I know, most formats like FBX do the same). Currently, we are forced to convert these into sRGB upon loading, then back into linear space inside shaders. This is especially cumbersome for vertex colors (which I haven’t even properly handled yet).

_Note: WGPU’s `-srgb` texture formats should be seen only as GPU-side EOTF/OETF helpers,, not as an indication of the texture’s inherent color space._ 

---

### 1. Introduces color space management and the concept of a working color space.

The logic borrows some ideas from three.js, with simplifications and adaptations for Pygfx’s API design and use cases.

**Let's clarifying the concept of color spaces first:**
Color representation can be based on different models (RGB, CMYK, YUV, HSL/HSV, etc.). These are models — mathematical coordinate systems — not bounded color spaces, and they do not define physical gamut ranges.

The RGB model is the most common and hardware-oriented; displays operate in RGB.
When rendering engines or display devices refer to a “color space,” they always mean within the RGB model, but with specific definitions of primaries, white point, and transfer function (EOTF). For example:

sRGB = RGB model + Rec.709 primaries + D65 white point + sRGB gamma curve

Display-P3 = RGB model + P3 primaries + D65 white point + sRGB gamma curve

Rec.2020-PQ = RGB model + Rec.2020 primaries + D65 white point + PQ curve

The primaries and white point define a triangle in the CIE 1931 xy chromaticity diagram, which determines the gamut’s size and shape -- And this is the "Color Space".
Thus, although RGB values are always normalized to [0–1], their meaning (location in the chromaticity diagram) differs across color spaces.

In this PR (currently focused on SDR displays):

- ColorSpace includes three values: `no_color_space`, `srgb`, and `linear_srgb`.

- It is extensible to future HDR color spaces (e.g. Display-P3).

- **Working color space**: can be `linear_srgb` or `srgb`. Default: `linear_srgb`.

- **Renderer output color space**: can be `srgb` or `linear_srgb`. Default: `srgb`.

     - If working in linear_srgb, output can be either linear_srgb (for rendering/baking textures) or srgb (for display).

     - If working in srgb, output can only reasonably be srgb (otherwise likely user misconfiguration).

- **Texture color space**: can be `no_color_space`, `srgb`, or `linear_srgb`.

     - `no_color_space`: data is not color (e.g. normal maps, metallic/roughness maps), and is uploaded raw regardless of working color space.

     - `srgb` or `linear_srgb`: treated as color data, converted according to the working color space.

     - Special case: float textures. Float textures are almost never stored in sRGB (e.g. .hdr, .exr are linear). The engine assumes float textures are `linear_srgb` or `no_color_space`, and issues warnings if they are marked otherwise.

**Q: Why `linear_srgb` instead of `linear` or `physical`?**
Because “linear” or “physical” are ambiguous. `linear_srgb` explicitly refers to the linearized version of the sRGB space. Similarly, Display-P3 has its own linearized form (maybe linear-P3 ?). All these are linear spaces,  and physically correct for computation. And they can be transformed into each other through a simple 3x3 matrix transformation.

**Q: Why remove YUV from texture color spaces?**
Because YUV is a color model, not a fully defined color space. Rendering engines manage RGB-based spaces (primaries, white point, EOTF), not mathematical transforms. For YUV video, it is typically decoded to RGBA for rendering. If GPU-side conversion is needed, YUV data should be uploaded as `no_color_space` and handled by custom shaders or materials (e.g. YuvTexture, YuvMaterial).

### 2. Improves the Color utility class.

Currently, Color values are forced in sRGB space, As mentioned earlier, this has many issues and is not very convenient.
This PR introduces the concepts of **color space management** and **working_colorspace**, 
With this, the working color space determines how Color values are interpreted:
- A Color’s values are simply data in the current working color space.
- The Color class adds convenient conversion methods and initialization parameters.

Example usage:
```python
pygfx.ColorManagement.working_color_space = ColorSpace.linear_srgb

c = Color(0.5, 0.5, 0.5)   # By default,  it is in the current working_color_space

print(c.rgb)  #  =(0.5, 0.5, 0.5)

c = Color(
    0.5, 0.5, 0.5, colorspace=ColorSpace.linear_srgb
)  # Set as a color in linear_srgb color space, consistent with the current working_color_space
print(c.rgb)  # =(0.5, 0.5, 0.5)

c = Color(
    0.5, 0.5, 0.5, colorspace=ColorSpace.srgb
)  # Set as color in sRGB space, automatically convert to current workspace (linear_sgb)
print(c.rgb)  # # ~=(0.214, 0.214, 0.214)

pygfx.ColorManagement.working_color_space = ColorSpace.srgb

c = Color(0.5, 0.5, 0.5)  # By default,  it is in the current working_color_space
print(c.rgb)  # =(0.5, 0.5, 0.5)

c = Color(
    0.5, 0.5, 0.5, colorspace=ColorSpace.linear_srgb
)  # Set as a color in linear_srgb color space, consistent with the current working_color_space (sRGB)
print(c.rgb)  # # ~=(0.735, 0.735, 0.735)

c = Color(
    0.5, 0.5, 0.5, colorspace=ColorSpace.srgb
)  #  Set as a color in srgb color space, consistent with the current working_color_space
print(c.rgb)  # =(0.5, 0.5, 0.5)
```
**When using CSS or Hex strings to set colors, because CSS web colors are usually located in sRGB space, we assume by default that CSS or Hex strings are located in sRGB space, which means:**
```python
pygfx.ColorManagement.working_color_space = ColorSpace.linear_srgb

c = Color("#808080")  # Default in sRGB space, automatically converted to the current workspace (linear_sgb)
print(c.rgb)  # ~=(0.216, 0.216, 0.216)

c = Color(
    "#808080", colorspace=ColorSpace.linear_srgb
)  # Set as a color in linear_srgb color space, consistent with the current working_color_space
print(c.rgb)  # ~=（0.502, 0.502, 0.502）

c = Color(
    "#808080", colorspace=ColorSpace.srgb
)  # Set as a color in srgb color space, automatically convert to current workspace (linear_sgb)
print(c.rgb)  # ~=(0.216, 0.216, 0.216)

pygfx.ColorManagement.working_color_space = ColorSpace.srgb

c = Color("#808080")  # Default in sRGB space, consistent with the current workspace
print(c.rgb)  # =(0.502, 0.502, 0.502)

c = Color(
    "#808080", colorspace=ColorSpace.linear_srgb
)  # Set as a color in linear_srgb color space, automatically convert to current workspace sRGB)
print(c.rgb)  # ~=（0.737, 0.737, 0.737）

c = Color(
    "#808080", colorspace=ColorSpace.srgb
)  # Set as a color in srgb color space, consistent with the current working_color_space
print(c.rgb)  # ~=(0.502, 0.502, 0.502)
```
---

## III. Impact on Pygfx applications

1. 3D rendering apps (linear workflow):

Almost no changes required. The default working color space is `linear_srgb`.

The corrected workflow is now:
```
Texture (sRGB)
    --> GPU EOTF decode to linear
    --> GPU interpolation sampling (linear space)
    --> Shading, blending (linear space)
    --> GPU OETF to sRGB for output
```
This fixes the errors in the old pipeline. Results are now correct.
_Note: User-defined Color values are now in linear_srgb (direct rgb assignment), which may require small adjustments._

2..Visualization apps (expected in sRGB workflow):
There are subtle changes. By default (linear_srgb working color space), textures are sampled in linear space, which differs from the old behavior (in sRGB space). This is more correct, but is in linear workflow, not match expectations for visualization cases.

If `ColorManagement.working_color_space = ColorSpace.srgb` is set, the workflow becomes:
```
Texture/Color (sRGB)
    --> GPU interpolation sampling (sRGB space)
    --> Shading, blending (sRGB space)
    --> Direct output
```
This matches expectations perfectly (all in srgb), with no any color space conversions, and aligns with the web color model behavior.

---
### IV. Task list

- [ ] Clarify and restructure ColorSpace, introduce ColorSpaceManagement.
- [ ] Improve Color utility class behavior.
- [ ] Modify and add unit tests related to Color and ColorSpace.
- [ ] Review and update examples.
- [ ] Documentation, including a ColorSpace guide.
- [ ] (Optional) Support for YUV textures (Do we really need this?).

### DIFF
--- a/examples/feature_demo/dynamic_env_map.py
+++ b/examples/feature_demo/dynamic_env_map.py
@@ -36,7 +36,11 @@ cube_size = env_img.shape[1]
 env_img.shape = 6, cube_size, cube_size, env_img.shape[-1]
 
 env_static = gfx.Texture(
-    env_img, dim=2, size=(cube_size, cube_size, 6), generate_mipmaps=True
+    env_img,
+    dim=2,
+    size=(cube_size, cube_size, 6),
+    generate_mipmaps=True,
+    colorspace="tex-srgb",
 )
 
 # Create the dynamic env map
--- a/examples/feature_demo/earth.py
+++ b/examples/feature_demo/earth.py
@@ -46,11 +46,11 @@ scene = gfx.Scene()
 earth_geometry = gfx.sphere_geometry(63.71, 100, 50)
 
 
-def load_texture(path, flip=False):
+def load_texture(path, flip=False, colorspace="tex-srgb"):
     img = iio.imread(path)
     if flip:
         img = np.ascontiguousarray(np.flipud(img))
-    tex = gfx.Texture(img, dim=2)
+    tex = gfx.Texture(img, dim=2, colorspace=colorspace)
     return tex
 
 
@@ -75,7 +75,7 @@ earth_material.emissive_intensity = 3.0
 # earth_material.light_map_intensity = 3.0
 
 earth_material.normal_map = load_texture(
-    model_dir / "planets" / "earth_normal_2048.jpg", flip=True
+    model_dir / "planets" / "earth_normal_2048.jpg", flip=True, colorspace="physical"
 )
 earth_material.normal_scale = 0.85, -0.85
 
--- a/examples/feature_demo/env_maps.py
+++ b/examples/feature_demo/env_maps.py
@@ -47,7 +47,11 @@ env_img.shape = 6, cube_size, cube_size, env_img.shape[-1]
 
 # Create environment map
 env_tex = gfx.Texture(
-    env_img, dim=2, size=(cube_size, cube_size, 6), generate_mipmaps=True
+    env_img,
+    dim=2,
+    size=(cube_size, cube_size, 6),
+    generate_mipmaps=True,
+    colorspace="tex-srgb",
 )
 
 teapot = trimesh.load(TEAPOT)
--- a/examples/feature_demo/gltf_viewer.py
+++ b/examples/feature_demo/gltf_viewer.py
@@ -81,14 +81,18 @@ env_img.shape = 6, cube_size, cube_size, env_img.shape[-1]
 
 # Create environment map
 env_tex = gfx.Texture(
-    env_img, dim=2, size=(cube_size, cube_size, 6), generate_mipmaps=True
+    env_img,
+    dim=2,
+    size=(cube_size, cube_size, 6),
+    generate_mipmaps=True,
+    colorspace=gfx.ColorSpace.srgb,
 )
 
 background = gfx.Background(None, gfx.BackgroundSkyboxMaterial(map=env_tex))
 background.visible = False
 scene.add(background)
 
-scene.add(gfx.Background.from_color((0.1, 0.1, 0.1, 1)))
+scene.add(gfx.Background.from_color(gfx.Color(0.1, 0.1, 0.1, 1).to_physical()))
 
 
 def add_env_map(obj, env_map):
--- a/examples/feature_demo/lightmap.py
+++ b/examples/feature_demo/lightmap.py
@@ -42,7 +42,7 @@ renderer = gfx.renderers.WgpuRenderer(canvas)
 meshes = gfx.load_gltf_mesh(model_dir / "lightmap" / "scene.gltf", materials=False)
 
 light_map = iio.imread(model_dir / "lightmap" / "lightmap-ao-shadow.png")
-light_map_tex = gfx.Texture(light_map, dim=2)
+light_map_tex = gfx.Texture(light_map, dim=2, colorspace="tex-srgb")
 
 # Create camera and controller
 camera = gfx.PerspectiveCamera(45, 1)
--- a/pygfx/__init__.py
+++ b/pygfx/__init__.py
@@ -17,6 +17,14 @@ from .animation import *
 from .renderers import *
 
 from .utils.color import Color
+
+from .utils.show import show, Display
+from .utils.viewport import Viewport
+from .utils.text import font_manager
+from .utils import cm, enums, logger
+from .utils.enums import *
+from .utils.color_management import ColorManagement
+
 from .utils.load import load_mesh, load_meshes, load_scene
 from .utils.load_gltf import (
     load_gltf,
@@ -25,11 +33,6 @@ from .utils.load_gltf import (
     load_gltf_mesh_async,
     print_scene_graph,
 )
-from .utils.show import show, Display
-from .utils.viewport import Viewport
-from .utils.text import font_manager
-from .utils import cm, enums, logger
-from .utils.enums import *
 
 # Temp fix for pyinstaller to pick up pylinalg
 import pylinalg
--- a/pygfx/animation/keyframe_track.py
+++ b/pygfx/animation/keyframe_track.py
@@ -23,7 +23,7 @@ class KeyframeTrack:
             # remove adjacent keyframes scheduled at the same times
             if time != prev_time:
                 if np.any(value != prev_value):
-                    if prev_time is not None:
+                    if prev_time != optimized_times[-1] if optimized_times else None:
                         optimized_times.append(prev_time)
                         optimized_values.append(prev_value)
                     optimized_times.append(time)
--- a/pygfx/renderers/wgpu/engine/renderer.py
+++ b/pygfx/renderers/wgpu/engine/renderer.py
@@ -33,8 +33,8 @@ from ....objects._lights import (
 from ....cameras import Camera
 from ....resources import Texture
 from ....resources._base import resource_update_registry
-from ....utils import Color
-from ....utils.enums import PixelFilter
+from ....utils import Color, ColorManagement
+from ....utils.enums import PixelFilter, ColorSpace
 
 from ... import Renderer
 from .blender import Blender
@@ -239,6 +239,7 @@ class WgpuRenderer(RootEventHandler, Renderer):
         sort_objects=True,
         enable_events=True,
         gamma_correction=1.0,
+        output_color_space=ColorSpace.srgb,
         ppaa="default",
         **kwargs,
     ):
@@ -274,24 +275,13 @@ class WgpuRenderer(RootEventHandler, Renderer):
         # Cache renderstate objects for n draws
         self._renderstates_per_flush = []
 
+        self.output_color_space = output_color_space
+
         # Get target format
         self.gamma_correction = gamma_correction
         self._gamma_correction_srgb = 1.0
-        if isinstance(target, AnyBaseCanvas):
-            self._canvas_context = self._target.get_context("wgpu")
-            # Select output format. We currently don't have a way of knowing
-            # what formats are available, so if not srgb, we gamma-correct in shader.
-            target_format = self._canvas_context.get_preferred_format(
-                self._shared.adapter
-            )
-            if not target_format.endswith("srgb"):
-                self._gamma_correction_srgb = 1 / 2.2  # poor man's srgb
-            # Also configure the canvas
-            self._canvas_context.configure(
-                device=self._device,
-                format=target_format,
-            )
-        else:
+
+        if not isinstance(target, AnyBaseCanvas):
             # Also enable the texture for render and display usage
             self._target._wgpu_usage |= wgpu.TextureUsage.RENDER_ATTACHMENT
             self._target._wgpu_usage |= wgpu.TextureUsage.TEXTURE_BINDING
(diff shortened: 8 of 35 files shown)


panxinmiao:
Regarding the errors in the current main branch caused by the color space conversion workflow, For 3D rendering using a linear workspace, the test case we mentioned earlier is a typical example: https://github.com/pygfx/pygfx/issues/557#issuecomment-2538233684

For visualization use cases in the sRGB space, consider the following simple example:

On a pure black (0,0,0) background, we place a pure white (255,255,255) layer with 0.5 transparency. After blending, what color should we expect?

Obviously, the expected result is a mid-gray (128,128,128). This is consistent with how Web colors are rendered in browsers. Here is an HTML example on jsfiddle: https://jsfiddle.net/fa9u2z1h/

As shown in the picture, you can use the screen color extraction tool to view it:

image


In Pygfx , We can use the following script:
```python
from rendercanvas.auto import RenderCanvas, loop
import pygfx as gfx

# gfx.ColorManagement.working_color_space = gfx.ColorSpace.srgb

canvas = RenderCanvas()
scene = gfx.Scene()

scene.add(gfx.Background.from_color("#000"))

plane = gfx.Mesh(
    gfx.plane_geometry(10, 10),
    gfx.MeshBasicMaterial(
        color="#fff",
        opacity=0.5,
    ),
)

scene.add(plane)

camera = gfx.PerspectiveCamera(0)
camera.show_object(scene)

canvas.request_draw(lambda: renderer.render(scene, camera))

if __name__ == "__main__":
    loop.run()
```

On our current main branch, the result is:
image

_Note that on the current main branch, sRGB color values obtained through texture interpolation may appear correct in some cases. For example, interpolating between (0,0,0) and (1,1,1) at 0.5 happens to produce (0.5,0.5,0.5). This is an instance of a “double error” resulting in an accidentally correct value (an incorrect workflow combined with interpolation in the wrong color space), which I have explained in detail in https://github.com/pygfx/pygfx/pull/1146#issuecomment-3117171517 before:
> In Workflow 2:_
Input 0.0 and 1.0 are treated as sRGB, then converted to linear in the shader. and 0,0 still 0.0, 1.0 still 1.0, After computation, the result is 0.5, then passed through the GPU’s OETF — 0.5 is transformed to around 0.73. But this is not what we want — it looks too bright.
In fact, if we get result by interpolating directly from srgb texture(that's exactly what this workflow does), we got 0.5, then convert that to linear (e.g., 0.5 → ~0.218), and then pass it through the GPU OETF again (0.218 → ~0.5), the final value happens to be correct — but only by coincidence, because we are in a wrong workflow and doing the wrong operation, which happened to give a seemingly correct result.

_However, when the result relies on calculations or blending, the color values—such as the example mentioned above—are incorrect._

With this PR:
When in default linear-srgb color space:
image


After correctly setting the working color space to sRGB (uncomment the line ` gfx.ColorManagement.working_color_space = gfx.ColorSpace.srgb`), the result becomes:
image

This matches our expectations and is consistent with the behavior on the Web.

Vipitis:
Not sure if helpful. But there is a color space visualizer implemented with pygfx https://github.com/colour-science/colour-visuals which I believe is even used in ocio somewhere. 

also more of a question/idea (as I am not a pygfx user myself): would it make to split gamma and gamut? So you don't end up needed to specify `linear-srgb` or `linear-rec709`(which I guess would be the same) when you just want the gamma to be linear. In a situation where later it changes to p3 for example? The applications I am used to (Resolve) make this split.

panxinmiao:
> Not sure if helpful. But there is a color space visualizer implemented with pygfx https://github.com/colour-science/colour-visuals which I believe is even used in ocio somewhere.

I took a look at this project, and it’s built on an older version of pygfx, so I can’t run tests on this PR branch.
That said, if I understand correctly, despite its name suggesting “color visualization,” it’s essentially just a demonstration and not fundamentally different from other visualization applications.
Since pygfx outputs in the sRGB color space, it’s not possible to accurately represent colors outside the sRGB gamut.


> also more of a question/idea (as I am not a pygfx user myself): would it make to split gamma and gamut? So you don't end up needed to specify `linear-srgb` or `linear-rec709`(which I guess would be the same) when you just want the gamma to be linear. In a situation where later it changes to p3 for example? The applications I am used to (Resolve) make this split.

Sorry, I don’t fully understand what you mean by "separating gamut and gamma", so please forgive me if my answer goes a bit off track. 😅 

From the rendering engine’s perspective, when we talk about “color space,” we consider both the gamut and the transfer function (EOTF, or gamma). From this standpoint, the answer would be “yes.”

Currently, we only support a single gamut (sRGB). The gamut of sRGB and linear-sRGB is identical (since they share the same primaries and white point).

Display-P3 and sRGB use the exact same EOTF (i.e., their gamma curve is identical), but P3 defines different primaries and therefore a wider gamut.

To switch from sRGB to P3, you first need a monitor that supports Display-P3. For our rendering engine, we also need to declare that the swapchain’s color space is Display-P3, which informs the display to interpret the output in Display-P3 mode.

In my opinion there are two possible approaches to supporting P3:

- Use P3 as the working space (linear-P3), then apply gamma correction for Display-P3 output. This is the most correct and optimal solution, but it only works best if the entire pipeline is P3-aware. At present, however, most input textures and assets are still authored in sRGB, we must convert them to P3 first. however, since their original colors occupy only a portion of the P3 gamut, they cannot fully take advantage of P3’s wider color range.

- Keep the working space in linear-sRGB, but convert the output to P3.
This is just a 3×3 matrix transform from linear-sRGB to linear-P3.  For PBR rendering,  even if we only do this, it can improve image quality: although the working space remains linear-sRGB, lighting and blending calculations can produce values outside the sRGB gamut (but still inside P3). If you don’t output to P3, those values would either need to be tone-mapped or simply clamped, both of which can cause color distortion.

I believe approach 2 already provides noticeable benefits, while approach 1 remains the long-term ideal if the full pipeline and assets are P3-aware.

Vipitis:
Yes I linked the library more as a coincidence. I believe their transfer matrices are mostly done in numpy.

I read through the upcoming changes with new wgpu-core releases and newer the webgpu headers. There is mention for display-p3 as a color space option instead of srgb. But only for the html canvas, so web. I don't think this will make its way to wgpu-native any time soon. But it also looks like you can specify a color space for the texture copy functions for output. I don't believe you can do hardware interpolated sampling in other color spaces. 

However if going for an intermediate working space, wouldn't it make sense to use something bigger like ACEScg? I suppose it depends on the kind of assets users want to support. Like I believe YUV as a model is supported because compressed video is a use case. 

Would color management entail you support various inputs, one working intermediate and then again various outputs?

panxinmiao:
> I don't believe you can do hardware interpolated sampling in other color spaces.

In the current WebGPU specification, the only defined color spaces are sRGB and Display-P3 (both using the same gamma curve). Therefore, for SDR content, using -srgb textures works correctly even for P3—the EOTF/OETF process applies properly.

**Texture sampling and interpolation:**

- When using -srgb textures, the GPU automatically decodes gamma to linear space and performs interpolation there. This is correct for both sRGB and Display-P3.

- Sampling in linear-sRGB or linear-P3 makes no difference as long as it is linear. Post-interpolation conversion via a linear 3×3 matrix is safe and physically correct.

**Example workflow (Working space: linear-P3):**

```
Display-P3 texture (-srgb) 
           → auto-converted to linear-P3 
           → shader computations 
           → output via -srgb texture (OETF applied)

sRGB texture (-srgb) 
           → auto-converted to linear-sRGB 
           → shader converts to linear-P3 
           → computations 
           → output via -srgb texture
```
If a color space has a different EOTF (e.g., PQ/HLG), `-srgb` textures cannot be used for automatic linearization and interpolation, and manual handling is required.

But, currently, WebGPU only defines sRGB and Display-P3. If future support for, e.g., Rec.2020-PQ (using the PQ curve) is added, I believe it would likely define a dedicated texture format (e.g., `-pq`) so that implementations like wgpu can automatically handle linear conversion and interpolation.


> However if going for an intermediate working space, wouldn't it make sense to use something bigger like ACEScg? I suppose it depends on the kind of assets users want to support. Like I believe YUV as a model is supported because compressed video is a use case.
> 
> Would color management entail you support various inputs, one working intermediate and then again various outputs?

Regarding your question about an “intermediate working space”: most current assets and their creation workflows are still authored in sRGB. Using a wider gamut working space provides limited benefit and adds unnecessary conversions (all sRGB inputs must be converted to the working space). 

Therefore, as I mentioned before, approach 2 might be preferable: keep the working space in linear-sRGB but allow P3 output, (It seems is also the approach adopted by most modern engines). That said, supporting other linear working spaces is not very difficult either, and we can provide it if users prefer.

almarklein:
Just a quick note that I've seen this PR. I'm trying to catch up with a lot of other stuff; I'll catch up with the colorspace stuff when some of that is wrapped up.

Just two things that come to mind as related (maybe you've already mentioned some of it):
* #1125 
* https://oklch.com/

PR #1168, opened 2025-08-22
EXP: try to implement stochastic transparency from Wyman paper

almarklein:
The paper: https://cwyman.org/papers/tvcg17_hashedAlphaExtended.pdf

During work on #1139 I made an attempt to make the noise patter move along with the object, which usually feels nicer than the noise being stuck to the screen. However, I ran into several issues, so I let go of that idea.

I recently ran into the above paper which goes to great lengths to do exactly that, solving the issues that I had and more.

This sounds great, but its mean caveat is that it uses white noise. See this (left is the paper, right is our current version, with blue noise anchored to the screen):

image
 
What I tried in this PR is to adjust the implementation in the paper to use blue noise. This works, as in it makes the noise more fine-grained, but the result is still more coarse than our current implementation. This is because of how the algorithm works: to avoid jumpy noise, it calculates two noise patterns, in log scale, and interpolates between them. That way the noise pattern gently goes from one pattern to another as objects move closer/further. However, interpolating two blue noise patterns results in a pattern that is less blue.

image

So I tried to disable the interpolation, but then there is still the problem that the noise pattern does not exactly match the screen resolution, so its still not as fine as what we currently have. Plus it introduces the problem that the noise is now very jumpy as objects move around.

In conclusion, this was a nice exercise, but I think we should stick to our current implementation.

There are some use-cases where noise attached to the object feels much better, and where I think we can also make it work properly, namely billboarding cases, such as points and markers. I may look into that that some other time.

### DIFF
--- a/examples/feature_demo/line_basic.py
+++ b/examples/feature_demo/line_basic.py
@@ -46,7 +46,7 @@ for i in range(len(positions)):
 
 line = gfx.Line(
     gfx.Geometry(positions=positions),
-    gfx.LineMaterial(thickness=22.0, color=(0.8, 0.7, 0.0), aa=False),
+    gfx.LineMaterial(thickness=22.0, color=(0.8, 0.7, 0.0), aa=False, opacity=0.5, alpha_mode='dither'),
 )
 scene.add(line)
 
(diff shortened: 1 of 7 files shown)


almarklein:
(Only made this PR to document this idea for future reference)

PR #1171, opened 2025-08-23
Turn off material.aa by default

almarklein:
Ref #1065

As planned, turning off shader-based aa. 

We can afford it, because we have pretty good post-processing anti-aliasing now.

It avoids a range of artifacts by avoiding objects that produce a mix of opaque and transparent fragments.

### DIFF
--- a/examples/feature_demo/image_stitching.py
+++ b/examples/feature_demo/image_stitching.py
@@ -52,7 +52,12 @@ for image_name in ["wood.jpg", "bricks.jpg"]:
     x += rgba.shape[1] - 200
 
 # Text is rendered as an overlay (using render_queue)
-text = gfx.Text("Image stitching", font_size=64, anchor="top-left")
+text = gfx.Text(
+    "Image stitching",
+    font_size=64,
+    anchor="top-left",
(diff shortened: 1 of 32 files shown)


Korijn (approved):

almarklein:
I did some testing on various monitors today, trying different combinations of `pixel_ratio`, `ppaa`, and `material.aa`.

My proposal is:
* Keep the behavior where we use SSAA on non-hiDPI displays
  * Well, actually fix/improve, see #1172 
  * It gives a good baseline for preventing aliasing for any object.
  * It seems 'fair' / consistent; the internal texture size is more or less the same, regardless of the monitor being used.
  * People can set `renderer.pixel_scale=1` if they want to squeeze performance.
* Enable `ppaa` by default.
  * It's cheap and helps a lot to get a smooth result.
* Disable `material.aa` by default.
  * It does not improve the result much if the above is used.
  * Except perhaps text, where slight variations in the perceived width of the stems of am H can stand out.
  * I'm still on the fence between enabling aa in text by default, or simply enabling it in all examples and docs.
  * Though tools that work in 2D like fastplotlib should probably enable aa on objects, because it's cheap.

PR #1172, opened 2025-08-25
Rename  alpha-method 'composite' to 'blended'

almarklein:
I propose to rename the alpha-method 'composite' to 'blended'. I feel that the word 'composite' is not an easy word; I find myself having to recall it every time that I write it, and that feels weird for the method that is very standard. The name 'blended' reflects clearly what is meant in a simpler way; its the word that we tend to use in discussions to describe this method, and I think it avoids confusion to just call it that.

FWIW: I originally chose 'composite' to free up the word 'blend' for `SomeMaterial(alpha_mode='blend')`.


Korijn (approved):

almarklein:
cc @panxinmiao I hope this helps

PR #1177, opened 2025-08-27
Support GLTF alphaMode

almarklein:
Replaces #1120

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -80,14 +80,16 @@ most appropriate). If you would prefer not to do that, please feel free to
 report to the Code of Conduct committee directly, or ask the committee for
 advice, in confidence.
 
-You can report issues to the Pygfx core team:
+You can report issues to the Pygfx Code of Conduct committee, at [conduct@pygfx.org](conduct@pygfx.org).
+Currently, the committee consists of the core team:
 
 [Korijn van Golen](https://github.com/Korijn)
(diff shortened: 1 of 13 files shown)


panxinmiao:
Not sure why #1120 was closed, but I think the opaque shader changes in it are necessary.

almarklein:
> Not sure why https://github.com/pygfx/pygfx/pull/1120 was closed, but I think the opaque shader changes in it are necessary.

Setting `alpha_mode='solid'` already does that: the blending is setup to ignore the source alpha:
https://github.com/pygfx/pygfx/blob/3d67dc37a6231ed25cd86e8f02ea1b95e48223b5/pygfx/renderers/wgpu/engine/blender.py#L232-L242

Korijn:
Is there a GLTF test case for this?

panxinmiao:
> Setting `alpha_mode='solid'` already does that: the blending is setup to ignore the source alpha:

No, here `bf.one` is the factor, and the final value depends on `src_alpha`.

`result_alpha = 1.0 * src_alpha + 0 * dst_alpha = src_alpha`

panxinmiao:
Current Result:

image


Correct result:

image

almarklein:
You're right. My bad! I updated the blender to fix the opaque alpha-method.

almarklein:
I found the example you posted, and that one still seems to fail. Looking into it.

almarklein:
Fixed.

Korijn (approved):

PR #1182, opened 2025-08-28
Access `alpha_config` will cause the bindings to be rebuilt?

panxinmiao:
While working on #974, I encountered a very strange bug.

In the `MeshShader`, simply accessing (without making any modifications) `material.alpha_config` causes the bindings to be rebuilt, which leads to unit test failures.
I suspect this is related to the tracker mechanism, but it’s definitely not the behavior we would expect.

### DIFF
--- a/pygfx/renderers/wgpu/shaders/meshshader.py
+++ b/pygfx/renderers/wgpu/shaders/meshshader.py
@@ -66,6 +66,9 @@ class MeshShader(BaseShader):
 
         color_mode = str(material.color_mode).split(".")[-1]
 
+        # access alpha config will cause the bindings to be rebuilt?
+        material.alpha_config
+
         self["color_mode"] = color_mode
         self["use_uniform_color"] = color_mode in ("auto", "uniform")
 


almarklein:
Accessing the `alpha_config` in the shader `__init__`  will trigger the shader from being re-compiled whenever the `alpha_config` is updated. Though I'm pretty sure this is also the case without referencing it in the shader.

When is the shader rebuild that is unexpected? I tried to reproduce and I can at least confirm its not rebuild on every draw.

panxinmiao:
> When is the shader rebuild that is unexpected? I tried to reproduce and I can at least confirm its not rebuild on every draw.

Here demonstrate that adding only one line of code to access `alpha_comfig` resulted in `test_wobject_updates.py` unit test failure.

```
FAILED tests/usecases/test_wobject_updates.py::test_updating_mesh_blending - assert {0: {0: <pygf..., ...}, 1: {}} == {0: {0: <pygf..., ...}, 1: {}}
  
  Omitting 1 identical items, use -vv to show
  Differing items:
  {0: {0: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f20182be740>, 1: <pygfx.renderers.wgpu.engine.bindin...ing.Binding object at 0x7f202ff74c90>, 3: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f202ff77460>, ...}} != {0: {0: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f4d0>, 1: <pygfx.renderers.wgpu.engine.bindin...ing.Binding object at 0x7f201840f230>, 3: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f620>, ...}}
  
  Full diff:
    {
        0: {
  -         0: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f4d0>,
  ?                                                                            ----
  +         0: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f20182be740>,
  ?                                                                           ++++
  -         1: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f070>,
  ?                                                                           ^^^^
  +         1: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f20182bc670>,
  ?                                                                           ^^^^
  -         2: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f230>,
  ?                                                                         ^^ ^^^^
  +         2: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f202ff74c90>,
  ?                                                                         ^^^^ ^^
  -         3: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f620>,
  ?                                                                         ^^ -- -
  +         3: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f202ff77460>,
  ?                                                                         ^^^^^
  -         4: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840f7e0>,
  ?                                                                         ^^^^  ^
  +         4: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f202ff74f30>,
  ?                                                                         ^^  ^^^
  -         5: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f201840d400>,
  ?                                                                          --- --
  +         5: <pygfx.renderers.wgpu.engine.binding.Binding object at 0x7f202ff751d0>,
  ?                                                                         +++++
        },
        1: {},
    }
···

**Comment by almarklein** [comment]
Ah, now I see what you mean.

So here's how it works:

For the alpha-methods "opaque", "stochastic", "weighted", changing the `alpha_config` requires a shader rebuild, because the parameters in the `alpha_config` dict affect the wgsl to produce the output fragment.

But for alpha-method "blended", the parameters are the blend equation params, which is a pipeline-level thing. In pipeline.py We take care to allow changing the `alpha_config` without requiring a shader rebuild if the alpha-method stays "blended". The failing test covers that.

So, in the shader you probably want to do something like:
```py
alpha_method = material.alpha_method
# and maybe
if alpha_method == 'x':
     alpha_config = material.alpha_config  # avoid getting this attribute unless needed
     param_of_interest = alpha_config['param_name'] 
```

*edit: can now use `material.alpha_method`*

almarklein:
Solved.

PR #1183, opened 2025-08-28
Dereference world-object in case its a weakproxy

almarklein:
This is needed to make PyGfx work with FPL again.

Previously the de-referencing happened in the `wobject.iter()`, which does `yield self`, leading to the same effect. But the renderer no longer uses `wobject.iter()`, because it uses its own generator to traverse the tree in an efficient way.

It's not pretty, but this is the best I came up with.


Korijn:
This PR really highlights how awkward weak refs are to work with... 

I wish we could do without them!

I'll approve in the name of backwards compatibility :)

Korijn (approved):

PR #1190, opened 2025-09-03
Remove that material.opacity affects alpha logic

almarklein:
I think that in hindsight, the fact the setting opacity <1 makes the behavior effectively "blend" does not add much. And without it the rules are easier to explain.

Also more docs improvements.

### DIFF
--- a/docs/transparency.rst
+++ b/docs/transparency.rst
@@ -62,16 +62,20 @@ Alpha what?
 
 The material has multiple properties related to transparency. Let's list them for clarity:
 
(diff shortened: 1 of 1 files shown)


Korijn (approved):
I am happy to see this!

almarklein:
> I am happy to see this!

I figured as much 😉   You and Pan were right.

almarklein:
I still think there's value in the 'auto' mode itself, b.c. it handles opaque objects well, and transparent objects reasonably well (depending on how well they can be sorted).

Korijn (approved):

PR #1199, opened 2025-09-22
Use new wgpu structs

almarklein:
Use the class-based structs from https://github.com/pygfx/wgpu-py/pull/748

I find that it makes the code easier to read; the classname makes it very explicit what you're dealing with, and easy to look up more info.

I also did a bit of benchmarking (took an example, and made it `renderer.render()` as fast as possible for 5 secs), but could not find a significant difference.

This PR cannot be merged until there is a new wgpu-py release that includes https://github.com/pygfx/wgpu-py/pull/748


almarklein:
*Blocked*

Korijn (approved):

PR #1200, opened 2025-09-22
Add Eye Dome Lighting (EDL) EffectPass

mortacious:
### Summary
This PR adds an Eye Dome Lighting (EDL) `EffectPass` to pygfx’s post-processing pipeline. EDL is a screen-space technique that enhances depth perception by darkening regions around depth discontinuities, particularly useful for point clouds and scientific visualization in general, especially if no additional attributes (e.g., colors or normals) are available ([link](https://www.kitware.com/eye-dome-lighting-a-non-photorealistic-shading-technique)).

edl

### Key Changes
- Added a new post-processing pass: `EDLPass`
- Added two examples in `examples/other/edl_pointcloud.py` and `examples/other/edl_mesh.py` demonstrating EDL on point clouds and meshes


Korijn:
This is awesome!

Would you mind summarizing how this technique is different from SSAO?

pygfx/renderers/wgpu/engine/edl.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+
+    def __init__(
+        self, *, strength=1.0, radius=1.5, num_samples=8, depth_edge_threshold=0.0
+    ):
+        super().__init__()
+        # Clamp/validate
+        num_samples = int(num_samples)
+        if num_samples not in (4, 8):
+            num_samples = 8
```
Korijn, line 120: where does num_samples go? it doesn't appear to be used at all

mortacious:
> This is awesome!
> 
> Would you mind summarizing how this technique is different from SSAO?

Although EDL and SSAO follow a similar approach - both are screen-space filters processing the depth buffer - their objectives are quite different.

EDL aims to improve the viewer's understanding of complex geometry. It achieves this by darkening pixels that have a large depth difference compared to their immediate neighbors. The result is a non-photorealistic outline effect that clearly defines shapes and edges.

In contrast, SSAO aims to simulate realistic lighting. It determines how much a point (pixel) is occluded by surrounding points. Points that are heavily occluded by nearby geometry closer to the camera are darkened while points standing out of their surroundings are brightened.

In practice EDL works well with a small radius and few samples (often just direct neighbors), while achieving good SSAO results requires a significantly larger number of samples (e.g., 64) across a wider area.

I have created a small example (in CloudCompare) below. The left image has EDL enabled, the right shows the same geometry with SSAO:

edl_ssao

pygfx/renderers/wgpu/engine/edl.py
mortacious: This indeed a leftover from my testing. I've just removed it!

Korijn:
On CI, open3d is not available, so that's causing the docs build to fail. You can exclude the examples from testing and the gallery (hidden or code) by following our [guide](https://github.com/pygfx/pygfx/blob/23d2bac93bba7688e4ecff577e4d460676f9bbd9/examples/HOWTO.md).

Also just want to let you know (in case you are eager) don't try to add open3d to the CI environment, open3d is notorious for lagging way behind newer python versions. We don't want to deal with that maintenance burden, so it's fine to keep the examples as they are and just skip them in our docs build.

PR #1207, opened 2025-10-03
Update ssaa.wgsl and ddaa.wgsl

### LINKED ISSUE #1202: Texture int offset out of range for ddaa shader
Hello,

I have a few github actions setup which works with Windows, macOS and Ubuntu. Everything was fine till I was running pygfx version 0.12.0

When the version was bumped to 0.13.0 it started having issues. The error is mentioned below.

```
ERROR    wgpu:_ffi.py:216 Shader translation error for stage ShaderStages(FRAGMENT): FXC D3DCompile error (Unspecified error (0x80004005)): (477,32-109): error X4566: offset texture instructions must take offset which can resolve to integer literal in the range -8 to 7
(512,32-110): error X4566: offset texture instructions must take offset which can resolve to integer literal in the range -8 to 7
```

PS: The project works fine on a regular windows setup.

Thanks in advance!

almarklein:
Closes #1202

* The ssaa will use a truncated kernel when using a cubic filter with a scale factor larger than 4.
* The ddaa algorithm has a few improvements:
  * It shifts the uv coord for sampling with `textureSampleLevel`, and limits the samples per step, so the int offset stays in range.
  * It looks further along edges (i.e. less jaggies).  
     * One sample is free because the info was there to calculate it. 
     * When the end of an edge is not found, the length is max_length + 1. So another free sample.
     * Uses 5x3 samples instead of 4 + 8.
     * So the max edge length is 16 rather than 12.
  * It is better optimized (faster on most hardware). 
  * Has a bugfix.


almarklein:
rdy

PR #1208, opened 2025-10-06
Per-vertex markers

almarklein:
* Add support for per-vertex markers by setting `material.marker_mode = "vertex"` and providing a `geometry.markers` buffer.
* Add `MarkerMode` and `MarkerInt` enum. The latter allows mapping a marker name to the corresponding int to put in the buffer.
* Updated two examples that used many different markers.
* Updated validation screenshot because the old version had a few `x if marker == .. else y`, which I removed because they don't work when the marker is defined per vertex.


Korijn:
The pygfx sdf looks messed up in the new screenshot?

almarklein:
> The pygfx sdf looks messed up in the new screenshot?

Yeah, there was a line `edge_width=0.1 if not marker == "custom" else 0.033333` in the original code, but that logic did not work with the new approach. Since the logo looks fine for the 'inner' and 'outer' modes I figured it fine.

kushalkolar:
Thanks!

PR #1216, opened 2025-10-11
remove assert in sprite, some cleanup

kushalkolar:
just removes an assert and sdf points have existed for a long while now :smile:


almarklein:
I think the note about `PointsSdfMaterial` refers to bitmap-based SDFs, like the glyphs in text rendering :)  But I'm fine with removing that comment. If anything, the issue tracker is meant for tracking feature requests like that.

kushalkolar:
there are screenshot diffs but only in python3.10 and not 3.13?

same lavapipe version on both, and there's nothing in this PR that should change screenshots.

almarklein:
That build seems to be glitching lately 🤔

PR #1217, opened 2025-10-12
add regular asterisk symobl

kushalkolar:
`*` is available on a US keyboard, whereas `✳` is not, and `*` is also what matplotlib uses.

by the way, it might be useful for the "alt_names" dict to be public, in fastplotlib I just copy pasted it. 

minor, but "asterix" is the french cartoon and "asterisk" is the symbol :sweat_smile:

### DIFF
--- a/pygfx/materials/_points.py
+++ b/pygfx/materials/_points.py
@@ -341,7 +341,7 @@ class PointsMarkerMaterial(PointsMaterial):
         Supported values:
 
         * A string from :obj:`pygfx.utils.enums.MarkerShape`.
-        * Matplotlib compatible characters: "osD+x^v<>".
+        * Matplotlib compatible characters: "osD+x^v<>*".
         * Unicode symbols: "●○■♦♥♠♣✳▲▼◀▶".
         * Emojis: "❤️♠️♣️♦️💎💍✳️📍".
         * A string containing the value "custom". In this case, the WGSL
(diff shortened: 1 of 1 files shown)


almarklein:
From what I can find on e.g. [wikipedia](https://en.wikipedia.org/wiki/Asterisk), the asterisk symbol can have different amount of legs. I suspect that I did not bind "*", because that symbols has 6 legs, and we don't have the 6-legged version yet. Maybe we can have a stab at adding it.

> "asterix" is the french cartoon and "asterisk" is the symbol 😅

🤦 we should fix that, and add "asterix"  in the `alt_names` dict ...

hmaarrfk:
Not sure what your computer shows, but for me, on a US keyboard, "shift + 8" shows 5 legs

image


But if i copy and paste the symbol that has 6 legs on wikipedia into the url bar, it shows 5 legs...

kushalkolar:
> on a US keyboard, "shift + 8" shows 5 legs

Same for me, but the character Almar posted has 6 legs 😂. EU keyboard?

hmaarrfk:
but in his latest message, on github, he typed a perfect 5 point character

image

kushalkolar:
On my phone Almar's is 6

hmaarrfk:
seems to be a font difference....

almarklein:
On the keyboard of my Macbook it has 6 legs, and for me it's also 6 in my comment 😂   So I guess this indeed also differs per font.

image

almarklein:
I added a six-legged marker, but could not commit to your branch, so I created a new PR: #1219

PR #1219, opened 2025-10-13
Add six-legged asterisk marker

almarklein:
Supersedes #1217

* [x] Rename ["asterix"](https://en.wikipedia.org/wiki/Asterix) to "asterisk".
* [x] Add six-legged marker as alt to the 8-legged marker.
    * The asterisk seems most common as a six-legged symbol, but 5 is also common (for some fonts).
* [x] Add "*" as a way to select the six-legged variant, just like in MPL.

### DIFF
--- a/pygfx/materials/_points.py
+++ b/pygfx/materials/_points.py
@@ -408,7 +408,7 @@ class PointsMarkerMaterial(PointsMaterial):
             resolved_name = alt_names.get(name, name).lower()
             if resolved_name not in MarkerShape:
                 raise ValueError(
-                    f"PointsMarkerMaterial.marker must be a string in {SizeMode}, or a supported characted, not {name!r}"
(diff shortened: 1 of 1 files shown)


kushalkolar:
Any chance we can have a bugfix release with this? Release and main are now incompatible due to https://github.com/pygfx/pygfx/pull/1219/files#diff-11be633f265d93744b9c27c6184bb9aa0feec67df47a8b4316dccbe3b7abe4baR141-R142

almarklein:
Yeah, we can do a small release. You wanna wait for #1220, I imagine?

kushalkolar:
> Yeah, we can do a small release. You wanna wait for #1220, I imagine?

yea I will finish #1220 tonight, did that at 4am so had to call it a day

might have a stab at #1221 too

PR #1220, opened 2025-10-21
account for transform of instanced objects for calculating bbox

kushalkolar:
Not sure if this is the best way to do it, works for my use case.

Minimal example:

```python
import numpy as np
import pygfx as gfx
import pylinalg as la

scene = gfx.Scene()

geometry = gfx.cylinder_geometry(10, 0, 100, 12)
material = gfx.MeshBasicMaterial(color="#336699")
cones = gfx.Group()

instance_count = 100
mesh = gfx.InstancedMesh(geometry, material, instance_count)
mesh.cast_shadow = mesh.receive_shadow = True

for i in range(instance_count):
    mesh.set_matrix_at(
        i,
        la.mat_compose(
            np.random.rand(3) * 4000 - 2000, (0, 0, 0, 1), np.random.rand() * 4 + 2
        ),
    )

scene.add(mesh)

print(mesh.get_world_bounding_box())

if __name__ == "__main__":
    disp = gfx.Display()
    disp.show(scene)
```

Without this PR the bbox is:

```
[[-10. -10. -50.]
 [ 10.  10.  50.]]
```

With this PR the bbox is around:

```
[[-1940.6628418  -1995.28747559 -1975.10424805]
 [ 1945.44067383  1944.54223633  1980.64733887]]
```

### DIFF
--- a/pygfx/objects/_instanced.py
+++ b/pygfx/objects/_instanced.py
@@ -81,7 +81,7 @@ class InstancedObject(WorldObject):
         transforms = self.instance_buffer.data["matrix"].transpose(0, 2, 1)
         aabbs = la.aabb_transform(aabb[None], transforms)
 
-        final_aabb = np.zeros((2, 3), dtype=float)
+        final_aabb = np.empty((2, 3), dtype=float)
         final_aabb[0] = np.min(aabbs[:, 0, :], axis=0)
         final_aabb[1] = np.max(aabbs[:, 1, :], axis=0)
 
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -553,7 +553,7 @@ class WorldObject(EventTarget, Trackable):
         # Combine
         if _aabbs:
             aabbs = np.stack(_aabbs)
-            final_aabb = np.zeros((2, 3), dtype=float)
+            final_aabb = np.empty((2, 3), dtype=float)
             final_aabb[0] = np.min(aabbs[:, 0, :], axis=0)
             final_aabb[1] = np.max(aabbs[:, 1, :], axis=0)
         else:


pygfx/objects/_instanced.py
```diff
@@ -73,6 +73,19 @@ def get_matrix_at(self, index: int):
         """get the matrix for the instance at the given index."""
         return self._store["instance_buffer"].data["matrix"][index].T
 
+    def get_bounding_box(self):
+        bbox = super().get_bounding_box()
+
+        # bbox of instanced objects
+        bbox_instance_min = self.instance_buffer.data["matrix"][:, 3, 0:3].min(axis=0),
+        bbox_instance_max = self.instance_buffer.data["matrix"][:, 3, 0:3].max(axis=0),
```
Korijn: You're just taking the translation components?

kushalkolar:
Updated to apply the transform to get the bbox. BTW vectorizing across all instances is not faster, the for loop is actually slightly faster with 500k instances.

This is not fast as I had hoped:
```python
aabb_transform_vec = np.vectorize(lambda matrix: la.aabb_transform(aabb, matrix), otypes=[np.float32], signature="(m, n) -> (2, 3)")
```

kushalkolar:
now very speedy with https://github.com/pygfx/pylinalg/pull/107 

will need to wait for the next release of `pylinalg` for tests to pass

Korijn:
pylinalg 0.6.8 has been released: https://github.com/pygfx/pylinalg/releases/tag/v0.6.8

kushalkolar:
should be gtg!

kushalkolar:
Thanks! @almarklein can we get a release, I will give #1221 a try later.

PR #1223, opened 2025-10-29
glTF: support accessor sparse

panxinmiao:
Sparse storage of accessor values that deviate from their initialization value.

See: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-accessor-sparse

Test Case:  [Simple Sparse Accessor](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/SimpleSparseAccessor)

image


panxinmiao:
CI failed, but it looks unrelated to this PR. 🤔

PR #1227, opened 2025-10-31
Physically Based Bloom

panxinmiao:
This PR add a Physically-based bloom effect pass, based on the technique from Call of Duty, presented at ACM Siggraph 2014.

This implementation uses progressive downsampling and upsampling without thresholding, making it suitable for HDR rendering pipelines. 

Reference: https://learnopengl.com/Guest-Articles/2022/Phys.-Based-Bloom

Examples:

https://github.com/user-attachments/assets/419bfe3e-294e-4faa-b124-12ab8f0770df

https://github.com/user-attachments/assets/f73c7f1d-7bb3-4642-b365-56a38c84fa96

Note: 
The current version of pygfx has some limitations. Since the rendering pipeline does not yet support HDR, colors with values greater than 1.0 (HDR colors) are clamped, causing the bright areas to lose color details and appear pure white. 
In the example video above, you can observe that increasing the “emissive strength“ causes the glowing parts of the object to turn white.

In one of my local branch, which uses an HDR rendering pipeline and tone mapping, the Bloom effects can produces a better results:

https://github.com/user-attachments/assets/57d155e1-4161-4d0d-bada-8b0c9ffae6c7

### DIFF
--- a/pygfx/objects/_base.py
+++ b/pygfx/objects/_base.py
@@ -278,13 +278,14 @@ class WorldObject(EventTarget, Trackable):
         This affects both opaque and transparent objects and can be used to resolve z-fighting,
         or control draw order beyond automatic depth sorting.
 
-        The effective render order is the sum of its render order and thet of all its parents.
+        If the object is part of a Group, the group's render_order is considered first (that is the group_order of the object).
 
         The order in wich objects are rendered is:
             1. the ``material.render_queue``.
-            2. the effective ``render_order``.
-            3. the distance to camera (if ``render.sort_objects==True``).
-            4. the position of the object in the scene graph.
+            2. the ``ob.parent.render_order`` (if ``isinstance(ob.parent, gfx.Group)``).
+            3. the ``ob.render_order``.
+            4. the distance to camera (if ``renderer.sort_objects==True``).
+            5. the position of the object in the scene graph.
 
         Also see ``material.render_queue``.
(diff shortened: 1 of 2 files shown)


pygfx/renderers/wgpu/engine/effectpasses.py
```diff
@@ -129,6 +129,23 @@ class FullQuadPass:
 
     wgsl = ""
 
+    # todo: configurable for each render target
+    load_op = wgpu.LoadOp.clear
+
+    # todo: configurable for each render target
+    blend_op = {
```
almarklein: More correct terminology:
```suggestion
    blend_state = {
```

panxinmiao:
It's strange — the CI indicates that some example screenshots have changed, but this PR should not affect the rendering results of those examples.

almarklein:
> It's strange — the CI indicates that some example screenshots have changed, but this PR should not affect the rendering results of those examples.

In other pr's as well. Maybe because it uses the new wgpu-py. Or another upstream lib had a new version. I'll have a look

almarklein (approved):

almarklein:
> It's strange — the CI indicates that some example screenshots have changed, but this PR should not affect the rendering results of those examples.

Looks like a glitch ... works fine again.

Korijn (approved):

PR #1230, opened 2025-11-07
Support KHR_materials_pbrSpecularGlossiness

panxinmiao:
This PR adds support for the [KHR_materials_pbrSpecularGlossiness](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Archived/KHR_materials_pbrSpecularGlossiness/README.md) extension in glTF. 

**It is important to note that this PR does not implement the Specular-Glossiness (SG) workflow itself (nor does pygfx intend to support the SG workflow)**. 

Instead, during model loading, materials that use the SG workflow are converted into an approximate Metallic-Roughness (MR) representation.

---
For those who are interested in background:

In the early days of PBR rendering, two primary workflows were commonly used: the Metallic-Roughness workflow and the Specular-Glossiness workflow. Some engines supported both workflows simultaneously. The early glTF specification also supported both workflows via extensions. However, with the introduction of extensions such as KHR_materials_ior and KHR_materials_specular, the gap between MR and SG workflows was significantly reduced.

Over the following years, the MR workflow became increasingly dominant. Many newer glTF extensions and advanced PBR features—such as clear coat, thin-surface transmission, volumetric effects, iridescence, and sheen—were designed around the MR workflow. In late 2021, Khronos officially archived the KHR_materials_pbrSpecularGlossiness extension and no longer recommended it for authoring new glTF assets. Many engines gradually dropped SG workflow support and focused solely on MR. pygfx’s PBR pipeline is based on the MR workflow and has no plans to add full SG workflow support.

However, a substantial number of earlier assets were authored using the SG workflow, and many models on platforms such as Sketchfab still export with KHR_materials_pbrSpecularGlossiness. For engines that do not support the SG workflow (such as pygfx), users typically need to convert SG assets to MR using external tools before loading them, which can be inconvenient. This PR streamlines the process by automatically converting SG workflow materials to approximate MR materials during loading, allowing SG-based glTF assets to be used directly.

---
Test Case: [SpecGlossVsMetalRough](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/SpecGlossVsMetalRough/README.md)

image

The bottle on the right was originally authored using the Specular-Glossiness workflow. During loading, we convert it to the Metallic-Roughness workflow for rendering. As shown, its appearance is almost indistinguishable from the bottle on the left, which was created directly using the MR workflow. The visual differences are minimal and can be considered negligible.

### DIFF
--- a/docs/advanced_shaders.rst
+++ b/docs/advanced_shaders.rst
@@ -42,7 +42,7 @@ The shader must implement a few methods. A typical shader is shown below:
             if material.has_some_value:
                 self["some_template_variable"] = True
 
-        def get_bindings(self, wobject, shared):
+        def get_bindings(self, wobject, shared, scene):
 
             # You can also set template-variables here. Again, when things that are used here change later, this
             # is detected, and this method will be called again. When a binding has changed (e.g. a colormap is replaced
@@ -118,7 +118,7 @@ to allow flexible code generation. Here's an example:
 
 .. code-block:: python
 
-        def get_bindings(self, wobject, shared):
+        def get_bindings(self, wobject, shared, scene):
             # Template variables can be set like this
             self["scale"] = 1.2
             ...
@@ -290,7 +290,7 @@ For images / volumes:
 
 .. code-block:: python
 
-        def get_bindings(self, wobjwect, shared):
+        def get_bindings(self, wobject, shared, scene):
             ...
             extra_bindings = self.define_img_colormap(material.map)
             bindings.extend(extra_bindings)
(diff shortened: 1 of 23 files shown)


almarklein (approved):
These techniques are all new to me 😄  Thanks for the explanation; sounds reasonable to me!

Korijn (approved):

Korijn:
Pretty awesome that we can provide this "backwards compatibility" so plainly :)

PR #1232, opened 2025-11-09
Direct GUI integration

Korijn:
Depends on https://github.com/pygfx/wgpu-py/pull/764 and https://github.com/pygfx/rendercanvas/pull/127

This example demonstrates how to integrate pygfx directly with wgpu and glfw, as a pure rendering engine.

With this PR I want to demonstrate that currently pygfx is tightly coupled to rendercanvas abstractions.

Of course this is super convenient if you want to get started with some viz right away, but it's also an obstruction if you want to integrate pygfx in your own application architecture.

I think pygfx can really shine if we cleanly separate the rendering engine from the interaction & GUI layers. This example shows that pygfx can be used as exactly that: just a mechanism to draw. But not without some modifications.

Again the purpose of this PR is to get your thoughts @almarklein I have a feeling you could see this PR coming from a mile away too. :D

A note on the changes:

* Firstly, I just made the absolute minimal set of changes to make the example run, to facilitate the discussion.
* I think in this situation, it is necessary that the application code (the code in the example file) calls `set_physical_size` and `set_pixel_ratio` on the renderer, also during resizes. And the renderer is then responsible for updating the context. Do you agree with this? (Update: applied this in the second commit)

The endgame I think is to provide one API here, and then users can choose to make the function calls on renderer themselves or let rendercanvas do it for them. Ideally that would mean there is no need for isinstance checks in the renderer anymore. Very similar to what was done in wgpu-py.

### DIFF
--- a/examples/other/gui_direct.py
+++ b/examples/other/gui_direct.py
@@ -81,9 +81,6 @@ glfw.set_framebuffer_size_callback(window, _on_size_change)
 glfw.set_window_content_scale_callback(window, _on_size_change)
 
 
-# Initialize physical size once. For robust apps update this on resize events.
-context.set_physical_size(*glfw.get_framebuffer_size(window))
-
 renderer = gfx.WgpuRenderer(context)
 camera = gfx.NDCCamera()
 
@@ -95,6 +92,9 @@ triangle = gfx.Mesh(
     ),
     gfx.MeshBasicMaterial(color_mode="vertex"),
 )
+
+# Initialize physical size once. For robust apps update this on resize events.
+_on_size_change()
 
 
 def main():


almarklein:
What about in https://github.com/pygfx/wgpu-py/pull/764 we add a property `physical_size` to `GPUCanvasContext`. 

Then, the `Renderer` accepts one of:
* `rendercanvas.WgpuContext`
  *  It has api to get physical size, pixel ratio, logical size, etc.
* `wgpu.GPUCanvasContext`: 
  * Renderer gets physical size from context.
  * User must set pixel ratio.
* `texture`:
  *  has a size. 
  * User must set pixel ratio.

PR #1234, opened 2025-11-11
fix docstring formatting

kushalkolar:
the performance tips in `pygfx.Texture` show up as arguments :joy: . I'm impressed at how it gets parse into this.

image

This puts them in a "Notes" section and formats them to show as a list.


kushalkolar:
seems like something weird is going on with pillow, I'll check this later.

Korijn (approved):

PR #1235, opened 2025-11-11
New lower bounds for rendercanvas and wgpu

almarklein:
I deliberately made the wgpu upper bound a bit higer; so we can make one more compatible release without having to update the bound here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,8 +21,8 @@ dependencies = [
     # Packages under the PyGfx umbrella.
     # For some of these we set an upper bound because (for now) they regularly break their API.
     # Packages that depend on PyGfx should not set an upper bound on e.g. wgpu, unless really needed!
-    "rendercanvas >= 2.2",
-    "wgpu >=0.24.0, <0.28",
+    "rendercanvas >= 2.4.0",
+    "wgpu >=0.28.0, <0.30",
     "pylinalg >=0.6.8,<0.7.0",
     # External dependencies
     "numpy",


almarklein:
I'm also creating a new branch `v0.15` from the `v0.15.0` tag, to give an upperbound to rendercanvas, and releasing `v0.15.1`.

hmaarrfk:
its funny that this reverts all the "best practices wisdom" that you tried so hard to impart!

the rule that "there are no rules in version 0" really holds...

almarklein:
> its funny that this reverts all the "best practices wisdom" that you tried so hard to impart!

How do you mean? From what I understand of the 'avoid upper-bounds-in-deps' reasoning, you do indeed expose yourself to unforeseen breakages, and the solution is to fix these asap.

almarklein:
Or do you mean that if we'd either have upper bounds on both rendercanvas *and* wgpu-py, or have upper bounds for neither, it was fine. But we now have trouble because only wgpu-py has an upper-bound ...

hmaarrfk:
It was a bit of a joke since I saw a few PRs loosening the upper bounds in the last few weeks. 

Then slight incompatibilities in strange test paths had you adding back upper bounds. 

Thx for following through and creating the 0.15.2 releases.

almarklein:
I guess its also a reminder that dependency management and API breakage are hard problems, no matter the strategy :)

hmaarrfk:
personally. locally i thought that rendercanvas 2.4 was compatible with pygfx 0.15.2 but it seems something is broken on at least one of my systems with


- pygfx 0.15.2
- wgpu-py 0.28.0
- rendercanvas 2.4.0

(I know pip shouldn't allow you to do this, but i was testing this out locally)

and it seems that this combo definitely still isn't allowed.

But it worked on "one machine".....

almarklein:
> it seems something is broken

I'm curious what is broken. If you have any tracebacks, please let me know. FWIW FPL at pygfx `main`  is suffering from a regression related to its example screenshots, which I am looking into.

hmaarrfk:
It seemed like new textures were just not getting sent to qt.

hmaarrfk:
Nothing appeared on the terminal

PR #1241, opened 2025-11-17
No blending needed for opaque

panxinmiao:
For opaque objects, the rendering pipeline does not need to enable any blending operations—simply writing the pixel values directly into the buffer is sufficient. This helps improve performance.

This change was originally included in https://github.com/pygfx/pygfx/pull/974, but it seems introduced slight screenshot differences in the CI results, and it was not directly related to the main topic of that PR. To clearly distinguish these differences from those potentially caused by https://github.com/pygfx/pygfx/pull/974 itself, I separated this into an independent PR.

### DIFF
--- a/pygfx/renderers/wgpu/engine/renderer.py
+++ b/pygfx/renderers/wgpu/engine/renderer.py
@@ -260,7 +260,10 @@ class WgpuRenderer(RootEventHandler, Renderer):
                 f"Render target must be a Canvas or Texture, not a {target.__class__.__name__}"
             )
         self._target = target
-        self.pixel_ratio = pixel_ratio
+        self._pixel_scale = None
+        self._pixel_ratio = None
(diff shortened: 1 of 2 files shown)


panxinmiao:
It's strange that the difference in screentshot has disappeared, 😓

almarklein (approved):

Korijn (approved):

panxinmiao:
Ah, the screenshot discrepancy has appeared again. It seems unrelated to the changes in this commit — there must be some unknown factor causing the CI screenshots to occasionally show slight differences.

almarklein:
> there must be some unknown factor causing the CI screenshots to occasionally show slight differences.

Yeah, still seems to glitch sometimes. I checked the diff screenshots last week or so. They had pixels being different in arbitrary places, so hard to pinpoint what's going on.

PR #1242, opened 2025-11-18
Fix renderer pixel scale

almarklein:
Fixes #1239

Before this PR, setting the renderer's `pixel_ratio`, calculates the `pixel_scale` and stores it. But when the pixel ratio of the canvas changes, the resulting pixel ratio is a different value. Similarly, using the default `pixel_scale`(the one where we use SSAA only when not on a HIDPI display), calculates the pixel-scale at instantiation, but does not update when the canvas's pixel ratio changes.

These problems already existed, but were exposed more because of recent changes in rendercanvas, most notably the one where setting the logical size cause cause a temporarily tiny pixel ratio (until the backend actually changes its size).

In this PR, the values for `pixel_scale` and `pixel_ratio` are always calculated on the fly, using the canvas config as a reference.

### DIFF
--- a/examples/other/integration_qt.py
+++ b/examples/other/integration_qt.py
@@ -31,7 +31,7 @@ class Main(QtWidgets.QWidget):
 
         # Create canvas, renderer and a scene object
         self._canvas = QRenderWidget(parent=self)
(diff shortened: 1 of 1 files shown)


almarklein:
ping

PR #1246, opened 2025-12-04
Add uniform.maprange to (almost) all objects that have .map

almarklein:
This adds ``uniform.maprange`` for all materials that have `.map`. Except for `BackgroundMaterial` where it makes little sense:
* line
* points
* mesh
* image (alias for clim)
* volume (alias for clim)

This is basically `ImageMaterial.clim`, but applied to texcoords. It makes it possible for scientific applications for the texcoords to represent a certain domain-specific value, e.g. temperature, force, or speed. And then use the `maprange` to project that value onto the colormap.

Note that this is different from TextureMap's transform, which defines what part of a texture should be considered the map.

Also see https://github.com/fastplotlib/fastplotlib/pull/953 and https://github.com/fastplotlib/fastplotlib/issues/960

### DIFF
--- a/pygfx/renderers/wgpu/engine/blender.py
+++ b/pygfx/renderers/wgpu/engine/blender.py
@@ -230,10 +230,7 @@ class Blender:
 
         # Get the color_blend mini-dict
         if alpha_method == "opaque":
-            color_blend = {
-                "alpha": blend_dict(bf.one, bf.zero, bo.add),
-                "color": blend_dict(bf.one, bf.zero, bo.add),
-            }
+            color_blend = None  # no blending needed for opaque
         elif alpha_method == "blended":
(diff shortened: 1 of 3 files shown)


examples/validation/validate_line_colormap.py
```diff
@@ -0,0 +1,92 @@
+"""
+Line colormapping
+=================
+
+Nearly all objects in Pygfx can have a colormap. And the texcoords
+can also be scaled using ``material.maprange``. This is useful
+in scientific applications, where texcoords can e.g. mean temperatire of force.
```
Korijn: Typos

Korijn (approved):

kushalkolar:
Thanks! This will be useful. :D

PR #1249, opened 2025-12-06
Add support for panning on macOS with trackpads

berendkleinhaneveld:
This allows for holding 'control' and clicking + dragging on macOS with trackpads.

There is a difference between how glfw and pyside handle this situation (control + mouse drag on trackpad on macOS):

glfw gets event: 'control+mouse1'
pyside gets event: 'meta+mouse2'

And the weird thing is: in Pyside, 'control+mouse1' will actually enable command + mouse drag 😕 
So with this solution you can use control _and_ command + mouse drag on Pyside to pan a scene.

### DIFF
--- a/pygfx/controllers/_orbit.py
+++ b/pygfx/controllers/_orbit.py
@@ -50,8 +50,10 @@ class OrbitController(PanZoomController):
         "mouse1": ("rotate", "drag", (0.005, 0.005)),
         "mouse2": ("pan", "drag", (1, 1)),
         # On macOS with trackpad, just 'mouse2' doesn't work
-        # Holding the control key will trigger the mouse2
-        # but uses 'meta' key as modifier
+        # so also register 'control' + click
+        # With glfw it is what you expect:
+        "control+mouse1": ("pan", "drag", (1, 1)),
+        # For Pyside it is something else:
         "meta+mouse2": ("pan", "drag", (1, 1)),
         "mouse4": ("quickzoom", "peek", 2),
         "wheel": ("zoom", "push", -0.001),


almarklein:
> There is a difference between how glfw and pyside handle this situation (control + mouse drag on trackpad on macOS):

mmm, I think we can consider this a bug in rendercanvas. I filed https://github.com/pygfx/rendercanvas/issues/154

I (eventually) came to the conclusion that checking both Meta and Control is the best solution for now. 

Apart from this Control/Meta stuff, is there a reason why you did not use Shift?

berendkleinhaneveld:
I didn't use shift because I wanted to keep it conceptually as close as possible to a 'right' mouse click/drag. 

I was also thinking that changing the default controls here would be better left out of scope: just make sure that 'meta' can be used as a modifier and not have pygfx throw an exception.

almarklein:
Ah I see what you mean. But this is a problem on any laptop (regardless of the OS), right?

berendkleinhaneveld:
> Ah I see what you mean. But this is a problem on any laptop (regardless of the OS), right?

That's true. We could just leave in the control+mouse1? Tgere is at least something that works on all systems and frameworks. I'm also ok with shift instead of control. Whatever you prefer.

almarklein:
I'm tempted to go with Shift, because control is simply a bit 'messy' right now. I felt a bit reluctant at first, because then shift cannot be used for other stuff, but I guess people would simply override shift+lmb in that case and all is well.

BTW: I just tried and it seems like a two-finger drag (on the touchpad) with the fingers pressed down will actually pan!

PR #1254, opened 2025-12-22
Post-processing: normal reconstruction

Korijn:
This post processing effect implements [this article](https://atyuwen.github.io/posts/normal-reconstruction/) to reconstruct normals from the depth buffer.

My goal is to learn more about shaders myself, and eventually I want to try my hand at [outlines](https://roystan.net/articles/outline-shader/) in a later PR.

https://github.com/user-attachments/assets/58638d2a-359a-4f5b-9621-aa7a3b29219b

### DIFF
--- a/examples/feature_demo/posteffect_normals.py
+++ b/examples/feature_demo/posteffect_normals.py
@@ -59,8 +59,6 @@ scene.add(boxes)
 
 controller = gfx.OrbitController(camera, target=scene_center, register_events=renderer)
 
-stats = gfx.Stats(viewport=renderer)
-
 normal_pass = NormalPass()
 renderer.effect_passes = [normal_pass]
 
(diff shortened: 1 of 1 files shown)


almarklein:
This looks good to me, but I have not checked the math thoroughly. Are the colors stable with the last commit? It looks stable when I run it, though I get a grayscale image.

> Debugging shaders is a skill I definitely need to work on. Outputting colors to the screen to figure out if the values are correct is really troublesome 😅

That's usually what I do 🤷   I never tries https://renderdoc.org/ yet, tbh.

Korijn:
🎉 It works now, I've updated the top post with a video demonstation. Check it out! Even the sharp corners of the cubes look perfect. :)

This took me a lot of time and about 6 different sessions to get it right. I learned so much in the process! 🥰

Korijn:
It seems like there are some flaky/broken tests in the main branch. :/

almarklein (approved):
Awesome work! And thanks for the clarifying comments in the shader regarding the coord system :)

The screenshot tests have been flaky for a while ... looks like lavapipe sometimes just decides to add weird patterns to the images 🤔

Korijn:
> Awesome work! And thanks for the clarifying comments in the shader regarding the coord system :)

I could not get this to work until I carefully worked through all the coordinate systems step by step :') And I did not know the position varying is automatically transformed in between the vertex shader to the fragment shader!

PR #1269, opened 2026-02-02
Fix an issue where keyframe optimization was not applied correctly.

panxinmiao:
As stated above, this was caused by a small typo.


panxinmiao:
These five examples seem to frequently fail the screenshot comparison tests:

 * `validate_image_colormap`
 * `validate_points_markers`
 * `validate_line_loop`
 * `validate_ppaa`
 * `validate_outpass`

 Some of the features they rely on appear to have subtle differences across platforms.

Korijn (approved):

Issue #650, opened 2024-02-17
Pyodide support


Korijn:
I've reviewed the requirements for making pygfx compatible with pyodide. It seems like all dependencies are already either built-in to pyodide (e.g. Jinja2 and cffi), or they are pure python. wgpu-py and pylinalg are already compatible I think. The only challenge lies in pygfx in (you guessed it) freetype-py and uharfbuzz which are most likely not available in emscripten format (I think).

almarklein:
I expect we need some work in wgpu-py too: https://github.com/pygfx/wgpu-py/issues/407

Although ... you mentioning that cffi is supported ... would wgpu-native be compilable with emscriptem and would it use webGPU then automatically?  Eeeh, probably not, but worth checking.

Korijn:
I don't think so. We'd have to use the JS API to talk to the browser's web GPU implementation.

legut2:
Okay. I was experimenting around with getting `pygfx` running in a browser this past weekend. I thought this would be a simple thing to do but it quickly got complicated. The first thing I tried was `pyodide`.  I hit a wall when I tried to render.

I got fairly far and it was a messy process and you can see the code over here:
[pyodide repo](https://github.com/legut2/pyodide-pygfx-messy-example)

You can inspect it through your browser at this
[static site](https://daniellegut.com) hosted on github pages. It computes, but doesn't render. I wanted to find a hacky way to render, but couldn't.

This involved tearing out anything `text` and `freetype` related in `pygfx` and building a custom wheel of `pygfx`. You can probably see all the code I had to comment out by opening the wheel. I think I even custom built uharfbuzz. I lost track of the many things I had to do just to get it to work as much as it does now.

With `pyodide` I'd usually hit this kind of error whenever initiating a renderer of pretty much any kind:
```
Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 596, in eval_code_async
    await CodeRunner(
  File "/lib/python312.zip/_pyodide/_base.py", line 410, in run_async
    coroutine = eval(self.code, globals, locals)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<exec>", line 12, in <module>
  File "/lib/python3.12/site-packages/pygfx/renderers/wgpu/engine/renderer.py", line 142, in __init__
    self._shared = get_shared()
                   ^^^^^^^^^^^^
[...]

PR #1273, opened 2026-02-12
Pyodide compatibility - pygfx in the browser [preview]

Vipitis:
Solves https://github.com/pygfx/pygfx/issues/650
Strongly related to https://github.com/pygfx/wgpu-py/pull/753

# [docs preview](https://pygfx--1273.org.readthedocs.build/1273/_gallery/index.html#feature-demos)

This PR includes all of the changes needed to run pygfx in the browser. This is very much parallel to the `wgpu-py` changes for browser compatibility: js ffi instead of wasm compiled wgpu-native https://github.com/pygfx/wgpu-py/pull/753

It also requires a few upstream changes. For example uharfbuzz https://github.com/harfbuzz/uharfbuzz/pull/275
Als well as changes to Imageio https://github.com/imageio/imageio/issues/1169

I think it might be sufficient to replace freetype by a canvas based Font atlast. This ain't done yet but I will look into this in the next few days. 

There is also a couple of issues with existing wgsl shaders. I hope to resolve those shortly. One of the problems is with fxaa/dpaa which is currently bypassed by env vars. A limitation in the text shader for example was https://github.com/gfx-rs/wgpu/issues/6990. And I hope the aa shader fixes are similarly simple. I think it might be useful to default to 1:1 pixels on the browser for performance. Will look into that.

Buffer copy is really slow and an issue for performance in picking. I want to look into async buffer download to reduce the latency. 

This PR should also eventually port the gallery to be interactive. Perhaps even Jupyter Lite. I want to have a minimal playground that let's users edit code. 

Plenty more things to try eventually.


Vipitis:
https://wgpu-py--753.org.readthedocs.build/en/753/gallery/pygfx_example.html

Vipitis:
went through all of the validation examples and looked for errors in the browser console. Here is a list of all that don't *run*... I haven't checked if they are semantically correct.
* [ ] line loop: `value nan cannot be represented as 'f32'` https://github.com/gpuweb/gpuweb/issues/5316 (can't be solved?)
* [x] ndc: `Failed to read the 'cullMode' property from 'GPUPrimitiveState': The provided value '0' is not a valid enum value of type GPUCullMode.`  Sounds like a hardcoded wgpu enum and should likely be `"none"`.
* [x] ppaa: `texture_2d<f32> cannot be used as the type of a 'let'`
* [x] ruler: font related, not sure anymore. the traceback points to the `run_sync` in the font loading logic... but the example works if I iterate through fewer children. it might be a race condition in the temporal cache... hard to tell.  I still think something is up with the race conditions (similar to the view office example), but avoiding the filename to face roundtrip just avoids the problem for now.
* [x] skybox: pyodide filesystem is not served.
* [ ] streamed blending: unsure yet which run_sync fails...
* [x] tiny image: `cannot assign to TypedArray of length 1 from buffer of length 4` might be some missing functionality in the wgpu PR.
* [ ] view offset: ~~`bytesPerRow (3200) is not a multiple of 256.` need to replace the lazy size by padding and handle offsets etc.~~ now there is some issue when doing `offscreen_canvas.draw` not sure why it has a problem during `mapAsync` yet... 
* [x] volume: `cannot assign to TypedArray of length 54 from buffer of length 56` same as above, maybe we don't need the alignment at all or I missed the offset in the calculation

Vipitis:
I added the preview in the docs for all examples... didn't apply the asset fetch patch yet and haven't looked into pyodide file system for trying to load the scenes - but quite a few examples do *function*. So try the docs preview, scroll down and click the button to load it in the browser. I didn't use a webworker/pyscript so the whole tab might become unresponsive on load or hangs.

you might need to click the canvas to get key events to work like here: 
https://pygfx--1273.org.readthedocs.build/1273/_gallery/feature_demo/fly_controller.html

Vipitis:
without the font crashes (still messed up tho), fpl is on the horizon 🤩

https://github.com/user-attachments/assets/ace2a68e-c8d0-420c-8344-598bb060349d

cc @kushalkolar

kushalkolar:
Jan you are epic!

Vipitis:
for anyone trying the docs preview: for some reason the performance and responsiveness is waaay better when you don't have the dev tools open - might be a chrome thing. [try here, now with assets for most examples, but not all](https://pygfx--1273.org.readthedocs.build/1273/_gallery/feature_demo/pbr_instanced.html)


https://github.com/user-attachments/assets/53878968-0b61-4022-a6c2-7384fadb20d4

pygfx/renderers/wgpu/wgsl/light_toon.wgsl
```diff
@@ -9,7 +9,7 @@ fn getGradientIrradiance( normal: vec3<f32>, light_dir: vec3<f32> ) -> vec3<f32>
     let coord = vec2f(dot_nl *0.5 + 0.5, 0.0);
 
     $$ if use_gradient_map is defined
-        return vec3f(textureSample(t_gradient_map, s_gradient_map, coord).r);
+        return textureSampleLevel(t_gradient_map, s_gradient_map, coord, 0.0).rrr;
```
Vipitis, line 12: the browser version seem to not get the the same conclusion during uniformity analysis than native. So the derivate based mipmap sampling can't run. Since this texture doesn't has mips (and it's not really 2D either...) explicit sampling level works instead. This might not matter here - but it's a semantic change that could be a problem elsewhere but I got lost in the templating to really understand the call tree here and perhaps solve the uniformity analysis

examples/validation/validate_view_offset.py
```diff
@@ -84,7 +84,13 @@ def animate1():
     renderer1.render(scene1, camera1)
 
 
-@canvas.request_draw
+# so in the current situation with run_sync and rendercanvas (2.6.2) having the two draw functions (which in turn call sync_waited promises somewhere) causes a problem with Pyodide.
```
Vipitis, line 87: @almarklein I haven't yet managed to minimize this issue yet, but perhaps you have a serendipitous idea why it might happen? Otherwise I will assume it's pyodide specific weirdness with their `run_sync` and move on for now.

Vipitis:
It's not hb-gpu yet, but there is now hb-raster in uharfbuzz which wasn't even that complex to implement. However it's not an signed distance field, so some examples are [not great](https://pygfx--1273.org.readthedocs.build/1273/_gallery/feature_demo/text_multiple_grid.html#interactive-example) - but it still works for most of the things... it might even be [more performant ?](https://pygfx--1273.org.readthedocs.build/1273/_gallery/feature_demo/text_waterfall.html#interactive-example).

Apart from all the hacks for waiting on upstream releases(like new rendercanvas styling) - I think this is sorta ready and useable.

I want to try implementing like an automated test runner. A bit like the [WebGL CST](https://registry.khronos.org/webgl/sdk/tests/webgl-conformance-tests.html) where you can see the doc screenshot next to a frame rendered by your browser. But this might be a challenge with using workers. Either case I should probably run the test suite against pyodide to find more failures.

PR #1282, opened 2026-03-06
Controller.handle_event: Redraw on resize

gselzer:
Since the resize will affect things like view aspect ratio, it seems natural that a resize should trigger a new draw to allow the controller to react. 

See pyapp-kit/ndv#239 for some related discussion.

I looked around for a natural spot to include a tests, but didn't see any spot that made sense. Let me know if you can think of one!

### DIFF
--- a/pygfx/controllers/_base.py
+++ b/pygfx/controllers/_base.py
@@ -256,6 +256,7 @@ class Controller:
             "key_up",
             "wheel",
             "before_render",
+            "resize",
         )
 
     def tick(self):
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks for looking into fixing the resize issue! I looked into rendercanvas and found that it indeed had a bug that it did not issue a redraw when resizing. This is a bug. I fixed it in https://github.com/pygfx/rendercanvas/pull/188, and rolled a new release (rendercanvas v2.6.2). So the changes in this PR should not be necessary.

gselzer:
Great, thank you! In hindsight rendercanvas does seem like the better spot for a fix! 👍

Issue #118, opened 2021-09-02
Support map projections


almarklein:
Non-linear projections, like map projections (e.g. Mercator) are are a use-case that we'd need to support. It's may also be tricky to get right, so it may be a good idea to consider this use-case early on.

almarklein:
Also see a relevant discussion in vispy (https://github.com/vispy/vispy/issues/1400).

PR #1286, opened 2026-04-01
Exploring nonlinear transformations (Log, Polar, and map projections)

almarklein:
Also see #118 and https://github.com/vispy/vispy/issues/1400

In this PR I explore a possible way to implement nonlinear projections.

### Design

The idea is that the camera gets a `.nonlinear` prop, that is a string with any of the supported projections (e.g. "ylog", "polar", "logpolar", "mercator").

Each nonlinear transform has a corresponding wgsl implementation, and some also a Python/numpy implementation.

In the shader, the nonlinear transform is applied to the world-space coordinates. In effect, we introduce a new space, the 'linear-space'. When there is no nonlinear transform, it is the same as the world-space. The camera operates in this linear space. This means that all panning and zooming works unchanged; it's `camera.position` is expressed in this linear space.

The `cmara.show_object()` can transform the bounding-box with the same non-linear transform to get a nice fit. This works for log plots, but not for any of the other transforms 🙃 

The ruler is made aware of these transforms as well, so that it can produce better auto-ticks. I implemented this for the log scale, which works petty well. It's currently broken for the polar plot. If a user provides explicit ticks, the ruler remains unaware of the nonlinear transform; it lives itself in the world space.

### Preliminary screenshots 

A log plot:
image

A polar plot similar to [this mpl example](https://matplotlib.org/stable/gallery/pie_and_polar_charts/polar_demo.html), but the rulers need work:

image

Drawing world coastal lines in lon/lat:

image

Drawing the same data with a spherical projection:

image

## Discussion

The high-level goal is that PyGfx is suited for (scientific) use-cases where these nonlinear transform play a role. We should in any case have examples that demonstrate how these cases are covered in PyGfx.

I'm not entirely sure if this PR is the best approach though. Let's also discuss alternatives. Maybe we can cover such cases with a few small features in the right places.

It's also worth mentioning that we don't necessarily need to cover log/polar/maps with the same solution. 

### The cool stuff

What's cool is that the data stays expressed in the native coordinate frame, and by changing the `camera.nonlinear`, you get a different view. You could show the same scene with different camera's using different transforms.

Any data that you add, like clouds, satellite imagery, plane trajectories, in the world map example would simply scale along.

### What's not so great

The new 'linear space' does introduce more complexity, and I'm not sure if all shaders will behave as expected, e.g. lights. We may need to make adjustments later on. But maybe that's fine, since these cases are kinda special?

The camera operates in this linear space. So in what space does e.g. `camera.look_at()` work?

The transform happens at the vertices. So for some objects we'd need to introduce a finer grid of vertices for this to work well, e.g. Image. Some objects, like the `Grid` cannot be transformed this way. It may be good to eventually have a grid based on a line object.

### Alternatives

For log transforms, users or plotting libs can easily transform their data with `np.log10()`, and set custom ticks. We can add an option to the ruler to show log10 ticks to make this use-case friendlier. Since this transform is relatively cheap, I think this is a very viable (and much simpler) alternative.

Polar transforms are rather special. Users and plotting libs could also construct them with some numpy transforms and  PyGfx primitives. We can add an option for the ruler to follow a curved path to accommodate this case. And maybe a grid that follows a specific curve, or applies a certain transform?

For maps, I think a similar story holds? Not sure, need to talk to some geo-persons.

Vispy took a pretty extreme route, in that every node can have any kind of transform. So you can have a `Group`, and everything in that group is transformed with log/polar/whatever. This can be very powerful but also the cause for complexity. I'm not sure how many users actually make use of this.

### DIFF
--- a/examples/feature_demo/log_plot.py
+++ b/examples/feature_demo/log_plot.py
@@ -1,0 +1,87 @@
+"""
+Line plot
+=========
+
+Show a line plot, with axii and a grid.
+"""
+
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
+import numpy as np
+from rendercanvas.auto import RenderCanvas, loop
+import pygfx as gfx
+from pylinalg import vec_transform, vec_unproject
+
+
+canvas = RenderCanvas()
+renderer = gfx.WgpuRenderer(canvas)
+
+scene = gfx.Scene()
+
+background = gfx.Background.from_color("#000")
+
+grid = gfx.Grid(
+    None,
+    gfx.GridMaterial(
+        major_step=1,
+        minor_step=0,
+        thickness_space="screen",
+        major_thickness=2,
+        minor_thickness=0.5,
+        infinite=True,
+    ),
+    orientation="xy",
+)
+grid.local.z = -10
+
+rulerx = gfx.Ruler(tick_side="right", tick_marker="tick_left", min_tick_distance=50)
+rulery = gfx.Ruler(tick_side="left", tick_marker="tick_right", min_tick_distance=40)
+
+x = np.linspace(20, 980, 200, dtype=np.float32)
+y = np.sin(x / 100) * 4 + 5 + np.random.uniform(-0.8, 0.8, x.shape).astype("f4")
+y = 2 ** y
+positions = np.column_stack([x, y, np.zeros_like(x)])
+
+line = gfx.Line(
+    gfx.Geometry(positions=positions),
+    gfx.LineMaterial(thickness=4.0, color="#aaf"),
+)
+scene.add(background, grid, rulerx, rulery, line)
+
+camera = gfx.OrthographicCamera(maintain_aspect=False,
+        nonlinear='ylog10'
+)
+camera.show_object(scene, match_aspect=True, scale=1.1)
+
+
+controller = gfx.PanZoomController(camera, register_events=renderer)
+
+
+# A user may prefer to hard-code the ticks, but the ruler does a pretty good
+# job at selection approproate ticks for log scales.s
+# rulery.ticks = [1, 5, 10, 50, 100, 500, 1000]
+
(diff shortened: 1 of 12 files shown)


almarklein:
After giving this more thought, and considering the comments in https://github.com/vispy/vispy/issues/1400, I think it'd be better that we do not support pre-defined nonlinear transforms, because in maps, the variety of transform is nearly endless, and also complex. Instead, we should provide a way for users to create custom non-linear transforms.

I can see three possible places to expose this:
* On the camera.
  * Like in the current PR, but instead of an enum it would be custom WGSL.
  * I like the ones below more.
* On the objects themselves.
  * This is the most practical to implement.
  * Maybe a bit odd to define the 'final' transform at the leaf node.
  * But who cares in use-cases like this? 
* On the scene.
  * A bit harder to implement, but should fit in the 'environment' shader logic that we also have for e.g. lights. 
  * Seems more appropriate, since the scene already represents the outermost parent.
  * Although we don't enforce that a scene cannot be used as a group; maybe we should disallow that.

Putting it on the scene seems the 'cleanest' solution, but it also introduces inconsistencies, e.g. bounding boxes don't take it into account. It will be a lot of work---and added complexity---to fully incorporate such a nonlinear transform into all the subsystems. Therefore I currently lean more towards putting it on the object, which marks that object as special, maybe bounding boxes can even ignore such objects. 

So a new proposal would be:

* Add a `WorldObject.nonlinear` or `Scene.nonlinear`, which is WGSL code that can be used to create a custom nonlinear transform.
* Repurpose the ruler logic in the current PR to make it easy to have automated ticks in log scale (since that's *relatively* common.
* Support rulers that follow a curved path so users can create a polar plot, even if one cannot really interactively navigate it as with other plots.
* Make examples for log, polar, and one map projection.

almarklein:
Made the first step for the proper solution (I think) in #1290.

almarklein:
I'm closing this, as it's superseded by #1290, #1291, and an upcoming PR for log and polar.

PR #1290, opened 2026-04-20
Add support for nonlinear transformations via ``WorldObject.nonlinear_transform``

almarklein:
See #1286 for the initial exploration to nonlinear transforms.

In this PR, support for nonlinear transformations is added to the world object. The transform is applied to the objects' vertices, not to that of its children. This makes that the transform get less "in the way" than in #1286. In the shader code, basically `load_s_positions(..)` is replaced with `nonlinear_transform(load_s_positions(..)`.

The transform is expressed as a WGSL function or expression that transform `vec3f -> vec3f`. There are no builtin transforms, but say a log transform is as simple as `vec3f(pos.x, log10(pos.y), pos.z)`. 

I added an example that adds a wobble effect to a mesh. In next PR's I will add examples for map transformations, log plots, and polar plots.


almarklein:
>  In next PR's I will add examples for map transformations

This one is getting really cool 😄 

*edit: see #1291*

pygfx/objects/_base.py
```diff
@@ -355,6 +359,33 @@ def receive_shadow(self) -> bool:
...
+
+    @nonlinear_transform.setter
+    def nonlinear_transform(self, value: str | None):
+        if not (value is None or isinstance(value, str)):
+            raise TypeError("nonlinear_transform must be str or None")
+        if value is not None:
+            value = value.strip() or None
+        if value is not None and not any(x in value for x in ("fn n", "return ", ";")):
+            value = f"fn nonlinear_transform(pos: vec3f) -> vec3f {{ return {value}; }}"
+        self._store.nonlinear_transform = value
```
almarklein, line 393: @Korijn this is the most important part; how this transform fits into the rest of the transform system.

examples/feature_demo/wobbly_mesh.py
```diff
@@ -0,0 +1,75 @@
...
+        fn nonlinear_transform(pos: vec3f) -> vec3f {
+            let a = u_wobject.amplitude;
+            let t = u_wobject.t;
+            return vec3f(
+                pos.x + a * sin(pos.x * 3.1 + t * 1.3),
+                pos.y + a * sin(pos.y * 7.2 + t * 2.1),
+                pos.z + a * sin(pos.z * 8.7 + t * 2.4)
+            );
+        }
+    """
```
kushalkolar, line 39: this is nice, it basically allows for arbitrary transforms!

pygfx/objects/_base.py
```diff
@@ -355,6 +359,33 @@ def receive_shadow(self) -> bool:
...
+        The WGSL must define a function ``fn nonlinear_transform(pos: vec3f) -> vec3f { ... }``.
+        Alternatively, the given WGSL can be a single expression (without ';')
+        using ``pos`` (a vec3f) as an existing variable, in which case the code is
+        wrapped in a function. E.g. ``vec3f(pos.x, pos.y*1.5, pos.z)``.
+
+        Note that the transform is applied to the raw geometry's vertex
+        positions of this object, and does not affect child objects.
+
+        Note that the transform is not taken into account in the calculation of
+        bounding boxes.
```
kushalkolar, line 381: I came across the same problem when editing the image shader to render only specific rows/cols, I resorted to just writing info about the new size to `ImageGraphic`. Perhaps users can override `WorldObject.get_bounding_box()` to account for the transform?

pygfx/renderers/wgpu/wgsl/image.wgsl
```diff
@@ -22,8 +22,10 @@ fn vs_main(in: VertexInput) -> Varyings {
     let index = i32(in.vertex_index);
     let i0 = geo.indices[index];
 
+    // Note: for a nonlinear transform to work well for an image, we'd want a much denser grid of vertices. Leaving that for another day.
```
kushalkolar, line 25: I think they're typically just used for the colormap which is different

pygfx/renderers/wgpu/wgsl/grid.wgsl
```diff
@@ -118,6 +118,8 @@ fn vs_main(in: VertexInput) -> Varyings {
         // Construct position using only the grid's rotation. Scale and offset are overridden.
         let pos = cam_pos_on_grid + coord1 * v1 + coord2 * v2;
 
+        // Note: the nonlinear_transform is currently ignored for the grid. It does not have geometry defined vertex positions so it does not make a lot of sense, it seems.
```
kushalkolar, line 121: grids are used for things like log plots, but we can leave it for later https://stackoverflow.com/a/3591526

pygfx/renderers/wgpu/wgsl/std.wgsl
```diff
@@ -53,6 +53,9 @@ fn is_orthographic() -> bool {
     return u_stdinfo.projection_transform[2][3] == 0.0;
 }
 
+// Nonlinear transform from WorldObject.nonlinear_transform.
+{{ nonlinear_transform or 'fn nonlinear_transform(pos: vec3f) -> vec3f { return pos; }' }}
```
kushalkolar, line 57: Curious, when the compiler sees that this function just returns the input, does it optimize it out?

kushalkolar (approved):
few small questions, gtg!

Had my thesis _proposal_ defense last week, I'll be able to keep up with this more from this week!

pygfx/objects/_base.py
almarklein: Yeah, or allow an overridden bounding box, e.g. `WorldObject.set_bounding_box()`.

The problem is that in order for bounding boxes to work, every transform needs an implementation in both wgsl and Python. And maybe for both the forward as the inverse transform. Vispy followed this through, which is pretty cool, but also adds complexity.

pygfx/renderers/wgpu/wgsl/grid.wgsl
almarklein: Yeah, I think we need a line-based grid.

pygfx/renderers/wgpu/wgsl/image.wgsl
almarklein: Yeah, but imagine morphing an image. You need more vertices to act as 'control points' for that.

pygfx/renderers/wgpu/wgsl/std.wgsl
almarklein: I would be *very* surprised if it doesn't :)

pygfx/objects/_base.py
Korijn: Looks like a sensible place to me!

pygfx/renderers/wgpu/wgsl/image.wgsl
kushalkolar: ah I think a way to get around that now is to use a mesh with an image :thinking:

pygfx/objects/_base.py
kushalkolar: I can see this being easy to do if you have a few pre-defined transformed (like log10, log2 etc.), but not if you allow arbitrary wgsl transforms.

pygfx/renderers/wgpu/wgsl/image.wgsl
almarklein: Yes, that's also an option. But the long-term solution would be to change the wgsl code where we generate the vertices on the fly. That way we can make the grid pretty dense without needing to update a buffer or anything.

PR #1291, opened 2026-04-21
Add examples for map projections

almarklein:
Uses ``ob.nonlinear_transform`` (from #1290) to implement map projections. Since there are *a lot* of map projections, we choose not to make any of them builtin. Instead we provide examples for how to apply them:

* `world_map_projection.py` shows the common Winkel Tripel projection often shown in modern atlases. Code for the well-known Mercator projection is also present.
* `world_map_projections.py` is an interactive example that morphs one projection into another :D, see video below.

Screen Recording 2026-04-21 at 12 34 12

Note: this PR is based off #1290, need that to merge first.


Korijn:
The demo is really breathtaking :) and great idea to document example transforms this way. I just worry it might be difficult to discover?

almarklein:
> I just worry it might be difficult to discover?

True. I could add a section to the docs to explain why we don't explicitly support map transforms and link to these examples.

kushalkolar (approved):
LGTM!

PR #1297, opened 2026-05-06
Don't use np.seterr

### LINKED ISSUE #1289: np.seterr performance impact
Hi everyone, 

i recently stumbled upon pygfx coming from fastplotlib and saw that the np.seterr in the following code snippet seems to be called several times at a per object level per frame https://github.com/pygfx/pygfx/blob/09889df43092d9b6d3edc22f9f234c3823faaaec/pygfx/objects/_base.py#L213-L219

This would mean that it would currently make Nx2 for e.g, 10k calls for 5k objects.

I was thinking to move this to renderer level instead so there would only be 2 calls made regardless of how many objects there are.

some benchmark code to showcase this with just numpy:

```python
import time
import numpy as np


def make_random_matrices(n):
    matrices = []
    for _ in range(n):
        m = np.random.randn(4, 4).astype(np.float32)
        m_inv = np.random.randn(4, 4).astype(np.float32)
        buf = {
            "world_transform": np.zeros((4, 4), dtype=np.float32),
            "world_transform_inv": np.zeros((4, 4), dtype=np.float32),
        }
        matrices.append((m, m_inv, buf))
    return matrices


# current: np.seterr called per object
def update_current(matrices):
    for matrix, inv_matrix, buf in matrices:
        orig = np.seterr(under="ignore")
        buf["world_transform"] = matrix.T
        buf["world_transform_inv"] = inv_matrix.T
        np.seterr(**orig)


# proposed fix: np.seterr called once
def update_fixed(matrices):
    orig = np.seterr(under="ignore")
    for matrix, inv_matrix, buf in matrices:
        buf["world_transform"] = matrix.T
        buf["world_transform_inv"] = inv_matrix.T
    np.seterr(**orig)


# no seterr at all
def update_bare(matrices):
    for matrix, inv_matrix, buf in matrices:
        buf["world_transform"] = matrix.T
        buf["world_transform_inv"] = inv_matrix.T


def bench(fn, matrices, warmup=3, repeats=50):
    for _ in range(warmup):
        fn(matrices)
    times = []
    for _ in range(repeats):
        t0 = time.perf_counter()
        fn(matrices)
        t1 = time.perf_counter()
        times.append((t1 - t0) * 1000)
    times.sort()
    return times[len(times) // 2]


if __name__ == "__main__":
    counts = [100, 500, 1_000, 2_000, 5_000, 10_000]
    print(f"{'objects':>8}  {'current (ms)':>13}  {'fixed (ms)':>11}  {'bare (ms)':>10}  {'speedup':>8}")
    print("-" * 65)
    for n in counts:
        matrices = make_random_matrices(n)
        t_current = bench(update_current, matrices)
        t_fixed = bench(update_fixed, matrices)
        t_bare = bench(update_bare, matrices)
        speedup = t_current / t_fixed
        print(f"{n:>8}  {t_current:>13.3f}  {t_fixed:>11.3f}  {t_bare:>10.3f}  {speedup:>7.1f}x")
    print()
    matrices_5k = make_random_matrices(5_000)
    t_cur = bench(update_current, matrices_5k)
    t_fix = bench(update_fixed, matrices_5k)
    print(f"  5K objects, current: {t_cur:.1f}ms ({t_cur/16.67*100:.0f}% of frame budget)")
    print(f"  5K objects, fixed:   {t_fix:.1f}ms ({t_fix/16.67*100:.0f}% of frame budget)")
    print(f"  saved per frame:     {t_cur - t_fix:.1f}ms")
```

and the output from the benchmark:
```markdown
 objects   current (ms)   fixed (ms)   bare (ms)   speedup
-----------------------------------------------------------------
     100          0.152        0.014       0.012     10.8x
     500          0.769        0.063       0.061     12.2x
    1000          1.541        0.128       0.125     12.1x
    2000          3.056        0.258       0.255     11.8x
    5000          7.802        0.653       0.647     12.0x
    10000         17.417       1.295       1.317     13.4x

  5K objects, current: 8.5ms (51% of frame budget)
  5K objects, proposed fix:   0.7ms (4% of frame budget)
  saved per frame:     7.8ms
```

## Proposed fix

The proposed fix would be to remove the np.seterr calls from _base.py:
```python
    def _update_world_transform(self):
        """This gets called right before being drawn, when the world transform has changed."""
        self.uniform_buffer.data["world_transform"] = self.world.matrix.T
        self.uniform_buffer.data["world_transform_inv"] = self.world.inverse_matrix.T
        self.uniform_buffer.update_full()
```

and place them in renderer.py as such:
```python
        # Suppress underflow warnings once and for all transform updates
        orig_err_setting = np.seterr(under="ignore")
        try:
            flat = FlatScene(scene, view_matrix, self._object_count)
        finally:
            np.seterr(**orig_err_setting)
        self._object_count = flat.object_count
        flat.sort()
```

Let me know what you think and if this is a feasible suggestion!

hmaarrfk:
closes https://github.com/pygfx/pygfx/issues/1289


almarklein (approved):
Awesome, thanks!

For future reference, this solution is a significant performance improvement. However, some benchmarks in #1289 suggested that we can squeeze some more performance by using `np.seterr` but one level higher up the call tree. However, that solution complicates the code more, which may not be worth the extra performance gain in practice. Will need some benchmarks with PyGfx actually running to establish that.

Issue #1294, opened 2026-05-03
Iso surface is rendered incompletely if step_size is set to values < 1.0 in VolumeIsoMaterial


tastenachim:
When determining the surface within the volume with a given threshold-value, the view-ray is sampled by stepping along the ray.
The step size is first determined by the volume shape. But for VolumeIsoMaterial, the actual step size can be additionally adjusted by the VolumeIsoMaterial.step_size parameter. However, the number of steps to take is not adjusted, thus the ray may not penetrate the volume as much if step_size < 1.0.

This leads to volume at the back not rendered.

See pygfx/renderers/wgpu/wgsl/volume_ray.wgsl , fn raycast() (of mode=='iso') starting at line 297:
```wgsl
        let nstepsf = f32(nsteps);     //  <-- taken from caller

        // Primary loop. The purpose is to find the approximate location where
        // the surface is.
        let iso_threshold = u_material.threshold;
        let actual_step_coord = u_material.step_size * step_coord;   // <-- adjusted step size
        var surface_found = false;
        var the_coord = start_coord;
        var the_value : vec4<f32>;
        for (var iter=0.0; iter<nstepsf; iter=iter+1) {             // <-- but not adjusted number or iterations.
            let coord = start_coord + iter * actual_step_coord;     // <-- possible smaller step size
            let value = sample_vol(coord, sizef);
            let reff = value.r;
            if (reff > iso_threshold) {
                the_coord = coord;
                the_value = value;
                surface_found = true;
                break;
            }
        }
```

While dividing  nsteps by the the step_size-factor could help, There may be some rounding  concernces, which may require a broader approach.

Not really related, but nearby:
Following this code segment is a refining loop with a user-defined but constant substep size. In my opinion, this can be exchanged into an divide-by-two approach as done for the 'mip' and 'minip' mode.

almarklein:
Yeah, that indeed looks like an oversight. Nice catch!

Apart from fixing this, it would be good to unify the implementation a bit; all raycasting methods could benefit from a user-settable relative step size. Similarly the refine-loop could be made more similar as you suggested.

Would you have time to prepare a PR?

PR #1299, opened 2026-05-15
Fix the lighting for the iso volume rendering

### LINKED ISSUE #1295: Wrong normal for VolumeIsoMaterial
In the shader for VolumeIsoMaterial, The algorithm to determine the normal at a given position within a volume looks suspicious:

See [wgsl](https://github.com/pygfx/pygfx/tree/main/pygfx/renderers/wgpu/wgsl) > volume_ray.wgsl, fn raycast() in mode =="iso" beginning at line: 356
```wgsl
        var normal : vec3<f32>;
        var positive_value : vec4<f32>;
        var negative_value : vec4<f32>;
        let gradient_coord = 1.5 * step_coord;   // <--- strange! step_coord is the view direction. not the shape of volume!

        negative_value = sample_vol(the_coord + vec3<f32>(-gradient_coord[0],0.0,0.0), sizef); // <-- is used here
        positive_value = sample_vol(the_coord + vec3<f32>(gradient_coord[0],0.0,0.0), sizef);
        normal[0] = positive_value.r - negative_value.r;

        // [..] the same for y, and z.

        normal = normalize(normal);
```
Now, if the view direction aligns with the x-axis (eg. step_coord = (1, 0, 0)), then the there will be no gradient in y and z, thus the normal would always be view-direction as well (independent of actual volume data). Also, view direction may have negative values --> that would also influence the normal direction. 

I think, that line is meant to adjust for the volume shape and scaling. And should better be based on sizef. However, some more brain cells must be activated to determine the correct factor and to analyse the implication on is_front and PhongMaterial, that is following that code.

Here is an example program, rendering a sphere, as Iso Surface, and as Mesh. The spheres are not expected to look identical (as the Iso material seems to use its own Light sources??) but the flat impression of the Iso-Sphere confirms the issue with the normals and reproduces the effect with the aligned normal=view-direction.

```python
import numpy as np
from rendercanvas.auto import RenderCanvas
import pygfx as gfx


# Create a volume with a sphere implicit surface
def create_sphere_volume(size=64, radius=20):
    """Create a 3D volume where each voxel value = R - distance_to_center"""
    center = size / 2.0
    x, y, z = np.ogrid[0:size, 0:size, 0:size]
    distance = np.sqrt((x - center)**2 + (y - center)**2 + (z - center)**2).astype(np.float32)
    return radius - distance

# Setup canvas and renderer
canvas = RenderCanvas()
scene = gfx.Scene()

max_radius = 20 # used for volume formula
radius = 17     # used for threshold -> radius/max_radius determines the actual radius.

# Create volume object
voldata = create_sphere_volume(size=64, radius=max_radius)
geometry = gfx.Geometry(grid=gfx.Texture(voldata, dim=3))
material = gfx.VolumeIsoMaterial(
    clim=(voldata.min(), voldata.max()),
    threshold=max_radius - radius  # Surface at R - distance == 0
)
volume = gfx.Volume(geometry, material)
scene.add(volume)

# Create and add mesh-based sphere
geometry_mesh = gfx.sphere_geometry(radius=radius)
material_mesh = gfx.MeshPhongMaterial(color=0.9, wireframe=False)
mesh_sphere = gfx.Mesh(geometry_mesh, material_mesh)
mesh_sphere.local.position = (30, 0, 0)  # Position on right
scene.add(mesh_sphere)
# it needs explicit light:
scene.add(gfx.DirectionalLight())

gfx.show(scene, canvas=canvas, renderer=renderer)
```
(Iso is at the top, mesh at the bottom)
Image

almarklein:
Fixes #1295

Comparison of a volumetric sphere and a mesh sphere:

Screen Recording 2026-05-17 at 12 32 55

* [x] Fixes the incorrect `gradient_coord` to calculate the surface normal.
* [x] Use the lights of the scene, using the same code paths as the mesh (phong shading).
* [x] Fall back to builtin lights when the scene has no lights, for backwards compat.

The methods that represent the different ray-casting modes not produce the `depth` value themselves. That gives them more control, allowing more sophisticated shaders in the future (e.g. iso that determines transparency based on the thickness of the space the ray is cast through).


pygfx/renderers/wgpu/wgsl/volume_ray.wgsl
```diff
@@ -342,21 +402,20 @@ $$ elif mode == 'iso'
             discard;
         }
 
-
         // Colormapping
         let color = sampled_value_to_color(the_value);
         // Move to physical colorspace (linear photon count) so we can do math
         $$ if colorspace == 'srgb'
-            let physical_color = srgb2physical(color.rgb);
+            let physical_albeido = srgb2physical(color.rgb);
```
Vipitis: isn't it *albedo*? I have found both variants across the codebase

almarklein: It is, I will correct to use albedo consistently.

almarklein:
Also: I just realized that the normal should be transformed with the world transform.

kushalkolar (dismissed):

almarklein:
> Also: I just realized that the normal should be transformed with the world transform.

Fixed. And calculate the `view_dir` in the same way (projecting from model-space to world space).

pygfx/renderers/wgpu/wgsl/volume_ray.wgsl
```diff
@@ -428,25 +428,24 @@ $$ elif mode == 'iso'
...
         normal[2] = positive_value.r - negative_value.r;
-        normal = normalize(normal);
 
-        // Get world and ndc pos from the calculated texture coordinate
-        let data_pos = the_coord * sizef - vec3<f32>(0.5, 0.5, 0.5);
-        let world_pos = u_wobject.world_transform * vec4<f32>(data_pos, 1.0);
-        let ndc_pos = u_stdinfo.projection_transform * u_stdinfo.cam_transform * world_pos;
+        // Project normal to world space
+        let normal_proj0 =  u_wobject.world_transform * vec4f(0.0, 0.0, 0.0, 1.0);
+        let normal_proj1 =  u_wobject.world_transform * vec4f(normal, 1.0);
```
Korijn: you can do this with just the 3x3 part of the world transform, saves a bunch of math operations, see e.g.

https://github.com/pygfx/pygfx/blob/eac8303fbd0e3e76d1089cfe21f53b1b9fae6648/pygfx/renderers/wgpu/engine/effectpasses.py#L701-L708

almarklein: Nice catch, thanks!

Vipitis (approved):
diff looks good and I also tried it just fine

Vipitis:
likely unrelated: choosing a `step_size` smaller than the `substep_size` for the material causes issues, I only saw part of the volume surface. Not sure there is any reason to chose such values - but perhaps a warning or a min could improve the user experience.

almarklein:
Yeah, the `step_size` logic should also be addressed and made more uniform accross the different ray methods, see #1294

PR #1300, opened 2026-06-02
Better support for texture interpolation

almarklein:
I ran into an issue where a uint16 texture had blocky pixels even though `vol.material.interpolation` was set to 'linear'. Took me a while to realize we simply have not implemented linear interpolation for textures that cannot use a sampler (samplers can only be used with float or unorm texture formats).

This fixes it:
* [x] Implement linear interpolation for `gfx.Image` regardless of format.
* [x] Implement linear interpolation for `gfx.Volume` regardless of format.
* [x] Implement cubic interpolation as well (except for yuv formats).
* [x] Prepare for supporting 16unorm and 16snorm formats, i.e. more efficient linear interpolation of uint16 images and volumes.


Vipitis:
Would this allow dropping the `float32-blendable` feature requirement? Apparently it's not supported by some phones which I found out when trying the pyodide docs on mine. 

I haven't tried this myself, but how does `textureGather` behave with a non filtering sampler?

almarklein:
Would this allow dropping the float32-blendable feature requirement?

I had not thought about that yet, but maybe that requirement can indeed be made optional. That said ... the primary target is desktop, and I don't care too much if cannot support certain phones with PyGfx 🤷

Related, I think it would be good to add support for 16unorm and 16snorm since 16bit data is quite common in scientific applications (I actually have a use-case of this right now in a client project). Looks like this needs changes in wgpu-py, and enabling another extension.

(Sorry edited instead of replied...)

Vipitis:
> Would this allow dropping the `float32-blendable` feature requirement?

> I had not thought about that yet, but maybe that requirement can indeed be made optional.

Not sure how much changes would be needed, perhaps if the check/fallback can be made programmatically to check for the existing sampler type instead of against a list of formats. 

> Related, I think it would be good to add support for `16unorm` and `16snorm` since 16bit data is quite common in scientific applications (I actually have a use-case of this right now in a client project). Looks like this needs changes in wgpu-py, and enabling another extension. 

The new texture formats are included here: https://github.com/pygfx/wgpu-py/pull/805 (to be fair I haven't tested them, perhaps we should have a test case for all texture formats somehow), the only breaking change is push constants which I don't think are used in pygfx... so it might *just work*.


I will find time for a review later today.

almarklein:
> Not sure how much changes would be needed, perhaps if the check/fallback can be made programmatically 

Something like this:
* Request the feature, but ok if not available (may need changes in `wgpu.preconfigure_default_device()`)
* Then in `imageshader.py` and `volumeshader.py` check if the feature is "loaded". If not, set the param to not use the sampler.
* Check colormaps too.

examples/validation/validate_volume_slice.py
```diff
@@ -0,0 +1,63 @@
+"""
+Validate volume slice
+=====================
+
+Same as validate_image2.py, but for volume slices.
```
Vipitis: `validate_image2.py` got removed (or renamed - which git didn't pick up on) to just `validate_image.py`

examples/validation/validate_volume.py
Vipitis: since the default is now "linear" for all dtypes this comment isn't fully correct

pygfx/materials/_image.py
```diff
@@ -1,6 +1,7 @@
 from ..resources import Texture, TextureMap
 from ._base import Material
 from ..utils import assert_type
+from ..utils.enums import InterpolationFilter
```
Vipitis, line 4: is the enum also meant to be user API? if so, it could be included in the docstring and typehint

pygfx/renderers/wgpu/shaders/imageshader.py
Vipitis:
```suggestion
            raise TypeError("Image.geometry.grid must be a 2D texture")
```

pygfx/renderers/wgpu/shaders/volumeshader.py
```diff
@@ -70,9 +75,10 @@ def get_bindings(self, wobject, shared, scene):
         ]
 
         tex_view = GfxTextureView(geometry.grid)
-        sampler = GfxSampler(material.interpolation, "clamp")
-        bindings.append(Binding("s_img", "sampler/filtering", sampler, "FRAGMENT"))
         bindings.append(Binding("t_img", "texture/auto", tex_view, vertex_and_fragment))
+        if self["interpolation"] == "via-sampler":
+            sampler = GfxSampler(material.interpolation, "clamp")
+            bindings.append(Binding("s_img", "sampler/filtering", sampler, "FRAGMENT"))
```
Vipitis: would it make sense to rename everything to _vol or does this break when using the included image code for volume sampling?

Vipitis (dismissed):
mostly just nits and couple questions.

went down a couple rabbit holes like *isn't `weight` just constant?* or *can you reuse some of the code?* but both lead nowhere: looks good to me

pygfx/renderers/wgpu/shaders/volumeshader.py
almarklein: good idea!

almarklein:
> isn't weight just constant?

Hehe, no they're not. If they were, you could do a lot of cool optimizations, like in our ssaa code: https://github.com/pygfx/pygfx/blob/82984011e61911ecfadcd32fd58a21dcf68a3d0d/pygfx/renderers/wgpu/wgsl/ssaa.wgsl#L256-L271

almarklein:
Thanks for the review @Vipitis ! Very helpful points!

Vipitis (approved):

PR #1302, opened 2026-06-08
Add weak event handlers and clear_event_handlers to EventTarget

hmaarrfk:
Working to find some memory leaks, these event handlers seem like strong candidates.

I need to cleanup the implementation. I don't really like the helper methods.
<details><summary>claude's comment</summary>
EventTarget stored every handler with a strong reference. When an object registers one of its own bound methods as a handler on an object it owns (for example a controller registering ``self._on_pointer_down`` on a scene it holds), this forms a reference cycle:

    owner -> child -> _event_handlers -> bound method -> owner

Reference counting can never break a cycle, so dropping the last external reference to the owner frees nothing: the objects, and any GPU resources such as textures they keep alive, are only reclaimed when the *cyclic* garbage collector runs. That collector is scheduled by generational allocation thresholds that count Python objects and are unaware of GPU memory, so VRAM can be exhausted long before it decides to run.

Add an opt-in ``weak=True`` argument to ``add_event_handler``. Weak handlers are stored as ``weakref.WeakMethod`` (for bound methods) or ``weakref.ref`` (for other callables), so the registration no longer keeps the callback - or, for a bound method, the instance it is bound to - alive. With it, the owner above is reclaimed by reference counting alone, without waiting for the cyclic collector. Dead handlers are skipped during dispatch and pruned via a weakref finalizer. The default (``weak=False``) preserves the existing strong-reference behaviour.

Also add ``clear_event_handlers(*types)`` to deterministically detach all handlers (optionally restricted to the given types), and teach ``remove_event_handler`` to match weakly-stored handlers.

</details>

### DIFF
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -93,7 +93,7 @@ an object registers one of its own bound methods on an object it owns (e.g. a
 controller registering ``self._on_pointer_down`` on a scene it holds), this
 forms a reference cycle::
 
-    owner -> child -> _event_handlers -> bound method -> owner
+    owner -> child -> owner._event_handlers -> bound method -> owner
 
 Reference counting cannot break a cycle, so dropping your last reference to the
 owner frees nothing until the *cyclic* garbage collector runs. That collector
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -91,7 +91,7 @@ Event handlers and memory
 By default ``add_event_handler`` keeps a *strong* reference to the handler. If
 an object registers one of its own bound methods on an object it owns (e.g. a
 controller registering ``self._on_pointer_down`` on a scene it holds), this
-forms a reference cycle::
+may form a reference cycle if the event handler is a class::
 
     owner -> child -> owner._event_handlers -> bound method -> owner
 
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -85,6 +85,34 @@ Doing a lot of work on certain events can affect performance. Especially events 
 Some seemingly hamless things, like using ``print()`` on every draw, are more expensive than you might think.
 
 
+Event handlers and memory
+-------------------------
+
+By default ``add_event_handler`` keeps a *strong* reference to the handler. If
+an object registers one of its own bound methods on an object it owns (e.g. a
+controller registering ``self._on_pointer_down`` on a scene it holds), this
+forms a reference cycle::
+
+    owner -> child -> _event_handlers -> bound method -> owner
+
+Reference counting cannot break a cycle, so dropping your last reference to the
+owner frees nothing until the *cyclic* garbage collector runs. That collector
+is scheduled by allocation thresholds that count Python objects and are unaware
+of GPU memory, so any textures and buffers the owner keeps alive can linger far
+longer than expected -- long enough to exhaust VRAM in workflows that create
+and discard many scenes.
+
+Two tools help reclaim such resources promptly:
+
+* Pass ``weak=True`` to ``add_event_handler``. The handler is then stored via a
+  weak reference, so the registration no longer keeps the callback (or, for a
+  bound method, its instance) alive, and the owner is reclaimed by reference
+  counting alone. Keep a reference to the callback yourself, otherwise it is
+  collected and never fires (see :meth:`~pygfx.objects.EventTarget.add_event_handler`).
+* Call :meth:`~pygfx.objects.EventTarget.clear_event_handlers` to deterministically
+  detach all handlers from an object before discarding it.
+
+
 Probably more
 -------------
 
--- a/pygfx/objects/_events.py
+++ b/pygfx/objects/_events.py
@@ -370,9 +370,17 @@ class EventTarget:
 
             obj.add_event_handler(self._on_pointer_down, "pointer_down", weak=True)
 
-        Note that with ``weak=True`` the caller is responsible for keeping
-        a reference to the callback alive; a handler whose only remaining
-        reference is this registration will be collected and never fire.
+        .. warning::
+
+            With ``weak=True`` the registration does not keep the callback
+            alive, so the caller must keep a reference to it. A handler whose
+            only remaining reference is this registration is collected and
+            never fires. This is safe for the common case above (a bound
+            method registered on an object the instance owns, since the
+            instance keeps the method's owner -- itself -- alive), but a
+            standalone function, a ``lambda``, or a bound method of an object
+            you do not otherwise hold on to will silently disappear. Use the
+            default ``weak=False`` for those.
         """
 
         decorating = not callable(args[0])
--- a/pygfx/objects/_events.py
+++ b/pygfx/objects/_events.py
@@ -324,6 +324,9 @@ class EventTarget:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._event_handlers = defaultdict(set)
+        # Finalizer shared by all weakly-stored handlers (see add_event_handler
+        # with weak=True). Created lazily, since most objects never use it.
+        self._weak_handler_remover = None
 
     def add_event_handler(self, *args, weak=False):
         """Register an event handler.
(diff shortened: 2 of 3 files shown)


pygfx/objects/_events.py
```diff
@@ -324,13 +325,19 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._event_handlers = defaultdict(set)
 
-    def add_event_handler(self, *args):
+    def add_event_handler(self, *args, weak=False):
```
hmaarrfk, line 331: @almarklein I would love your thoughts on making this weak by default. i'm torn.

pygfx/objects/_events.py
```diff
@@ -365,14 +393,35 @@ def my_handler(event):
...
             return decorator
         return decorator(callback)
 
+    def _wrap_weak_handler(self, callback):
+        # Wrap a callback in a weak reference so that storing it as a handler
+        # does not keep it (or, for a bound method, its instance) alive. The
+        # finalizer prunes the dead wrapper from every type it was registered
+        # for. It only holds a weak reference to self, so it does not turn the
+        # registration into a reference cycle of its own.
+        selfref = ref(self)
```
hmaarrfk: apparently, this helps get a little more "cleaning" done eagerly, but there is already a "cleanup zombies" later on. not sure which is best.

docs/performance.rst
```diff
@@ -85,6 +85,34 @@ Doing a lot of work on certain events can affect performance. Especially events
...
 
+Event handlers and memory
+-------------------------
+
+By default ``add_event_handler`` keeps a *strong* reference to the handler. If
+an object registers one of its own bound methods on an object it owns (e.g. a
+controller registering ``self._on_pointer_down`` on a scene it holds), this
+forms a reference cycle::
+
+    owner -> child -> _event_handlers -> bound method -> owner
```
hmaarrfk:
```suggestion
    owner -> child -> owner._event_handlers -> bound method -> owner
```

docs/performance.rst
```diff
@@ -85,6 +85,34 @@ Doing a lot of work on certain events can affect performance. Especially events
 Some seemingly hamless things, like using ``print()`` on every draw, are more expensive than you might think.
 
 
+Event handlers and memory
+-------------------------
+
+By default ``add_event_handler`` keeps a *strong* reference to the handler. If
+an object registers one of its own bound methods on an object it owns (e.g. a
+controller registering ``self._on_pointer_down`` on a scene it holds), this
+forms a reference cycle::
```
hmaarrfk:
```suggestion
may form a reference cycle if the event handler is a class::
```

pygfx/objects/_events.py
```diff
@@ -365,14 +396,42 @@ def my_handler(event):
...
         return decorator(callback)
 
+    def _get_weak_handler_remover(self):
+        # The finalizer for weakly-stored handlers. It receives the dead
+        # weakref and discards it from every event type, so dead handlers do
+        # not pile up. Following pygfx.utils.weak, it captures only a weak
+        # reference to self (via the default argument), so it does not turn the
+        # registration into a reference cycle of its own.
+        remover = self._weak_handler_remover
+        if remover is None:
```
hmaarrfk:
```suggestion
        if self._weak_handler_remover is None:
```

pygfx/objects/_events.py
```diff
@@ -365,14 +396,42 @@ def my_handler(event):
...
+
+            def remove(dead_ref, selfref=ref(self)):  # noqa: B008
+                self = selfref()
+                if self is not None:
+                    for handlers in self._event_handlers.values():
+                        handlers.discard(dead_ref)
+
+            remover = self._weak_handler_remover = remove
+        return remover
```
hmaarrfk:
```suggestion
            self._weak_handler_remover = remove
        return self._weak_handler_remover
```

pygfx/objects/_events.py
```diff
@@ -381,7 +440,36 @@ def remove_event_handler(self, callback, *types):
...
+            # The handler may have been registered weakly; find the wrapper
+            # whose referent is the given callback and remove that.
+            for handler in handlers:
+                if isinstance(handler, ref) and handler() == callback:
+                    handlers.discard(handler)
+                    break
+            else:
+                raise KeyError(callback)
+
+    def clear_event_handlers(self, *types):
```
hmaarrfk, line 456: do we actually want this????

almarklein: I think so

almarklein:
Using weakrefs for the handlers has been discussed a long time ago. Then we decided to not do that because if people use a lambda or a locally defined function, it's immediately cleaned up.

But I do agree that these handlers are prone to cause memory leaks, so I'm very open to the idea to use weakrefs in some way. One thing though, is that we should probably adopt the same strategy in rendercanvas. For consistency, and because I still plan to simplify Pygfx's event handling, unifying the two more.

Also interested in @Korijn's ideas.

pygfx/objects/_events.py
almarklein: Maybe default True for class-objects (`WeakMethod`) and False otherwise (possibly lambda or local func).

hmaarrfk: Interesting idea. Will wait for your final word when it comes

Korijn:
> Also interested in @Korijn's ideas.

Well, in python, weak event handlers are the intended solution to this specific problem.

In general however, to solve extern resource management (files, VRAM, etc), you need to define ownership explicitly. This is a design problem.

Who is the one true owner of a resource? They are responsible for disposal of external resources. Even if there are still event handlers around (!).

So to do this, dispose needs to be an explicit method, independent of the deconstructor(`__del__`). Because resource management and memory management are different problems. If you decouple these concerns, it won't matter when the GC collects a cycle.

In Three.js this is actually not solved at all, because it's a library, not an application. And different kinds of applications need to manage their resources in different ways.

I really regret adding interaction layers like controllers and events to pygfx, because it violates this boundary between library and application, and causes issues like this to come up in the pygfx repo, even though they should really be application concerns.

What I would do today, if I could do it over, is add explicit `dispose()` methods to objects that hold GPU resources (Geometry, Material and Texture only), and leave it at that, never calling dispose automatically.

It's really up to the application to decide what is appropriate. You could build a ResourceManager class that holds references to all GMT objects and organize them however you like. In a game it's probably appropriate to group them by level, so you can dispose a whole group of objects when the level unloads. For example.

hmaarrfk:
> Who is the one true owner of a resource? They are responsible for disposal of external resources.

I feel like we had the same discussion internally yesterday ^_^. We definitely agree.

One challenge is with python you expect "well my object no longer exists, `__del__` should be called" for me.

> I really regret adding interaction layers like controllers to pygfx

I think the controllers are really critical to any graphics library. It could be "better separated" I agree, and we are trying to untangle these ourselves given these recent findings.

> What I would do today, if I could do it over, is add explicit dispose() methods to objects that hold GPU resources (Geometry, Material and Texture only), and leave it at that, never calling dispose automatically.

We have in fact added this to our code outside of pygfx.

---------------------------------------------------------------

@Korijn, i think the general question is: 

Do you think it would be a mistake given the state today to add the ability to have weak handlers in Pygfx?

Korijn:
> One challenge is with python you expect "well my object no longer exists, `__del__` should be called" for me.

Yeah, so, I think that should be decoupled from GPU resource management.

> I think the controllers are really critical to any graphics library. It could be "better separated" I agree, and we are trying to untangle these ourselves given these recent findings.

I agree with their criticality, I just think they should be more like a piece of code that you can link up to other components yourself, instead of it being integrated out of the box. The math is the hard part that you want to offer as a library, the application glue is not.

> Do you think it would be a mistake given the state today to add the ability to have weak handlers in Pygfx?

It's the right move! Do it!

hmaarrfk:
> It's the right move! Do it!

thanks for confirming!

Issue #1303, opened 2026-06-11
Striping artifacts in Volume Rendering


gselzer:
Hi!

I've started seeing some striping artifacts when I display volumes using pygfx. I figured maybe it was just my machine, but with pyapp-kit/scenex#88 it seems clear that it's not just me.

<img width="800" height="598" alt="Image" src="https://github.com/user-attachments/assets/d6c4d86c-5c9a-43bf-9591-2d3549a555f6" />

I've narrowed things down to a pure-pygfx script, which I've inserted below

<details><summary>Script</summary>

```python

"""Plain pygfx equivalent of basic_volume.py for debugging."""

# /// script
# requires-python = ">=3.10"
# dependencies = [
#   "numpy==2.4.6",
#   "pygfx @ git+https://github.com/pygfx/pygfx.git@main",
#   "rendercanvas==2.6.3",
#   "pyqt6==6.11.0",
#   "qtpy==2.4.3",
#   "imageio[tifffile]==2.37.3",
# ]
# ///

import numpy as np
import pygfx
from imageio.v2 import volread
from rendercanvas.auto import RenderCanvas, loop

url = "https://gitlab.com/scikit-image/data/-/raw/2cdc5ce89b334d28f06a58c9f0ca21aa6992a5ba/cells3d.tif"
data = np.asarray(volread(url)).astype(np.uint16)[:, 0, :, :]

canvas = RenderCanvas(size=(800, 600), title="Volume (pygfx)")
renderer = pygfx.renderers.WgpuRenderer(canvas)

scene = pygfx.Scene()
vol = pygfx.Volume(
    pygfx.Geometry(grid=pygfx.Texture(data, dim=3)),
    pygfx.VolumeMipMaterial(
        clim=(float(data.min()), float(data.max())),
    ),
)
scene.add(vol)

orbit_center = np.array([
    (data.shape[2] - 1) / 2,
    (data.shape[1] - 1) / 2,
    (data.shape[0] - 1) / 2,
])

camera = pygfx.PerspectiveCamera(fov=70, depth_range=(1, 1_000_000))

cam_pos = orbit_center + np.array([300.0, 0.0, 0.0])
camera.local.position = cam_pos
camera.world.reference_up = (0.0, 0.0, 1.0)
camera.look_at(tuple(orbit_center))


@canvas.request_draw
def animate():
    renderer.render(scene, camera)
    canvas.request_draw()

loop.run()

```
</details>

## Alterations that avoid this behavior

[...]

PR #1304, opened 2026-06-12
Use the determinant to compute the camera sign

gselzer:
As noted by @Vipitis, this avoids rendering artifacts when the diagonal of the world matrix contains a zero. This can happen with camera rotations. This PR changes the shader to use the sign of the determinant instead.

This PR also adds a new validation example. We could alternatively bake that transformation matrix into another validation example - let me know which would be preferred.

I do still need to upload the validation PNG - my understanding is that it can be downloaded, verified, and added from the CI build here?

AI use: I used Claude Sonnet to pair down the example script posted in #1303 down to a validation example. The only important thing is that camera's world matrix, which would probably be about all that is needed to carry over to another example if we don't want a whole example just for testing this.


Vipitis:
> I do still need to upload the validation PNG - my understanding is that it can be downloaded, verified, and added from the CI build here?

Yes, you can get them from the build artifacts here https://github.com/pygfx/pygfx/actions/runs/27447371673 (or get them locally, best using llvmpipe on Linux - see the bottom of the readme)

gselzer:
Thanks, screenshot added.

Thoughts on adding the important parts of this example to another validation script vs. keeping separate?

almarklein:
> Thoughts on adding the important parts of this example to another validation script vs. keeping separate?

Not sure what you mean. We now have a separate validation example for this particular case already, right?

I tried the new validation example on `main` and the result looks fine, so it looks like the example does no capture the error case. *Edit: it seems to actually capture the error case now, not sure what happened when I tried it before I wrote this.*

gselzer:
> Not sure what you mean. We now have a separate validation example for this particular case already, right?

Yes, this PR adds a separate validation case. I was simply stating that it may not be necessary to have a separate one - we could incorporate the important parts of this issue (namely the camera transform) into another example if we didn't want a separate example just for this.



> I tried the new validation example on `main` and the result looks fine, so it looks like the example does no capture the error case.

Really? I just tried that as well on `main` (although I may have done something wrong, it took me a few hours of struggling to get `test_examples` running on WSL), and the new example fails on my machine. It produces a blank image (which is a little strange because that wasn't what I saw on Windows)

almarklein:
Taking your original code in #1303 and adapting it a bit so it does not need an external volume does the trick to reproduce the erorr:
```py


import numpy as np
import pygfx
from rendercanvas.auto import RenderCanvas, loop

data = np.zeros((60, 256, 256), np.uint8)
data[10:-10, 30:-30, 30:-30] = 70
data[20:-20, 40:-40, 40:-40] = 90

canvas = RenderCanvas(size=(800, 600), title="Volume (pygfx)")
renderer = pygfx.renderers.WgpuRenderer(canvas)

scene = pygfx.Scene()
vol = pygfx.Volume(
    pygfx.Geometry(grid=pygfx.Texture(data, dim=3)),
    pygfx.VolumeMipMaterial(clim=(0, 100)),
)
scene.add(vol)

orbit_center = np.array([
    (data.shape[2] - 1) / 2,
    (data.shape[1] - 1) / 2,
    (data.shape[0] - 1) / 2,
])

camera = pygfx.PerspectiveCamera(fov=70, depth_range=(1, 1_000_000))

cam_pos = orbit_center + np.array([300.0, 0.0, 0.0])
camera.local.position = cam_pos
camera.world.reference_up = (0.0, 0.0, 1.0)
camera.look_at(tuple(orbit_center))


@canvas.request_draw
def animate():
    canvas.request_draw()

loop.run()
```

On `main`:
image

WIth the proposed fix:
image

gselzer:
Yeah, I see something very similar to you on `main` with that code. But does the new `validate_volume_ray_cam_sign` validation example not look like this for you on `main`?

image

almarklein:
I would have sworn that yesterday it did not. But it does now. I suspect I did have your change activated from a previous time of trying it our or something ... Sorry for the noise!

almarklein (approved):

almarklein:
Thanks for the fix!

almarklein:
> Yes, this PR adds a separate validation case. I was simply stating that it may not be necessary to have a separate one 

Let's stick to this specific one, otherwise we risk that this case is factored out at some point.

PR #1305, opened 2026-06-22
Use unorm for 16bit textures

almarklein:
Requires https://github.com/pygfx/wgpu-py/pull/805.

Also `texture-format16bit-norm` is a native-only feature, which `wgpu.preconfigure_default_device(` does not allow (need to comment a check in wgpu to be able to use it). The texture features are going to change a bit soon, with two tiers to enable a bunch of features at once. 

So we probably need to wait until wgpu 0.30 and then wgpu-py updating to that. Or we can allow wgpu-py to enable this feature so we can start using it earlier.

### DIFF
--- a/examples/validation/validate_image.py
+++ b/examples/validation/validate_image.py
@@ -27,10 +27,10 @@ scene = gfx.Scene()
 
 data = np.array(
     [
-        [0, 1, 1, 1],
-        [1, 1, 1, 1],
-        [1, 1, 1, 1],
-        [1, 1, 1, 2],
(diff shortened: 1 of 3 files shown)


Vipitis:
maybe a future warning, I believe the native 16bit norm texture formats are still filterable as "float" which is the default. But in the webgpu spec they are [`unfilterable-float`](https://www.w3.org/TR/webgpu/#texture-format-caps). So I had to update sampling type in two places: https://github.com/Vipitis/wgpu-py/commits/browser/examples/cube.py but this now works native and browser

PR #1306, opened 2026-07-01
View reshape

gselzer:
The [newest version of numpy](https://github.com/numpy/numpy/releases/tag/v2.5.0) deprecated assignments to [view.shape]. This PR cleans up all of those usages, using `np.reshape` instead.

### DIFF
--- a/pygfx/resources/_buffer.py
+++ b/pygfx/resources/_buffer.py
@@ -90,7 +90,7 @@ class Buffer(Resource):
             # Store data and view, and do some basic checks.
             # The view is a numpy array, but we go via memoryview to ensure data follows the buffer protocol.
             self._data = data
-            self._view = view = np.asarray(memoryview(data))
+            view = np.asarray(memoryview(data))
             if self._force_contiguous:
                 check_data_is_clean_for_performance("buffer", view)
             the_nbytes = view.nbytes
@@ -103,12 +103,12 @@ class Buffer(Resource):
                 the_nitems = view.shape[0]
             else:
                 the_nitems = 1  # A scalar, e.g. a uniform struct
-            reshape_array(view, the_nitems)
+            self._view = reshape_array(view, the_nitems)
             # Establish format
             detected_format = None
-            element_format = get_element_format_from_numpy_array(view)
+            element_format = get_element_format_from_numpy_array(self._view)
             if element_format:
-                elements_per_item = int(np.prod(view.shape[1:], initial=1))
+                elements_per_item = int(np.prod(self._view.shape[1:], initial=1))
                 detected_format = (f"{elements_per_item}x" + element_format).lstrip(
                     "1x"
                 )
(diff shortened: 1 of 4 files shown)


pygfx/resources/_buffer.py
```diff
@@ -459,4 +459,5 @@ def reshape_array(view, n):
         if n == 0:
             elements_per_item = int(np.prod([max(i, 1) for i in view.shape], initial=1))
         # This can fail if the data is not contiguous and strides don't work out.
-        view.shape = (n, elements_per_item)
+        view = np.reshape(view, (n, elements_per_item), copy=False)
```
Vipitis: why is `copy=False` only needed here?

gselzer: Without knowing the other places where you're looking (I'm assuming maybe you're talking about `pygfx.geometries._toroidal`), I just assumed this was a spot where we would not want copies to be made. Admittedly I do not understand the situations where a copy *would* be made - if you have suggestions I'm open to changing!

Either way, we should probably change the other non-`tests` examples to follow whatever is decided here

almarklein: Since these were all shape assigments, a copy is not necessary, but I'm fine with omitting the copy arg to keep things concise.

I do prefer `view.reshape(n, elements_per_iterm)` though (and all other places) for it being much shorter.

gselzer: This sounds like a plan! Let me know if any other concerns come up!

almarklein (approved):

almarklein:
Thanks!

hmaarrfk:
i think many of these needed `copy=False` in them. the old `arr.shape = ` force `copy=False` implicitely, guaranteeing a shared view. the view is now no longer guaranteed.

hmaarrfk:
but perhaps i'm wrong

almarklein:
My thought was that since they were already cases where a copy was not needed, the copy are was also not needed to enforce it. But now that you mention it, I agree that it would be good to enforce it, to make sure it keeps working as expected in case the surrounding logic is changes later. And maybe also as a check the given data.

@gselzer would you mind adding a `copy=False` to these cases? Then I'll go have a look into rendercanvas-wx-stalling now 😄

hmaarrfk:
since the time i wrote the message, i learned that this would bump your minimum numpy version to 2.1 
* 2.0 does NOT have the `copy` parameter 
https://numpy.org/doc/2.0/reference/generated/numpy.reshape.html
* 2.1 does have the `copy` parameter https://numpy.org/doc/2.1/reference/generated/numpy.reshape.html

almarklein:
Thanks for the heads-up. Same for `ndarray.reshape()`

stanbot8:
Just a heads-up, there are still several direct assignments to NumPy array `.shape`, mainly in the examples as well as one in `Texture.send_data()`.

I collected the remaining replacements in #1313. That PR was closed, but the diff may still be useful if someone wants to finish the cleanup.

PR #1307, opened 2026-07-02
InstancedMesh and InstancedLine now honor instance_buffer.draw_range

almarklein:
This makes it possible to allocate room for say 1M instances and then only use a selection.

### DIFF
--- a/pygfx/renderers/wgpu/shaders/lineshader.py
+++ b/pygfx/renderers/wgpu/shaders/lineshader.py
@@ -314,11 +314,11 @@ class LineShader(BaseShader):
     def get_render_info(self, wobject, shared):
         # Determine how many vertices are needed
         offset, size = self._get_n(wobject.geometry.positions)
-        n_instances = 1
+        inst_offset, inst_size = 0, 1
         if self["instanced"]:
-            n_instances = wobject.instance_buffer.nitems
+            inst_offset, inst_size = wobject.instance_buffer.draw_range
         return {
-            "indices": (size, n_instances, offset, 0),
+            "indices": (size, inst_size, offset, inst_offset),
         }
 
     def get_code(self):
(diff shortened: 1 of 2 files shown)


Vipitis:
Hey, I am on a trip through next week. Will be able to read a few diffs or leave comments but no in depth reviews

kushalkolar:
What's an example use case?

almarklein:
The `draw_range` could already be used on a buffer for e.g. points or lines, but also mesh objects. It tells the renderer to use a selection of the buffer instead of the whole buffer. You can then use that buffer as a collection where elements can be added and removed without having to re-allocate a whole buffer that precisely matches the size.

This PR allows the same for instanced objects. So you can e.g. in a game prepapre 10.000 enemy models, and use a selection of that, matching the number of current enemies. So spawning a new enemy is a matter of increasing the draw_range, without a need to create a new mesh object.

kushalkolar (approved):

kushalkolar:
> The `draw_range` could already be used on a buffer for e.g. points or lines, but also mesh objects. It tells the renderer to use a selection of the buffer instead of the whole buffer. You can then use that buffer as a collection where elements can be added and removed without having to re-allocate a whole buffer that precisely matches the size.
> 
> This PR allows the same for instanced objects. So you can e.g. in a game prepapre 10.000 enemy models, and use a selection of that, matching the number of current enemies. So spawning a new enemy is a matter of increasing the draw_range, without a need to create a new mesh object.

ah maybe a simpler example is if you want to draw a forest of trees with different densities?

almarklein:
> ah maybe a simpler example is if you want to draw a forest of trees with different densities?

Sure that works as well, though it's especially useful in cases where the number of objects changes a lot and you want to avoid allocating/deleting new objects continuously.

PR #1308, opened 2026-07-06
Add copy=False back to reshape calls

gselzer:
Follow-up PR for #1306 - as @hmaarrfk noted (and listed in the [changelog](https://github.com/numpy/numpy/releases?page=4)), I've bumped the minimum bound for numpy to `2.1`. There was no previous minimum bound, so adding *some* minimum bound seems good, but want to draw attention as this version is not quite 2 years old...


almarklein (approved):
Since we need to move away from shape assignment but still want to make sure no copy is made, I think indeed need to require numpy 2.1. In september it will be 2 years old. Not super-old, but also not *that* new.

@Korijn thougths?

almarklein (approved):

PR #1314, opened 2026-08-25
Texcoord interpolation for lines

kushalkolar:
This adds texture coord interpolation options ("flat" and "perspective"). Not sure if `map_interpolation` is the best name for this in the `LineMaterial` and I guess you want an enum for this? 

I think this is only relevant for lines. It's not applicable to `Points` since every displayed fragment is sampled from its own vertex so there's no interpolation. And I don't know much about meshes, but I also think this doesn't really apply to how meshes are used with texture maps?

Almar: Dug into this since you asked a while ago why we're rendering our own colors instead of using texcoords, this is also one reason why :laughing: 

The top line is the desired behavior when using a colormap with categorical/qualitative data and the bottom is what we currently get (perspective interpolation).
image

Things get a bit weird when you rotate and the plane of the fragment is no longer orthogonal to the view direction, but I think it makes sense since it's basically a tube :thinking: 

https://github.com/user-attachments/assets/c4fcfe07-5d60-402b-8d7e-2c0a88541a83

LLM Disclosure: I did use claude code to help with the shader edits but I babysat it through everything and read everything.

### DIFF
--- a/pygfx/renderers/wgpu/shader/resolve.py
+++ b/pygfx/renderers/wgpu/shader/resolve.py
@@ -231,10 +231,21 @@ class VaryingResolver:
             used_builtins = used_varyings.intersection(builtin_varyings)
             used_slots = used_varyings.difference(used_builtins)
             used_slots = list(sorted(used_slots))
+            # Varyings tagged with a `// flatvarying <name>` directive are declared
+            # with @interpolate(flat), so they are not interpolated across the primitive.
+            flat_prefix = "// flatvarying "
+            flat_varyings = {
+                line.strip()[len(flat_prefix) :].split("//")[0].strip()
+                for line in lines
+                if line.strip().startswith(flat_prefix)
+            }
             # Build struct
             struct_lines = ["struct Varyings {"]
             for slotnr, name in enumerate(used_slots):
-                struct_lines.append(f"    @location({slotnr}) {name} : {types[name]},")
+                interpolation = "@interpolate(flat) " if name in flat_varyings else ""
+                struct_lines.append(
+                    f"    @location({slotnr}) {interpolation}{name} : {types[name]},"
+                )
             for name in sorted(used_builtins):
                 struct_lines.append(f"    @builtin({name}) {name} : {types[name]},")
             struct_lines.append("};\n")
--- a/examples/feature_demo/line_map_interpolation.py
+++ b/examples/feature_demo/line_map_interpolation.py
@@ -19,18 +19,18 @@ import cmap
 from rendercanvas.auto import RenderCanvas, loop
 import pygfx as gfx
 
-
-canvas = RenderCanvas(size=(1000, 500))
+canvas = RenderCanvas()
 renderer = gfx.WgpuRenderer(canvas)
 
-# A line whose points carry categorical labels, e.g. cluster ids.
-x = np.linspace(0, 4 * np.pi, 50)
+# A line that represents categorical data, e.g. cluster labels
+x = np.linspace(0, 4 * np.pi, 10)
 y = np.sin(x)
-labels = np.repeat([0, 2, 4, 6, 8], 10)  # five of the ten tab10 colors
+positions = np.column_stack([x, y, np.zeros(10)]).astype(np.float32)
 
-positions = np.array([x * 100, y * 100, np.zeros_like(x)], np.float32).T.copy()
+labels = np.repeat([0, 2, 4, 6, 8], 2).astype(np.float32)
 # Place each label at the center of its texel in the 10-color map.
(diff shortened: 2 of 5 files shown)


Vipitis:
generally looks good, but a couple points:

* There is also an `@interpolate(linear)` option which is linear in screen space. https://gpuweb.github.io/gpuweb/wgsl/#interpolation-type-linear

* In wgpu v30 which we will hopefully get in a few weeks, there is a change that requires flat to be specified for integers. so while not a problem here - it might need more fixes

* More of a comment, and I would also be interested in your opinions. Since the shadercode is generated by all kinds of templating. It's often something you want to inspect while making changes to it. Would it make sense to keep the output code for at least some text or examples in clear text tracked in the repository? Wgpu does something like this for naga. It bloats the repo and the diff, but it it makes it easier to read? We have a few tests for the shader composer. But no whole shader.

kushalkolar:
> There is also an @interpolate(linear) option which is linear in screen space. https://gpuweb.github.io/gpuweb/wgsl/#interpolation-type-linear

Is there a usecase for this with lines though? 

> In wgpu v30 which we will hopefully get in a few weeks, there is a change that requires flat to be specified for integers. so while not a problem here - it might need more fixes

I think we can work that in pre-emptively? If `texcoords.dtype is np.integer`, force interpolate to `"flat"`, could have also an `"auto"` mode that determines based on the `texcoords`?

> Would it make sense to keep the output code for at least some text or examples in clear text tracked in the repository? Wgpu does something like this for naga. It bloats the repo and the diff, but it it makes it easier to read? We have a few tests for the shader composer. But no whole shader.

I can see how this could be useful for checking diffs, but it would be a lot of code for all the permutations! And I do know about people who complain about repo size :sweat_smile:

Vipitis:
> Is there a usecase for this with lines though?

I won't be able to try it for two weeks as I am on a trip right now. But I guess it does make sense over perspective if the line is meant to be viewed orthogonally anyway. It could be faster (or slower) but it likely very hardware and implementation depending. My point is more about including all options 

> I think we can work that in pre-emptively? If `texcoords.dtype is np.integer`, force interpolate to `"flat"`, could have also an `"auto"` mode that determines based on the `texcoords`?

I think that's what we will have to do. I believe the default/assumed is perspective either way. It's probably not a loss to always declare one of the three. Which would simplify the templating logic a bit. 


> I can see how this could be useful for checking diffs, but it would be a lot of code for all the permutations! And I do know about people who complain about repo size 😅

I wouldn't want to write out all permutations. Just maybe all composed shadercode from the existing examples. The two solution I can think of are either host tests in a different repository and use submodule (which I think is awful). Or have some CI script to obtain a diff, post it like a bot comment. And that seems limiting too, as you don't really have it in a proper diff viewer and it's spammy to have bot replies. 
If you include large parts of generated code that changes upon a few small tweak. The LOC number can be misleading. Plus you would need to run it locally before pushing a commit which adds more friction.

pygfx/renderers/wgpu/shader/resolve.py
```diff
@@ -231,10 +231,21 @@ def update_lines(self, lines):
...
             used_slots = used_varyings.difference(used_builtins)
             used_slots = list(sorted(used_slots))
+            # Varyings tagged with a `// flatvarying <name>` directive are declared
+            # with @interpolate(flat), so they are not interpolated across the primitive.
+            flat_prefix = "// flatvarying "
+            flat_varyings = {
+                line.strip()[len(flat_prefix) :].split("//")[0].strip()
+                for line in lines
+                if line.strip().startswith(flat_prefix)
+            }
```
almarklein, line 241: Right now, all lines in the wgsl, of all shaders, are checked for this "flatvarying" hint, which adds significant overhead.

Maybe we can specify that these comments must occur before a certain spot that all shaders have, e.g. `@vertex`. That way the loop above can break early instead of checking all 1000 lines of wgsl :)

almarklein:
> Is there a usecase for this with lines though?

Scientists will find a use-case 😉. Makes sense to add the option for linear now, I think.

About an enum ... yeah, we could. Or just type with a Literal for now. The use of enums and types needs to be straightened out throughout Pygfx at some point...


> Would it make sense to keep the output code for at least some text or examples in clear text tracked in the repository?

I think it would be interesting to see how changes like the current one affect the final wgsl. It would increase the size of the repo a bit, but not *that* much; much less than the images of the examples, for instance.

pygfx/renderers/wgpu/shader/resolve.py
kushalkolar: Is there a way we can also limit it to just the line shader?

PR #1315, opened 2026-08-31
Fix dashed line loops

### LINKED ISSUE #1103: Dash pattern looks weird for looped lines
Image

Notice how the pattern for the `loop`'ed segment is much "tighter" than the others.

```patch
diff --git a/examples/validation/validate_line_loop.py b/examples/validation/validate_line_loop.py
index 76dd3207..3bb3c0c3 100644
--- a/examples/validation/validate_line_loop.py
+++ b/examples/validation/validate_line_loop.py
@@ -107,14 +107,28 @@ positions = np.vstack(
 
 line1 = gfx.Line(
-    gfx.LineMaterial(thickness=14, color="red", opacity=0.7, loop=True),
+    gfx.LineMaterial(
+        thickness=14,
+        color="red",
+        opacity=0.7,
+        loop=True,
+        dash_pattern=[2, 2],
+        dash_offset=2,
+    ),
 )
 scene.add(line1)
 
 
 line2 = gfx.Line(
     gfx.Geometry(positions=rect_points * 10),
-    gfx.LineMaterial(thickness=20, color="cyan", opacity=0.7, loop=True),
+    gfx.LineMaterial(
+        thickness=20,
+        color="cyan",
+        opacity=0.7,
+        loop=True,
+        dash_pattern=[2, 2],
+        dash_offset=2,
+    ),
 )
 scene.add(line2)
 

```

Since 0.10.0

hmaarrfk:
Closes https://github.com/pygfx/pygfx/issues/1103

worked with claude to make this workout.

I think claude is pretty good at fixing these types of bugs when the strong architecture is in place!

image
image


hmaarrfk:
> 🤖 This comment was written by **Claude** (Claude Code), working in @hmaarrfk's checkout and on their behalf. The commits on this branch are authored by them; the analysis, code and the images below are mine. Treat the claims here as machine-generated and check them — every number below comes from a command whose recipe is given, so they should all be reproducible.

## Before / after

### 1. The closing segment (#1103)

The segment that closes a loop was measured as if it spanned the whole loop, so its dashes came out `loop_length / closing_segment_length` times too dense — for a square, 4× too dense, which reads as a solid line. Every shape below is a loop; the affected edge is the one leading back to the first node.

dashed loops, before and after

Closer, on the square in the top-left:

closing segment close-up

The green square in the bottom-right of the first image is the check that the arithmetic is exact, not merely better: its perimeter is a whole multiple of the dash period, so if the closing distance is right the corner where the loop closes must be *indistinguishable* from the other three. It is — including when a dash wraps around it (`dash_offset=1`), where all four corners render identically.

### 2. The dash phase of later pieces (drifting on zoom)

Separate defect, found while looking at the first. The cumulative distance ran continuously across nan-separated pieces, so each piece's dash phase was the accumulated length of every piece before it. Because that length is measured in the thickness space, zooming a `thickness_space="screen"` line changes it — and the later a piece appears in the buffer, the faster its dashes race.

Six loops in one buffer, dash phase at the start of each, over a 10% zoom range (period is 4.0):

| zoom | loop0 | loop1 | loop2 | loop3 | loop4 | loop5 |
|---|---|---|---|---|---|---|
| 1.00 | 0.00 | 2.87 | 3.46 | 1.06 | 1.34 | 0.11 |
| 1.10 | 0.00 | 0.75 | 2.60 | 1.97 | 0.68 | 1.32 |
| **drift, before** | 0.00 | **3.62** | **3.37** | **3.02** | **2.93** | **3.68** |
| **drift, after** | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 |

3.68 out of a 4.0 period is 92% — the pattern sweeps nearly all the way round while you scroll. Each piece now restarts at zero, which is also what SVG does at each subpath.

The same loop rendered at zoom 1.00 → 1.10, before (top) and after (bottom):

zoom drift

## Screenshots

`validate_line_2d.png` is the only existing reference this PR moves. To be sure of that rather than assume it, all 51 `compare` examples were rendered on `main` and on this branch inside one identical lavapipe container: the other 50 come out byte-identical.

validate_line_2d

The change is the dash phase of the piece after the nan gap, which now restarts. The committed PNG is taken verbatim from this PR's own `Screenshots / Regenerate` artifact, so it is CI's render, not a local one.

`validate_line_dash_loop.png` is new. It was generated locally in an `ubuntu:24.04` + `mesa-vulkan-drivers` container; CI's later `Regenerate` artifact agrees with it to `maxdiff=1`, inside the comparison's own `atol`.

## Tests

`tests/renderers/test_line_dashing.py` asserts on the baked cumdist buffer rather than on pixels, because the expected values are exact — a regular n-gon of radius r has known side lengths and a known perimeter. Five of its six tests fail against `main`.

## Two pre-existing bugs, fixed along the way

Both were hit while testing the above, and neither is related to loops:

- `_bake_line_distance` crashed outright for a dashed line with nan positions and `thickness_space="world"` — the fix-up array was allocated with 2 columns while the world-space transform yields 3.
- The nan indices in the loop bake were computed relative to the draw range but used as absolute indices, so loops were marked on the wrong nodes whenever `positions.draw_range` had a nonzero offset.

## Three commits that are not about lines at all

CI was red on this branch for reasons that predate it — dependency drift since main last ran. Each is split into its own commit so it can be dropped or cherry-picked:

- `flit_core` 4.0.2 released, and the `<4` cap makes the non-isolated `python -m build` in *Build release* refuse to run. Widened to `<5`; verified `build -n -s`, `build -n -w`, `twine check`, and that the sdist installs and imports.
- `ruff` grew RUF036, which flags two pre-existing `None | ...` annotations in `renderer.py`.
- `ruff format` now formats Python inside Markdown, and the README example is unformatted.

## One thing that is still red, and is not this PR

*Test examples* failed once on 3.10 with five examples off by 2–382 pixels: `validate_image_colormap`, `validate_line_loop`, `validate_outpass`, `validate_points_markers`, `validate_ppaa`. Four of the five cannot be touched by a line-shader change at all. main's own last CI run fails the **same five with the identical pixel counts** (2, 10, 208, 4, 382) and the same maxdiffs, just on the 3.13 job instead — so it is environment-dependent noise that moves between Python versions. It has since passed on a re-run here, and I deliberately did not regenerate those five: doing so would chase driver noise and would likely break the other version.

## Still open, for a maintainer to decide

Each loop keeps one seam where the dash phase wraps, visible at e.g. the triangle's apex. That is inherent unless the period is stretched to fit the loop a whole number of times — which is exactly what the green square demonstrates when it happens to fit. It could be done in the bake with no shader change, but it would alter dash length by a few percent on every loop, so it wants a deliberate API decision rather than being slipped in behind a bugfix. Happy to implement it either way.

<sub>The images above are hosted on the `pr-1315-media` branch of hmaarrfk/pygfx; it holds nothing but these four PNGs and can be deleted when this PR closes.</sub>

hmaarrfk:

Restructured, so the two kinds of change can be merged separately.

- The three unrelated CI fixes (flit_core, RUF036, README formatting) have moved to #1316.
- Everything about lines is now a **single commit** here, including the two PNGs.

This branch is stacked on #1316. Neither branch can serve as a base in this repo, so until #1316 merges this PR shows its three commits as well; once it does, this rebases onto main and shows only its own commit. Nothing about the content changed in the restructure — the tree at the tip is byte-identical to the one CI already went green on.

The before/after images and the analysis are in the [comment above](https://github.com/pygfx/pygfx/pull/1315#issuecomment-5479150795), still accurate.

Vipitis:
I can't comment in the changes specifically. But will leave a comment about a likely future issue. Wgsl doesn't allow nan at compile time, so there I already a workaround with uniforms. But that also might break in the future, as it has been a problem with the pyodide branch. Not sure how easy it is to avoid semantic nan without introducing other sentinel

hmaarrfk:
> Wgsl doesn't allow nan at compile time, so there I already a workaround with uniforms. 

I don't think i use nan in any "new" way that isn't used already. it mostly just changes how the computation is done for the "accumulation"

hmaarrfk:

Follow-up opened as #1317 (draft): dashes that split instead of sliding when you zoom.

It is stacked on this branch. #1315 fixed the phase error that shows up *between* nan-separated line pieces; #1317 addresses the part that shows up *within* one piece, which re-anchoring cannot reach — the drift is proportional to a dash's distance from its anchor, so there is always some.

Reviewing order is #1316, then this, then #1317. Nothing in #1317 changes existing rendering: all 51 `compare` examples are byte-identical to this branch unless the new `dash_scaling` is set.

pygfx/renderers/wgpu/wgsl/line.wgsl
```diff
@@ -181,15 +190,31 @@ fn vs_main(in: VertexInput) -> Varyings {
...
+            // because they were set before node_index was moved to the loop's start.
             node_index = node_index - loop_node_count;
             node_index_next = node_index + 1;
             is_connecting_node_in_loop = true;
         }
     } else {
         node_index = min(u_renderer.last_i, node_index);
+        $$ if dashing and line_type == 'line'
+        cumdist_index = node_index;
+        $$ endif
```
almarklein, line 217: This is already the default value, so these 3 lines can (IIUC) be removed.

hmaarrfk: We no longer read anything from node_index below and we read from cumdist_index. 

See line 365

hmaarrfk: oh i see what you mean. 
1. On new line 177, we define `cumdist_index = node_index`
2. We redefine it

However, you are missing the fact that top of this exact line, we do `node_index = min(u_rendered.last_i, node_index)`

which can change the result.

almarklein: Ah, makes sense! My bad.

almarklein (approved):

hmaarrfk:
Thanks Almar!

PR #1316, opened 2026-08-31
Fix three CI failures caused by dependency drift

hmaarrfk:
Fix some CI problems that popup up in Aug 2026.

> 🤖 This PR was prepared by **Claude** (Claude Code) acting as @hmaarrfk's agent, in their checkout and at their direction. The commits are authored by them; the diagnosis and the changes are mine. Please read it as machine-generated work — each claim below came from a command that is named, so it is all reproducible.

Three unrelated one-line-ish fixes for CI failures that have nothing to do with any feature work. They are all dependency drift: main's last CI run predates these releases, so main was green when it ran and would be red if it ran today. Split out of #1315, which was carrying them only because it happened to be the branch that noticed.

Each is a separate commit so any one can be dropped.

### 1. `flit_core` 4 broke *Build release*

The job runs `pip install -U flit` and then `python -m build -n`, so the build uses whatever `flit_core` is current. 4.0.2 is now out, and the `<4` cap makes the non-isolated build refuse to start:

```
ERROR Unmet dependencies (checked against .../bin/python):
	flit_core<4,>=3.2
		wanted: <4,>=3.2
		found: 4.0.2
```

Widened to `<5`. Verified against `flit_core` 4.0.2 that `flit_core.buildapi` is still present, that `python -m build -n -s` and `-n -w` both succeed, that `twine check` passes on both artifacts, and that the resulting sdist installs and imports. The same commands fail on main with the error above.

### 2. `ruff` grew RUF036, which fails *Linting*

```
pygfx/renderers/wgpu/engine/renderer.py:381:40: RUF036 `None` not at the end of the type union.
pygfx/renderers/wgpu/engine/renderer.py:412:40: RUF036 `None` not at the end of the type union.
```

`None | int | float` → `int | float | None`, and the same for `None | float`. Annotation order only; no behaviour change.

### 3. `ruff format` now formats Python inside Markdown, which also fails *Linting*

The example block in `README.md` is not in that style, so `ruff format --check .` fails on a file no feature branch touches. This commit is `ruff format README.md` verbatim: two blank lines before the top-level `def` and `if`, and no trailing blank line inside the fence.

---

After this: `ruff check .` and `ruff format --check .` are both clean, `pytest tests` is 346 passed, and *Build release* goes green.

One note on what is **not** here. *Test examples* fails intermittently with five examples off by 2–382 pixels — `validate_image_colormap`, `validate_line_loop`, `validate_outpass`, `validate_points_markers`, `validate_ppaa`. main's own last run fails the same five with identical pixel counts (2, 10, 208, 4, 382) and the same maxdiffs, just on the 3.13 job rather than 3.10, and it passes on re-run. That looks like driver noise that migrates between Python versions, so regenerating those references would chase the noise and likely break whichever version is not currently failing. Left alone deliberately; happy to dig into it separately if you want.

#1315 is stacked on this branch. Since neither branch can be a base in this repo, #1315 shows these three commits too until this merges; it will be rebased to just its own commit afterwards.

</details?>

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,12 +69,13 @@ cube = gfx.Mesh(
 
 rot = la.quat_from_euler((0, 0.01), order="XY")
 
+
 def animate():
     cube.local.rotation = la.quat_mul(rot, cube.local.rotation)
 
+
 if __name__ == "__main__":
     gfx.show(cube, before_render=animate)
-
 ```
 <img src="./docs/_static/guide_rotating_cube.gif" alt="drawing" width="400"/>
 
--- a/pygfx/renderers/wgpu/engine/renderer.py
+++ b/pygfx/renderers/wgpu/engine/renderer.py
@@ -378,7 +378,7 @@ class WgpuRenderer(RootEventHandler, Renderer):
             return pixel_scale
 
     @pixel_scale.setter
-    def pixel_scale(self, pixel_scale: None | int | float):
+    def pixel_scale(self, pixel_scale: int | float | None):
         self._pixel_scale = None
         self._pixel_ratio = None
         if pixel_scale is not None:
@@ -409,7 +409,7 @@ class WgpuRenderer(RootEventHandler, Renderer):
         return self.pixel_scale * target_pixel_ratio
 
     @pixel_ratio.setter
-    def pixel_ratio(self, pixel_ratio: None | float):
+    def pixel_ratio(self, pixel_ratio: float | None):
         self._pixel_scale = None
         self._pixel_ratio = None
         if pixel_ratio is not None:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ Repository = "https://github.com/pygfx/pygfx"
 
 # Flit is great solution for simple pure-Python projects.
 [build-system]
-requires = ["flit_core >=3.2,<4"]
+requires = ["flit_core >=3.2,<5"]
 build-backend = "flit_core.buildapi"
 
 # ===== Tooling


hmaarrfk:

*Test examples 3.10* is red here, and it is worth saying plainly that it cannot be this PR: the diff is three type-annotation reorderings, a version cap in `pyproject.toml`, and whitespace in a README code fence. Nothing here can change a rendered pixel.

It turns out this PR is a good control for it. The same five examples fail with a **byte-identical fingerprint** in three different places:

| example | main's own last run (3.13) | #1315 (3.10) | this PR (3.10) |
|---|---|---|---|
| `validate_image_colormap` | 2px / max 75 | 2px / max 75 | 2px / max 75 |
| `validate_line_loop` | 10px / max 75 | 10px / max 75 | 10px / max 75 |
| `validate_outpass` | 208px / max 83 | 208px / max 83 | 208px / max 83 |
| `validate_points_markers` | 4px / max 75 | 4px / max 75 | 4px / max 75 |
| `validate_ppaa` | 382px / max 75 | 382px / max 75 | 382px / max 75 |

Same examples, same pixel counts, same maxdiffs — on main, on a rendering PR, and on a PR with no rendering in it. The only thing that moves is *which* Python job draws the short straw; both #1315 and this one have also passed 3.10 on a re-run of the identical commit.

So it reads as one fixed rasterisation difference between two environments that the job selects between non-deterministically, rather than anything anyone's branch did. Counts in the low hundreds of pixels at most, out of a million.

I have deliberately not regenerated those five references. Whichever environment they were regenerated from, the other one would then fail — it would move the red, not remove it. Worth its own issue; the fix probably lies in pinning whatever differs between the two, or in a per-example tolerance, and neither belongs in this PR.

A re-run should turn this green.

Vipitis (approved):
Diff is fine.

Although I am not a fan of the verbosity of these descriptions. wgpu upstream changed their LLM policy to require human written PR descriptions. Which also increases the likelihood that the author really understands what's going on. 

We dropped python 3.10 in wgpu-py already and it's end of life next month. I didn't check which wgpu-py release is on CI for the 3.10 job, so it could be an older version. (Odd that the agent didn't consider this tbh, maybe it wasn't allowed to read upstream commits...)

hmaarrfk:
> wgpu upstream changed their LLM policy to require human

Yeah i'm with that. I try to really make sure that things are clear.

The AI thinks it is doing really "critical work sometimes".


I mean, without this, i would have left my other PR mix in these fixes....

hmaarrfk:
Do you want me to drop Python 3.10, and see if moving it to Python 3.11 will fix thigns?

Vipitis:
> Do you want me to drop Python 3.10, and see if moving it to Python 3.11 will fix thigns?

In the end not my call, but my suggestion would be: drop 3.10 and move the test example jobs to 3.11 and 3.14 as the low and hi bounds. Technically 3.15 could/should work but I haven't tried it with wgpu-py yet.

3.10 should still work since we don't use any 3.11 features as far as I know. But it's just not tested. 


E: also checked the CI logs and both jobs did install wgpu 0.32

Ref for `wgpu-py` dropping 3.10 https://github.com/pygfx/wgpu-py/pull/813#issue-4673267150

hmaarrfk:
> E: also checked the CI logs and both jobs did install wgpu 0.32 red https://github.com/pygfx/wgpu-py/pull/813#issue-4673267150

this is the part that is confusing me, I feel like wgpu "next" that required python 3.11 didn't get released yet.....

Vipitis:
The commit history has the bump later... I can't check the wheel myself atm

hmaarrfk:
I mean the error is:

```
>           np.testing.assert_allclose(img, stored_img, atol=atol)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=1
E           
E           Mismatched elements: 421 / 4000000 (0.0105%)
E           First 5 mismatches are at indices:
E            [41, 1697, 2]: 211 (ACTUAL), 213 (DESIRED)
E            [43, 1696, 2]: 207 (ACTUAL), 209 (DESIRED)
E            [43, 1697, 2]: 193 (ACTUAL), 195 (DESIRED)
E            [44, 1697, 2]: 183 (ACTUAL), 185 (DESIRED)
E            [46, 1706, 2]: 166 (ACTUAL), 168 (DESIRED)
E           Max absolute difference among violations: 2

```

so if we are OK increasing the tolerance to 2. it seems like a driver bug.

almarklein:
I'm +1 on dropping Python 3.10, if that fixes things, for the same reason as we dropped it in wgpu.

almarklein (approved):

hmaarrfk:
It does not fix things in this case

PR #1317, opened 2026-08-31
Dashes that split instead of sliding when you zoom

### LINKED ISSUE #1103: Dash pattern looks weird for looped lines

hmaarrfk:
I would like to consider an update to the dashing algorithm to help make things "more natural" presently the consistent "changing" of the path is dizzying.

#1315 and #1316 are in, so this is now rebased onto plain `main` and ready to look at.

Video recording has audio

https://github.com/user-attachments/assets/ee593927-af6e-4e4f-a2dd-f9e313cbd853

<details>
<summary>claude</summary>

> 🤖 This description was written by **Claude** (Claude Code) acting as @hmaarrfk's agent, in their checkout and at their direction. The commits are authored by them; the design, code and measurements are mine. Read it as machine-generated work and check it — every number below came from a command that is named, so it is reproducible.

**Draft, for evaluation.** The mechanism works and is measured; the API shape is the part that wants a maintainer's opinion.

## Where this sits

#1316 (CI fixes) and #1315 (the dashed-loop fixes — closes #1103, follows on from #1047) have both landed. History is now squashed to two commits — **the feature** and **the examples** — on top of one:

| | |
|---|---|
| #1324 | bake functions get the viewport size, not the canvas size |
| **this** | dashes that split instead of sliding, and dashes that fit the piece |

#1324 is included here because the split-view examples need it: without it the two panes disagree by thousands of pixels on settings that should be identical. It stands on its own and can merge first, after which this rebases down to its own two commits.

This is a genuine follow-on rather than a separate topic. #1315 established that a dash's phase error is proportional to its distance from the anchor, and fixed the part of that which showed up *between* nan-separated pieces. This addresses the part that shows up *within* one piece, which no amount of re-anchoring can reach.

## The problem

With `thickness_space="screen"` the cumulative distance is measured in screen pixels, so for view scale `s`:

```
dash_progress(L) = s * L / thickness   (mod D)
d(phase)/ds      = L / thickness
```

A dash drifts in proportion to its distance from the start of its line piece. Zooming makes the far end of a long line race while the near end sits still, and a dashed circle's pattern rotates as you scroll. Restarting the phase per piece (#1315) removed the drift *between* pieces; by construction it cannot touch the drift *within* one.

dash_scale_step sweep

## The approach

Anchor the pattern in model space and let only its period follow the view, snapped:

```
one dash unit = step ** floor(log_step(thickness * model_units_per_pixel) + bias)
```

Dash edges then sit at multiples of `dash_size * step**level` in model space. When the step is a whole number, the edges of one level are a strict subset of the next level's, so **no edge ever moves**: each dash splits into `step` pieces, and new dashes appear in the gaps.

Measured on a straight line over a smooth 2x zoom in 17 steps:

| | edges surviving frame to frame | worst shift of a surviving edge |
|---|---|---|
| continuous | 14.2% | 0.84 model units |
| quantized | **100%** | **0.000** |

At a level change the edge count went 7 → 13 with all 7 originals intact.

## Two parameters, with the old behaviour as a point in the family

The first draft had only a "quantized or not" switch, which is the wrong shape: it made the existing behaviour a separate mode rather than a value. The base of the quantization fixes that, because `step ** floor(log_step(x) + bias)` tends to `x` as `step -> 1`.

**`dash_scale_step`** (>= 1, default 2) — the factor the size jumps by:

| step | worst \|quantized / continuous - 1\| over a zoom sweep |
|---|---|
| **1.0** | **0.00000012** — float32 noise, i.e. exactly continuous |
| 1.01 | 0.00496 |
| 1.1 | 0.0486 |
| 1.5 | 0.2218 |
| 2.0 | 0.4076 |

**`dash_max_scale`** (default `None` = centred) — where the size range sits. The range is always exactly one step wide, from `dash_max_scale / step` to `dash_max_scale`, and only its placement is free. Over four octaves of zoom (the observed overshoot is the hysteresis factor, 2^0.1 = 1.072):

| `dash_max_scale` | ideal | observed | meaning |
|---|---|---|---|
| `step` | [1.000, 2.000] | [1.000, 2.140] | never finer than asked |
| `sqrt(step)` (default) | [0.707, 1.414] | [0.759, 1.515] | closest to asked |
| 1 | [0.500, 1.000] | [0.500, 1.070] | never coarser than asked |

the four steps

## Hysteresis (`dash_scale_hysteresis`)

The level snap is a Schmitt trigger. A plain threshold flickers badly when the view parks near a boundary — measured at a boundary over 200 frames:

| jitter | level changes without hysteresis | with |
|---|---|---|
| 0.5% | 108 | 0 |
| 5.0% | 108 | 0 |

How far past a boundary it holds is **`dash_scale_hysteresis`**, in fractions of a step, default 0.1 — a material property, alongside the rest of the pattern's appearance. Being able to set it to 0 is the only way to see the flicker it prevents, which is what the test does: 0 size changes at the default against more than 10 at zero. The price is that the dashes may stray that much further from the size asked for — the range widens by `step ** hysteresis` at each end, so at the default step of 2 the bound goes from 1.41 to 1.52.

A monotonic zoom still steps exactly once per level, about 7% later than the nominal point.

## What this does not do, and the things to argue with

* **Only a whole-number step gives dashes that truly never move.** The nesting needs levels to differ by a whole factor. At 1.25 or 1.6 the dashes still jump rather than slide, but they land somewhere new when they do. The dial is continuous; the "never moves" property is not.
* **Within one strongly foreshortened segment the dashes still bunch toward the far end.** The phase is linear in model space and the projection is not. At the nodes quantized and continuous agree to about 1e-6; between them ink coverage matches to 0.2% and the dashes sit a fraction of a period out. Removing that needs the scale chosen per fragment rather than per segment, using the `cumdist_per_pixel` varying the shader already carries.
* **The quantized bake no longer short-circuits.** An earlier version skipped it while one object-wide level held; a per-segment scale depends on the camera, so it runs every frame, at about the cost of the continuous path. Correctness was worth more than the saving.
* **There is a visible pop at each level change**, since the dash size changes by the step. The pop-free version morphs dash lengths across the change instead. I costed it and did not build it: the blocker is that a zero-length stroke does not vanish but renders as a dot, because round caps are added around it (measured: `dash_pattern=[0, 4]` gives 0.206 ink coverage, not 0). Making an inserted dash fade in would mean letting stroke length go negative by the cap radius, which varies per fragment — that is a rework of the dash/cap SDF, not a tweak.
* **A true "pinch in the middle" split does not work.** At level *k* strokes occupy `[0,2], [4,6], …`; at *k-1* they occupy `[0,1], [2,3], [4,5], …`. So `[0,2]` must become `[0,1]` with a *new* stroke at `[2,3]`. Pinching `[0,2]` in the middle converges to `[0,1]` and `[1,2]` — the wrong pattern, half a period out of phase, which would then have to slide to correct.
* **Screen thickness space only.** In model and world space the pattern is anchored to the object already, and quantizing would only make its size wrong; the shader falls back to continuous, and a test asserts it.
* **`LineMaterial` only, not `LineSegmentMaterial`**, which computes its cumdist in the shader rather than in the bake.
* **`dash_offset` is in whole periods** under this mode. A fractional offset shifts the pattern by half a period at each level change — reintroducing exactly the sliding this removes.

## A second knob: `dash_fit`

Added after the above, on the same theme of "the pattern should belong to the line rather than to the frame". A dash pattern has a period and a line piece has a length, and the two are almost never a whole multiple of each other, so the pattern stops wherever the piece ends. On a closed piece that is a **seam**: the pattern comes back round to its own start mid-period, and the stroke that closes the loop is cut short or lands on top of the first one. Every dashed circle in pygfx has one today.

`dash_fit="stretch"` scales each piece by the small factor that makes a whole number of periods span it. It is entirely in the bake -- the cumdist buffer *is* the dash phase, so this is a per-piece multiply, with no WGSL change and no pipeline rebuild.

Open and closed pieces get different targets. A closed piece is fitted to `n` whole periods so the pattern meets itself. An open piece is fitted to `n` periods *less the final gap*, so it begins and ends with a full stroke instead of trailing off into empty space -- which is what "fit the dashes to the path" means everywhere else it exists (Illustrator's *Aligns dashes to corners and path ends*).

The two features compose rather than fight, but only because the count is snapped differently under each scaling. Under `continuous` it is the nearest whole number. Under `quantized` that would break the whole point, since a level change must *split* each dash, which needs the count to be divisible by the step -- and 7 cannot halve to 3.5. So the count is snapped to a power of the step:

```
count = step ** round(log_step(wanted))
```

which is level-independent by construction, since `wanted` is proportional to `step**-level`. Measured over a smooth 16x zoom on a 99-gon:

| | count around the loop |
|---|---|
| `exact` | 5.183 … 82.924 — never whole, so the seam is always there |
| `stretch` | **4, 8, 16, 32, 64** — always whole, always exactly doubling |

So the loop closes at every zoom *and* nothing slides. The cost is one more factor of `sqrt(step)` of slack in the on-screen dash size, on top of quantization's own, and that pieces of different lengths end up with different dash sizes — which is inherent to fitting, not a limitation.

**This could reasonably be split into its own PR.** It is a separate user-facing knob and touches no shader code. It is here because the quantized interaction above is the interesting part of it, and that only exists in the context of this branch.

## Prior art

Worth knowing that nobody does this. Every library surveyed sits at one of two poles — pattern in object space (SVG, Skia, Cairo, Three.js: anchored, unbounded density) or in screen space (Cesium, OpenLayers, matplotlib: constant density, slides). matplotlib has this exact bug. The nearest thing is Mapbox GL JS, whose `line-dasharray` *is* a power-of-two LOD cross-fade — and which is also the one people file issues about, because it quantizes the bake but still stretches the pattern continuously in between.

The in-house precedent is `Ruler` (`pygfx/objects/_ruler.py:508-536`), which picks a step from a 1-2-2.5-5 ladder each frame and places ticks at absolute world multiples, so ticks appear and vanish but never translate. Powers of two rather than 1-2-5 here: the ladder breaks nesting at 5→2, and the payoff of 1-2-5 is round *labels*, which dashes do not have.

## Verification

* `tests/renderers/test_line_dashing.py` — 28 tests, asserting on the baked buffer rather than on pixels, since the expected values are exact. `dash_fit="exact"` is asserted bit-identical to the unfitted cumdist.
* Full suite: 374 passed. `ruff check .` and `ruff format --check .` clean.
* All 51 `compare` examples render **byte-identical** to #1315, on both this machine and in a lavapipe container, so nothing changes unless `dash_scaling` is set.
* `examples/feature_demo/line_dash_scaling.py` steps 1 → 1.25 → 1.6 → 2 on identical geometry.
* `examples/feature_demo/line_dashing_interactive.py` — every parameter on an imgui control, as a split view with pygfx's defaults on the left, since what these parameters change is *motion* and the shape of a closed path. Each pane holds a line, triangle, square, pentagon and 99-gon, so five piece lengths are on screen at once; the seam is visible on the left without moving anything. A **continuous** preset button returns the right pane to today's behaviour — verified pixel-identical to a genuine `dash_scaling="continuous"` material, 0–2 differing pixels out of ~12k across four zooms. Driven offscreen through 1536 control combinations.
* `examples/feature_demo/line_dashing_3d.py` — the same controls on a perspective camera, on the wireframe-cubes scene from `validate_line_thickness_space.py`, with an `fov` slider (0 is orthographic) and a flat square as the on-screen control for the depth-extent rule above. 2304 combinations.
* `examples/feature_demo/line_dash_scaling_interactive.py` puts every parameter on an imgui control, against a line pinned to `dash_scaling="continuous"` as a control, and animates the zoom — because what these parameters change is *motion*, which a static row of screenshots cannot show. Driven offscreen through every combination of the controls (5760 renders, each over a zoom sweep crossing several levels) with the GUI callback exercised on both branches of every widget.

The example also carries a toggle between drawing the five shapes as one nan-separated `Line` and as five separate `Line` objects sharing a material. That is a check rather than a demo — flipping it should not move anything — and it turned up one thing worth recording: the two differ by about 2e-7 relative, roughly 190 antialiased pixels out of 630,000, all at the loop seams. The cause is float32, not logic: a later piece's cumdist is summed straight through the lengths of the pieces before it and only then has its own offset subtracted, so it carries a couple of ulps that an object starting from zero does not. Pre-existing, harmless, and `dash_fit="stretch"` normalises exactly that quantity away (the seam phase comes out at 0.000000 both ways).

## The scale is per segment

`model_units_per_pixel` is not one number. The moment a line has any extent along the view direction, a segment running away from the camera covers far fewer pixels per model unit than one across the view. The first version of this took one factor per object, which is wrong in a way you can see: turn a cube edge-on and its receding edges fill with dashes the near edges do not have. Two segments of equal model length, one across the view and one receding, with a pattern asking for a 32 px period:

| | continuous | per-object | per-segment |
|---|---|---|---|
| face-on | 32.0 px | 35.9 px | **32.0 px** |
| foreshortened | 32.0 px | **14.1 px** | **28.2 px** |

Each segment gets its own level and its own hysteresis. The dashes still hold still, because a zoom changes every segment's scale by the same factor so the levels step together; a rotation changes them by different factors, which is exactly when the dashes *should* be redistributed. The segment closing a loop is foreshortened independently and gets its own scale too.

This also removes the limitation the first version had. `step=1` now reproduces continuous for *any* geometry from *any* camera:

| step=1 vs continuous, max deviation | per-object | per-segment |
|---|---|---|
| flat square, face-on | 0.00% | 0.000098% |
| cube, four camera angles | **49.87%** | **0.000235%** |

<sub>The images are hosted on the `pr-dash-scaling-media` branch of hmaarrfk/pygfx, which holds nothing but those two PNGs and can be deleted when this PR closes.</sub>

</details>

### DIFF
--- a/examples/feature_demo/line_dash_scaling.py
+++ b/examples/feature_demo/line_dash_scaling.py
@@ -1,0 +1,118 @@
+"""
+Dash scaling
+============
+
+How the dash pattern responds to zoom. Scroll to zoom, drag to pan.
+
+All four rows are the same geometry with the same ``dash_pattern``, and differ
+only in ``dash_scale_step``: the factor by which the dash size is allowed to
+jump as the view scale changes.
+
+* ``dash_scale_step=1`` (top) is the old behaviour, and is what
+  ``dash_scaling="continuous"`` does. The size follows the view exactly, so the
+  dashes always have the size asked for -- but their number along the line has
+  to change continuously, so they slide. A dash drifts in proportion to its
+  distance from the start of its line piece, which is why the far end of the
+  long line races while the near end barely moves.
+
+* ``dash_scale_step=2`` (bottom) snaps the size to powers of two. Nothing
+  slides: each time the level changes, every dash splits in two (or pairs
+  merge), and no dash moves. The cost is that the size is only right to within
+  a factor of two.
+
+The two rows between are the same mechanism at 1.25 and 1.6, to show that this
+is one continuous dial rather than two modes. Only a whole-number step gives
+dashes that truly never move, because the dash starts of one level are a subset
+of the next level's only when the levels differ by a whole factor; at 1.25 and
+1.6 the dashes still jump rather than slide, but they land somewhere new when
+they do. The smaller the step, the smaller and more frequent the jumps, until
+at 1 they merge into a continuous slide.
+
+A second parameter, ``dash_max_scale``, chooses where the size range sits: at
+``dash_scale_step`` the dashes are never finer than asked, at 1 never coarser,
+and the default (None) centres it.
+
+Watch the far end of the long line for the sliding, and watch any one dash on
+the bottom row to see it split rather than move.
+"""
+
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
+import numpy as np
+from rendercanvas.auto import RenderCanvas, loop
+import pygfx as gfx
+
+
+canvas = RenderCanvas(size=(1000, 700))
+renderer = gfx.WgpuRenderer(canvas)
+scene = gfx.Scene()
+scene.add(gfx.Background.from_color("#000"))
+
+
+def circle(n, x=0.0, y=0.0, r=55.0):
+    t = np.linspace(0, 2 * np.pi, n, endpoint=False)
+    return np.stack(
+        [x + r * np.sin(t), y + r * np.cos(t), np.zeros_like(t)], axis=1
+    ).astype(np.float32)
+
+
+nanpoint = np.full((1, 3), np.nan, np.float32)
+
+
+def geometry(y):
+    """A long straight line plus two loops, so drift shows up at both ends."""
+    straight = np.array([[-430, y + 70, 0], [430, y + 70, 0]], np.float32)
+    return gfx.Geometry(
+        positions=np.vstack(
+            [
+                straight,
+                nanpoint,
+                circle(96, -150, y),
+                nanpoint,
+                circle(6, 150, y),
+            ]
+        ).astype(np.float32)
+    )
+
+
+# dash_scale_step, colour, label
+VARIANTS = [
+    (1.0, "#f55", "dash_scale_step=1: the old behaviour, exact size but slides"),
+    (1.25, "#5f5", "dash_scale_step=1.25: small, frequent jumps"),
+    (1.6, "#59f", "dash_scale_step=1.6: larger, rarer jumps"),
+    (2.0, "#fd5", "dash_scale_step=2: dashes split in two, and never move"),
+]
+
+for i, (dash_scale_step, color, label) in enumerate(VARIANTS):
+    y = 240 - i * 170
+    material = gfx.LineMaterial(
+        thickness=8,
+        color=color,
+        loop=True,
+        aa=True,
+        dash_pattern=[2, 2],
+        thickness_space="screen",
+        dash_scaling="quantized",
+    )
+    material.dash_scale_step = dash_scale_step
+    scene.add(gfx.Line(geometry(y), material))
+
+    text = gfx.Text(
+        text=label,
+        font_size=15,
+        screen_space=True,
+        anchor="middle-left",
+        material=gfx.TextMaterial(color=color),
+    )
+    text.local.position = (-430, y + 95, 0)
+    scene.add(text)
+
+camera = gfx.OrthographicCamera(1000, 700)
+controller = gfx.PanZoomController(camera, register_events=renderer)
+
+canvas.request_draw(lambda: renderer.render(scene, camera))
+
+if __name__ == "__main__":
+    print(__doc__)
+    loop.run()
--- a/examples/feature_demo/line_dashing_3d.py
+++ b/examples/feature_demo/line_dashing_3d.py
@@ -1,0 +1,483 @@
+"""
+Dashing in 3D, side by side
+===========================
+
+The dashing controls of ``line_dashing_interactive.py``, but on a perspective
+camera and as a split view: **left is pygfx's current behaviour, right is
+whatever the panel says**. One camera drives both halves, so they are always
+looking at the same thing from the same place and the only difference is the
+material.
+
+The scene is the wireframe-cubes scene from
+``validation/validate_line_thickness_space.py`` -- three cubes of the same
+apparent size, but built at model scales of 1, 1/2 and 1/3.3 inside containers
+scaled to match. That is what makes ``thickness_space`` interesting: in model
+space the three are drawn at three different dash sizes even though they look
+the same size on screen.
+
+Each cube is one ``Line``: two square loops joined by four uprights, as
+nan-separated pieces. ``LineSegmentMaterial`` is deliberately not used, because
+it computes its cumulative distance in the shader rather than in the bake, so
+``dash_scaling`` and ``dash_fit`` do not apply to it.
+
+**What 3D adds.** Everything the 2D example shows still holds. The extra
+question here is what one dash unit is *worth* on screen, which stops being one
+number the moment a line has any extent along the view direction: a segment
+running away from the camera covers far fewer pixels per model unit than one
+across the view.
+
+The dash scale is therefore taken per segment. Press **continuous**
+(``dash_fit=exact``, ``dash_scaling=quantized``, ``dash_scale_step=1``) and the
+two halves agree: exactly at ``fov=0``, and to within a fraction of a dash under
+perspective. Turn a cube edge-on with quantization on and its receding edges keep
+roughly the dash size the pattern asks for, instead of being crammed with the
+same number of dashes as the edges facing you.
+
+What perspective still costs, and it is much smaller: the phase is linear in
+model space, while the projection is not, so within one strongly foreshortened
+segment the dashes bunch slightly toward the far end. At the nodes the two
+halves match to about 1e-6; between them the ink coverage matches to 0.2% and
+the dashes sit a fraction of a period out. Removing that too would need the
+scale chosen per fragment rather than per segment.
+
+The flat square floating above the cubes is a useful reference for the opposite
+reason: it has no extent along the view direction when you face it, so it is the
+one shape for which nothing subtle is happening at all.
+
+Orbit with the left mouse button and zoom with the wheel, in **either** pane;
+one camera drives both, so the halves always show the same view.
+"""
+
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
+import math
+
+import numpy as np
+import pylinalg as la
+from rendercanvas.auto import RenderCanvas, loop
+import pygfx as gfx
+from wgpu.utils.imgui import ImguiRenderer
+from imgui_bundle import imgui
+
+
+TITLE = "Dashing in 3D, side by side"
+CANVAS_SIZE = 1500, 820
+PANEL_WIDTH = 360
+LABEL_WIDTH = -170
+
+DASH_PATTERNS = [
+    ("2, 2", [2, 2]),
+    ("4, 2", [4, 2]),
+    ("1, 3", [1, 3]),
+    ("6, 2, 2, 2", [6, 2, 2, 2]),
+]
+THICKNESS_SPACES = ["screen", "world", "model"]
+DASH_SCALINGS = ["continuous", "quantized"]
+DASH_FITS = ["exact", "stretch", "spread"]
+PPAA_MODES = ["none", "fxaa", "ddaa"]
+PIXEL_FILTERS = ["nearest", "linear", "tent", "disk", "bspline", "mitchell", "catmull"]
+PIXEL_RATIOS = [1.0, 2.0]
+
+
+def cube_pieces(size=50.0):
+    """A cube as nan-separated pieces: two square loops and four uprights.
+
+    Drawn this way rather than as twelve loose segments so that there are real
+    joins and closed pieces to dash, which is where the pattern's behaviour
+    actually shows.
+    """
+    s = size
+    corners = [(-s, -s), (s, -s), (s, s), (-s, s)]
+    bottom = np.array([[x, y, -s] for x, y in corners], np.float32)
+    top = np.array([[x, y, s] for x, y in corners], np.float32)
+    uprights = [np.array([[x, y, -s], [x, y, s]], np.float32) for x, y in corners]
+    return [bottom, top, *uprights]
+
+
+def joined(parts):
+    nanpoint = np.full((1, 3), np.nan, np.float32)
+    stacked = [parts[0]]
+    for part in parts[1:]:
+        stacked += [nanpoint, part]
+    return np.vstack(stacked).astype(np.float32)
+
+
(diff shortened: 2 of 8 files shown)


hmaarrfk:

Two red checks, neither attributable to this branch. Both clear on a re-run, which needs write access to this repo.

**`Test py313` — cancelled, not failed.** The job's `Unit tests` step has `conclusion: cancelled`; it was killed mid-run. The identical step passed on py310, py311, py312, py314 and pyinstaller in this same run. Not concurrency either: there is one CI run on this branch and nothing newer superseded it.

**`Test examples 3.10` — the known environment flake.** Same five examples, byte-identical fingerprint, in three unrelated places:

| example | main's own last run (3.13) | #1316 (3.10) | this PR (3.10) |
|---|---|---|---|
| `validate_image_colormap` | 2px / max 75 | 2px / max 75 | 2px / max 75 |
| `validate_line_loop` | 10px / max 75 | 10px / max 75 | 10px / max 75 |
| `validate_outpass` | 208px / max 83 | 208px / max 83 | 208px / max 83 |
| `validate_points_markers` | 4px / max 75 | 4px / max 75 | 4px / max 75 |
| `validate_ppaa` | 382px / max 75 | 382px / max 75 | 382px / max 75 |

#1316 is a useful control there: it changes three type annotations, a version cap and some README whitespace, and fails the same five identically. Only *which* Python job draws the short straw moves.

Two further checks that this is not the new example: `Test examples 3.13` passed on this very commit with the same code, and `examples/feature_demo/line_dash_scaling.py` is marked `test = 'run'`, so it is not in the compare set at all. The `Regenerate` job also passed, so the new example runs clean in CI.

I have deliberately not regenerated those five references: whichever environment they were regenerated from, the other would then fail, which moves the red rather than removing it. Probably deserves its own issue.

Everything this PR is responsible for is green: Linting, Build release, Regenerate, Docs, readthedocs, `Test examples 3.13`, and five of six interpreters.

hmaarrfk:
@almarklein @Vipitis feel free to listen to the video, i would recommend NOT reading the code, i haven't, but I wanted to have the conversation about how we can improve this behavior.

if this is "beyond the scope of PyGFX" happy to move it elsewhere.

Vipitis:
thanks for the video, I think with these visual topics it quite helpful to provide them. And I don't really enjoy reading the very verbose language model output either.

I don't really have a say for pygfx, but it does seem like there is improvements to provide here - so I can share an awesome solution to a somewhat similar problem using fractals and surface derivatives. Here is a great explanation that derives how it works: https://youtu.be/HPqGaIMVuLs and I feel like this in some way solves your problem too. (let me know if see the similarities and how that might be applied to a dashed line instead of halftone dithering).

plus I am finally back from my trips and got a computer in front of me - so I can take a look at code, try examples and do reviews again.

hmaarrfk:
Thanks for sharing, that was a super cool video to watch.

nobody likes to read the AI's long verbosity. 

AI does move so fast that the old strategies of collaborating don't always hold (i'm still trying to find the right balance).

But it does condense what feels like 6 weeks of "back and forth" into 1 post. so I find it hard to remove part of the content.

I mostly ask myself "is this total trash", and usually the answer (with what I am able to get out of Opus) is no.


--------------------------------------------------------------------------------

Please don't read the code until you give me time to read it. I haven't.

However, I feel like many conversations would stall at "well it is a good idea, but it would be hard to do", and AI, makes it possible to free ourselves from "hard to do" and get into the real content.

From the video you shared, the "effect" of new dots appearing between the old ones is really awesome, and much smoother than anything I could have thought of. I'm personally not going to diminish the work that went into that dithering effect, nor am I going to ignore the fact that this took about 2 days of "concentrated" mental work to get things to the point where they are here.

-------------------------------------------------------------------------------- 

Perhaps this is in fact the limit of "shaders that should be part of pygfx". It is quite a cool effect, but of course as with anything, has tradeoffs, and isn't a "clear win" for everybody. Perhaps putting this in a different library (my own, perhaps closed source) is the way to go.

This is part of the conversation I would like to have before nit-picking the code.

almarklein:
I really like the idea of stretching things a bit so the ends fit nicely! And maybe this is by itself enough to fix the dizzying effect of the racing dashes? 

> AI does move so fast that the old strategies of collaborating don't always hold (i'm still trying to find the right balance).

I probably have a more conservative view on this than you. I want to still be able to understand all code added to PyGfx. So if a lot of complexity is added for a small feature, I'd be against it. But glancing over the code, this does not seem too bad in this case, actually.

hmaarrfk:
> I want to still be able to understand all code added to PyGfx.

I think this is fair. This is ultimately up to you.

I want to make sure I avoid any unnecessary review.

> I want to still be able to understand all code added to PyGfx.

You should definitely play with this yourself to see what you like.

```
python examples/feature_demo/line_dashing_interactive.py
```

then set 

* `dash_scale_step` to `1` (and not 2).
* `dash_fit` to `stretch`
* `dash_scaling` continuous is also `useful` if you want to really see what "minimal" changes would do.

I think that without the `dash_scale_step` to 2, it still moves around too much for me. The video from @Vipitis also had an other developer find that this "scaling as a factor of 2" was useful when doing these kinds of effects.

Vipitis:
I had a little play with this, done some sketches and tried to come up with an alternative. The fractal idea doesn't really work out - since you end up with huge gaps. Meaning you would have to do something where you add/remove elements via a fractal pattern and then move them back to the regular internal (or one requested) with a smooth-ish motion.

The one idea I got to was to "fix" the vertices and do the interpolation for each line segment based on it's screen space length, and then stepping to the closest integer number of elements. And then you "add" them in the center of that line segment or if you you go with powers of two, you split them etc. you can design some fun and smooth looking animation on how to interpolate between these. Point being, you avoid the sliding over vertices - but the problem will be that dashes will be uneven when the shapes are not regular. or when some lines are nearly or exactly linear but still use multiple vertices.

Didn't find the time to read the articles or the lineshader code, so no implementation or demo yet and my notes aren't explaining much.

hmaarrfk:
Thanks for taking a look. 

It seems that you are interested in taking some time to think about an improved implementation or algorithm overall?

I honestly don't mean to nerd snipe, and I am certainly respectful of "this is just alot of changes, and these shader optimizations are too niche for pygfx, please find an other package"

PR #1318, opened 2026-08-31
Investigate the machine-dependent validate_* screenshot failures

hmaarrfk:

> 🤖 This PR was prepared by **Claude** (Claude Code) acting as @hmaarrfk's agent, in their checkout and at their direction. The commits are authored by them; the diagnosis and the changes are mine. Please read it as machine-generated work — every claim below came from a command or a CI job that is named, so it is all reproducible.

Stacked on #1316 (`maintenance-ci-fixes`). Since a fork branch cannot be a PR base in this repo, this PR is opened against `main` and therefore also shows #1316's three commits; it will be rebased to just its own once #1316 merges.

This is the one CI failure #1316 deliberately left alone: the five `validate_*` screenshot comparisons that fail on one of the two *Test examples* jobs and pass on the other, in the same run, at the same commit.

**Short version: it is the runner's CPU vendor.** `rcpps` returns different bits on Intel and AMD — it is allowed to — and llvmpipe uses it. The renders are bit-for-bit deterministic given the vendor; across 16 runners the split is perfect and no llvmpipe setting changes it. So the fix is on the comparison side.

---

## Step 1 — the two *Test examples* jobs are the same environment

`.github/workflows/ci.yml` gives the matrix `pyversion: '3.10'` and `'3.13'`, but the `Set up Python` step under it hardcoded `python-version: 3.12`. Both jobs ran CPython 3.12.14; the matrix value only named the job and the artifact. Comparing the two jobs of run [33401555767](https://github.com/pygfx/pygfx/actions/runs/33401555767) — *3.10* red, *3.13* green:

| | 3.10 (failed) | 3.13 (passed) |
|---|---|---|
| runner image | `ubuntu24/20260823.283` | `ubuntu24/20260823.283` |
| mesa | `25.2.8-0ubuntu0.24.04.2` | `25.2.8-0ubuntu0.24.04.2` |
| adapter | `llvmpipe (LLVM 20.1.2, 256 bits)` | `llvmpipe (LLVM 20.1.2, 256 bits)` |
| `Successfully installed` (41 pkgs) | — | byte-identical to 3.10 |

Same commit, same image, same driver, same interpreter, same wheels — one red, one green.

## Step 2 — there are exactly two outcomes, and both are deterministic

The `screenshot-diffs-py*` artifacts from run [29693084673](https://github.com/pygfx/pygfx/actions/runs/29693084673) (2026-07-19, *3.13* failed) and run 33401555767 (2026-08-31, *3.10* failed) have identical sha256 for all five rendered PNGs. Six weeks and a numpy/imageio/wgpu bump apart, the failing render is bit-for-bit the same. It is always the same five examples, and no other job in the workflow has ever been the one to fail (runs 25909892011, 28777492015, 29693084673, 33401555767).

## Step 3 — the differences are one interpolation step

`atol` is 1, so only a difference of ≥ 2 fails. Measured against the checked-in references:

| example | values over atol | of | max diff |
|---|---|---|---|
| `validate_image_colormap` | 2 | 1,440,000 | 2 |
| `validate_line_loop` | 11 | 3,200,000 | 2 |
| `validate_outpass` | 208 | 4,000,000 | 3 |
| `validate_points_markers` | 4 | 5,760,000 | 2 |
| `validate_ppaa` | 421 | 4,000,000 | 2 |

Alpha is never affected. Individual pixels:

```
validate_image_colormap  y=306 x=149  ref=[138 121   0 255]  got=[140 121   0 255]
validate_image_colormap  y=306 x=450  ref=[  0 121 138 255]  got=[  0 121 140 255]
validate_line_loop       y=514 x=492  ref=[234  93  93 255]  got=[232  93  93 255]
validate_ppaa            y=43  x=1697 ref=[154 154 195 255]  got=[154 154 193 255]
```

`validate_image_colormap` is the clearest case: exactly two pixels in a 600×600 image, mirror images of each other about the centre column (149 + 450 = 599), on the row through the 256-entry red→green→blue colormap where one entry is about 2/255 wide. A last-bit difference in the sampled coordinate puts the sample on the other side of a texel boundary and moves the output by exactly one colormap entry. The rest are the same story in an antialiasing coverage weight.

## Step 4 — it is the CPU vendor

Probe runs [33434283344](https://github.com/hmaarrfk/pygfx/actions/runs/33434283344) (8 runners × 4 llvmpipe configurations) and [33434596876](https://github.com/hmaarrfk/pygfx/actions/runs/33434596876) (16 runners), each printing `/proc/cpuinfo` and a sha256 per render:

| runners | CPU | render variant | values over atol |
|---|---|---|---|
| 11 | AMD EPYC 7763 (Zen 3) / EPYC 9V74 (Zen 4) | **A** — matches the references | 0 |
| 13 | Intel Xeon 8370C (Ice Lake) / 8573C (Emerald Rapids) / 6973P-C | **B** — the CI failure, exactly | 646 |

24 runners, 5 microarchitectures, zero exceptions. Variant B's per-example counts are 2 / 11 / 208 / 4 / 421 — the same numbers as the failing CI jobs.

It is **not** AVX-512: two of the AMD runners have `avx512f`/`avx512vl` and still produce variant A. Both variants report `llvmpipe (LLVM 20.1.2, 256 bits)`.

And **nothing configurable moves it**. `LP_NUM_THREADS=1`, `LP_NATIVE_VECTOR_WIDTH=128` (which does change the adapter string to `128 bits`), `GALLIVM_PERF=no_filter_hacks,no_quad_lod`, and `GALLIVM_PERF=nopt` (LLVM optimisation passes off) each produce byte-identical hashes to `default` on every machine. Not tile scheduling, not vector width, not the sampler fast paths, not what the optimiser chose to emit.

## Step 4b — it is not the python version

Worth ruling out explicitly, since the two jobs are named after python versions. On `maintenance-ci-fixes`, i.e. **without** the tolerance fix, 24 runners split evenly over the two versions the matrix now uses ([33447402324](https://github.com/hmaarrfk/pygfx/actions/runs/33447402324), [33447593774](https://github.com/hmaarrfk/pygfx/actions/runs/33447593774)):

| python | AMD | Intel |
|---|---|---|
| 3.11 | 8 pass | **4 fail** |
| 3.13 | 8 pass | **4 fail** |

Every one of those eight failures is the same five examples. The version makes no difference at all; the vendor makes all of it. Moving the matrix off 3.10 fixes the job names, not this.

## Step 5 — the actual instruction

That leaves instruction semantics. `rcpps` and `rsqrtps` are approximations; the ISA only guarantees about 12 bits and leaves the rest implementation defined. A tiny C program compiled with `-msse2` on each runner:

| | `rcpps(3.0, 7.0, 0.1, 1234.5)` | `rsqrtps(...)` |
|---|---|---|
| AMD (8 runners) | `3eaaa800 3e124000 41200000 3a545800` | `3f13c800 3ec18000 404a6800 3ce92800` |
| Intel (8 runners) | `3eaaa000 3e124000 41200000 3a545000` | `3f13c800 3ec18000 404a6000 3ce92800` |

`3eaaa800` vs `3eaaa000` for 1/3 — one ulp of a 12-bit approximation, about 2⁻¹². That is the whole story: llvmpipe uses these in its interpolation and sampling paths, the difference propagates to a texture coordinate or a coverage weight, and where the colormap is steep it comes out as 2 levels in the final 8-bit image. It splits exactly along the same line as the renders.

None of this is under pygfx's control, and none of it is a bug in pygfx, mesa, or the CPUs.

## The fix

Requiring bit-exactness from a JIT that targets the host CPU is asking for something the toolchain does not promise, so `test_examples_compare` now allows a value to exceed `atol` — but only slightly, and only for a small part of the image:

- up to **0.1%** of values may exceed `atol`;
- **no** value may exceed **3**.

The observed noise is at most 3 levels on at most 0.01% of values, so there is a factor of ten of headroom on the count. Checked against the real Intel-variant artifacts (all five now pass) and against synthetic regressions:

| | verdict |
|---|---|
| Intel variant vs. references | pass |
| one pixel off by 4 | **fail** — `1 by more than 3 (0 allowed)` |
| 0.2% of values off by 2 | **fail** — `4139 … (4000 allowed)` |
| whole image off by 2 | **fail** — `1343506 … (4000 allowed)` |

The amplitude ceiling is what keeps this honest: the noise is small-amplitude and sparse, while a real rendering regression moves the pixels it touches much further. The assertion message now also reports both counts and the largest difference, which the old `assert_allclose` message did not.

### Two related things the investigation turned up

**The *Test examples* matrix did not test the versions it named.** It named its two jobs 3.10 and 3.13 while the `setup-python` step under it hardcoded 3.12, so both jobs ran 3.12. That is what made this failure look like it depended on the python version when it does not. The step now uses `${{ matrix.pyversion }}`.

**3.10 cannot be one of them, and it is not a timeout.** `imgui-bundle` has published cp311+ wheels only since 1.6.0, and pygfx requires `>=1.92` because that is what `wgpu`'s own `imgui` extra pins (`imgui-bundle>=1.92.0,<2`). On 3.10 pip therefore falls back to the sdist. That build first fails for a missing apt package:

```
CMake Error at external/glfw/glfw/src/CMakeLists.txt:186 (message):
  RandR headers not found; install libxrandr development package
```

which looked fixable, so [probe run 33437386359](https://github.com/hmaarrfk/pygfx/actions/runs/33437386359) installed `xorg-dev` and friends and gave the build a 60 minute budget. It got to 357 of 390 objects and then failed to compile:

```
external/imgui/bindings/pybind_imgui.cpp:69:59: error:
  cannot convert 'nanobind::detail::ndarray_handle*' to 'nanobind::detail::nb_internals*'
external/hello_imgui/bindings/pybind_hello_imgui.cpp:76:21: error: (same)
```

A nanobind signature mismatch in the 1.92.900 sdist — five minutes in, not a timeout, and not something more time or more apt packages fixes. So the matrix is **3.11 and 3.13**, with the reason in a comment. pygfx itself still supports 3.10 and the unit-test matrix still covers it; this is only about the `examples` extra.

Also worth flagging, not changed here: `pyproject.toml` declares `requires-python >= 3.10`, but the `examples` and `docs` extras cannot be installed on 3.10. Upstream imgui-bundle declares `requires-python >= 3.10` on every release while shipping no cp310 wheel, which is why pip tries the source build instead of reporting an unsupported version.

**The references are a coin flip too.** `.github/workflows/screenshots.yml` regenerates them on a plain `ubuntu-latest` runner, so which vendor's rendering gets committed depends on which machine that job drew. The current references are the AMD variant. Without a tolerance the project is one regeneration away from flipping which half of the fleet fails; with it, either variant is acceptable.

## Verification

Two rounds of the real `pytest -v examples` on 16 `ubuntu-latest` runners each — [33434930667](https://github.com/hmaarrfk/pygfx/actions/runs/33434930667) and [33435240990](https://github.com/hmaarrfk/pygfx/actions/runs/33435240990). **32 for 32 green**, `194 passed` on every one. Four of them drew Intel boxes (three 8573C, two 6973P-C), which is where it matters; the AMD ones confirm the change did not loosen anything they were already passing.

That is on top of the offline check: the new comparison was run against the actual failing artifacts downloaded from CI run 33401555767, i.e. against the exact bytes an Intel runner produced.

The temporary `probe` workflow and `probe_render.py` are removed in the last commit. What remains for review is the tolerance change in `test_examples.py` and the examples matrix in `ci.yml`.

### Not fixed here

Nothing else. The remaining diff is deliberately small.

</details>


hmaarrfk:
@Vipitis what do you think of this investigation, the TLDR is that it believes it is at the heart of JIT + CPU

Vipitis:
well, if it's a CPU bug in llvmpipe that's distinct to the GH runners that's well out of our reach. That matches the call to drop 3.10

hmaarrfk:
>  That matches the call to drop 3.10

i'm not sure that is the right conclusion.

The bug seems to be related by the underlying runner, what processor do they have.

so this isn't a python 3.10/3.11 thing.

Vipitis:
Okay, I read the newest edit and assume it's factually correct. Two suggestions:

* The `test examples py3xx` jobs should run exactly that python version. Hard coded 3.12 is clearly a bug and should be fixed. I checked in wgpu-py CI workflows too and there we don't have a specified version for that job its - still 3.12 hardcoded. I don't think FPL copies the job script (it's also possible to reuse the job script... So the pygfx infra cloud be a bit centralized with less duplication and desync, it might be as simple as `using: pygfx/pygfx/ci/test-examples` but I am not sure here, also out of scope.)

* Comparison test with an epsilon of allowed error are problematic in the first place. Since it's arbitrarily chosen. There was an article I read a while ago https://lisyarus.github.io/blog/posts/its-ok-to-compare-floating-points-for-equality.html which explains several cases where comparison is fine and where epsilon is fine. Since the main result of my [undergrad thesis](https://doi.org/10.1109/LLM4Code66737.2025.00017) was that comparing images for semantic similarity only works for exact clones and available metrics are usually for reconstruction (think image compression). There have been a few new ideas which include learning based comparison metrics that look for computer graphics glitches instead of visually apparent compression artifacts [conference talk by Intel](https://youtu.be/DRMoCXrDO24). Also out of scope. Point being, that proving two code snippets are semantically identical is basically solving the halting problem. So you need to exhaustively check all fuzzy tails which is way out of scope again. So increasing the error tolerance will be fine, if you can motivate it's epsilon, that would be best. These test are meant to 1. Spot major differences and 2. notify maintainers that the rendering code has changed.

almarklein:
> The test examples py3xx jobs should run exactly that python version. 

👍 

> Comparison test with an epsilon of allowed error are problematic in the first place.

It is not perfect, but since we know that different implementation don't use the exact same logic (different precision for some operations) we need to do something, and a tolerance + percentage of pixels that is allowed this tolerance seems like a good trade off. In any case, I don't know of a better way. Maybe per-platform screenshots, but that's not-ideal for other reasons. Maybe use another software renderer that does not have this limitation, but out of scope for now, I guess.

@hmaarrfk how does this explain that the CI errors were flaky? Would it be a slightly different running that picked up the job, for instance?

Let's also bump `requires-python` in `pyproject.toml` to 3.11

Korijn:
Just chiming in here to confirm the diagnosis, I already confirmed as much on my own hardware when debugging this years ago. Floating point operations are not equivalent across CPU vendors. Emulation like QEMU (as used commonly with x86 docker containers on apple silicon) can further compound this effect. I also ran into this in the past when storing golden truth test data for floating point math algorithms.

hmaarrfk:
> @hmaarrfk how does this explain that the CI errors were flaky? Would it be a slightly different running that picked up the job, for instance?

Some runners pick up AMD, some runners pick up Intel. They approximate the `pow` function differently, this causes "2 bits of error".

hmaarrfk:
at this point, my recommendation to improve on this is two fold:

1. Relax the tolerance to 3
2. Remove the usage of the transcendental for a faster, more accurate approximation.

hmaarrfk:
Will do each of these in independent PRs.

PR #1320, opened 2026-09-01
improve offscreen rendering tests by changing the implementation of the rgb2physical conversion

hmaarrfk:

**Stacked on #1321.** That is the tolerance bump, which you want in first; since a branch cannot be a PR base here, this PR is opened against `main` and therefore also shows #1321's commit. It will be rebased to just its own three once #1321 merges.

> 🤖 Prepared by **Claude** (Claude Code) acting as @hmaarrfk's agent, at their direction. Commits are authored by them; the diagnosis and the changes are mine. Every number below came from a named command or CI run.

Screenshots are regenerated, by the Screenshots workflow on this PR (run 33512056320) so they match the driver stack CI actually uses — Ubuntu mesa 25.2.8 / LLVM 20.1.2 — rather than whatever a developer happens to have installed.

## Why

`Test examples` fails on roughly a quarter of runs, always the same five `validate_*` comparisons, and it depends on nothing but which machine GitHub hands out. Rendering to an `rgba8unorm-srgb` target hands the linear→sRGB conversion to the driver, and llvmpipe's version of it is built on `rsqrtps` — the ~12-bit approximate reciprocal square root whose low bits x86 leaves implementation-defined. Measured on the runners themselves:

```
AMD    rsqrtps(3.0, 7.0, 0.1, 1234.5) = 3f13c800 3ec18000 404a6800 3ce92800
Intel  rsqrtps(3.0, 7.0, 0.1, 1234.5) = 3f13c800 3ec18000 404a6000 3ce92800
                                                          ^^^^^^^^
```

mesa's own comment on that code:

> *"the constants are magic values. They were found empirically ... This function has an error of max +-0.17. Not sure this is actually enough, we require +-0.6 but that may include the +-0.5 from integer conversion."*

Pinned at mesa `8e12d6000b247715b7c9bfaac67bca565dd8b9d8`:

| what | where |
|---|---|
| the approximation + that comment | [`lp_bld_format_srgb.c:244-292`](https://gitlab.freedesktop.org/mesa/mesa/-/blob/8e12d6000b247715b7c9bfaac67bca565dd8b9d8/src/gallium/auxiliary/gallivm/lp_bld_format_srgb.c#L244-292) |
| what `lp_build_fast_rsqrt` emits | [`lp_bld_arit.c:2616-2639`](https://gitlab.freedesktop.org/mesa/mesa/-/blob/8e12d6000b247715b7c9bfaac67bca565dd8b9d8/src/gallium/auxiliary/gallivm/lp_bld_arit.c#L2616-2639) |
| mesa's deterministic encode, unused by the JIT | [`format_srgb.h:92-126`](https://gitlab.freedesktop.org/mesa/mesa/-/blob/8e12d6000b247715b7c9bfaac67bca565dd8b9d8/src/util/format_srgb.h#L92-126) |

That reference block is reproduced in `colorspace.wgsl`, so it travels with the code.

### How this was isolated

Everything else was ruled out first, AMD vs Intel on identical drivers, 12-16 runners per round:

| probe | vendor-dependent? |
|---|---|
| `(x-r0)/(r1-r0)`, `mat3*vec3`, `x*a+b`, `1.0/(x+1)` | no |
| `pow(x,2.4)`, `pow(x,1/2.4)`, `exp2`, `log2`, `sqrt`, `inverseSqrt` | no |
| `textureSample` nearest, rasterizer interpolation | no |
| `store_rgba8unorm` | **no** |
| **`store_rgba8unorm-srgb`** | **yes — 516/32768** |

Then a scene bisection: solid colours identical, anything with a *range* of colours differs. Sweeping a gradient through the sRGB store, 43 of 2048 values encode differently, and every one has its exact value within 0.125 of a `.5` rounding boundary. Neither vendor is right — against the exact encode, AMD matches on 1913/2048 and Intel on 1898/2048.

## What changed

**Both conversions avoid `pow()`.** Not only for llvmpipe: Vulkan defines `pow(x,y)` as `exp2(y*log2(x))` and allows `exp2`/`log2` 3 ULP outside [0.5, 2.0], so a `pow()` result isn't reproducible between implementations either. Of the float ops only `OpFAdd`, `OpFSub`, `OpFMul` must be correctly rounded.

| | implementation | max error |
|---|---|---|
| `srgb2physical` | degree-6 minimax polynomial in `x`, Horner, mul/add only | 6.0e-6 |
| `physical2srgb` (new) | mesa's basis — powers of a square root — with two extra terms and refitted coefficients | **0.0013** of an 8-bit level |

A plain polynomial can't do the encode direction: the curve is near-vertical at the low end and even degree 12 only reaches ~0.94 of a level. mesa's own version of the same basis is 0.165; this is ~130× better for two more `sqrt`. The old `pow()` one-liners are kept verbatim in comments above each.

**Intermediates move to `rgba16float`.** That's what stops the driver converting the intermediates as well — fixing only the final store left 61% of the divergence in place, which the 16-runner check caught. It was already the stated ideal in that comment (`TODO: use half-floats`), and it gives *more* precision than either 8-bit option, not less. Costs 4 bytes per pixel.

**Result, AMD vs Intel, 7 scenes, verified on CI:** 23,506 differing values → **0**.

## Screenshot impact

44 of the 51 references are within 2 levels of the old ones, which is the expected size of the change. **Four change visibly**, all for one reason, and this is the part of the PR that is a judgement call rather than a fix:

| example | values changed | max |
|---|---|---|
| `validate_light_shadow` | 1572 | 255 |
| `validate_line_thickness` | 40 | 255 |
| `validate_blend_additive` | 5134 | 193 |
| `validate_line_2d` | 9766 | 100 |

The intermediates are `rgba16float` now, so a blend result is no longer clamped to [0,1] at every step the way an 8-bit unorm target clamped it. Values accumulate past 1.0 and saturate only at the final store — e.g. in `validate_blend_additive` a pixel goes `[0, 221, 0, 184]` → `[0, 255, 0, 255]`.

I checked the alternative rather than assume. Plain `rgba8unorm` intermediates keep the clamping but lose precision in the darks, and measured against main are worse overall: **15 of 51 examples within 2 levels, versus 44**. Hence `rgba16float`. If you would rather this PR not touch blending at all, the fallback is to leave the intermediates as `rgba8unorm_srgb` — that keeps every render identical but leaves ~39% of the vendor divergence in place, so the tests get less flaky rather than deterministic.

## Bugs caught while regenerating — none of which a green CI run would have shown

- **Double encode on texture targets.** The old code applied sRGB handling *only* for canvas targets; a `gfx.Texture` target received raw linear output. Generalising it encoded twice for `validate_outpass`, which renders into `gfx.Texture(format="4xu1")` and then displays those textures — `32→96`, `128→187`, exactly `srgb()` applied twice. Now scoped back to canvas targets, with the old texture-target behaviour preserved and commented.
- **Alpha premultiplication order.** The driver encoded the *premultiplied* value the shader returns, and `srgb(rgb)·a ≠ srgb(rgb·a)`. The output pass now encodes `rgb * a`.
- **`snapshot()` leaked float16.** That dtype branch already existed but was dead, because the colour texture was always 8-bit. With `rgba16float` the method started returning float16, which broke `examples/introductory/offscreen.py` with `TypeError: Cannot handle this data type: (1, 1, 4), <f2` from PIL. It converts to sRGB uint8 now, which is the contract it has always had.

## A pre-existing NaN this exposes

`validate_line_thickness` produces **one NaN fragment**, all four channels, at internal pixel (140, 198). The float→unorm conversion maps NaN to zero, so the 8-bit colour buffer was silently swallowing it; a float buffer keeps it and the resolve filter spreads one bad texel into a 4×4 block, which is where that example's `max 255` comes from — `[115,115,115,255]` → `[0,0,0,0]`, alpha included.

This is **not** caused by this PR. main renders bit-identically across two runs on one machine (0 differences across all 51 examples), so the NaN is deterministic and already there. It is not any single line either: the thin material alone, each thickness alone, and each dash pattern alone are all clean — it needs the whole scene. **Worth its own issue.** The output pass now drops non-finite values so a bad fragment stays local, but that is a guard, not a fix.

## Tests

`tests/renderers/test_colorspace.py` renders all 256 8-bit values through the real renderer and asserts each comes back as itself, plus that the round trip is monotonic. Both approximations are constrained so their coefficients sum to exactly 1, so white stays white.

Two existing tests changed, both because the colour texture is no longer 8-bit:

- `test_renderer_picking` asserted **exact** float equality on picked colours. That only held because the readback was quantised to 8 bits; it now compares with `atol=1e-5`.
- `test_renderstate` set a blender format to `rgba16float` specifically to *differ* from the default. That is the default now, so it uses `rgba8unorm`.

## Things a reviewer should push back on if they disagree

- **`get_pick_info()["rgba"]` is now unquantised.** Same values, more precision — but code comparing it exactly to a literal will need a tolerance.
- **+4 bytes/pixel** for the internal buffers.
- **Every reference screenshot changes.** They also become regenerable on any runner, which they were not before: `screenshots.yml` runs on a plain `ubuntu-latest`, so which vendor's rendering got committed was previously a coin flip.
- The `-srgb` suffix is now stripped from the canvas format unconditionally. A canvas that only offers an sRGB format would break; I have not found one, but I have not proven none exists.

</details>


pygfx/renderers/wgpu/engine/blender.py
```diff
@@ -42,27 +42,26 @@ def blend_component(src_factor, dst_factor, operation):
...
+    # llvmpipe that conversion is built on rsqrtps and is not reproducible
+    # between CPU vendors, which made the offscreen screenshot tests flaky. See
+    # the comment above srgb2physical() in std.wgsl.
     #
-    # This is 4 bytes per pixel.
-    # TODO: use half-floats
+    # This is 8 bytes per pixel, up from 4.
     "color": (
-        wgpu.TextureFormat.rgba8unorm_srgb,
+        wgpu.TextureFormat.rgba16float,
```
hmaarrfk, line 58: @almarklein i have a feeling that this is at least in part part of the "limitations" that were happening. the loss of precision here (repeated color conversion at 8 bits) can cause this drift.

pygfx/renderers/wgpu/wgsl/colorspace.wgsl
```diff
@@ -0,0 +1,75 @@
...
+    let x = (color + 0.055) / 1.055;
+    let f = (((((-0.073726393 * x
+        + 0.315543950) * x
+        + -0.614526184) * x
+        + 0.960801971) * x
+        + 0.428587764) * x
+        + -0.017192921) * x
+        + 0.000511814;
+    let t = color / 12.92;
+    return select(f, t, color <= vec3<f32>(0.04045));
```
hmaarrfk, line 46: This is the new "approximation" to get away from pow.

I'm not quite sure on the performance merits, but it is MUCH less readable, but again, `pow` and hopefully the comments, help better justify the merits.

Vipitis: To me this seems awful. Pow is problematic in shader code because it contains undefined behavior with negative numbers, which results in gives vendor specific results and causes all kinds of bugs... (Non Nvidia GPU and screen space bloom is a common one) but I don't feel like an approximation like this for the purpose of CPU error is the proper solution. We should just go with the existing function and pow as it were. I think even the color science package uses it (or a matrix variant). 

Fixing the immediate and increasing the tolerance will do fine. I can't try it myself this week but I think the current wgpu-py release should support even 16bit unorm texture formats (with the native specific features). 

It's not like pygfx is aiming to be a critically accurate renderer at this point. (Although it's fun to explore spectral path tracing, I have done this recently and a repo is on my account).

hmaarrfk: I don’t understand what you are saying. I would like to learn what you have in mind. So please let me rephrase what I understand:

* you don’t think pow should be in shader code
* You don’t like the current implementation because I’m focusing on “software” rendering

I have made the PR to increase the tolerance to be merged independently. It amounts to roughly a few line change. 


I believe that even without software rendering, using repeated multiplication has benefits on many hardware GPUs. However my knowledge is outdated. 

If using a different texture format is preferred that would be great. Do let me know.

almarklein: I think @Vipitis means that this modification seems rather big to solve this specific error for CPU-based renderers.

Also, as soon as one shaders uses pow somewhere, we're back to the same problem. So if we'd do it right, we'd have to forbid `pow` and make all shaders use `our_pow_aprox` instead, and that seems silly.

I think we just need to accept that the output is not always pixel-perfect and deal with that specifically as good was we can.

hmaarrfk:

> 🤖 Written by **Claude** (Claude Code) at @hmaarrfk's direction. Every number below is reproducible from the formulas in `colorspace.wgsl`.

On the complexity concern — here is what the two approximations actually do.

error against the exact sRGB definition, at 8, 10 and 12 bits

The grey band is ±½ an output step: anything inside it cannot change the stored image at that bit depth.

The important thing to notice is what is being replaced. It is **not** `pow()`. It is the driver's fixed-function encode, which never appears in pygfx source — the red curve. On llvmpipe that is `a*x^0.375 + b*x^0.5 + c` with both powers from `rsqrtps`, and mesa's own comment on it reads *"the constants are magic values. They were found empirically ... This function has an error of max +-0.17. Not sure this is actually enough"*.

So this is a **more accurate approximation replacing a less accurate one**, not an approximation replacing an exact formula. And the less accurate one is also the one that disagrees between Intel and AMD.

## Why the bit depth matters

At 8 bits the driver's error stays (just) inside the invisible band, which is why nobody noticed it until CI started comparing images byte-for-byte. It does not stay there:

| output depth | driver today | `physical2srgb` (ours) | `srgb2physical` (ours) |
|---|---|---|---|
| 8-bit | ±0.164 | **±0.0012** | ±0.019 |
| 10-bit (HDR10) | ±0.658 | **±0.0050** | ±0.077 |
| 12-bit (Dolby Vision) | **±2.63** | **±0.0200** | ±0.309 |
| 16-bit | **±42.1** | **±0.320** | ±4.94 |

By 10 bits the driver's encode is off by more than half a step and starts to be visible; by 12 it is off by two and a half steps, which is banding you can see. Ours is still two orders of magnitude inside the invisible band at 12 bits. If pygfx ever renders to a higher-depth target, that column is the difference between a correct gradient and a visibly quantised one.

## srgb → physical: polynomial vs `pow()`

```
srgb2physical: how far the polynomial is from pow(), measured in 8-bit
levels of the FINAL image (i.e. after it is converted back for display)

   peak error : 0.01917 levels
   that is    : 1 part in 52 of one 8-bit step

Every 8-bit value, decoded then re-encoded, must come back as itself:

   pow()      : 256/256 exact
   polynomial : 256/256 exact

Worst-case individual values (8-bit in -> linear -> 8-bit out):

     in   pow() linear    poly linear        diff    out pow  out poly
     20    0.006995410    0.006989483   -5.93e-06         20        20
     50    0.031896033    0.031901957   +5.92e-06         50        50
     96    0.116970668    0.116964744   -5.92e-06         96        96
    152    0.313988713    0.313994636   +5.92e-06        152       152
    153    0.318546778    0.318552702   +5.92e-06        153       153
    206    0.617206562    0.617200638   -5.92e-06        206       206
```

Peak disagreement with `pow()` is **1/52nd of one 8-bit level**, and every one of the 256 values still round-trips exactly. The right-hand columns show it: at the six worst inputs in the whole range, the 8-bit output is identical either way.

## Why not just keep `pow()`

`pow()` is one line and looks exact, but it is not a primitive. Vulkan defines it as `exp2(y * log2(x))` and allows `exp2` and `log2` 3 ULP outside [0.5, 2.0], so its result is implementation-defined too, just less visibly. Of the float operations only `OpFAdd`, `OpFSub` and `OpFMul` are required to be correctly rounded, which is what the Horner form uses.

That said, `pow()` was **never** the source of the CI flakiness — I probed it directly across AMD and Intel on identical drivers and it was bit-identical. Changing it is defence in depth, not the fix.

## If this is still too much

#1322 is the minimal alternative: it keeps `pow()` in both directions, adds no new file, and changes no test. It fixes the flakiness just as completely (23,506 differing values → 0), at the cost of the accuracy in the table above and some precision in the darks. The two PRs are meant to be considered against each other.
</details>

pygfx/renderers/wgpu/wgsl/colorspace.wgsl
hmaarrfk: I think i'm understanding that this isn't being "targeted" only to software renderer but to ALL GPU implementations.

This is probably heavy handed.

hmaarrfk:
I think i understand, i'll have to rethink this. thanks

PR #1321, opened 2026-09-01
Bump screenshot tolerance to 3

hmaarrfk:
We found that CPU hardware affected the floating point optimizations by this much. We are investigating more robust, faster rendering, but for now, this is the approach that makes sense to get reliable CIs

https://github.com/pygfx/pygfx/pull/1318

### DIFF
--- a/examples/tests/test_examples.py
+++ b/examples/tests/test_examples.py
@@ -32,7 +32,18 @@ examples_to_run = [ex[0] for ex in examples_info if ex[2] == "run"]
 examples_to_compare = [ex[0] for ex in examples_info if ex[2] == "compare"]
 
 
-CUSTOM_TOLERANCES = {"validate_text_md": 2}
+# The maximum difference allowed for any single value.
+#
+# Rendering is deterministic on a given machine, but it is not bit-identical
+# across machines. Llvmpipe JITs its shaders for the host CPU, and some of the
+# instructions it leans on (the approximate reciprocals in particular) are only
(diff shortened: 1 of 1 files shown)


examples/tests/test_examples.py
```diff
@@ -32,7 +32,18 @@
...
+# Rendering is deterministic on a given machine, but it is not bit-identical
+# across machines. Llvmpipe JITs its shaders for the host CPU, and some of the
+# instructions it leans on (the approximate reciprocals in particular) are only
+# specified to about 12 bits, so their low bits are implementation defined. On
+# CI that shows up as a few hundred values, out of millions, moving by 2 or 3
+# levels on the examples that have a steep colormap or a lot of antialiased
+# edges: the exact same commit renders one way on every AMD runner and another
+# way on an Intel one. See #1318.
+# https://github.com/pygfx/pygfx/pull/1318
+ATOL = 3
```
almarklein, line 47: I think in a previous iteration of this code it also had a max on the percentage of pixels that was allowed to deviate at all?

hmaarrfk: yeah, i blame AI for overcomplicating things in the previous version

Lets just keep things simple, and bump ATOL to 3 and breathe.

hmaarrfk: The long term fix is to move away from pow. so like this just lets us "move on" from this noise while we get to the pow fix.

almarklein (approved):

PR #1323, opened 2026-09-06
Characterise the seam a line draws across a broken join

hmaarrfk:
Dashed lines draw a thin dark line straight across the inside of a sharp corner. It has always been there; it is easiest to notice on a dashed polygon, where a dash landing on a vertex puts the overlap in the middle of the ink rather than at the end of it.

Found while looking at something else in #1317, but entirely independent of it — it reproduces on plain `main` with a default `LineMaterial`.

**This PR does not fix it.** It pins the defect down with a measurement and an example, and records why the obvious one-line fix is wrong. Left as a draft for that reason.

## The defect, and why `depth_compare="<="` is not the fix

A row of corners from 90° down to 30°, translucent, rendered offscreen with `ppaa` off at 1:1 so every pixel is a real fragment, cropped to the corner and blown up 8×.

**Top row — `depth_compare="<"`, i.e. today.** A dark hairline across the inside of each corner from 70° down.
**Bottom row — `depth_compare="<="`, the obvious candidate.** The dark seam is gone and a *bright* lens has taken its place.

translucent comparison

Every inked pixel of a uniformly translucent line over a flat background must composite to the same value. At a 60° corner:

| `depth_compare` | interior min | interior max | |
|---|---|---|---|
| `"<"` (today) | **70** | 170 | dark seam |
| `"<="` (candidate) | 170 | **210** | bright patch |

`<=` moves the error rather than removing it. Both fragments now survive instead of one being dropped — and surviving means both composite, so the overlap is painted twice. On an *opaque* line that bright patch is invisible, which is exactly why an opaque metric must not be trusted here. A passing test asserts this, so the candidate does not get re-proposed.

## Cause

A line overlaps itself wherever a join is *broken* — a corner too sharp to be mitred, which the shader covers with the two segments' own caps instead. Those faces are coplanar and disagree about coverage where they overlap, so the depth test arbitrates, and the fragment that wins may be the one carrying a partial antialiasing alpha.

Dashing is what makes these corners broken joins at all: the shader mitres up to `max_vec_mag`, which is 100 when solid but drops to 1.5 (~90°) when dashing. So a *solid* 60° corner is properly mitred and is clean — that is the control in the test file, not a second symptom.

## Measuring it

The tests assert **uniformity** rather than a reference value, which keeps them adapter-independent: any spread across the eroded interior is a defect, and its direction says which one — darker means a fragment was dropped, brighter means one was composited twice.

This matters, because the opaque seam is a resampling coincidence: invisible at pixel ratio 1, `224/255` at 2, `241/255` at 4. A test built on it can pass by accident. The translucent one shows at every pixel ratio, and a test asserts that.

The requirement is marked `xfail(strict=True)`, so whoever fixes this is told by a failing XPASS rather than having to find the file.

## Where a real fix has to go

Any fix that leaves both faces drawing will hit the double-composite, so this is not reachable by patching the SDF each face evaluates. It points at not emitting overlapping geometry at a broken join at all — clipping the two contributions at the angle bisector so they tile rather than overlap.

An attempt at the union-of-capsules approach (pass the turn angle as a varying, have each cap take the `min` of the two capsule SDFs) was made and abandoned: it cleans up the opaque metric but still reads 70 on the translucent one, and it bleeds ink into gaps when the corner lands mid-gap, because the union ignores the neighbour's dash state.

One useful finding for whoever picks this up: `load_s_positions(index)` is a free indexed loader, so the vertex shader can already read nodes i−2 and i+2. A face knowing both of its bounding nodes' turn angles is reachable without new buffers.

## Known gap

A *solid* corner at ~30° or sharper has a small defect of its own (a spread of about 18), unaffected by `depth_compare` either way. Different mechanism, not understood, deliberately excluded rather than folded in.

## Verification

Suite 360 passed, 4 xfailed. Examples 125 passed, 71 skipped. `ruff check .` and `ruff format --check .` clean. No rendering code is changed, so the `compare` screenshots are untouched by construction.

`examples/validation/validate_line_join_seam.py` draws the row for eyeballing. It is marked `run` rather than `compare` deliberately: it currently renders *incorrectly* on purpose, and pinning that as a reference screenshot would be pinning the bug.

<sub>Images are on the `pr-line-seam-media` branch of hmaarrfk/pygfx, which holds nothing else and can be deleted when this PR closes.</sub>

---

🤖 This description was written by **Claude** (Claude Code) acting as @hmaarrfk's agent, in their checkout and at their direction. The commit is authored by them; the diagnosis, code and measurements are mine. Every number above came from a command that is named, so it is reproducible.

https://claude.ai/code/session_01PrzhWvnA9xhCauwXfx3WXc

### DIFF
--- a/examples/validation/validate_line_join_seam.py
+++ b/examples/validation/validate_line_join_seam.py
@@ -1,0 +1,94 @@
+"""
+Line join seam
+==============
+
+* A row of increasingly sharp corners, dashed on top and solid below.
+* The dash phase is chosen so that a dash straddles each corner.
+* The lines are translucent, which is what makes the defect plain.
+
+**This example currently renders incorrectly**, on purpose: it is the picture of
+an open bug. Every pixel of a uniformly translucent line over a flat background
+must composite to the same value, and on the dashed row the inside of each
+corner from 70 degrees down is crossed by a darker hairline.
+
+A line overlaps itself wherever a join is "broken" -- a corner too sharp to be
+mitred, which the shader covers with the two segments' own caps instead. Those
+faces are coplanar and disagree about coverage where they overlap, so the depth
+test arbitrates, and the fragment that wins may be the one carrying a partial
+antialiasing alpha. That is the seam.
+
+Dashing is what makes these corners broken joins at all: the shader mitres up to
+`max_vec_mag`, which is 100 when solid but drops to 1.5 (about 90 degrees) when
+dashing. Hence the solid row below is clean at the same angles -- it is the
+control, not a second example of the bug.
+
+The corners are drawn separately rather than as one polygon so that each gets
+the same dash phase, which is what makes the row comparable.
+"""
+
+# sphinx_gallery_pygfx_docs = 'screenshot'
+# sphinx_gallery_pygfx_test = 'run'
+
+import numpy as np
+from rendercanvas.auto import RenderCanvas, loop
+import pygfx as gfx
+
+
+canvas = RenderCanvas(size=(1000, 500))
+renderer = gfx.WgpuRenderer(canvas)
+renderer.ppaa = "none"  # the seam is the shader's own, not the AA pass's
+
+scene = gfx.Scene()
+scene.add(gfx.Background.from_color("#000"))
+
+THICKNESS = 14.0
+# The dash phase at the corner is leg_length / THICKNESS, in dash units, so this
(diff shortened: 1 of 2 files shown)


hmaarrfk:
I think the actual fix needs to involve creating an SDF. but i feel like this is already done. i'll have to think about this more deeply.

PR #1325, opened 2026-09-07
Add physical_offset to stdinfo_data accessible in shader

hmaarrfk:
I've found that this was weird since the physical size doesn't really make sense when you use viewports (which I have been using to do side by side comparisons) without this.

Regression test to come.

All code was AI generated (Claude) from the AI noticing the inability to use the physical information.

This message was written by a human.

We generate a canvas with 2 viewports, the triangle shader is the same, but since there is no offset, it is hard for the two triangles to use the stdinfo information

On main:
main_image

This branch:
this_branch_image


examples/validation/validate_viewport_offset.py
```diff
@@ -0,0 +1,126 @@
...
+
+        @fragment
+        fn fs_main(varyings: Varyings) -> FragmentOutput {
+            // The viewport's size and its origin, both in physical pixels.
+            let size = u_stdinfo.physical_size.xy;
+            let offset = u_stdinfo.physical_offset.xy;
+
+            // position.xy is measured from the framebuffer's corner, so the
+            // offset must come off to make it relative to this viewport.
+            let here = varyings.position.xy - offset;
```
hmaarrfk, line 85: The real question is wehther or not we want to allow users to utilize `varying.position.xy` in a meaningful way. I think the answer is yes.

Thus we need the offset.

hmaarrfk: presently, no shader uses it.

